PR #35, opened 2018-10-01
`pack run` to build and immediately run a container from source

scothis:
Opening this PR to collect feedback. It's not yet ready to merge.

Introduces a `pack run` command which builds a container and then
immedatitly runs the newly built container.

A subset of the `pack build` args are supported, including:
- `--path`
- `--builder`
- `--run-image`

And `--port` is introduced as the local port the running container will
receive requests for.

The run command internally leverages the build command's infrastructure.

The reponame is the md5 hash of the working directory.


scothis:
Known issues:
- [ ] need to add tests
- [ ] `ctl-c` is only impactful after the container starts running, it is ignored prior to that

josegonzalez:
Should this also have a method of pulling in a `.env` file for environment variables? Or perhaps flags for setting env vars?

scothis:
@josegonzalez can you expand on what you would like to see?

scothis:
There are two modifications we should make to this PR before merging:
- [ ] use a common prefix for the built image so that images are identifiable
- [ ] instead of using 8080 inside the container, get the default port from the docker expose metadata for the container

@josegonzalez env vars can be set by individual buildpacks, is that sufficient for you?

josegonzalez:
@scothis not sure what you mean by that? Do you mean a user would have to create an `env-buildpack` to do what I am suggesting here? 

Note: I don't think this is blocking, but more of a usability improvement for those utilizing the tool locally to verify their apps will work as expected on deployment.

run.go
```diff
@@ -0,0 +1,198 @@
...
+
+type RunConfig struct {
+	Port  string
+	Build Task
+	// All below are from BuildConfig
+	AppDir   string
+	Builder  string
+	RunImage string
+	RepoName string
+	// Above are copied from BuildFlags are set by init
```
jchesterpivotal: I find the "Above" comments a little confusing, especially with the "All below" comment at the top.

scothis: Would you believe that I copied that comment from BuildConfig, which the first comment is trying to describe? Yea, it's confusing, I'll remove the extra comments

scothis:
@jchesterpivotal ptal

jchesterpivotal:
Thanks. LGTM after change.

run.go
```diff
@@ -0,0 +1,194 @@
...
+		Cli:             bc.Cli,
+		Stdout:          bc.Stdout,
+		Stderr:          bc.Stderr,
+		Log:             bc.Log,
+		FS:              bc.FS,
+		Config:          bc.Config,
+		Images:          bc.Images,
+		WorkspaceVolume: bc.WorkspaceVolume,
+		CacheVolume:     bc.CacheVolume,
+	}
```
ekcasey: I not sure why some of these flags are needed on RunConfig. It seems like `AppDir`, `Builder`, `RunImage`,  `Config`, `Images`, `WorkspaceVolume` and `CacheVolume` aren't needed by the implementation of Run. I think they are only needed to generate the BuildConfig, which you are already saving on RunConfig. Let me know if I am missing something...

scothis: You're right, they are not used. I'll remove them.

scothis:
@ekcasey please take another look

PR #55, opened 2018-10-15
Fix tests relative to URIs support in builder.toml files

ericbottard:
I fixed the tests relative to URIs support.
What needs to be tested in the case of absolute paths is an absolute path appearing in the `builder.toml` file (as opposed to the path to that file being passed as absolute).

To achieve that, I generated the contents of the `builder.toml` files in the tests. Same applies for the http case and dynamic random port.

Please be advised that I focused the relevant tests in this PR.


ericbottard:
Ping @jchesterpivotal @sclevine 
Please take a look at this and if you're ok with it, we can then build on top of it. In particular, we may want to warn users about wrongful usage of `file://` URIs and consider caching the extracted archives between invocations

jchesterpivotal:
LGTM.

PR #72, opened 2018-11-20
Use docker/pkg/archive tar

dgodd:
Both create tar file and untar


dgodd:
While this change passes all tests on my machine it fails in travis on both linux and macos. I am going to close this PR unmerged.

PR #73, opened 2018-11-22
Package a binary suitable for containerizing.

nicerobot:
The last time I used Go binaries in containers they required static builds and this was the technique.
I wasn't able to get the released linux binary to work in a container but if there's a better solution, I look forward to learning it. Also, i've never done this with travis so I hope it works :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ script:
 - |
   pack_version="$TRAVIS_TAG (git sha: $(git log -1 --pretty=%H))"
   pack_artifact_version="${TRAVIS_TAG:-$TRAVIS_BUILD_NUMBER}"
+  GOOS=linux CGO_ENABLED=0 go build -ldflags "-s -w -X 'main.Version=${pack_version}'" -o "artifacts/pack-${pack_artifact_version}-docker" ./cmd/pack
   GOOS=linux go build -ldflags "-X 'main.Version=${pack_version}'" -o "artifacts/pack-${pack_artifact_version}-linux" ./cmd/pack
   GOOS=darwin go build -ldflags "-X 'main.Version=${pack_version}'" -o "artifacts/pack-${pack_artifact_version}-darwin" ./cmd/pack
   GOOS=windows go build -ldflags "-X 'main.Version=${pack_version}'" -o "artifacts/pack-${pack_artifact_version}-windows" ./cmd/pack


sclevine:
Thanks for pointing this out @nicerobot.

The released Linux binary does work in containers, just not containers that use Alpine (due to [libmusl](https://en.wikipedia.org/wiki/Musl)). Assuming we don't need cgo, I support making this change to the Linux binary in the next release (although I'm not sure if we want to strip the binary using the additional linker flags).

That said, the pack CLI is intended to be used on a local workstation with a Docker daemon. It won't work from within a container unless you run a Docker daemon nested in the container, bind mount a host Docker daemon through, or run a daemon over the network. If you're building on a cloud container platform, I'd suggest [using the lifecycle directly](https://github.com/buildpack/lifecycle/tree/master/images) instead. The pack CLI is intended to orchestrate Docker containers to implement the CNB model. If you're orchestrating your own containers, you don't actually need it. :smile:

sclevine:
> Part of why i do this is to avoid the need to manage local binaries and to ensure that the team doesn't inadvertently apply an update to tooling that hasn't been tested yet.

>  another reason I run it in a container is because I'm on a mono-repo and I want to be able to mount other folders into container so that I can share code across the services

Seems totally reasonable. We only have [sample v3 buildpacks](https://github.com/buildpack/samples) available right now, so most of our users are buildpack authors or cloud platform maintainers. Just wanted to make sure the lifecycle wasn't a better fit for your use case.

If you add a sign-off on your commit for the DCO (`git commit -s`), I can merge this PR. (Although I might adjust the flags to [match the lifecycle](https://github.com/buildpack/lifecycle/blob/master/Dockerfile#L8) after merging.) Alternatively, I'm happy to make the change on master directly if you'd prefer.

>  I've built several Maven and Node.js repositories while testing this. It's quite nice. 

Glad to hear that! Worth noting that the buildpacks that ship with pack right now are examples for buildpack authors. If you're considering using pack in production, you may want to wait until there are production-ready v3 buildpacks available.

(With my Cloud Foundry hat on: production-ready CF v3 Node.js and Java buildpacks will likely be the first v3 buildpacks that we ship.)

sclevine:
The Linux binary is now statically linked (baaa1d3e04cfcabe2a071b2f621c85fb9be706a2).

This should apply to the release that we're planning to cut today. 😄

ekcasey:
Hi @nicerobot we changed `master` to make future builds of the linux binary static. Thanks for the contribution

PR #75, opened 2018-11-26
Escape '/' chars in buildpack ID when creating builder

jkutner:
I'm using `_` to replace `/` because the former is not allowed per the spec.


jkutner:
@ekcasey I added a new commit here to do the same escapeID stuff in builder.go. My tests are in bad shape locally, so i'm unsure what will pass on Travis.

PR #76, opened 2018-12-07
Added support for CMD and ENTRYPOINT

jkutner:
This change depends on https://github.com/buildpack/lifecycle/pull/44. It does the following:

* Sets `ENTRYPOINT` of the exported image to `/lifecycle/launcher`. After lifecycle is updated, we can read the entrypoint from origImage and use that value.
* Adds support for custom `CMD` from the command-line with the `--cmd` flag.


jromero:
@jkutner it seems like this PR has grown a bit stale. Are we still interested in providing such a feature in pack?

jkutner:
paging @hone. will this help with the entrypoint/cmd stuff you've been working on?

hone:
Based off of the Buildpack WG meeting today, going to close this and open a RFC around adding a `--process` flag to `pack`.

PR #128, opened 2019-03-29
Fix Default App Directory

hone:
per lifecycle#77 the default app directory shouldn't be prefixed with `/app`

### DIFF
--- a/build/phase.go
+++ b/build/phase.go
@@ -94,7 +94,7 @@ func (p *Phase) Run(context context.Context) error {
 		return errors.Wrapf(err, "failed to create '%s' container", p.name)
 	}
 	p.appOnce.Do(func() {
-		appReader, _ := archive.CreateTarReader(p.appDir, launchDir+"/app", p.uid, p.gid)
+		appReader, _ := archive.CreateTarReader(p.appDir, launchDir, p.uid, p.gid)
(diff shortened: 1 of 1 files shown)


jkutner:
i don't know why, but when i pull this branch locally, i see:

```
$ git diff master..fix-default-app-directory
...
@@ -419,8 +403,7 @@ func (b *BuildConfig) build(ctx context.Context, lifecycle *build.Lifecycle) err
                "builder",
                build.WithArgs(
                        "-buildpacks", buildpacksDir,
-                       "-layers", layersDir,
-                       "-app", appDir,
+                       "-layers", launchDir,
                        "-group", groupPath,
                        "-plan", planPath,
                        "-platform", platformDir,
```

unsure why Github doesn't show that for me

jkutner:
i think it just needs a rebase

sclevine:
@ekcasey I see some other places that reference `/workspace/app` or look like they might need to be updated for the directory rename: https://github.com/buildpack/pack/blob/master/build.go#L77

hone:
my bad, the rebase fixed it locally for me. sorry :( going to close this.

ekcasey:
@sclevine probably shouldn't have put https://github.com/buildpack/lifecycle/issues/77 in done. The lifecycle changes are done which was higher priority b/c we wanted to finish lifecycle beta first so that folks could create their beta builders.

 More work need to be done to allow `pack` to use the new defaults. We need to create a second volume for the app. Until that is done `pack` need to use the flags to explicitly set values to the original default.

PR #131, opened 2019-04-01
Skip restore and analyze when clearing cache

### LINKED ISSUE #101: Don't restore launch metadata when `--clear-cache`

jromero:
Resolves #101


ekcasey:
This looks great!

 Once the lifecycle support the persistent store https://github.com/buildpack/spec/pull/37, we will probably need to change this to pass a flag to `analyzer` so that `analyzer` can restore only `store.toml` but not  launch layer metadata in the `--clear-cache` case. But given the feedback from multiple users I don't think proper `--clear-cache` behavior should have to wait for those analyzer changes.

PR #140, opened 2019-04-03
'set-default-builder' should suggest builders

### LINKED ISSUE #137: `set-default-builder` should suggest builders
If I run `set-default-builder` with the help flag, or without providing a builder, I'd like a hint at which builders I could potentially use. Whether we display that in the help text, or from another command specific to list suggested builder, that information would be useful for users who already know they need to set a default builder but don't know what builders are available.

jromero:
Resolves #137 


jromero:
Requested changes made. Here's the latest view of the output:

Screen Shot 2019-04-03 at 12 49 31 PM

ekcasey:
thanks for the contribution!

PR #142, opened 2019-04-03
Proposal: support remote daemon using docker volume to store vm cert path

micahyoung:
**Not ready to merge. Hoping for feedback.**

From what I can see, there's no simple way to use TLS-secured remote docker daemon without somehow proxying the `$DOCKER_CERT_PATH` certs contents to lifecycle phase containers on the remote daemon. Without this, there are various failures like. 

```
...
===> RESTORING
[restorer] Error: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

My thought is to use a Docker volume on the remote daemon with the cert path contents prior to running pack commands. Then a custom environment variable would tell pack which volume to use and it would reconstruct the correct environment variables to talk back to the remote daemon.

Workflow example:
```bash
export DOCKER_HOST="tcp://<remote daemon host>:2376"
export DOCKER_TLS_VERIFY="1"
export DOCKER_CERT_PATH="~/.docker"
export PACK_DOCKER_CERT_VOLUME="pack-docker-cert-volume"

docker volume create $PACK_DOCKER_CERT_VOLUME
docker run -d --name docker-certs -v $PACK_DOCKER_CERT_VOLUME:/docker-cert-path alpine tail -f /dev/null
tar cC ~/.docker ca.pem cert.pem key.pem | docker cp - docker-certs:/docker-cert-path
docker rm docker-certs

pack build ...
```

This works with the changes below though they'd of course need be cleaned up and tested. Do you all have thoughts? Is there another approach that I'm missing?

### DIFF
--- a/build/phase.go
+++ b/build/phase.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/buildpack/lifecycle/image/auth"
@@ -71,6 +72,16 @@ func WithArgs(args ...string) func(*Phase) (*Phase, error) {
 func WithDaemonAccess() func(*Phase) (*Phase, error) {
 	return func(phase *Phase) (*Phase, error) {
 		phase.ctrConf.User = "root"
+
+		if dockerVolumeName, ok := os.LookupEnv(`PACK_DOCKER_CERT_VOLUME`); ok {
+			phase.ctrConf.Env = []string{
+				fmt.Sprintf(`DOCKER_HOST=%s`, os.Getenv(`DOCKER_HOST`)),
+				fmt.Sprintf(`DOCKER_TLS_VERIFY=%s`, os.Getenv(`DOCKER_TLS_VERIFY`)),
+				fmt.Sprintf(`DOCKER_CERT_PATH=%s`, `/pack-docker-cert-path`),
+			}
+			phase.hostConf.Binds = append(phase.hostConf.Binds, fmt.Sprintf(`%s:/pack-docker-cert-path`, dockerVolumeName))
+		}
+
 		phase.hostConf.Binds = append(phase.hostConf.Binds, "/var/run/docker.sock:/var/run/docker.sock")
 		return phase, nil
 	}


ekcasey:
@micahyoung If we were going to support this I would want it to "just work" rather than requiring users to create a volume and set an env variable in advance. I would propose that pack create the volume with the certs if it finds a DOCKER_CERT_PATH env var.

Alternatively we could create an ephemeral image with the certs baked in (just add a layer to the ephemeral builder image that pack is already creating). We would only want to use the image with the certs for `analyzer` and `exporter` phases to avoid exposing the credentials to buildpacks.

micahyoung:
Thanks @ekcasey and apologies for the slow reply. I like the just-works philosophy though considering it's potentially relaxing one's security posture a bit by distributing your client credentials, I wonder if this needs to be very opt-in, though still easy. Maybe: 
```
pack build --copy-tls-cert-path ...
```

Also, we could perhaps get clever with the implementation to ensure that the certs don't ever stay in a layer - perhaps tar+base64 encode them into an environment variable have the pack binaries unpack and use them from there.

ekcasey:
@micahyoung I don't think there is a risk of the exported app image containing the client credentials. Only files in the layers directory are ever exported to the app image. Additionally, we would only want to provide the credentials to the containers running the `restorer` `analyzer` `exporter` and `cacher` lifecycle phases. Only those phases will potentially talk to a docker daemon anyways. Containers that invoke the buildpacks themselves (`detector` and `builder` phases) should never have the credentials (presumably there is a lower level of trust for buildpack code than lifecycle code).

But you are right, however we expose the certs to the lifecycle, whether it's through a volume, layer or env var we would be exporting the certs to the docker daemon itself (even if only for a brief time window). Presumably you need those very same credentials to read them back from the docker daemon, but maybe opt in is still a good policy. I don't think base64 encoded env vars are any more secure than an image layer however. In either case, if you have access to the image or container you have access to the credentials.

If we simply added the certs and set the `DOCKER_CERT_PATH` env var, only `pack` would need to understand the complexities of this feature and the lifecycle binaries could simply do the expected thing, making them more portable.

PR #143, opened 2019-04-04
Verify builder exists and is valid when setting default builder

### LINKED ISSUE #135: `pack set-default-builder` should check if builder exists and issue a warning if not
Check remotely and locally perhaps?

jromero:
Resolves #135 

Output:
image

image

image

image

### DIFF
--- a/commands/set_default_builder.go
+++ b/commands/set_default_builder.go
@@ -8,7 +8,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func SetDefaultBuilder(logger *logging.Logger) *cobra.Command {
+func SetDefaultBuilder(logger *logging.Logger, client PackClient) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "set-default-builder <builder-name>",
 		Short: "Set default builder used by other commands",
@@ -21,6 +21,26 @@ func SetDefaultBuilder(logger *logging.Logger) *cobra.Command {
 				return nil
 			}
 
+			imageName := args[0]
+
+			logger.Verbose("Verifying local image...")
+			info, err := client.InspectBuilder(imageName, true)
+			if err != nil {
+				return err
+			}
+
+			if info == nil {
+				logger.Verbose("Verifying remote image...")
+				info, err := client.InspectBuilder(imageName, false)
+				if err != nil {
+					return err
+				}
+
+				if info == nil {
+					return fmt.Errorf("builder %s not found", style.Symbol(imageName))
+				}
+			}
+
 			cfg, err := config.NewDefault()
 			if err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


commands/set_default_builder.go
```diff
@@ -21,6 +21,26 @@ func SetDefaultBuilder(logger *logging.Logger) *cobra.Command {
 				return nil
 			}
 
+			imageName := args[0]
+
+			logger.Verbose("Verifying locally...")
+			info, err := inspector.InspectBuilder(imageName, true)
```
ameyer-pivotal: Does this do any validation on the builder other than existence? I need to look at what scenario a nil info object is returned...

jromero: Yes, it's basically the same mechanism used in `pack inspect-builder` where it retrieves the build info via labels/metadata. 

The info would be nil if the image does not exist. It's not considered an error. That's how the `pack inspect-builder` command denotes "Not present".

I'll add a few snapshots of the output in various scenarios.

jromero:
Just realized I need to also update the acceptance tests. I'll do that as soon as I have some free time.

commands/set_default_builder.go
ameyer-pivotal: I think though that we're still missing the cases where a user sets an existing but terrible image as their builder. For instance, someone recently set `java` as their builder and pack gave no feedback that it wouldn't work.
We should add an additional check that looks for something that only exists on a valid builder (like the "io.buildpacks.builder.metadata" label, possibly even validating the JSON in that label's value, though I'm not sure how far we should go).

jromero: That's already in there. :) See output example 2. I used 'java' explicitly thinking about that use case.

ameyer-pivotal: Ah I see. I think checking for stack ID is sufficient for now.

I did notice some weird output in the error message: `stackID`. Could you change it to `stack ID`? And also have it style the first builder with `style.Symbol` rather than single quotes?
I don't think that was your doing, but it's still in the domain of this story, while we're in that area.

jromero: Requested changes made.

ameyer-pivotal: This is minor too: May want to adjust the verbiage to be consistent between local and remote:
```
Verifying local image...
Verifying remote image...
```

OR

```
Verifying locally...
Verifying remotely...
```

I somewhat prefer the first example because it seems clearer, but that's just me

jromero: I had gone with the latter yesterday (already changed) but I can update it with the former. I do like that better.

Do you know if the rebase will be merged back in? I went ahead and merged the broken master yesterday due to code conflicts before I noticed it was broken. It seems now like master was reverted. Should I revert as well or wait for rebase to go into master?

ameyer-pivotal: Sounds good.

Yes, assume the rebase refactor will be there. I'm about to fix the issue (Windows build fails), and I'll remerge back. You can wait until then to avoid rebase hell.

ekcasey:
Acceptance tests are fixed on master now. Travis will likely pass if you rebase.

PR #152, opened 2019-04-07
suggest-builders command

### LINKED ISSUE #144: pack users can view suggested builders
As a pack cli user
I want the option to reference builders suggested for use by the CNB project
So that I can feel confident about the builder I use when I run ```pack set-default-builder```

**Acceptance Criteria**

_Given_
I have saved the pack binary in an executable path

**Case 1:**  _Display list of suggested builders_

**When** I run ```pack suggest-builders``` 
**Then** I see the list of suggested buildpacks I would see if I tried to run pack build for the first time w/o setting a default-builder

**Case 2:**   ```pack suggest-builders``` _appears in "help" output_

**When** I run ```pack``` 
**Or** ```pack --help```
**Then** I see ```pack suggest-builders``` included in the  list of commands
**And** the associated text reads ```Display list of builders recommended by the Cloud Native Buildpacks project```

**Notes:**

* I don't think users need to be able to run --help on ```pack suggest-builders``` because the text they see in the initial help output should be sufficient for them to understand the purpose of the command

jromero:
Resolves #144 

Output:
image

Help:
image


commands/build.go
```diff
@@ -90,16 +90,16 @@ func suggestSettingBuilder(logger *logging.Logger) {
 
 func suggestBuilders(logger *logging.Logger) {
 	logger.Info("Suggested builders:\n")
+
 	tw := tabwriter.NewWriter(logger.RawWriter(), 10, 10, 5, ' ', tabwriter.TabIndent)
-	for len(suggestedBuilders) > 0 {
-		n := rand.Intn(len(suggestedBuilders))
-		builders := suggestedBuilders[n]
+	for _, i := range rand.Perm(len(suggestedBuilders)) {
```
jromero: This replaces the need to mutate `suggestedBuilders` and prevent consecutive tests to fail due to change of state.

commands/suggest_builders.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/buildpack/pack/logging"
+	"github.com/spf13/cobra"
+)
+
+func SuggestBuilders(logger *logging.Logger) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "suggest-builders",
+		Short: "Display list of builders recommended by the Cloud Native Buildpacks project",
```
ameyer-pivotal: Maybe shorten to: `Display list of recommended builders`?

commands/build.go
```diff
@@ -90,16 +90,16 @@ func suggestSettingBuilder(logger *logging.Logger) {
 
 func suggestBuilders(logger *logging.Logger) {
```
ameyer-pivotal: While you're here, could you move `suggestBuilders` to `commands.go`? Now that it's shared, it should go with the other shared functions.

commands/suggest_builders.go
jromero: I put exactly what the issue acceptance criteria states. Maybe @mgibson1121 can weigh in on this. I'm good to change it to whatever makes the most sense.

commands/build.go
jromero: Will do.

commands/suggest_builders.go
ameyer-pivotal: Ah, my bad. Didn't realize that. I'll leave it to @mgibson1121 then :)

mgibson1121: I'm fine with @ameyer-pivotal's suggestion

ekcasey:
Acceptance tests are fixed on master now. Travis will likely pass if you rebase.

PR #161, opened 2019-04-12
Display error when --path is invalid

### LINKED ISSUE #122: No error when --path is invalid
When passing `--path foobar` to `pack build`, no error is generated, and buildpacks simply fail detection.

jromero:
Resolves #122

**Output:**
image

image

### DIFF
--- a/build.go
+++ b/build.go
@@ -123,6 +123,16 @@ func (bf *BuildFactory) BuildConfigFromFlags(ctx context.Context, f *BuildFlags)
 	if err != nil {
 		return nil, err
 	}
+	stat, err := os.Stat(appDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, errors.Errorf("app directory %s does not exist", style.Symbol(appDir))
+		} else {
+			return nil, err
+		}
+	} else if !stat.IsDir() {
+		return nil, errors.Errorf("provided app directory %s is not a directory", style.Symbol(appDir))
+	}
 
 	f.RepoName = calculateRepositoryName(appDir, f)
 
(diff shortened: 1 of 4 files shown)


build.go
```diff
@@ -191,45 +201,6 @@ func (bf *BuildFactory) BuildConfigFromFlags(ctx context.Context, f *BuildFlags)
 	return b, nil
 }
 
-func Build(ctx context.Context, outWriter, errWriter io.Writer, appDir, builderImage, runImage, repoName string, publish, clearCache bool) error {
```
jromero: I took the liberty of removing this as it seemed unused and duplicate of what the command does.

run.go
```diff
@@ -46,45 +42,6 @@ func (bf *BuildFactory) RunConfigFromFlags(ctx context.Context, f *RunFlags) (*R
 	return rc, nil
 }
 
-func Run(ctx context.Context, outWriter, errWriter io.Writer, appDir, buildImage, runImage string, ports []string) error {
```
jromero: I took the liberty of removing this as it seemed unused and duplicate of what the command does. Can revert if I was a bit hasty.

jromero: Reverted as I just realized this is intended to be used as a library as well.

build.go
jromero: Reverted and adapted as I just realized this is intended to be used as a library as well.

ekcasey:
@jromero Thanks for the contribution!

So far, we have intentionally avoided abstracting the filesystem, instead creating tmp dirs or using test fixtures when necessary for testing. I would like to stick to that pattern if possible.

jromero:
> @jromero Thanks for the contribution!
> 
> So far, we have intentionally avoided abstracting the filesystem, instead creating tmp dirs or using test fixtures when necessary for testing. I would like to stick to that pattern if possible.

Good to know. Made the necessary changes using fixtures.

PR #165, opened 2019-04-19
Resolve symbolic links for app dir

### LINKED ISSUE #93: pack build fails when app directory is symlinked
running pack compiled from master

```
~/riff/squared$ ls -l ../squared
lrwxr-xr-x  1 joeuser  staff  15 Jan 23 12:55 ../squared -> node-fun-square

~/riff/squared$ ~/pack/pack build myrepo/foo:v1 --builder $PFS_BUILDER_IMAGE --run-image $PFS_PACKS_RUN_IMAGE

Defaulting app directory to current working directory /Users/joeuser/riff/squared (use --path to override)

...

===> DETECTING

ERROR: copy app to workspace volume: Error response from daemon: Error processing tar file(exit status 1): cannot overwrite directory "/workspace/app" with non-directory "/"
```

jromero:
Resolves #93 


ekcasey:
@jromero Apologies, we basically did a wholesale rewrite of build as part of our effort to move all the major commands onto the `pack` client. But it should be easier to test now 😅

build_test.go
```diff
@@ -184,6 +184,69 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+			it("resolves relative symbolic links", func() {
+				appDir := filepath.Join("testdata", "some-app")
+				absoluteAppDir, err := filepath.Abs(appDir)
+				h.AssertNil(t, err)
+
+				relLinkRef := "./some-app"
+				relLink := filepath.Join("testdata", "some-app.link")
+
+				_ = os.Remove(relLink)
+				h.AssertNil(t, os.Symlink(relLinkRef, relLink))
```
ekcasey: If we are going to make temporary file for testing we should make them in a temp dir instead of in the repo. We generally only use the testdata dir if we are going to commit the files.

build_test.go
```diff
@@ -184,6 +184,69 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+				linkRef1 := absoluteAppDir
+				absoluteLink1 := filepath.Join("testdata", "some-app-abs-1.link")
+
+				linkRef2 := "some-app-abs-1.link"
+				relLink2 := filepath.Join("testdata", "some-app-rel-2.link")
+
+				_ = os.Remove(absoluteLink1)
+				_ = os.Remove(relLink2)
+
+				h.AssertNil(t, os.Symlink(linkRef1, absoluteLink1))
```
ekcasey: Same as above

testdata/.gitignore
```diff
@@ -0,0 +1 @@
+*.link
```
ekcasey: We wouldn't need this is we created the temporary files in a temp dir

build.go
```diff
@@ -167,20 +167,41 @@ func (c *Client) validateImageReference(imageName string) (name.Reference, error
 	return ref, nil
 }
 
+func mapAppDirError(dir string, err error) error {
```
ekcasey: Pulling out the error logic feels a little awkward to me. I think we could just return the raw error from processAppDir even in the NotExist case. The os.NotExist error will contain the path that we are writing in this custom error, making it seem unnecessary.

jromero: Yeah. I did this prior to the refactor when the error wasn't already prefixed with the path. 

After making the suggested change this is what it looks like:

```
ERROR: invalid app dir '/tmp/does-not-exists': lstat /private/tmp/does-not-exists: no such file or directory
```

build_test.go
jromero: _Full disclosure: I did feel a tad dirty doing it this way._  

I initially attempted to follow the prior practice of using fixtures but when it came time to submit the PR I remembered that git doesn't play nice with symlinks. I obviously should have choose a different path. Thanks for the redirection. Hope it's better now.

ekcasey:
@jromero thanks for the contribution!

PR #176, opened 2019-05-08
Support builder description in builder.toml

ameyer-pivotal:
- Shown in inspect-builder output
- Shown in suggested builder output (use default descriptions if not present)

[buildpack/pack#172]

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -78,6 +78,10 @@ func GetBuilder(img imgutil.Image) (*Builder, error) {
 	}, nil
 }
 
+func (b *Builder) Description() string {
+	return b.metadata.Description
+}
+
 func (b *Builder) GetLifecycleVersion() string {
 	return b.metadata.Lifecycle.Version
 }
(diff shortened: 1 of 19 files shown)


builder/builder.go
```diff
@@ -194,6 +198,10 @@ func hasBPWithVersion(bps []BuildpackMetadata, version string) bool {
 	return false
 }
 
+func (b *Builder) SetDescription(description string) {
```
sclevine: This is starting to look like Java. Why do we need getters/setters here? I see we're returning a concrete `*Builder` in the constructor. If we need to change these properties via an interface,  could we refactor to pull the state out into data structs?

ameyer-pivotal: @sclevine This type follows a builder-esque (not to be confused with a CNB builder lol) pattern so is stateful. If you still think it should be changed, I'm not sure I'm following your proposed solution. Could you clarify?

ameyer-pivotal: Per side conversation, I'll capture this in a chore and merge as is

PR #182, opened 2019-05-10
Logging Refactor

murphybytes:
This PR addresses this [RFC](https://github.com/buildpack/rfcs/pull/6) with the goal of removing project specific logging and replacing it with a minimal logging interface.  The logging interface should allow pack to be used as a separate package with pluggable logging compatible with any number of popular packages such as Go standard library log, [apex](https://github.com/apex/log), [logrus](https://github.com/sirupsen/logrus) and others.  I have attempted to retain backward compatibility with existing logging used by the `pack` command including optional verbose logging, colors and timestamps.  This is accomplished by a shim implemented in `cmd/pack/logging.go`.  The logging interface is defined along with a few helper methods in `github.com/buildpack/pack/logging`.


commands/inspect_builder_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/buildpack/pack/internal/mocks"
```
sclevine: Move to lower import group.

logging/logger.go
```diff
@@ -1,116 +1,68 @@
...
 }
 
-func (l *Logger) RawVerboseWriter() io.Writer {
-	if !l.verbose {
-		return ioutil.Discard
-	}
-	return l.out.rawOut
+// Writes bytes to the embedded log function
+func (w *Writer) Write(buf []byte) (int, error) {
+	w.logMethod(w.prefix + string(buf))
```
sclevine: This prefixes each write instead of prefixing each line. We shouldn't assume that each underlying write corresponds to a log line. Could we allow the logger itself to provide the `io.Writer`?

cmd/pack/logging_test.go
```diff
@@ -0,0 +1,60 @@
+package main
```
sclevine: We have a pattern in this codebase of not testing unexported types inside of the package under test (i.e., we always use `_test` packages).

Could this be an acceptance test (see the `acceptance` package)? Alternatively, could we move these tests and the associated code out of `cmd/pack` entirely?

logging/logger.go
sclevine: Nevermind. This problem exists in the original implementation, and should be fixed by a different PR.

murphybytes:
@sclevine Thanks for the review.  The decision to use github.com/apex/log was mostly because it is so easy to implement a custom log message handler function in cmd/pack/logging.go and more closely approximate the functionality of the pack cli.

build/phase.go
```diff
@@ -129,8 +129,8 @@ func (p *Phase) Run(context context.Context) error {
 		context,
 		p.docker,
 		p.ctr.ID,
-		p.logger.VerboseWriter().WithPrefix(p.name),
-		p.logger.VerboseErrorWriter().WithPrefix(p.name),
+		logging.NewWriter(p.logger.Debug).WithPrefix(p.name),
+		logging.NewWriter(p.logger.Error).WithPrefix(p.name),
```
ameyer-pivotal: Will this still print only when verbose?

commands/build.go
```diff
@@ -57,7 +57,7 @@ func Build(logger *logging.Logger, config *config.Config, packClient *pack.Clien
 			}); err != nil {
 				return err
 			}
-			logger.Info("Successfully built image %s", style.Symbol(imageName))
+			logger.Infof("Successfully built image %s", style.Symbol(imageName))
```
ameyer-pivotal: Do these preserve newlines that the old logger did?

cmd/pack/main.go
```diff
@@ -3,51 +3,65 @@ package main
...
+			cfg = initConfig(logger)
+			packClient = initClient(&cfg, logger)
 		},
 	}
-	rootCmd.PersistentFlags().BoolVar(&color.NoColor, "no-color", false, "Disable color output")
-	rootCmd.PersistentFlags().BoolVar(&timestamps, "timestamps", false, "Enable timestamps in output")
-	rootCmd.PersistentFlags().BoolVarP(&quiet, "quiet", "q", false, "Show less output")
+	rootCmd.PersistentFlags().Bool("no-color", false, "Disable color output")
+	rootCmd.PersistentFlags().Bool("timestamps", false, "Enable timestamps in output")
+	rootCmd.PersistentFlags().Bool("quiet", false, "Show less output")
```
ameyer-pivotal: It seems we lost the `q` short form

cmd/pack/main.go
```diff
@@ -3,51 +3,65 @@ package main
...
+
 	cobra.EnableCommandSorting = false
 	rootCmd := &cobra.Command{
 		Use: "pack",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			logger = *logging.NewLogger(os.Stdout, os.Stderr, !quiet, timestamps)
-			cfg = initConfig(&logger)
-			packClient = initClient(&cfg, &logger)
+			if fs := cmd.Flags(); fs != nil {
+				if flag, err := fs.GetBool("no-color"); err != nil {
```
ameyer-pivotal: Binding to variables seems a bit simpler than calling `GetBool`, though this is pretty straightforward too. Is there a reason this way is preferred?

commands/commands.go
```diff
@@ -74,16 +75,20 @@ func createCancellableContext() context.Context {
 	return ctx
 }
 
-func suggestSettingBuilder(logger *logging.Logger, client PackClient) {
-	logger.Info("Please select a default builder with:\n")
-	logger.Info("\tpack set-default-builder <builder image>\n")
+func suggestSettingBuilder(logger logging.LoggerWithWriter, client PackClient) {
+	out := logger.Writer()
```
ameyer-pivotal: `logger.Infof` seems simpler than this. Is there something I'm missing?

commands/inspect_builder.go
```diff
@@ -48,60 +49,61 @@ func InspectBuilder(logger *logging.Logger, cfg *config.Config, client PackClien
...
 		return
 	}
 
 	if info.Description != "" {
-		logger.Info("\nDescription: %s", info.Description)
+		logger.Infof("\nDescription: %s", info.Description)
 	}
 
-	logger.Info("\nStack: %s\n", info.Stack)
+	_, _ = fmt.Fprintf(out, "\nStack: %s\n\n", info.Stack)
```
ameyer-pivotal: Same question as above about using `Infof` instead. Here and pretty much everywhere there is a `fmt.Fprintln`.

commands/inspect_builder_test.go
```diff
@@ -69,13 +70,11 @@ func testInspectBuilderCommand(t *testing.T, when spec.G, it spec.S) {
 
 				h.AssertContains(t, outBuf.String(), `Remote
 ------
-
-ERROR: some remote error
+some remote error
```
ameyer-pivotal: We should try to preserve the `ERROR` prefix for errors if we can (everywhere, not just here)

downloader_test.go
```diff
@@ -45,7 +40,7 @@ func testDownloader(t *testing.T, when spec.G, it spec.S) {
 			cacheDir, err = ioutil.TempDir("", "")
 			h.AssertNil(t, err)
 
-			subject = NewDownloader(&emptyLogger{}, cacheDir)
+			subject = NewDownloader(mocks.NewMockLogger(ioutil.Discard), cacheDir)
```
ameyer-pivotal: Nice :)

image/fetcher.go
```diff
@@ -82,8 +82,9 @@ func (f *Fetcher) pullImage(ctx context.Context, imageID string) error {
 	if err != nil {
 		return err
 	}
-	termFd, isTerm := term.GetFdInfo(f.logger.RawVerboseWriter())
-	err = jsonmessage.DisplayJSONMessagesStream(rc, &colorizedWriter{f.logger.RawVerboseWriter()}, termFd, isTerm, nil)
+	writer := logging.NewWriter(f.logger.Debug)
+	termFd, isTerm := term.GetFdInfo(writer)
```
ameyer-pivotal: Is this validated empirically? I remember having trouble getting `GetFdInfo` to work on a custom writer.

internal/logging/logging.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/apex/log"
+)
+
+// Terminal colors
+const (
```
ameyer-pivotal: I wonder if these should move to `styles.go`. That tends to be where we aggregate all the color-related tuning.

commands/commands.go
murphybytes: These are normal informational messages that shouldn't be prefixed with INFO/DEBUG/ERROR and are always displayed no matter the desired log level, so we just output the text as formatted.

internal/logging/logging_test.go
```diff
@@ -0,0 +1,60 @@
...
+			logger.Info("test")
+			expected := "2019/05/15 01:01:01.000000 \x1b[34mINFO  \x1b[0m test                     \n"
+			if expected != log.String() {
+				t.Fatalf("actual %q expected %q", log.String(), expected)
+			}
+		})
+
+		it("it has no time and color by default", func() {
+			logger.Info("test")
+			expected := "\x1b[34mINFO  \x1b[0m test                     \n"
```
ameyer-pivotal: This seems to reimplement a bit of the work done by `fatih/colors`. I wonder if there's a way we could utilize that instead.

acceptance/testdata/inspect_builder_output.txt
```diff
@@ -1,15 +1,15 @@
 Inspecting builder: '%s'
 
-Remote
-------
+Remote
```
ameyer-pivotal: Are these whitespace changes intentional?

acceptance/acceptance_test.go
```diff
@@ -28,6 +27,7 @@ import (
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
+	"github.com/buildpack/lifecycle/metadata"
```
ameyer-pivotal: This should be in the second group of imports (technically it's third-party to pack)

acceptance/acceptance_test.go
```diff
@@ -440,9 +440,7 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
 				cmd := packCmd("build", repoName, "-p", filepath.Join("testdata", "mock_app"))
 				output, err := h.RunE(cmd)
 				h.AssertNotNil(t, err)
-				h.AssertContains(t, output, `Please select a default builder with:
-
-	pack set-default-builder <builder image>`)
+				h.AssertContains(t, output, `Please select a default builder with:`)
```
ameyer-pivotal: Curious about the reason for this change. To make the tests less brittle?

acceptance/acceptance_test.go
```diff
@@ -716,14 +712,12 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
 			configuredRunImage := "some-registry.com/pack-test/run1"
 			cmd := packCmd("set-run-image-mirrors", "pack-test/run", "--mirror", configuredRunImage)
 			output := h.Run(t, cmd)
-			h.AssertEq(t, output, "Run Image 'pack-test/run' configured with mirror 'some-registry.com/pack-test/run1'\n")
+			h.AssertContains(t, output, "Run Image 'pack-test/run' configured with mirror 'some-registry.com/pack-test/run1'\n")
```
ameyer-pivotal: Should be able to go back to equality when prefixes are removed for the CLI right?

build/phase.go
```diff
@@ -129,8 +129,8 @@ func (p *Phase) Run(context context.Context) error {
 		context,
 		p.docker,
 		p.ctr.ID,
-		p.logger.VerboseWriter().WithPrefix(p.name),
-		p.logger.VerboseErrorWriter().WithPrefix(p.name),
+		logging.NewPrefixWriter(p.logger.Writer()).WithPrefix(p.name),
```
ameyer-pivotal: Wonder if the prefix should just be part of the constructor now that there's a constructor, rather than a separate method

cmd/pack/main.go
```diff
@@ -3,51 +3,63 @@ package main
...
+			cfg = initConfig(logger)
+			packClient = initClient(&cfg, logger)
 		},
 	}
-	rootCmd.PersistentFlags().BoolVar(&color.NoColor, "no-color", false, "Disable color output")
-	rootCmd.PersistentFlags().BoolVar(&timestamps, "timestamps", false, "Enable timestamps in output")
-	rootCmd.PersistentFlags().BoolVarP(&quiet, "quiet", "q", false, "Show less output")
+	rootCmd.PersistentFlags().Bool("no-color", false, "Disable color output")
+	rootCmd.PersistentFlags().Bool("timestamps", false, "Enable timestamps in output")
+	rootCmd.PersistentFlags().BoolP("quiet", "q", false, "Show less output")
```
ameyer-pivotal: Thanks for putting the `q` back 👍

commands/inspect_builder.go
```diff
@@ -48,60 +53,64 @@ func InspectBuilder(logger *logging.Logger, cfg *config.Config, client PackClien
...
+func inspectBuilderOutput(logger logging.Logger, client PackClient, imageName string, local bool) {
 	info, err := client.InspectBuilder(imageName, local)
 	if err != nil {
 		logger.Error(err.Error())
 		return
 	}
 
 	if info == nil {
-		logger.Info("\nNot present")
+		logger.Info("")
```
ameyer-pivotal: Not opposed to using control characters instead of separate calls, but it's a matter of personal preference. I wonder if we should be consistent throughout our codebase though.

commands/inspect_builder_test.go
ameyer-pivotal: I think we still want the ERROR prefix

ameyer-pivotal: And the newline

commands/set_default_builder_test.go
```diff
@@ -109,7 +111,7 @@ func testSetDefaultBuilderCommand(t *testing.T, when spec.G, it spec.S) {
 				command.SetArgs([]string{imageName})
 
 				h.AssertNotNil(t, command.Execute())
-				h.AssertContains(t, outBuf.String(), "ERROR: failed to inspect image nonbuilder/image")
+				h.AssertContains(t, outBuf.String(), "failed to inspect image nonbuilder/image")
```
ameyer-pivotal: Same comment as before about keeping ERROR prefix

commands/suggest_stacks.go
```diff
@@ -1,16 +1,17 @@
...
 )
 
-func SuggestStacks(logger *logging.Logger) *cobra.Command {
+func SuggestStacks(logger logging.Logger) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "suggest-stacks",
 		Short: "Display list of recommended stacks",
 		Args:  cobra.NoArgs,
-		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
+		RunE: logError(nil, func(cmd *cobra.Command, args []string) error {
```
ameyer-pivotal: I don't think we should be passing nil here. `logError` may need to handle/log a syntax error

internal/logging/logging.go
```diff
@@ -0,0 +1,123 @@
...
+
+type logWithWriters struct {
+	log.Logger
+	out io.Writer
+	errOut io.Writer
+	handler *handler
+}
+
+// Writer returns stdout if level is verbose (debug)
+func(lw *logWithWriters) Writer() io.Writer {
```
ameyer-pivotal: Suggest calling this `DebugWriter` (similar to the old `VerboseWriter`)

internal/logging/logging.go
```diff
@@ -0,0 +1,123 @@
...
+	if lw.Level == log.DebugLevel {
+		return lw.out
+	}
+	return ioutil.Discard
+}
+
+// ErrorWriter - returns stderr if level is verbose (debug)
+func(lw *logWithWriters) ErrorWriter() io.Writer {
```
ameyer-pivotal: Maybe `DebugErrorWriter` similar to the old `VerboseErrorWriter`?

internal/logging/logging.go
```diff
@@ -0,0 +1,123 @@
...
+		return lw.errOut
+	}
+	return ioutil.Discard
+}
+
+func(lw *logWithWriters) WantTime(f bool) {
+	lw.handler.wantTime = f
+}
+
+func(lw *logWithWriters) WantQuiet(f bool) {
```
ameyer-pivotal: This should use the `f` bool

internal/logging/logging_test.go
```diff
@@ -0,0 +1,129 @@
...
+		it.After(func() {
+			log.Reset()
+			errlog.Reset()
+		})
+
+		it("can enable time in logs", func() {
+			logger.WantTime(true)
+			logger.Error("test")
+			expected := "2019/05/15 01:01:01.000000 \x1b[31;1mERROR  \x1b[0mtest                     \n"
+			if expected != log.String() {
```
ameyer-pivotal: We have some custom assertions in `test_helpers.go` that you can use instead. We typically import them under the alias `h`. Look for `h.AssertEqual` for instance.

logging/default_logger_test.go
```diff
@@ -0,0 +1,80 @@
...
+			logger = New(&w)
+		})
+
+		it.After(func(){
+			w.Reset()
+		})
+
+		it("should print debug messages properly", func(){
+			logger.Debug("test")
+			if f, _ := regexp.MatchString(debugMatcher, w.String()); !f {
```
ameyer-pivotal: Same comment as before about test helpers

acceptance/testdata/inspect_builder_output.txt
murphybytes: Yes

internal/logging/logging.go
murphybytes: I'd disagree on this one. I think you're correct that `DebugWriter` is more clear for internal purposes however the tradeoff is that `Writer` is supported by external logging packages that a developer might want to plug into the pack module so `Writer` is part of the logging.Log contract.

ameyer-pivotal: Ah, makes sense. I wonder if this method should always return the writer then, and we provide some other means (method, function) of getting one that's tied to the verbosity/debug setting.

PR #185, opened 2019-05-13
Make pack use launch cache and volume caches when lifecycle supports it

ameyer-pivotal:
[buildpack/pack#178]

Signed-off-by: Danny Joyce <djoyce@pivotal.io>


build/lifecycle.go
```diff
@@ -50,16 +57,34 @@ type LifecycleOptions struct {
 }
 
 func (l *Lifecycle) Execute(ctx context.Context, opts LifecycleOptions) error {
-	cacheImage := cache.New(opts.Image, l.docker)
-	l.logger.Verbose("Using cache image %s", style.Symbol(cacheImage.Image()))
+	l.Setup(opts)
+	defer l.Cleanup()
+
+	lifecycle020OrLater := l.hasVersion020OrLater()
```
ekcasey: maybe we can call this check `supportsLaunchCache` or something like that instead of `hasVersion0200OrLater`

build/lifecycle.go
```diff
@@ -50,16 +57,34 @@ type LifecycleOptions struct {
...
+	var buildCache, launchCache Cache
+	if lifecycle020OrLater {
+		buildCache = cache.NewVolumeCache(opts.Image, "build", l.docker)
+		launchCache = cache.NewVolumeCache(opts.Image, "launch", l.docker)
+		l.logger.Verbose("Using build cache volume %s", style.Symbol(buildCache.Name()))
+		l.logger.Verbose("Using launch cache volume %s", style.Symbol(launchCache.Name()))
```
ekcasey: I wonder if exposing the idea of the launch cache to users is confusing (especially in comparison to build cache). I'm not sure they necessarily need to know about it

build/lifecycle.go
```diff
@@ -134,3 +163,20 @@ func randString(n int) string {
 	}
 	return string(b)
 }
+
+func (l *Lifecycle) hasVersion020OrLater() bool {
```
ekcasey: I wonder if we need to be this strict about the version format, it would be easy to do `strings.Compare` instead (semvers are lexicographically increasing) but i'm open to parsing the semver if we feel it is better.

image/fetcher.go
```diff
@@ -68,10 +65,7 @@ func (f *Fetcher) fetchDaemonImage(name string) (imgutil.Image, error) {
 		return nil, err
 	}
 
-	found, err := image.Found()
-	if err != nil {
-		return nil, err
-	}
+	found := image.Found()
```
ekcasey: inline?

build/lifecycle.go
```diff
@@ -50,16 +57,34 @@ type LifecycleOptions struct {
...
 	if opts.ClearCache {
-		if err := cacheImage.Clear(ctx); err != nil {
-			return errors.Wrap(err, "clearing cache")
+		if err := buildCache.Clear(ctx); err != nil {
+			return errors.Wrap(err, "clearing build cache")
+		}
+		l.logger.Verbose("Build cache %s cleared", style.Symbol(buildCache.Name()))
+
+		if lifecycle020OrLater {
+			if err := launchCache.Clear(ctx); err != nil {
```
ekcasey: There might be some weird performance implications of this. Even when `--clear-cache` is passed, the `exporter` still reads the metadata from the previous image and will try to reuse layers if the digests match (note: there shouldn't be metadata only layers in the `--clear-cache` case b/c analyze has been skipped). This will make things slow if we clear the `launch-cache`.

One could argue that the launch cache is just a more performant previous image rather than a cache? (maybe a stretch)

Unless we played something like this https://github.com/buildpack/lifecycle/issues/20 which would result in zero layer reuse in the `--clear-cache` case (and which, as an aside, actually fixes an interesting race condition) or did something else to prevent the exporter from trying to reuse layers, I don't think we should clear the launch cache (cc @sclevine).

build/lifecycle.go
ameyer-pivotal: Will do!

build/lifecycle.go
ameyer-pivotal: Agreed. We'll remove it.

build/lifecycle.go
ameyer-pivotal: True, though it compares incorrectly for things like `0.2.0-rc.1` vs `0.2.0`. That said, for the purpose of determining whether 0.2.0 features are supported, both `0.2.0-rc.1` and `0.2.0` satisfy `>= 0.2.0`. We'll make the change.

image/fetcher.go
ameyer-pivotal: Done.

build/lifecycle.go
ameyer-pivotal: I agree about thinking of launch cache as a for of "previous image" that happens to be performant. We'll make the change.

PR #188, opened 2019-05-16
Update README with more builder.toml details

ameyer-pivotal:
- Fully documents `builder.toml` structure
- Updates `create-builder` image to represent new changes
- Updates `Builders explained` section with info about lifecycle
- Other stylistic changes

Please fact check and offer any comments/suggestions.

**Rendered preview here**: https://github.com/buildpack/pack/tree/readme-update#pack---buildpack-cli-


ekcasey:
Technically I don't believe the `buildpacks` or `groups` are required in the builder configuration, but perhaps they should be (a builder isn't particularly useful without them). So I think it still makes sense to document them this way.

ameyer-pivotal:
@ekcasey My thought exactly when I wrote that.

README.md
```diff
@@ -225,10 +225,114 @@ that they appear in `builder.toml`. This image's base will be the build image as
 > It's important to note that the buildpacks in a builder are not actually executed until
 > [`build`](#building-explained) is run.
 
+#### Lifecycle
+
+A builder also includes a layer with a series of binaries that are executed during [`build`](#building-explained).
+These binaries represent a specific implementation of the buildpack **lifecycle**, as defined by the
+[Cloud Native Buildpacks API Specification](https://github.com/buildpack/spec#buildpack-api-v3---specification). While
+most users will find the [default lifecycle implementation](https://github.com/buildpack/lifecycle#lifecycle)
+sufficient, users are free to implement their own. See
```
ekcasey: `pack` does rely on unspecified flags that are provided by the default lifecycle. So there is no guarantee that any given lifecycle is compatible with pack, even if it complies with the spec (Another way to put it: this particular platform has a contract with the lifecycle that is a superset of the specified contract). Perhaps we shouldn't mention custom lifecycle implementations in the `pack` docs.

ameyer-pivotal: Great point. I agree we should change the wording. It's updated now.

ameyer-pivotal: The previous wording was perhaps too much detail anyway, so I like the way it is now.

README.md
```diff
@@ -225,10 +225,111 @@ that they appear in `builder.toml`. This image's base will be the build image as
...
+  - **`id`** _(string, required)_
+    <br>
+    An identifier for the buildpack
+    
+  - **`uri`** _(string, required)_
+    <br>
+    Either a fully-qualified URL to a `.tgz` file, or a path to a local buildpack's `.tgz` file or directory (relative to
+    `builder.toml`)
+    
+  - **`latest`** _(boolean, optional, default: `false`)_
```
josegonzalez: Do we also need to specify the `version` here? Version is specified in `groups.buildpacks`, but it doesn't appear to me that there is a way to specify where to get a specific version.

ameyer-pivotal: @josegonzalez Funny you should mention that. We just had a discussion about that earlier this afternoon. The version is actually specified in the buildpack's own `builder.toml`.

That said, we agree that it can make writing a builder.toml less intuitive, so we [created an issue](https://github.com/buildpack/pack/issues/187) to add `version` to the top-level buildpacks (and enforce some validation).

PR #194, opened 2019-05-23
Add goimports to Makefile and use the Makefile in travis CI

ekcasey:


build/testdata/fake-lifecycle/phase.go
```diff
@@ -136,12 +136,12 @@ func testBuildpacks() {
 
 func testProxy() {
 	fmt.Println("proxy test")
-	fmt.Println("HTTP_PROXY="+os.Getenv("HTTP_PROXY"))
-	fmt.Println("HTTPS_PROXY="+os.Getenv("HTTPS_PROXY"))
-	fmt.Println("NO_PROXY="+os.Getenv("NO_PROXY"))
-	fmt.Println("http_proxy="+os.Getenv("http_proxy"))
-	fmt.Println("https_proxy="+os.Getenv("https_proxy"))
-	fmt.Println("no_proxy="+os.Getenv("no_proxy"))
+	fmt.Println("HTTP_PROXY=" + os.Getenv("HTTP_PROXY"))
```
ameyer-pivotal: Super minor, but I think the way it was before is actually the convention, right?

PR #208, opened 2019-06-07
Fix build not failing when CreateTarReader returns an error

neerfri:
I've been hunting a weird issue for a while where files were missing from my workspace.

Eventually I tracked it down to a `archive/tar: sockets not supported` error returned from `(tar.Writer* )WriteHeader(header)`. This is caused by the unix domain sockets that are often used by ruby servers.

That specific scenario is covered by the test added in the PR.
The code that fixes the problem is more broad than what is tested and also includes other types of files that will be filtered from the tar such as `fifo` named pipes, device files (for example if someone symlinked to `/dev/whatever`.

It would be a very good idea to implement some `.gitignore`-like mechanism, maybe in a `.packignore` file, but that's a story for another change. I'll work on that one in parallel cause it's very important to me (otherwise developer credentials may find themselves in the production build)

### DIFF
--- a/build/lifecycle_test.go
+++ b/build/lifecycle_test.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"math/rand"
+	"net"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


jromero:
+1 the idea of `.packignore`. Might be useful for a slew of other common scenarios such as sensitive files, caches, generated code, etc.

neerfri:
@jromero see https://github.com/buildpack/pack/issues/210

neerfri:
@ekcasey tests are passing now. I'd appreciate your review.

PR #211, opened 2019-06-10
Allow create-builder on Windows

### LINKED ISSUE #181: pack create-builder on Windows
### `pack create-builder`

**Given** My `builder.toml` only contains buildpack uris that points to buildpack archives
**When** I run `pack create-builder -b builder.toml` on windows
**Then** It succeeds and the file permission encoded in the archive are preserved (i.e. we don't lose executable info)

**Given** My `builder.toml` contains a buildpack uris that points to directory
**When** I run `pack create-builder -b builder.toml` on windows
**Then** It fails with an error reporting that buildpack directories are not supported on windows

### `pack build --buildpack`

Treat buildpack uris the same way `pack create-builder` does on Windows

jromero:
- Also allow tgz buildpacks on build command
- Still disallowing directory buildpacks on Windows

Resolves #181


buildpack/fetcher_test.go
```diff
@@ -49,9 +49,26 @@ func testBuildpackFetcher(t *testing.T, when spec.G, it spec.S) {
...
+				Return(downloadPath, nil)
+
+			out, err := subject.FetchBuildpack(downloadPath)
+			h.AssertNil(t, err)
+			h.AssertEq(t, out.ID, "bp.one")
+			h.AssertEq(t, out.Version, "some-buildpack-version")
+			h.AssertEq(t, out.Stacks[0].ID, "some.stack.id")
+			h.AssertEq(t, out.Stacks[1].ID, "other.stack.id")
+			h.AssertNotEq(t, out.Path, "")
+			h.AssertOnTarEntry(t, out.Path, "./bin/detect", h.ContentEquals("I come from a directory"))
```
ekcasey: The contents of this test string are a bit confusing in the tgz test case context.

commands/create_builder.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"net/url"
 	"path/filepath"
-	"runtime"
 
 	"github.com/BurntSushi/toml"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
+	"github.com/buildpack/pack/internal/paths"
+
```
ekcasey: group this with pack imports

create_builder_test.go
```diff
@@ -242,6 +239,24 @@ func testCreateBuilder(t *testing.T, when spec.G, it spec.S) {
 			assertTarHasFile(t, layerTar, "/lifecycle/cacher")
 			assertTarHasFile(t, layerTar, "/lifecycle/launcher")
 		})
+
+		if runtime.GOOS == "windows" {
```
ekcasey: same comments regarding if/else as above

downloader.go
```diff
@@ -9,10 +9,12 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/pkg/errors"
 
-	"github.com/buildpack/pack/internal/archive"
+	"github.com/buildpack/pack/internal/paths"
+
```
ekcasey: pack imports can be grouped together

downloader.go
```diff
@@ -21,61 +23,58 @@ type Downloader struct {
...
-	}
-}
-
-func (d *Downloader) handleFile(bpURL *url.URL) (string, error) {
-	path := bpURL.Path
-
-	if filepath.Ext(path) != ".tgz" {
-		return path, nil
-	}
+	hasScheme := schemeRegexp.MatchString(pathOrUri)
```
ekcasey: I'm curious why we broke the no scheme case out of the switch statement. Was there some edge case here I am not comprehending?

downloader_test.go
```diff
@@ -5,9 +5,10 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"runtime"
 	"testing"
 
+	"github.com/buildpack/pack/internal/paths"
+
```
ekcasey: same comment about the imports

internal/archive/archive_test.go
```diff
@@ -44,59 +43,99 @@ func testArchive(t *testing.T, when spec.G, it spec.S) {
 		}
 	})
 
-	when("force exec is true", func() {
-		it("writes a tar to the dest dir with 0777", func() {
-			fh, err := os.Create(filepath.Join(tmpDir, "some.tar"))
+	when.Focus("#ReadTarEntry", func() {
```
ekcasey: Accidental Focus here?

acceptance/acceptance_test.go
```diff
@@ -747,7 +769,7 @@ func createBuilder(t *testing.T, runImageMirror string) string {
 			t.Fatal("LIFECYCLE_PATH must be an absolute path")
 		}
 		t.Logf("Adding lifecycle path '%s' to builder config", lifecyclePath)
-		_, err = builderConfigFile.Write([]byte(fmt.Sprintf("uri = \"%s\"\n", lifecyclePath)))
+		_, err = builderConfigFile.Write([]byte(fmt.Sprintf("uri = \"%s\"\n", strings.ReplaceAll(lifecyclePath, `\`, `\\`))))
```
ekcasey: what is the purpose of this? I think I'm missing something

builder/builder.go
```diff
@@ -464,6 +481,125 @@ func (b *Builder) buildpackLayer(dest string, bp buildpack.Buildpack) (string, e
...
+	return nil
+}
+
+func (b *Builder) embedLifecycleTar(tw *tar.Writer, srcTar string) error {
+	var (
+		tarFile    *os.File
+		gzipReader *gzip.Reader
+		fhFinal    io.Reader
+		err        error
+		regex      = regexp.MustCompile(`^[^/]+/([^/]+)$`)
```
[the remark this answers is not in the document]
ekcasey: Nevermind, I see that this is being checked in the lifecycle fetcher

builder/builder.go
```diff
@@ -422,37 +432,44 @@ func (b *Builder) buildpackLayer(dest string, bp buildpack.Buildpack) (string, e
...
 
-	if err := archive.WriteDirToTar(
-		tw,
-		bp.Dir,
-		fmt.Sprintf("%s/%s/%s", buildpacksDir, bp.EscapedID(), bp.Version),
-		b.UID,
-		b.GID,
-		false,
-	); err != nil {
+	if filepath.Ext(bp.Path) == ".tgz" {
```
ekcasey: I wonder if we can have the buildpack object provide a TarReader in both the `.tgz` and dir case so we don't have to worry about dirs vs archives in so many separate places

ekcasey:
I think in tests that need a `.tgz` test fixture file we should tar up a dir before the test and clean up after the test. If we commit `.tgz` test fixtures, changing them will be a pain, as is exploring to understand the tests.

acceptance/acceptance_test.go
jromero: This is necessary for TOML with Windows paths.

build_test.go
```diff
@@ -499,36 +499,50 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+						h.AssertEq(t, bldr.GetBuildpacks(), []builder.BuildpackMetadata{
+							{ID: "buildpack.id", Version: "buildpack.version", Latest: true},
+							{ID: "some-buildpack-id", Version: "some-buildpack-version"},
+						})
+
 					})
-				})
+				}
 
 				// TODO: support other uris
```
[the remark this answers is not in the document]
jromero: NTS: Need to add successful `.tgz` tests

builder/builder.go
jromero: Deferred

builder/builder.go
```diff
@@ -514,33 +650,24 @@ func (b *Builder) lifecycleLayer(dest string) (string, error) {
...
-	}); err != nil {
-		return err
-	}
+func unixClean(path string) string {
+	cleaned := filepath.Clean(path)
 
-	if _, err := tw.Write([]byte(buf)); err != nil {
-		return err
+	if runtime.GOOS == "windows" {
+		return strings.ReplaceAll(cleaned, `\`, "/")
```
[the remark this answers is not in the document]
jromero: NTS: Investigate

builder/builder_test.go
```diff
@@ -257,72 +258,90 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
-						layerTar,
-						"/buildpacks/other-buildpack-id/latest",
-						h.SymlinksTo("/buildpacks/other-buildpack-id/other-buildpack-version"),
-					)
-					h.AssertOnTarEntry(t,
-						layerTar,
-						"/buildpacks/other-buildpack-id/latest",
-						h.HasOwnerAndGroup(0, 0),
+					h.AssertOnTarEntry(t, layerTar, "/buildpacks/tgz-buildpack-id/tgz-buildpack-version",
+						h.IsDirectory(),
```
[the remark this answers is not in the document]
jromero: NTS: Add to others

downloader.go
jromero: This is because when we rely on `url.Parse(pathOrUri)` the result for `c:\user\test\` _incorrectly_ picks up `c` as the scheme later falling into the `unsupported protocol`. This new check is more for determining if it's a uri or file path so we can be certain `url.Parse` will yield more accurate results. 

Would it help to rename the boolean to `isUri`?

Issue #201, opened 2019-05-29
Suggestion: use "Squash and merge" instead of a merge for pull request


neerfri:
Hi CNB members!

Great appreciation for the project. Thank you for contributing your time to improve everyones workflow.

I'd like to suggest an admin of this repository will [disable the normal "merge button" in GitHub](https://help.github.com/en/articles/configuring-commit-squashing-for-pull-requests) so that pull requests that are merged in this repository are merged as a single change and are shown on the git history of the master in the right point in time.
This makes it much easier for your users to understand the changes that happened between versions.

For example see the [current changes from `v0.2.0` to `master`](https://github.com/buildpack/pack/compare/v0.2.0...master):
![image](https://user-images.githubusercontent.com/21379/58579304-95b7b900-81ff-11e9-9f3b-b46b04d18a2f.png)

The counter argument has always been that defaulting to squash merge "deletes" the journey that the developers working on the changes have gone. In the case of Github that is not really correct since the original PR and it's associated branch are always available with all their commits, comments and checks. Frankly it's more common that the commits are not a good representation of the parts that compose the pull request, with lots of "fix typo" and revert commits.

As a user of many OSS projects I can say that those that switched to squash merges are much easier for me to navigate and contribute to.

sclevine:
Thanks for the suggestion @neerfri! I'm glad that you appreciate the project.

Up to this point, our philosophy has been "don't re-write other people's work."

To me, avoiding rebasing/squashing on merge is less about preserving the user journey, and more about preserving the sanctity of original author's contributions. Consider that contributors may PGP sign their commits and/or expect that the authorship of each commit be preserved as-is. It's difficult to know what a PR submitter is comfortable with, and it isn't practical to ask each submitter. 

Additionally, the CNCF requires that each commit into the project contain a DCO sign-off from whomever authored the code. I assume that these are supposed to be preserved after the merge (but I'm not certain).

neerfri:
@sclevine Thanks for your response.

As you can see below, kubernetes, that has similar requirements and similar mindset is using a form of squash-merge for their master:
![image](https://user-images.githubusercontent.com/21379/58605870-45b41300-824e-11e9-84ce-4ef0ad1366db.png)


Just to make sure I am understood. We are not expecting any developer to rebase or squash their commits. We are talking about an automatic button in GitHub that does something very similar to `git merge --squash` (not the same as squash via `git rebase -i`)
[...]

PR #218, opened 2019-06-20
ignore unix domain socket files in archive.WriteDirToTar

neerfri:
At the moment `archive.WriteDirToTar` will fail if there are any unix domain socket files in the directory tree being archived (the app).
This change skips those files assuming they should not be copied, [since they can't be](https://github.com/golang/go/blob/release-branch.go1.12/src/archive/tar/common.go#L658)

### DIFF
--- a/build/lifecycle_test.go
+++ b/build/lifecycle_test.go
@@ -6,14 +6,12 @@ import (
 	"io"
 	"io/ioutil"
 	"math/rand"
-	"net"
 	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
 	"time"
 
-	"github.com/buildpack/imgutil"
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
@@ -21,10 +19,12 @@ import (
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
+	"github.com/buildpack/imgutil"
 	"github.com/buildpack/pack/build"
 	"github.com/buildpack/pack/builder"
 	"github.com/buildpack/pack/internal/archive"
 	"github.com/buildpack/pack/internal/mocks"
+	"github.com/buildpack/pack/logging"
 	h "github.com/buildpack/pack/testhelpers"
 )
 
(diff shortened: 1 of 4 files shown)


jromero:
@neerfri seems like a previous test added to make sure tarring fails for non-standard files is failing now that we are simply ignoring the files. Can you please make the updates to the tests so that they may pass?

@ekcasey what is your thought on this PR proposal? The basic idea is to silently skip over files that can't be tarred. We could keep this strategy at least until the [RFC with ignored files](https://github.com/buildpack/rfcs/pull/10) is implemented.

sclevine:
@neerfri I believe this change breaks symlinks, which Go's `os` package does not consider to be regular files.

In fact, I don't suspect this bug extends to special file types aside from sockets. See: https://golang.org/src/archive/tar/common.go?s=22088:22153#L619

A number of special file types are supported by `archive/tar` (as well as gnu/bsd tar). I think the ideal behavior here is to only read the contents of regular files (as is currently implemented), and skip sockets (which definitely needs to be implemented).

neerfri:
@jromero @sclevine You are both right.
The PR is breaking the test I added in #208  because it's fixing the source of the error. I worked on new code yesterday that would produce the error again. What I came up with is creating a directory without list permissions. which makes the `Walk` process return an error and the test code validates that error.

@sclevine You are also right about this code breaking the symlink behavior, I found that while I was trying to write a new test, I'll fix that too. I really like the idea to limit less files. I'll see how to integrate it into the code change

I will try to finish that change and submit it here today

Thanks

neerfri:
@jromero @sclevine all fixed

jromero:
Looking good. Much more simplistic change.

I've made some proposed improvements in this patch: [pr-218.patch.txt](https://github.com/buildpack/pack/files/3325893/pr-218.patch.txt)

The changes include:
1. Create restricted directory in a temp directory to keep testdata/fixtures untouched while testing.
2. Change test restricted directory perms to `0222` to allow for system deletion.
3. Add relevant test for `WriteDirToTar` silently ignoring sockets.

neerfri:
@jromero Thanks! committed the patch.
If this branch is merged you can squash merge it, the individual commits are not important [:-)](https://github.com/buildpack/pack/issues/201)

jromero:
@neerfri Since we haven't fully agreed on squash merging (#201) I can't take it upon myself do it.

If you'd like to rebase (aka cleanup the history) to build a better story for this change please feel free to do so. Otherwise, we'll merge it as is within the next 24 hours.

Sorry for the inconvenience.

sclevine:
@jromero If @neerfri agrees to us squash-merging this PR, I think that's totally fine. My reasoning in #201 is that we shouldn't clobber commits without the consent of the original author.

PR #219, opened 2019-06-20
Handle context cancelling when executing container

dmathieu:
When cancelling a running build (for example with Ctrl-C in the CLI), we should be able to cancel any running container right away.
With this change, we will properly respect when the context is cancelled by killing the currently running container, and therefore stopping the build.

### DIFF
--- a/build/testdata/fake-lifecycle/vendor/github.com/docker/docker/api/types/container/host_config.go
+++ b/build/testdata/fake-lifecycle/vendor/github.com/docker/docker/api/types/container/host_config.go
@@ -7,7 +7,7 @@ import (
 	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/docker/api/types/strslice"
 	"github.com/docker/go-connections/nat"
-	"github.com/docker/go-units"
+	units "github.com/docker/go-units"
 )
 
 // Isolation represents the isolation technology of a container. The supported
--- a/container/run.go
+++ b/container/run.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Run starts a docker container, streams its logs and waits for it to have finished running
 func Run(ctx context.Context, docker *client.Client, ctrID string, out, errOut io.Writer) error {
 	bodyChan, errChan := docker.ContainerWait(ctx, ctrID, dcontainer.WaitConditionNextExit)
 
@@ -39,7 +40,17 @@ func Run(ctx context.Context, docker *client.Client, ctrID string, out, errOut i
 			return fmt.Errorf("failed with status code: %d", body.StatusCode)
 		}
 	case err := <-errChan:
+		if err == context.Canceled {
+			return docker.ContainerStop(context.Background(), ctrID, nil)
+		}
+
 		return err
 	}
-	return <-copyErr
+
+	err = <-copyErr
+	if err == context.Canceled {
+		return docker.ContainerStop(context.Background(), ctrID, nil)
+	}
+
+	return err
 }
--- a/container/run_test.go
+++ b/container/run_test.go
@@ -1,0 +1,91 @@
+package container_test
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	dockertypes "github.com/docker/docker/api/types"
+	dcontainer "github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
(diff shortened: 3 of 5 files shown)


dmathieu:
I'm not sure what the tests failure is about. But master appears to be failing too.

jromero:
@dmathieu It appears that there is a formatting issue in this PR regarding the order of imports. It's not very obvious so we'll add a ticket to improve the output. 

For now, you can run `make format imports` and update the PR with the changes it creates. That should get it through CI.

dmathieu:
Thank you for the formatting tip. It looks like that solved it.

The linux tests are still failing, but that still seems unrelated to this change.

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x80c801]
goroutine 9 [running]:
testing.tRunner.func1(0xc0000cb700)
	/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/testing/testing.go:830 +0x392
panic(0x88af40, 0xcd87a0)
	/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/runtime/panic.go:522 +0x1b5
github.com/buildpack/pack/app_test.assertOnRunningContainer(0xc0000cb700, 0xc0001f54e0, 0x0, 0x0, 0x0, 0xc0003121b0, 0xc000318000, 0xc000135dc0)
	/home/travis/gopath/src/github.com/buildpack/pack/app/run_test.go:212 +0x391
github.com/buildpack/pack/app_test.testApp.func1.2.2()
	/home/travis/gopath/src/github.com/buildpack/pack/app/run_test.go:66 +0x7c
github.com/sclevine/spec.run(0xc0000cb700, 0xc000135e58, 0x1, 0x1)
	/home/travis/gopath/src/github.com/buildpack/pack/vendor/github.com/sclevine/spec/spec.go:269 +0x44
github.com/sclevine/spec.Run.func3(0xc0000cb700, 0xc000256f00, 0x3, 0x4, 0xc000260d00, 0x3, 0x4, 0x5d0d0326, 0x1, 0x1, ...)
	/home/travis/gopath/src/github.com/buildpack/pack/vendor/github.com/sclevine/spec/spec.go:262 +0x375
github.com/sclevine/spec.node.run.func1(0xc0000cb700)
	/home/travis/gopath/src/github.com/buildpack/pack/vendor/github.com/sclevine/spec/parser.go:133 +0x79
testing.tRunner(0xc0000cb700, 0xc0001f53e0)
	/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/testing/testing.go:916 +0x35a
FAIL	github.com/buildpack/pack/app	12.442s
```

jromero:
The panic is related to what appears to be a `nil` `error` that we are trying to extract the `err.Error()` from.

This should fix it:
```patch
diff --git a/app/run_test.go b/app/run_test.go
index d4c8a78..98a7e1f 100644
--- a/app/run_test.go
+++ b/app/run_test.go
@@ -209,7 +209,7 @@ loop:
                }
        }
        cancel()
-       if err := <-done; !strings.Contains(err.Error(), context.Canceled.Error()) {
+       if err := <-done; err != nil && !strings.Contains(err.Error(), context.Canceled.Error()) {
                t.Fatalf("expected canceled context, failed with a different error: %s", err)
        }
 }
```

Also, in investigating that issue I did discover that the other failing test `TestContainerRun/ContainerRun/#Run/can_cancel_a_running_container` is probably failing due to a race condition based on the exact timing necessary by that test. Furthermore, when the changes to `container/run.go` are reverted they tend to have no effect on said test. Are we sure that test is working as expected?

dmathieu:
Thank you for the feedback @jromero. I do see a difference in the test between before and after the change.
Without this change, when cancelling the request, the container would keep running (the HTTP request to wait for it is canceled, but the container itself isn't killed). As such, it returns a "context cancelled" error. After this change, it doesn't return an error, and kills the container.

ekcasey:
@dmathieu Can you describe the behavior you were seeing that motivated this PR and #225 ? When I attempted to reproduce the underlying issue with pack v0.2.1. the running containers appeared to be properly cancelled.

dmathieu:
@ekcasey you're right, my bad. Somehow, my previous builds weren't cancellable but they are now.

PR #221, opened 2019-06-21
Skip layer analysis when clearing cache

jromero:
Skip layer analysis for analyzer when clearing cache
- For older lifecycles, continue to skip analyze altogether

Relates to buildpack/lifecycle#138

### DIFF
--- a/build/lifecycle.go
+++ b/build/lifecycle.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/buildpack/pack/builder"
 	"github.com/buildpack/pack/cache"
+	"github.com/buildpack/pack/lifecycle"
 	"github.com/buildpack/pack/logging"
 	"github.com/buildpack/pack/style"
 )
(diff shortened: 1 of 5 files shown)


ekcasey:
@jromero I believe that if we are clearing the cache, we don't want to reuse layers from the analyzed image. Therefore it would be appropriate to skip analyze. I would assume that the absence of an `analyzed.toml` would simply result in the `exporter` never reusing layers (as it would on a first build).

build/lifecycle.go
```diff
@@ -76,7 +76,8 @@ func (l *Lifecycle) Execute(ctx context.Context, opts LifecycleOptions) error {
 		l.logger.Debugf("Build cache %s cleared", style.Symbol(buildCache.Name()))
 	}
 
-	if lifecycleVersion := l.builder.GetLifecycleVersion(); lifecycleVersion == nil {
+	lifecycleVersion := l.builder.GetLifecycleVersion()
+	if lifecycleVersion == nil {
 		l.logger.Debug("Warning: lifecycle version unknown")
```
ekcasey: Now that we want to treat an unknown lifecycle version like the default lifecycle, we should probably set lifecycleVersion to the default in this case so that it is easy to be consistent.

build/lifecycle.go
```diff
@@ -94,11 +95,11 @@ func (l *Lifecycle) Execute(ctx context.Context, opts LifecycleOptions) error {
 		return err
 	}
 
-	l.logger.Debug(style.Step("ANALYZING"))
-	if opts.ClearCache {
+	if opts.ClearCache && lifecycleVersion != nil && lifecycleVersion.LessThan(semver.MustParse("0.3.0")) {
 		l.logger.Debug("Skipping 'analyze' due to clearing cache")
 	} else {
-		if err := l.Analyze(ctx, opts.Image.Name(), opts.Publish); err != nil {
+		l.logger.Debug(style.Step("ANALYZING"))
```
ekcasey: We should go back to printing this even if we skip analyze

ekcasey:
@jromero and I are going to merge this w/o waiting for Travis b/c of degraded Travis availability.

PR #224, opened 2019-06-26
Improve contribution process

jromero:
- Add CONTRIBUTING.md with setup and process instructions
- Distinguish between verification of format and actual formatting
- Use Makefile for osx on travis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,42 +1,53 @@
 # Go parameters
-GOCMD=go
+GOCMD?=go
 GOENV=CGO_ENABLED=0
 PACK_VERSION?=dev
 PACK_BIN?=pack
+PACKAGE_BASE=github.com/buildpack/pack
+PACKAGES:=$(shell $(GOCMD) list ./... | grep -v /testdata/)
+SRC:=$(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
-all: clean test build
+all: clean verify test build
 
-build: format imports vet
+build: format vet
+	@echo "> Building..."
 	mkdir -p ./out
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,44 +1,53 @@
...
+GOCMD?=go
 GOENV=CGO_ENABLED=0
 PACK_VERSION?=dev
 PACK_BIN?=pack
+PACKAGE_BASE=github.com/buildpack/pack
+PACKAGES:=$(shell $(GOCMD) list ./... | grep -v /testdata/)
+SRC:=$(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
-all: clean test build
+all: clean verify test build
```
jromero: These changes make it so it's more obvious what the problem is when CI runs.

Example output:

```bash
$ make verify
> Installing goimports...
go install -mod=vendor golang.org/x/tools/cmd/goimports
> Verifying format...
ERROR: Format verification failed!

diff -u ./lifecycle/fetcher.go.orig ./lifecycle/fetcher.go
--- ./lifecycle/fetcher.go.orig 2019-06-25 22:52:03.000000000 -0500
+++ ./lifecycle/fetcher.go      2019-06-25 22:52:03.000000000 -0500
@@ -63,9 +63,9 @@
        return Metadata{Version: version, Path: path}, nil
 }
 
-func validateTarEntries (tarPath string, entryPath ...string) error {
+func validateTarEntries(tarPath string, entryPath ...string) error {
        var (
-               tarFile     *os.File
+               tarFile    *os.File
                gzipReader *gzip.Reader
                fhFinal    io.Reader
                err        error
make: *** [verify-format] Error 1

```

PR #225, opened 2019-06-26
Stop the lifecycle if the context moved to done

dmathieu:
This goes in the same vein as #219, to allow cancelling a running build.
A context can move to done at any of those step, meaning we shouldn't execute the following ones.

`Execute()` has no tests at all, and it would be quite difficult to write a tests which cancel the context at a specific step. I'm happy to look into writing tests if a suggestion is provided though.


ekcasey:
@dmathieu I am confused as to which problem this PR is solving. I just tried to Ctrl-C a running pack build with the latest release of `pack` (v0.2.1) and I observed that the container was stopped and the lifecycle did not continue running (as expected).

PR #227, opened 2019-06-28
Expose builder config loading method

### LINKED ISSUE #214: Expose functions to load builder config files
The logic necessary to load config files is locked up with the CLI commands and is not usable by the API client. For example, to create a builder it is necessary to have a BuilderConfig which for the CLI is defined in a builder.toml file. There is processing the CLI does on the toml file after it is parsed in order to make the file usable. This [logic is locked in unexpected methods](https://github.com/buildpack/pack/blob/b914032c63f6927813f2accb523cf584ea25a556/commands/create_builder.go#L62-L103) exposed only to the CLI commands.

I'd like to be able to use the API to create a builder from a builder.toml without needing to install the CLI (see https://github.com/projectriff/builder/issues/46)

ameyer-pivotal:
Resolves #214

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -1,5 +1,17 @@
 package builder
 
+import (
+	"io"
+	"net/url"
+	"os"
+	"path/filepath"
+
+	"github.com/BurntSushi/toml"
+	"github.com/pkg/errors"
+
+	"github.com/buildpack/pack/internal/paths"
+)
+
 type Config struct {
 	Description string            `toml:"description"`
 	Buildpacks  []BuildpackConfig `toml:"buildpacks"`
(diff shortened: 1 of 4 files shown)


files.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/BurntSushi/toml"
+	"github.com/pkg/errors"
+
+	"github.com/buildpack/pack/builder"
+	"github.com/buildpack/pack/internal/paths"
+)
+
+// ReadBuilderConfigFile reads a builder configuration from the file path provided
+func ReadBuilderConfigFile(path string) (builder.Config, error) {
```
jromero: Minor: Feels like now that these are external there should be some tests associated with them for certain guarantees. Given that this is mostly simply reorganization we could probably live without them.

ameyer-pivotal: That makes sense. Since it's still tested indirectly via other tests, I'm OK deferring.

files.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return uri, nil
+}
+
+// ReadEnvFile reads environment variables from the file path provided. The file should have one variable per line,
+// where each variable is of the form 'VAR=VALUE' or 'VAR'. When using the latter value-less form, value will be taken
+// from the current environment. Additional variables of the same form may be provided (and will have precedence over
+// those specified in the file).
+func ReadEnvFile(path string, additionalVars ...string) (map[string]string, error) {
```
ameyer-pivotal: I know I said I wasn't going to add the additional vars to this method, but I ran into issues due to splitting the associated methods between the `commands` and `pack` package, so I move them all over. It seemed like the simplest way to avoid duplicating the `addEnvVar` functionality.

files.go
ekcasey: I agree with @jromero  that we should add unit tests if we are exposing this as part of the library API.

files.go
ekcasey: I still feel like this interface is confusing with the additionalVars are a parameter. I feel like something called ReadEnvFile should just concern itself with the file. Why can't the command package have a private helper function that parses the env flags and adds the vars to the map?

ameyer-pivotal:
@ekcasey A private helper is not a problem. It would just have to be duplicated. `pack` can't import the `commands` package, and `commands` cant use it if it lives privately in `pack`. I could create a third package to decouple, but feels a little tedious.

files.go
ameyer-pivotal: @ekcasey A private helper is not a problem. It would just have to be duplicated. pack can't import the commands package, and commands cant use it if it lives privately in pack. I could create a third package to decouple, but feels a little tedious.

files.go
ameyer-pivotal: @ekcasey A private helper is not a problem. It would just have to be duplicated. pack can't import the commands package, and commands cant use it if it lives privately in pack. I could create a third package to decouple, but feels a little tedious.

files.go
ameyer-pivotal: @ekcasey OK, I'll write some tests to get us started. They probably won't be as exhaustive as the cases we go through for create builder, but we can slowly migrate them over perhaps.

ameyer-pivotal: Move to internal package and added tests

files.go
ameyer-pivotal: Moved to internal package and added tests

commands/commands.go
```diff
@@ -9,9 +9,11 @@ import (
 	"syscall"
 	"text/tabwriter"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/buildpack/pack"
+	env "github.com/buildpack/pack/internal"
```
ekcasey: This import alias is a little awkward. I think you meant to make an env package in internal?

builder/config.go
```diff
@@ -26,3 +38,65 @@ type LifecycleConfig struct {
...
+	config, err := parseConfig(file, builderDir)
+	if err != nil {
+		return Config{}, errors.Wrapf(err, "parse contents of '%s'", path)
+	}
+
+	return config, nil
+}
+
+// parseConfig reads a builder configuration from reader and resolves relative buildpack paths using `relativeToDir`
+func parseConfig(reader io.Reader, relativeToDir string) (Config, error) {
```
ameyer-pivotal: We decided not to export this for now because it felt a little awkward. Can export it later if needed.

PR #231, opened 2019-07-05
Store download cache in versioned directories

### LINKED ISSUE #229: fix backwards incompatibility in download cache structure
When using the latest `pack` built from from master, errors like the following occur when `pack` attempts to use cached assets if they were cached by the previous release.

```
$ pack create-builder -b builder.toml my/builder
....
Using cached version of "https://github.com/buildpack/lifecycle/releases/download/v0.2.1/lifecycle-v0.2.1+linux.x86-64.tgz"
ERROR  fetching lifecycle: invalid lifecycle: failed to open tar '<home>/.pack/download-cache/79999e36a09222a4052d6eb5678f003c17916330771a3a2d8c4cfed209fa090e.tgz' for validation: open '<home>/.pack/download-cache/79999e36a09222a4052d6eb5678f003c17916330771a3a2d8c4cfed209fa090e.tgz: no such file or directory
```

This problem can be fixed by clearing the download cache. We should fix `pack` such that either the old cache can be used or, more likely, `pack` seamlessly removes entries with an invalid format and redownloads those assets as necessary.

jromero:
Resolves #229 

### DIFF
--- a/downloader_test.go
+++ b/downloader_test.go
@@ -23,144 +23,156 @@ func TestDownloader(t *testing.T) {
 func testDownloader(t *testing.T, when spec.G, it spec.S) {
 	when("#Download", func() {
 		var (
-			err      error
-			tmpDir   string
-			tgz      string
-			cacheDir string
-			subject  *Downloader
+			cacheDir   string
+			err        error
+			originalWd string
+			subject    *Downloader
+			tgz        string
+			tmpDir     string
 		)
 
 		it.Before(func() {
-			tmpDir, err = ioutil.TempDir("", "")
+			tmpDir, err = ioutil.TempDir("", "test-downloader")
 			h.AssertNil(t, err)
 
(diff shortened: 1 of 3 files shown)


downloader_test.go
```diff
@@ -23,144 +23,156 @@ func TestDownloader(t *testing.T) {
...
+				h.AssertOnTarEntry(t, out, "file.txt", h.ContentEquals("some file contents"))
 			})
-			defer server.Close()
 
-			out, err := subject.Download(server.URL() + "/downloader/somefile.tgz")
-			h.AssertNil(t, err)
-			h.AssertMatch(t, out, `\.tgz$`)
-			h.AssertOnTarEntry(t, out, "file.txt", h.ContentEquals("some file contents"))
-		})
+			it("caches to versioned directory", func() {
```
jromero: This is the only added test. All other changes here are just clean up/organizational.

PR #232, opened 2019-07-05
Improve client constructors

ekcasey:
* Improves client constructors
* Prevents pack client from writing a config file when it is being consumed as a library
* Make `config.Config` a CLI only concepts and removes config field from `pack.Client`
* Refactors `config.Config` into a value object and provides package level functions for writing & reading config files

Outstanding Issues:
Download cache path should be configurable (is is currently always `$HOME/.pack/download-cache`). This PR does not fix that issue.


build_test.go
```diff
@@ -1,4 +1,4 @@
-package pack_test
+package pack
```
jromero: Was it intentional to make this "whiteboxed"? It doesn't seem like any methods are not exported.

commands/build.go
```diff
@@ -26,7 +26,7 @@ type BuildFlags struct {
 	Buildpacks []string
 }
 
-func Build(logger logging.Logger, config *config.Config, packClient *pack.Client) *cobra.Command {
+func Build(logger logging.Logger, cfg config.Config, packClient *pack.Client) *cobra.Command {
```
jromero: It doesn't seem like `AdditionalMirrors` is being passed in to `Build`. How do local run image mirrors get set?

jromero: Did our acceptance tests not catch that or are we missing something?

rebase.go
```diff
@@ -4,45 +4,58 @@ import (
...
 type RebaseOptions struct {
-	RepoName string
-	Publish  bool
-	SkipPull bool
-	RunImage string
+	RepoName          string
+	Publish           bool
+	SkipPull          bool
+	RunImage          string
+	AdditionalMirrors map[string][]string
```
jromero: ~Can we confirm that these are being passed in from the command?~

It seems like these aren't being passed from the `Rebase` command.

jromero: We probably need to create or modify a test so that this doesn't happen frequently. 

We were discussion/thinking a unit test of cmd arguments to options would be sufficient and quick. There is also an argument to just modifying an acceptance test to do it.

commands/set_run_image_mirrors_test.go
```diff
@@ -0,0 +1,55 @@
+package commands
+
+//when("Config#SetRunImageMirrors", func() {
```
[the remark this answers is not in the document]
ekcasey: oops, clearly I threw this up without reading through the changes carefully (apologies for rushing)

rebase.go
ekcasey: I modified a build and rebase acceptance test so we can have confidence we are wiring the mirrors all the way through.

commands/run.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/buildpack/pack/logging"
 )
 
-func Run(logger logging.Logger, config *config.Config, packClient *pack.Client) *cobra.Command {
+func Run(logger logging.Logger, cfg config.Config, packClient *pack.Client) *cobra.Command {
```
jromero: Looks like we're missing passing the additional mirrors here too (and tests, likely) -- says @ameyer-pivotal

jromero: Looks like we're missing passing the additional mirrors here too (and tests, likely) (@ameyer-pivotal)

ekcasey: Given that `pack run` can't be executed against a registry and given that users can't supply the image name to `pack run` selecting a mirror based on the registry in our autogenerated name (always implicitly dockerhub) feels a little awkward

ameyer-pivotal: Agreed

PR #233, opened 2019-07-08
Update the Heroku buildpack image tag and description

jkutner:
/cc @hone 

Signed-off-by: Joe Kutner <jpkutner@gmail.com>


commands/commands.go
```diff
@@ -36,7 +36,7 @@ var suggestedBuilders = [][]suggestedBuilder{
 		{"Cloud Foundry", "cloudfoundry/cnb:cflinuxfs3", "larger base image with Java, Node.js & Python"},
 	},
 	{
-		{"Heroku", "heroku/buildpacks", "heroku-18 base image with official Heroku buildpacks"},
+		{"Heroku", "heroku/buildpacks:18", "heroku-18 base image with buildpacks for Ruby, Java, & more"},
```
hone: should we just list out all 6 languages?

ekcasey: I you add a description to your builder config this hardcoded message will be irrelevant. I think we are hoping remove these hardcoded strings eventually. https://buildpacks.io/docs/using-pack/working-with-builders/#builder-configuration

ekcasey:
I am noticing that the `heroku/buildpacks` and `heroku/buildpacks:18` builders have an "Unknown" lifecycle version according to `pack inspect-builder`

```
$pack inspect-builder heroku/buildpacks:18
Inspecting builder: heroku/buildpacks:18

Remote
------

Stack: heroku-18

Lifecycle Version: Unknown
...
```

Is this actually a custom lifecycle version or was it simply built with an old version of `pack`? `pack` can work with an unknown version but having the lifecycle version metadata can be useful to users and it helps `pack` know which lifecycle features it can use.

ameyer-pivotal:
@jkutner I think (going along with @ekcasey's comment) that you can avoid this PR altogether and just recreate the builder with a `description` (and also a `lifecycle.version`) :)

jkutner:
I've updated the `heroku/buildpacks` image with the description (and lifecycle.version), but I think we still need to merge this to make sure the `:18` tag gets in.

PR #234, opened 2019-07-09
Add support for zip-formatted files for the app path

### LINKED ISSUE #106: Support a zip file of an app dir with `pack build`
**Acceptance Criteria**

_Zip file_

**Given** my application is located at the root of a zip-formatted file
**When** I run `pack build -p <path to zip> my-app`
**Then** The build succeeds

_Non-zip file_

**When** I run `pack build -p <path to file> my-app`
**Then** The build fails with a message:
```
ERROR: the provided application path must be a directory or zip-formatted file
``` 

**Notes**
- Should work regardless of file extension (check first four bytes for `PK\x03\x04`)
- Should also work for `pack run`

jromero:
Resolves #106


build.go
```diff
@@ -181,33 +182,49 @@ func (c *Client) validateImageReference(imageName string) (name.Reference, error
...
+	}
+
+	if !fi.IsDir() {
+		fh, err := os.Open(resolvedAppPath)
+		if err != nil {
+			return "", errors.Wrap(err, "read file")
+		}
+		defer fh.Close()
+
+		isZip, err := archive.IsZip(fh)
```
jromero: @sclevine Decided to check early for a better user experience. We had talked offline about just letting the reader fail, but this will notify the user earlier (before pulling images, trying to start a build, etc).

internal/archive/archive.go
```diff
@@ -205,3 +204,95 @@ func WriteDirToTar(tw *tar.Writer, srcDir, tarDir string, uid, gid int, mode int
 		return nil
 	})
 }
+
+func WriteZipToTar(tw *tar.Writer, srcZip, basePath string, uid, gid int, mode int64) error {
```
jromero: A lot of this code is duplicated from `WriteDirToTar` and can be merged with a little bit more work. For the sake of trying to get this feature in, we are deferring it for now.

build/phase.go
```diff
@@ -118,20 +120,17 @@ func (p *Phase) Run(context context.Context) error {
 		return errors.Wrapf(err, "failed to create '%s' container", p.name)
 	}
 	p.appOnce.Do(func() {
-		var mode int64 = -1
-		if runtime.GOOS == "windows" {
-			mode = 0777
+		var appReader io.ReadCloser
+		appReader, err = p.createAppReader()
+		if err != nil {
+			err = errors.Wrapf(err, "create tar archive from '%s'", p.appPath)
```
ekcasey: We should return from this function if creating the tar reader fails instead of continuing to the copy

build/phase.go
```diff
@@ -148,3 +147,21 @@ func (p *Phase) Run(context context.Context) error {
 func (p *Phase) Cleanup() error {
 	return p.docker.ContainerRemove(context.Background(), p.ctr.ID, types.ContainerRemoveOptions{Force: true})
 }
+
+func (p *Phase) createAppReader() (io.ReadCloser, error) {
```
ekcasey: Similar to my comments on the original buildpack tgz PR, it would be nice if the different formats of input parameters were abstracted away from lower level components. Given the `client.Build` handles validating inputs, perhaps it can provide the Lifecycle with an App struct of some sort that the Phase can just read from (so we don't need to think about file types in many different places). I am okay merging this w/o making that change now but I would like to drive toward that design in the future.

internal/archive/archive.go
```diff
@@ -21,19 +21,28 @@ func init() {
...
+		defer func() {
+			w.CloseWithError(err)
+		}()
 
 		tw := tar.NewWriter(w)
 		defer tw.Close()
 
-		err := WriteDirToTar(tw, srcDir, tarDir, uid, gid, mode)
-		errChan <- err
+		err = writeFn(tw, src, basePath, uid, gid, mode)
```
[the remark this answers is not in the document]
sclevine: I would keep this as-is. The tar writer should be closed first, and `writeFn` could theoretically panic.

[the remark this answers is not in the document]
jromero: `defer` also handles panics right? So deferring would be the safest way to ensure it's _always_ closed without thinking too much about it. Is there a drawback to using `defer` that you can see?

.travis.yml
```diff
@@ -27,4 +27,8 @@ jobs:
         - NO_DOCKER=true
         - GO111MODULE=on
       script: go test -mod=vendor -count=1 -parallel=1 -v ./...
-      after_success: go build -mod=vendor -o pack ./cmd/pack
\ No newline at end of file
+      after_success: go build -mod=vendor -o pack ./cmd/pack
+
+branches:
+  only:
+    - master
```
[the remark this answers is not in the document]
jromero: Correct. Before this change, we were getting double the build jobs for any PR. We discussed our typical workflow and decided that there is no value in building individual branches aside from `master`.

Here are some quick references we reviewed:

* https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests
* https://stackoverflow.com/a/31882307

acceptance/acceptance_test.go
```diff
@@ -149,87 +149,95 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
...
-					t.Log("exporter reuses unchanged layers")
+			for pathType, appPath := range map[string]string{
+				"zip": filepath.Join("testdata", "mock_app.zip"),
+				"dir": filepath.Join("testdata", "mock_app"),
+			} {
+
+				pathType := pathType
+				appPath := appPath
+
+				it("creates image on the daemon from app "+pathType, func() {
```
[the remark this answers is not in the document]
ameyer-pivotal: Hmm... the code paths are fairly different when copying the app into the container, so I'd say that it working for zip doesn't necessarily imply it works for a dir, nor vice versa.

build/phase.go
```diff
@@ -118,20 +120,17 @@ func (p *Phase) Run(context context.Context) error {
...
-		appReader, errChan := archive.CreateTarReader(p.appDir, appDir, p.uid, p.gid, mode)
 		if err = p.docker.CopyToContainer(context, p.ctr.ID, "/", appReader, types.CopyToContainerOptions{}); err != nil {
 			err = errors.Wrapf(err, "failed to copy files to '%s' container", p.name)
 		}
 
-		err = <-errChan
-		if err != nil {
-			err = errors.Wrapf(err, "create tar archive from '%s'", p.appDir)
-		}
+		_ = appReader.Close()
```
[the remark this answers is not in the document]
ameyer-pivotal: Makes sense. Not sure why we didn't honestly...

build/phase.go
ameyer-pivotal: I like that idea. Agree that it's not 100% necessary, but that we should do it soon. I'll make a chore for it.

ameyer-pivotal: Chore link: https://github.com/orgs/buildpack/projects/1#card-23809969

.travis.yml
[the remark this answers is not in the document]
ameyer-pivotal: The reason that this works for PRs is that they show up as `master` when they run the build ;)

acceptance/acceptance_test.go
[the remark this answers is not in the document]
ameyer-pivotal: @jromero @ekcasey I think we should duplicate this for-loop test, strip it down, and restore the original journey style (for just one case, not all the enumerations of app path types). Thoughts?

build/phase.go
ameyer-pivotal: Good catch!

acceptance/acceptance_test.go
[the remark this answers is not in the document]
ameyer-pivotal: In fact, there's no need for a loop anymore, since dir and zip are the only cases

build/phase.go
ameyer-pivotal: @ekcasey Resolved by adding err check around the copy. We can't return the error because we're in a closure, but the effect should be the same. Feel free to double-check us.

ekcasey: we can't return an error but we can just `return` right?

build/phase.go
```diff
@@ -118,19 +120,18 @@ func (p *Phase) Run(context context.Context) error {
...
 
-		appReader, errChan := archive.CreateTarReader(p.appDir, appDir, p.uid, p.gid, mode)
-		if err = p.docker.CopyToContainer(context, p.ctr.ID, "/", appReader, types.CopyToContainerOptions{}); err != nil {
-			err = errors.Wrapf(err, "failed to copy files to '%s' container", p.name)
+		if err != nil {
+			err = errors.Wrapf(err, "create tar archive from '%s'", p.appPath)
 		}
 
-		err = <-errChan
 		if err != nil {
```
ekcasey: I think you want the opposite conditional here

acceptance/acceptance_test.go
```diff
@@ -145,99 +145,108 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		when("default builder is set", func() {
+
```
ekcasey: Minor nitpick: this extra space conflicts with our style conventions

build/phase.go
ameyer-pivotal: Haha, yeah... We saw that too

acceptance/acceptance_test.go
ameyer-pivotal: Oops! Artifact of c/p. Done.

build/phase.go
ameyer-pivotal: Yeah that's what we did

ameyer-pivotal: Done.

PR #244, opened 2019-07-23
Fix: Tests hanging on MacOS when testing error on phase

jromero:

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -38,7 +38,12 @@ func readAsTar(src, basePath string, uid, gid int, mode int64, writeFn func(tw *
 		}()
 
 		tw := tar.NewWriter(w)
-		defer tw.Close()
+		defer func() {
+			// only close if no errors have occurred
+			if err == nil {
+				tw.Close()
+			}
+		}()
 
 		err = writeFn(tw, src, basePath, uid, gid, mode)
 	}()


jromero:
> I think it's important that the tar writer be closed before the pipe. Otherwise, the tar footer will be cut off: https://golang.org/src/archive/tar/writer.go?s=14125:14156#L456

Interesting, was not aware that the footer would be written on `Close`. Will investigate further, including any potential alternative solutions.

jromero:
@sclevine I dug into a bit further and it seems to be that the issue only arises when we call `tw.Close()` when an error while writing has occurred.

I've made additional changes to address your concerns as well. Let me know what you think and if the solution makes sense.

internal/archive/archive.go
```diff
@@ -34,11 +34,16 @@ func readAsTar(src, basePath string, uid, gid int, mode int64, writeFn func(tw *
 	go func() {
 		var err error
 		defer func() {
-			w.CloseWithError(err)
+			_ = w.CloseWithError(err)
```
ekcasey: apologies for the nitpick, but if we aren't using any of the return values I think we should omit the assignment to `_`

internal/archive/archive.go
```diff
@@ -34,11 +34,16 @@ func readAsTar(src, basePath string, uid, gid int, mode int64, writeFn func(tw *
...
 		}()
 
 		tw := tar.NewWriter(w)
-		defer tw.Close()
+		defer func() {
+			// only close if no errors have occurred
+			if err == nil {
+				_ = tw.Close()
```
ekcasey: same thing

Issue #245, opened 2019-07-29
pack --buildpack interprets url as a buildpack ID


ekcasey:
Observed:

```
pack build --buildpack https://example.com/buildpack.tgz -p ../some-app some/app
...
No version for https://example.com/buildpack.tgz buildpack provided, will use https://example.com/buildpack.tgz@latest
setting custom order
ERROR  failed to set custom buildpack order: no versions of buildpack https://example.com/buildpack.tgz were found on the builder
```

Expected:

pack should download the buildpack from the the provided url

PR #246, opened 2019-07-29
Fix issue where buildpack uris are treated like ids

ekcasey:


jromero:
Resolves #245

PR #250, opened 2019-08-02
Adds basic compatibility with buildpack/lifecycle#149

### LINKED ISSUE #243: Provide basic compatibility with lifecycle#149
This issue is to provide basic compatibility with buildpack/lifecycle#149

> The following document fully reflects compatibility between pack 0.3.0 and 0.4.0 _(upcoming release)_ against lifecycle 0.3.0 and 0.4.0 _(upcoming release)_.
https://docs.google.com/spreadsheets/d/12pF7xsCLvxSoaSlh8mHjjhorVAzacllz38BXyr-ubJI/view

See rows 4, 5, 14 and 15 of document above as they pertain to basic compatibility.

jromero:
Resolves #243
Relates to buildpack/lifecycle#149


acceptance/testdata/mock_buildpacks/v1/simple-layers-buildpack/buildpack.toml
```diff
@@ -0,0 +1,9 @@
+schema = "1"
```
jkutner: do we need to keep some documentation of the previous schema versions? Is `schema` a required element? If not, what is the default when it's not present? We could probably sort this all out in an RFC.

acceptance/testdata/mock_buildpacks/v1/builder.toml
```diff
@@ -1,3 +1,5 @@
+schema = "2"
```
jkutner: i noticed this is in a dir `v1`. is that related to schema version (and is it correct)?

acceptance/testdata/mock_buildpacks/v1/simple-layers-buildpack/buildpack.toml
ameyer-pivotal: Will be removing this for this PR

acceptance/testdata/mock_buildpacks/v1/builder.toml
ameyer-pivotal: This is the schema of the builder.toml, which is 2 in both cases (v1 dir and v2 dir) because it's always using the latest pack.

ameyer-pivotal: We'll be removing references to `schema` in this PR for now, until we can discuss further

builder/config.go
```diff
@@ -15,16 +15,47 @@ import (
...
+	}
+
+	return order
+}
+
+type GroupConfig struct {
+	Group []BuildpackRefConfig `toml:"group"`
+}
+
+type BuildpackRefConfig struct {
```
ekcasey: This is exactly the same as `BuildpackMetadata` and `BuildpackRefMetadata`. I think we just need one `BuldpackRef`; that would be less confusing.

ekcasey: It could also be `GroupEntry` or something like that

builder/metadata.go
```diff
@@ -1,44 +1,118 @@
...
 	Version string `json:"version"`
-	Latest  bool   `json:"latest"`
+	Latest  bool   `json:"latest"` // deprecated
 }
 
-type GroupMetadata struct {
-	Buildpacks []GroupBuildpack `json:"buildpacks" toml:"buildpacks"`
+type OrderMetadata []GroupMetadata
+
+func (o OrderMetadata) ToConfig() OrderConfig {
```
ekcasey: This appears to be translating between two identical objects

builder/builder.go
```diff
@@ -379,14 +396,23 @@ func (b *Builder) rootOwnedDir(path string, time time.Time) *tar.Header {
...
 
 func (b *Builder) orderLayer(dest string) (string, error) {
-	orderTOML := &bytes.Buffer{}
-	err := toml.NewEncoder(orderTOML).Encode(OrderTOML{Groups: b.metadata.Groups})
+	buf := &bytes.Buffer{}
+	lifecycleVersion := b.GetLifecycleVersion()
+
+	var tomlData interface{}
+	if lifecycleVersion != nil && lifecycleVersion.LessThan(semver.MustParse("0.4.0")) {
+		tomlData = v1OrderTOMLFromOrderTOML(orderTOML{Order: b.metadata.Groups.ToConfig()})
```
ekcasey: why not just make one directly from an Order instead of creating an intermediate `orderTOML`?

acceptance/acceptance_test.go
```diff
@@ -328,7 +339,7 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
 						)
 						txt, err := h.RunE(cmd)
 						h.AssertNotNil(t, err)
-						h.AssertContains(t, txt, "buildpack 'other/stack/bp' version 'other-stack-version' does not support stack 'pack.test.stack'")
+						h.AssertContains(t, txt, "buildpack 'other/stack/bp@other-stack-version' does not support stack 'pack.test.stack'")
```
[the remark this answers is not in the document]
jromero: I'm curious about that. I know we use it in the output of `inspect-builder` so my thought is that if we use it in a place where it's intended to be displayed normally surely it would make sense to display it in an error (edge) case.

acceptance/acceptance_test.go
```diff
@@ -780,57 +791,84 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
...
-		h.AssertNil(t, err)
+	if v, ok := os.LookupEnv("LIFECYCLE_VERSION"); ok {
+		lifecycleVersion = semver.MustParse(v)
 	}
 
+	// DETERMINE TEST DATA
+	t.Log("Using buildpacks from: ", testBuildpacksDir())
+	h.RecursiveCopy(t, testBuildpacksDir(), tmpDir)
+
+	// COPY builder.toml
```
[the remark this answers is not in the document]
jromero: True. Maybe "AMENDING" would be a better term?

build/phase_test.go
```diff
@@ -129,7 +129,7 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
 				h.AssertContains(t, outBuf.String(), "failed to read file")
 			})
 
-			when("is posix", func() {
+			when.Pend("is posix", func() {
```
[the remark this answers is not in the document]
jromero: It was there due to issue #244 but can be removed now. Will remove.

builder/builder.go
```diff
@@ -49,6 +48,19 @@ type Builder struct {
 	replaceOrder  bool
 }
 
+type orderTOML struct {
+	Order OrderConfig `toml:"order"`
+}
+
+type stackTOML struct {
```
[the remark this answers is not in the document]
jromero: The purpose is mostly an organizational as @ameyer-pivotal and I came to the consensus that having a clear separation between what toml format and metadata format made a lot more sense. Particularly as we diverge between what `builder.toml` contents look like and what we store on the `label` (for compatibility purposes).

The cost of having a few extra objects vs a clearer picture and less probably of inadvertently changing the format of either was worth it to us when we discussed it.

builder/metadata.go
jromero: Not necessarily, the inner object might have that resemblance but the parent `Order` differs in inner properties. `Buildpacks` vs `Groups`.

acceptance/acceptance_test.go
```diff
@@ -276,21 +276,32 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S) {
...
 							"--buildpack", notBuilderTgz,
+
+							// with version
 							"--buildpack", "simple/layers@simple-layers-version",
+
+							// without version
 							"--buildpack", "noop.buildpack",
+
+							// latest (for backwards compatibility)
+							"--buildpack", "read/env@latest",
```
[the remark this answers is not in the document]
ekcasey: We can leave this for now, but let's take out support for this syntax before the `v0.5.0` release.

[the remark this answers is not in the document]
ekcasey: We can print a deprecated warning instead of removing it.

PR #259, opened 2019-08-13
Additional compatibility with lifecycle#149 changes

### LINKED ISSUE #252: Additional compatibility with lifecycle#149
This issue is to provide additional compatibility with buildpack/lifecycle#149

> The following document fully reflects compatibility between pack 0.3.0 and 0.4.0 (upcoming release) against lifecycle 0.3.0 and 0.4.0 (upcoming release).
https://docs.google.com/spreadsheets/d/12pF7xsCLvxSoaSlh8mHjjhorVAzacllz38BXyr-ubJI/view

See rows 10-12 of document above as they pertain to compatibility.

jromero:
Resolves #252 


Makefile
```diff
@@ -40,6 +40,10 @@ acceptance: format vet
 	@echo "> Running acceptance tests..."
 	$(GOCMD) test -mod=vendor -v -count=1 -parallel=1 -timeout=0 -tags=acceptance ./acceptance
 
+acceptance-all: format vet
+	@echo "> Running acceptance tests..."
+	ACCEPTANCE_SUITE_CONFIG=$$(cat ./acceptance/testconfig/all.json) $(GOCMD) test -mod=vendor -v -count=1 -parallel=1 -timeout=0 -tags=acceptance ./acceptance
```
jromero: @ameyer-pivotal mentioned that there might have been some discussions about this being a path to a file instead of json structure. Is there a preference?

acceptance/testconfig/all.json
```diff
@@ -0,0 +1,7 @@
+[
```
sclevine: Could this just be code in the acceptance test? Seems like the normal table-testing idioms should apply.

builder/builder.go
```diff
@@ -79,7 +79,7 @@ func GetBuilder(img imgutil.Image) (*Builder, error) {
 
 	label, err := img.Label(MetadataLabel)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to find run images for builder %s", style.Symbol(img.Name()))
+		return nil, errors.Wrapf(err, "failed to get metadata label from image %s", style.Symbol(img.Name()))
```
ekcasey: We should probably give the label name in this error message so folks know what to look for.

builder/builder_test.go
```diff
@@ -202,16 +215,152 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
 
-			when("order is invalid", func() {
-				it("produces an error", func() {
-					subject.SetOrder(builder.Order{{
-						Group: []builder.BuildpackRef{
-							{BuildpackInfo: buildpack.BuildpackInfo{ID: "missing-buildpack-id", Version: "missing-buildpack-version"}}},
-					}})
+			when("validating order", func() {
+				when("has single buildpack", func() {
+
```
ekcasey: remove empty line here

builder/builder_test.go
```diff
@@ -202,16 +215,152 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								Version: "buildpack-1-version",
+							},
+							Path: buildpackTgz,
+							Stacks: []buildpack.Stack{
+								{ID: "some.stack.id"},
+							},
+						})
+					})
+
+					it("should resolve unset version", func() {
```
ekcasey: We should probably check the `order.toml` in this test as well.

builder/builder_test.go
```diff
@@ -202,16 +215,152 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								Version: "buildpack-1-version",
+							},
+							Path: buildpackTgz,
+							Stacks: []buildpack.Stack{
+								{ID: "some.stack.id"},
+							},
+						})
+					})
+
+					it("should error", func() {
```
ekcasey: We can probably move this test into the "has a single buildpack" context, to avoid repeating the same `Before`

builder/builder_test.go
```diff
@@ -202,16 +215,152 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+								{BuildpackInfo: buildpack.BuildpackInfo{ID: "missing-buildpack-id"}}},
+						}})
+
+						err := subject.Save()
+
+						h.AssertError(t, err, "no versions of buildpack 'missing-buildpack-id' were found on the builder")
+					})
+				})
+
+				when("order points to missing buildpack version", func() {
```
ekcasey: Same comment as above, we could skip respecifying the same Before steps if we move the its to the other `when` block.

inspect_builder.go
```diff
@@ -16,7 +16,7 @@ type BuilderInfo struct {
 	RunImage         string
 	RunImageMirrors  []string
 	Buildpacks       []builder.BuildpackMetadata
-	Groups           []builder.V1Group // TODO: change to v2 order type when order label is added
+	Groups           []builder.OrderEntry
```
ekcasey: Should we call this field `Order` for consistency?

internal/archive/archive.go
```diff
@@ -143,6 +143,49 @@ func ReadTarEntry(tarPath string, entryPath ...string) (*tar.Header, []byte, err
 	return nil, nil, fmt.Errorf("could not find entry path '%s' in tar", entryPath)
 }
 
+func ListTarContents(tarPath string) ([]tar.Header, error) {
```
ekcasey: It seems like this function is only used by tests. We should probably move it to `testhelpers`.

acceptance/testconfig/all.json
jromero: There is something that we are trying to achieve in CI that makes it preferable to not have the full configuration statically set in the `acceptance_test`.

After talking to @ekcasey, we are leaning towards having a single environment variable that combines the paths and combinations to make it clearer.

Something like this:

```json
[
  {"pack": "", "pack_create_builder": "", "lifecycle": ""},
  {"pack": "", "pack_create_builder": "./pack-0.3.0", "lifecycle": "./lc-0.4.0"},
  {"pack": "./pack-0.3.0", "pack_create_builder": "", "lifecycle": "./lc-0.4.0"},
  {"pack": "./pack-0.4.0", "pack_create_builder": "./pack-0.3.0", "lifecycle": "./lc-0.4.0"}
]
```

We are also exploring passing a path to the suite config file instead of json through an environment variable.

Thoughts?

Makefile
ekcasey: Lets not hang this PR up on this question or the test config question. But I would like to improve the experience of running the compatibility tests in the future.

acceptance/testconfig/all.json
ekcasey: Let's punt this to a future PR.

PR #266, opened 2019-08-22
new default paths on builder

ekcasey:
* add symlinks and copies at old paths for buldpack compatibility with pack v0.3.0


builder/builder_test.go
```diff
@@ -973,7 +928,7 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			it("adds the env vars as files to the image", func() {
-				layerTar, err := baseImage.FindLayerWithPath("/platform/env/SOME_KEY")
+				layerTar, err := baseImage.FindLayerWithPath("/cnb/platform/env/SOME_KEY")
```
jromero: Is this remanent of moving the platform and then reverting it? Shouldn't it be `/platform/env/SOME_KEY`?

build/phases.go
```diff
@@ -7,25 +7,18 @@ import (
...
 	cacheDir       = "/cache"
 	launchCacheDir = "/launch-cache"
+	platformDir    = "/platform"
 )
 
 func (l *Lifecycle) Detect(ctx context.Context) error {
 	detect, err := l.NewPhase(
 		"detector",
 		WithArgs(
-			"-buildpacks", buildpacksDir,
```
jromero: If I understand this change correctly, we are no longer explicitly setting paths and strictly relying on the defaults set in the lifecycle for most paths. Is that right? 

If so, why can't the other directories also be defaulted?

ekcasey: I left the other directories b/c, for every build, we do things like mount volumes or write env vars to those paths. Therefore, the lifecycle defaults could result in a failure if they don't agree w/ `pack`. `lifecycle` and `buildpacks` paths only matter when you are creating a builder and we have more flexibility if `pack build` doesn't try to override the defaults (unless... you use `--buildpack` during `build` and clean separation breaks down).

builder/builder.go
```diff
@@ -416,6 +427,10 @@ func (b *Builder) defaultDirsLayer(dest string) (string, error) {
 		return "", errors.Wrapf(err, "creating %s dir in layer", style.Symbol(layersDir))
 	}
 
+	if err := tw.WriteHeader(b.rootOwnedDir(cnbDir, now)); err != nil {
+		return "", errors.Wrapf(err, "creating %s dir in layer", style.Symbol(buildpacksDir))
```
ekcasey: error message should use `cnbDir` instead of `buildpacksDir`

builder/compat.go
```diff
@@ -51,18 +66,104 @@ func (o Order) ToV1Order() V1Order {
...
+		if err := addSymlink(tw, compatLink, bpDir); err != nil {
+			return err
+		}
+
+		if lifecycleVersion := b.GetLifecycleVersion(); lifecycleVersion != nil && lifecycleVersion.LessThan(semver.MustParse("0.4.0")) {
+			if err := symlinkLatest(tw, bpDir, bp, b.metadata); err != nil {
+				return err
+			}
+		}
+
```
ekcasey: delete this empty line

builder/compat_test.go
```diff
@@ -0,0 +1,373 @@
...
+	"github.com/buildpack/pack/builder"
+	"github.com/buildpack/pack/buildpack"
+	"github.com/buildpack/pack/internal/archive"
+	"github.com/buildpack/pack/lifecycle"
+	h "github.com/buildpack/pack/testhelpers"
+)
+
+func TestCompat(t *testing.T) {
+	color.NoColor = true
+	spec.Run(t, "Builder", testCompat, spec.Parallel(), spec.Report(report.Terminal{}))
```
ekcasey: `"Compat"` instead of `"Builder"`

PR #274, opened 2019-08-31
update latest suggested stacks for cloudfoundry

drnic:
/cc @sclevine confirming that these are the new suggested images for `pack suggest-stacks`?


drnic:
@ekcasey thanks; fixed

PR #275, opened 2019-09-02
Fix pack completion command

### LINKED ISSUE #272: .pack/completion: no such file or directory
I tried to run the tutorial but seems there's something missing in completion setup

https://buildpacks.io/docs/install-pack/#auto-completion

```
suztomo@suxtomo24:~$ pack version
v0.3.0 (git sha: 3123e32ea7c570527c6e2cdb31f4e5c002334287)
suztomo@suxtomo24:~$ pack completion
ERROR  open /usr/local/google/home/suztomo/.pack/completion: no such file or directory
```

Am I missing something?

tbalthazar:
The `pack completion` command throws an error if the `PACK_HOME`
directory doesn't exist. (as described in #272)

This PR makes sure the directory is created if it doesn't exist yet,
and adds a test for the `CompletionCommand` to verify it works both
when the `PACK_HOME` directory exists and when it doesn't.

FIXES #272

### DIFF
--- a/commands/completion.go
+++ b/commands/completion.go
@@ -25,6 +25,10 @@ To configure your bash shell to load completions for each session, add the follo
 			if err != nil {
 				return errors.Wrap(err, "getting pack home")
 			}
+
+			if err = config.MkdirAll(packHome); err != nil {
+				return errors.Wrapf(err, "creating pack home: %s", packHome)
+			}
 			completionPath := filepath.Join(packHome, "completion")
 
 			if err := cmd.Parent().GenBashCompletionFile(completionPath); err != nil {
--- a/commands/completion_test.go
+++ b/commands/completion_test.go
@@ -1,0 +1,81 @@
+package commands_test
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+	"github.com/spf13/cobra"
+
+	"github.com/buildpack/pack/commands"
+	"github.com/buildpack/pack/internal/mocks"
+	"github.com/buildpack/pack/logging"
+	h "github.com/buildpack/pack/testhelpers"
+)
+
+func TestCompletionCommand(t *testing.T) {
+	spec.Run(t, "Commands", testCompletionCommand, spec.Random(), spec.Report(report.Terminal{}))
+}
+
+func testCompletionCommand(t *testing.T, when spec.G, it spec.S) {
+	var (
+		command    *cobra.Command
+		logger     logging.Logger
+		outBuf     bytes.Buffer
+		packHome   string
+		missingDir string
+	)
+
+	it.Before(func() {
+		logger = mocks.NewMockLogger(&outBuf)
+		// the CompletionCommand calls a method on its Parent(), so it needs to have
+		// one.
+		command = &cobra.Command{}
+		command.AddCommand(commands.CompletionCommand(logger))
+		command.SetArgs([]string{"completion"})
+		var err error
+		packHome, err = ioutil.TempDir("", "")
+		h.AssertNil(t, err)
+		missingDir = filepath.Join(packHome, "not-found")
+	})
+
+	it.After(func() {
+		os.RemoveAll(packHome)
+	})
+
+	when("#CompletionCommand", func() {
+		when("PackHome directory does exist", func() {
+			it.Before(func() {
+				os.Setenv("PACK_HOME", packHome)
+			})
+
+			it.After(func() {
+				os.Unsetenv("PACK_HOME")
+			})
+
+			it("creates the completion file", func() {
+				h.AssertNil(t, command.Execute())
+				h.AssertContains(t, outBuf.String(), filepath.Join(packHome, "completion"))
+			})
+		})
+
+		when("PackHome directory does not exist", func() {
+			it.Before(func() {
+				os.Setenv("PACK_HOME", missingDir)
+			})
+
+			it.After(func() {
+				os.Unsetenv("PACK_HOME")
+			})
+
+			it("creates the completion file", func() {
+				h.AssertNil(t, command.Execute())
+				h.AssertContains(t, outBuf.String(), filepath.Join(missingDir, "completion"))
+			})
+		})
+	})
+}
--- a/config/config.go
+++ b/config/config.go
@@ -49,7 +49,7 @@ func Read(path string) (Config, error) {
 }
 
 func Write(cfg Config, path string) error {
-	if err := os.MkdirAll(filepath.Dir(path), 0777); err != nil {
+	if err := MkdirAll(filepath.Dir(path)); err != nil {
 		return err
 	}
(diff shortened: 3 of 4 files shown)


commands/completion_test.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+
+	it.After(func() {
+		os.RemoveAll(packHome)
+	})
+
+	when("#CompletionCommand", func() {
+		when("PackHome directory does exist", func() {
+			it.Before(func() {
+				os.Setenv("PACK_HOME", packHome)
```
jromero: Although the risk is minimal it does leak the environment variable to further tests. We should really be unsetting the env var in an `it.After`.

commands/completion_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+			it("creates the completion file", func() {
+				h.AssertNil(t, command.Execute())
+				h.AssertContains(t, outBuf.String(), filepath.Join(packHome, "completion"))
+			})
+		})
+
+		when("PackHome directory does not exist", func() {
+			it.Before(func() {
+				os.Setenv("PACK_HOME", missingDir)
```
jromero: Ditto

commands/completion.go
```diff
@@ -25,6 +26,10 @@ To configure your bash shell to load completions for each session, add the follo
 			if err != nil {
 				return errors.Wrap(err, "getting pack home")
 			}
+
+			if err = os.MkdirAll(packHome, 0755); err != nil {
```
jromero: Seems like we are also doing this logic in `config.Write`. I wonder if it would be a better solution to simply create the directory structure as part of `main.initConfig` with the default config. This would hence remove the responsibility of having to create `packHome` from all commands.

commands/completion_test.go
tbalthazar: Good catch, thanks!

commands/completion.go
tbalthazar: Thanks for your feedback @jromero.

I agree that it would be nice to make sure to create the `PACK_HOME` directory in a central place, but I'm not sure (yet?) what this place would be.

I'm not sure creating `PACK_HOME` in [`main.initConfig`](https://github.com/buildpack/pack/blob/master/cmd%2Fpack%2Fmain.go/#L79) would allow us to remove the [`MkdirAll` call in `config.Write`](https://github.com/buildpack/pack/blob/master/config%2Fconfig.go/#L52), because this method doesn't only create the `PACK_HOME` directory, but also any sub-directory present in the `path` (as validated [by this test](https://github.com/buildpack/pack/blob/master/config%2Fconfig_test.go/#L99)).

Also, creating `PACK_HOME` in `main.initConfig` would make it hard to unit-test the `CompletionCommand`, which is a smell to me 😄 .

My understanding is that config paths are created in a _central_ place already (the `config.Write` method), and that the `CompletionCommand` is not dealing with config here, it wants to add a file in the `PACK_HOME` directory.

We could make sure the `PACK_HOME` directory exists in the [`config.PackHome`](https://github.com/buildpack/pack/blob/master/config%2Fconfig.go/#L29) method, but it feels a bit like a side effect to me to do that there.

In summary, I can't come up with a central place I like to create that `PACK_HOME` directory 😄 .

I guess I like the idea that it's the responsibility of the function that uses `PACK_HOME` to make sure it exits before trying to write to it. (the `Downloader` does that for instance, it is [given packHome](https://github.com/buildpack/pack/blob/master/client.go/#L66) and it makes sure [the directory exists](https://github.com/buildpack/pack/blob/master/downloader.go/#L75).

But I don't have a clear mental model of the whole `pack` command, so I might be missing something.

What do you think?

jromero: > I'm not sure creating PACK_HOME in main.initConfig would allow us to remove the MkdirAll call in config.Write, because this method doesn't only create the PACK_HOME directory, but also any sub-directory present in the path (as validated by this test).
>
> Also, creating PACK_HOME in main.initConfig would make it hard to unit-test the CompletionCommand, which is a smell to me 😄 .

I think that it simply makes it more clear that it's not, and in our case maybe shouldn't be, the responsibility of the `CompletionCommand` and instead the responsibility of the pack command as a whole. But that's not to say that's the only way to achieve what we want.

> We could make sure the PACK_HOME directory exists in the config.PackHome method, but it feels a bit like a side effect to me to do that there.

Completely, agree. That's not ideal or expected.

<hr/>

Overall, I think the biggest hang up is just the fact that we have different permissions at this call site versus `config.Write`. This is a relatively a simple issue and we could work through it by creating a single method to create the directories in a predictable fashion. 

For example, abstracting away the need to know the details of `PackHome`. Anything that needs a path from within PackHome would just use something like this.

```
func config.Path(path string) (string, error) {
	// mkdirall PackHome + path
        // return PackHome + path
}
```

tbalthazar: > Overall, I think the biggest hang up is just the fact that we have different permissions at this call site versus config.Write

Agreed.

> This is a relatively a simple issue and we could work through it by creating a single method to create the directories in a predictable fashion.
> For example, abstracting away the need to know the details of PackHome. Anything that needs a path from within PackHome would just use something like this.

The `config.Write` method currently allows to write a config file anywhere (it's not scoped to `PACK_HOME`).

A minimal change we could do to address the _permission_ issue is have a method like this:

```go
func config.Mkdirall(path string) (string, error) {
  // os.MkdirAll(path, 0777)
}
```

and use it in `config.Write` and `CompletionCommand`.

I'm wondering if the `config.Write` method should be changed to be scoped to `PACK_HOME` (or if there's a use case for writing config outside of `PACK_HOME`).
AFAICT, `config.Write` is currently only called with `config.DefaultConfigPath`.

What I would be tempted to do is to add a method that looks like what you suggested:

```go
func config.Path(path string) (string, error) {
	// mkdirall PackHome + path
        // return PackHome + path
}
```

and modify `config.Write` and `CompletionCommand` to use that method.

But since it would modify the behavior of the `config.Write` method, I think it might be worth discussing it in another PR (to keep this one very simple)

What do you think?

jromero: I think we are on the same page. Either option sounds good and acceptable AFAIAC. Whichever you feel more comfortable with.

tbalthazar: Ok, I'll finish this PR by adding a `config.Mkdirall` method because I think it's simple, it fixes a real issue and leaves the code in a better situation.
I'll open a separate PR to discuss the option of modifying `config.Write` to have it always write in `PACK_HOME` and see if it is an issue that we want to address now or at all.

Thanks for all the feedback on this PR <3

commands/completion.go
```diff
@@ -25,6 +25,10 @@ To configure your bash shell to load completions for each session, add the follo
 			if err != nil {
 				return errors.Wrap(err, "getting pack home")
 			}
+
+			if err = config.MkdirAll(packHome); err != nil {
+				return errors.Wrap(err, "creating pack home: "+packHome)
```
jromero: Minor: We have a preference to use `Wrapf` instead of string contamination.

tbalthazar: Updated.

suztomo:
Thank you

PR #279, opened 2019-09-03
Allow tar file buildpacks on Windows

### LINKED ISSUE #270: Support tar buildpacks for Windows
### Motivation

As part of buildpack/docs#43 we have discovered that `tar` archives are inherently supported for buildpacks on posix and would like to update changes related to #181 to allow for `tar` archives instead of solely checking against `tgz`.

Additional considerations:
* We should stop relying on extension and instead look at the header bytes to determine file type.

jromero:
Resolves #270

### DIFF
--- a/build.go
+++ b/build.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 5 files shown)


build.go
```diff
@@ -255,6 +257,37 @@ func isBuildpackID(bp string) bool {
 	return false
 }
 
+func ensureBPSupport(bpPath string) (err error) {
```
ekcasey: I feel like this type of logic belongs in the blob package. The idea was to encapsulate all logic about data formats in that package.

For now it looks like directory blobs are never allowed on windows (b/c of the perm issues). If we wanted to support directory blobs for an app dir or adding a buildpack at build time we could create strict and non-strict blob constructor options to that package.

jromero: Deferred to #286

Issue #280, opened 2019-09-04
Bump `DefaultLifecycleVersion` in `pack`


ameyer-pivotal:

jromero:
Resolved by #287

PR #282, opened 2019-09-05
Validate buildpack API version

### LINKED ISSUE #254: Validate buildpack API version
### Buildpack incompatible with lifecycle
**Given** I have a buildpack.toml for buildpack X with an `api` string field `"0.x"`
**Given** A lifecycle with a descriptor file specifying that the lifecycle has Buildpack API other than `0.x`
**When** I run `pack create-builder` while targeting the given lifecycle
**Then** I see an error message:
```
buildpack X (Buildpack API version 0.x) is incompatible with lifecycle X (Buildpack API version 0.y)
```

### Buildpack compatible with lifecycle
**Given** I have a buildpack.toml for buildpack X with an `api` string field `"0.x"`
**When** I run `pack create-builder` while targeting lifecycle with a matching Buildpack API version
**Then** the command completes successfully

### Buildpack incompatible with builder (lifecycle)
**Given** I have a `buildpack.toml` for buildpack X with an `api` string field `"0.x"`
**Given** A builder that specifies a lifecycle with Buildpack API other than `"0.x"` in it's metadata
**When** I run `pack build` while targeting the given builder
**Then** I see an error message:
```
buildpack X (Buildpack API version 0.x) is incompatible with builder X (Buildpack API version 0.y)
```

### Buildpack compatible with builder (lifecycle)
**Given** I have a `buildpack.toml` for buildpack X with an `api` string field less than `"0.x"`
**Given** A builder that specifies a lifecycle with Buildpack API `"0.x"` in it's metadata
**When** I run `pack build` while targeting the given builder
**Then** the command completes successfully

## Missing api metadata

**Given** a `buildpack.toml` with a missing `api` field
**Then** the `api` shall be assumed to be `0.1`

**Given** a lifecycle with a missing descriptor file
**Then** the lifecycles Buildpack API shall be assumed to be `0.1`

**Given** a builder with missing lifecycle api metadata
**Then** the lifecycles Buildpack API shall be assumed to be `0.1`

### LINKED ISSUE #284: Use the lifecycle's Buildpack API Version to determine order.toml format
We should use the lifecycle's Buildpack API Version instead of the lifecycle version now.

_Originally posted by @ekcasey in https://github.com/buildpack/pack/diffs/53_

jromero:
Also:

* Default lifecycle during `create-builder` if no version or URI is specified
* Use BP API Version to determine `order.toml` format

Resolves #254
Resolves #284 

### DIFF
--- a/build_test.go
+++ b/build_test.go
@@ -15,6 +15,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Masterminds/semver"
 	"github.com/buildpack/imgutil/fakes"
 	"github.com/docker/docker/client"
 	"github.com/fatih/color"
@@ -22,6 +23,7 @@ import (
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
+	"github.com/buildpack/pack/api"
 	"github.com/buildpack/pack/blob"
 	"github.com/buildpack/pack/builder"
 	ifakes "github.com/buildpack/pack/internal/fakes"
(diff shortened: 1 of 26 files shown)


build_test.go
```diff
@@ -570,6 +600,51 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 					h.AssertNil(t, os.Remove(buildpackTgz))
 				})
 
+				when("buildpack is an older api version", func() {
```
ekcasey: Given that we cover this in the builder package tests I don't see the point in duplicating this test coverage here, there  will just be more tests to maintain.

create_builder_test.go
```diff
@@ -160,6 +168,31 @@ func testCreateBuilder(t *testing.T, when spec.G, it spec.S) {
 				err := subject.CreateBuilder(context.TODO(), opts)
 				h.AssertError(t, err, "buildpack from URI 'https://example.fake/bp-one.tgz' has version '1.2.3' which does not match version '0.0.0' from builder config")
 			})
+
+			it("should fail when buildpack is not compatible with lifecycle", func() {
```
ekcasey: Same comment as for `build_test.go`, as long as we have one test where we error if saving the builder errors, I don't think we need to retest all of the validations from `builder_test.go`.

builder/lifecycle.go
```diff
@@ -16,44 +16,42 @@ import (
...
 
-var lifecycleBinaries = []string{
-	"detector",
-	"restorer",
-	"analyzer",
-	"builder",
-	"exporter",
-	"cacher",
-	"launcher",
+func DefaultLifecycleDescriptor() LifecycleDescriptor {
```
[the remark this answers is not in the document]
jromero: We have an upcoming story in the works that will make this not equal to each other. #280

builder/lifecycle.go
```diff
@@ -16,44 +16,42 @@ import (
...
-
-var LatestLifecycleDescriptor = LifecycleDescriptor{
-	Info: LifecycleInfo{
-		Version: lifecycleVersionLatest,
-	},
-	API: LifecycleAPI{
-		PlatformVersion:  apiVersionLatest,
-		BuildpackVersion: apiVersionLatest,
-	},
+func AssumedLifecycleDescriptor() LifecycleDescriptor {
```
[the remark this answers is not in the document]
jromero: > I'm not sure what the motivation for moving this to a function is?

The motivation for the function is to handle pointer references in cases where we use the value as a "default" before deserialization. Through the process of "decoding" it would update the pointer reference instead of creating a new reference hence affecting multiple objects. Basically, we decided to make this a function to ensure we always have a new reference. This would hopefully prevent us from running into this issue again.

That said, we could probably just go with the separate constants as strings to prevent the same issue.

create_builder_test.go
jromero: We couldn't find the duplicate test. Could you clarify? There are other errors tested in builder_test.go but not this one specifically.

jromero: Did you mean `build_test.go`?

ekcasey: I guess it isn't currently in `builder_test.go` but it seems analogous to https://github.com/buildpack/pack/blob/feature/254-validate-buildpack-api/builder/builder_test.go#L437 to me. This is where I would expect to look for this level of coverage.

jromero: Makes sense. Moved test.

PR #290, opened 2019-09-12
`pack build --no-pull` should never checks remote.

### LINKED ISSUE #249: `pack build --no-pull` should only check local registry. Currently: checks remote; if that fails, it exits
Please see attached screenshots showing:
1. pack inspect-builder --> shows an error checking remote registry, success checking local
2. pack build --no-pull --> shows failure with same error as remote check in screenshot 1, then exit

Expected behavior:
pack build no-pull would not check remote at all, only local, and successfully use the local copy of the builder

Image 1:
Screen Shot 2019-08-01 at 11 44 32 AM

Image 2:
Screen Shot 2019-08-01 at 11 47 20 AM

dgageot:
Fix #249


dgageot:
@ekcasey This is pretty straightforward. It makes working with `--no-pull` a few seconds faster. And makes is possible to re-run `pack build` without an internet connection.

PR #291, opened 2019-09-12
pack build allows multiple `--env-file`

### LINKED ISSUE #264: pack build should allow multiple `--env-file` declarations
Currently if `--env-file` is passed in multiple times to `pack build`, only the final one is available within the build container.  The command should allow multiple occurrences of `--env-file` (and `--env` if that is not already the case).

dgageot:
Fixes #264

### DIFF
--- a/commands/build_test.go
+++ b/commands/build_test.go
@@ -2,6 +2,7 @@ package commands_test
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


commands/build_test.go
```diff
@@ -0,0 +1,159 @@
...
+				command.SetArgs([]string{"--builder", "my-builder", "image"})
+				h.AssertNil(t, command.Execute())
+			})
+		})
+
+		when("an env file is provided", func() {
+			var envPath string
+
+			it.Before(func() {
+				envfile, err := ioutil.TempFile("", "envfile")
```
ekcasey: the temp file should be cleaned up in an `it.After`

commands/build_test.go
```diff
@@ -0,0 +1,159 @@
...
+				h.AssertNil(t, err)
+				defer envfile.Close()
+
+				envfile.WriteString(`KEY=VALUE`)
+				envPath = envfile.Name()
+			})
+
+			it("builds an image with a builder", func() {
```
ekcasey: It would be better if the description strings in each `it` described what this particular test was asserting on (i.e. env vars in this case)

dgageot:
Thanks. I'm on it

PR #296, opened 2019-09-13
Fix the TestBuildCommand unit tests

dgageot:
Those are fixes to #291:
 + Delete temp files after the test
 + Give better description for test cases


jromero:
Thanks for the quick turn around! 💯 

Merging as soon as test pass.

PR #297, opened 2019-09-13
Small changes to make the code simpler

dgageot:

### DIFF
--- a/build/phases.go
+++ b/build/phases.go
@@ -122,21 +122,7 @@ func (l *Lifecycle) newExport(repoName, runImage string, publish bool, launchCac
 			),
 		)
 	}
-	if launchCacheName != "" {
-		return l.NewPhase(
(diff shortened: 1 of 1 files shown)


build/phases.go
```diff
@@ -126,42 +121,37 @@ func (l *Lifecycle) newExport(repoName, runImage string, publish bool, launchCac
...
-					"-image", runImage,
-					"-layers", layersDir,
-					"-app", appDir,
-					"-daemon",
-					repoName,
-				),
-			)
-		}
 	}
+	if launchCacheName != "" {
```
ekcasey: There are further simplifications we could make here. Looking at the code `launchCacheName` will never be an empty string so we probably don't need this guard/check.

dgageot: Thanks for the feedback. I removed some more code!

PR #298, opened 2019-09-13
Show progress of downloads

### LINKED ISSUE #236: show progress of downloads
## Motivation
Some buildpacks are large and therefore a buildpack download during `pack create-builder` can take a long time. As a user I want visual feedback about the progress of a download so that I have confidence `pack` is still working have some idea how long I will have to wait for the download to complete.

## Suggested UX
**When** `pack` is downloading an asset (e.g. a buildpack, or a lifecycle archive)
**Then** I should see a progress bar that updates as the download proceeds

Example:
```
$pack create-builder
...
Downloading from "https://example.com/my/buildpack.tgz"
[=======>          ] 50%
...
```

**When** the download is complete
**Then** the progress bar should be replaced with the message `Download Complete`

dgageot:
Had to vendor an additional (small) dependency.

download-progress

Fix #236

### DIFF
--- a/blob/downloader.go
+++ b/blob/downloader.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/mitchellh/ioprogress"
 	"github.com/pkg/errors"
 
 	"github.com/buildpack/pack/internal/paths"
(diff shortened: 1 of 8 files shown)


ameyer-pivotal:
We'll want this to be more visual in the future, and probably hide when `--verbose` is not supplied (currently, only the URL hides), or always show both URL and progress regardless of `--verbose` (whichever seems more consistent and necessary). This is a still a great addition!

Accepted.

PR #299, opened 2019-09-13
Add a linter

dgageot:
I wanted to start introducing a meta-linter. It won't run on Travis yet, nor does it make sure that the required tool is installed.


build.go
```diff
@@ -97,7 +97,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 	lifecycleVersion := descriptor.Info.Version
 	if lifecycleVersion == nil {
 		c.logger.Warnf("lifecycle version unknown, assuming %s", style.Symbol(builder.AssumedLifecycleVersion))
-		lifecycleVersion = builder.VersionMustParse(builder.AssumedLifecycleVersion)
+		// lifecycleVersion = builder.VersionMustParse(builder.AssumedLifecycleVersion)
```
jromero: This doesn't look right.

internal/logging/logging.go
```diff
@@ -112,12 +112,10 @@ func (lw *logWithWriters) WantQuiet(f bool) {
 }
 
 // NewLogWithWriters creates a logger to be used with pack CLI.
-func NewLogWithWriters() *logWithWriters {
+func NewLogWithWriters() *logWithWriters { //nolint:golint
```
jromero: Can we ignore specific errors instead of all rules from a certain linter?

Makefile
```diff
@@ -30,6 +30,10 @@ format: install-goimports
 	@echo "> Formating code..."
 	@goimports -l -w -local ${PACKAGE_BASE} ${SRC}
 
+lint:
+	@echo "> Linting code..."
+	@golangci-lint run -c golangci.yaml
```
jromero: We should be able to [automatically install](https://github.com/golangci/golangci-lint#windows-macos-and-linux) this similar to `goimports` and `mockgen`. Is there a reason we wouldn't want to do that?

jromero: Re-read your description and noticed that you called out that this aspect isn't done yet. 👍

build.go
dgageot: yeah. Exactly the kind of bugs that a linter catches :-)

internal/logging/logging.go
dgageot: Yes, I think. I'll try it. The doc says something like that is valid: `//nolint:golint,unused`

build.go
jromero: Sorry, I probably should have been more specific. I was calling out the fact that this line is commented out. What lint warning or error was the linter providing?

dgageot: It says, lifecycleVersion is not used:
```
ineffassign: Detects when assignments to existing variables are not used
```

jromero: Ah! That makes sense. Probably remnants of prior logic. The message still makes sense but might even need to move to a more appropriate location of the code base.

So this should probably now look like this for the time being:

```go
if descriptor.Info.Version == nil {
	c.logger.Warnf("lifecycle version unknown, assuming %s", style.Symbol(builder.AssumedLifecycleVersion))
} else {
	c.logger.Debugf("Executing lifecycle version %s", style.Symbol(descriptor.Info.Version.String()))
}
```

Agree with your sentiment. Lint would have caught this. 😄

internal/logging/logging.go
sclevine: We should fix this as well -- definitely better to export (or return through an interface).

dgageot:
@jromero I've updated the PR to make sure `golangci-lint` is installed by the Makefile. However, it pulls so many dependencies that I chose not to pollute the main `go.mod`. You tell me what you think

PR #301, opened 2019-09-13
when reading archives treat ./file like file

ekcasey:
Signed-off-by: Micah Young <myoung@pivotal.io>

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/archive/archive.go
```diff
@@ -143,7 +143,7 @@ func ReadTarEntry(rc io.Reader, entryPath string) (*tar.Header, []byte, error) {
 			return nil, nil, errors.Wrap(err, "failed to get next tar entry")
 		}
 
-		if header.Name == entryPath {
+		if filepath.Clean(header.Name) == entryPath {
```
jromero: Do we want to clean both sides?

ekcasey: I was imagining that since we are the only users of this function we can always pass in paths w/o `./`, whereas we should be able to handle any reasonable tar we get.

PR #314, opened 2019-09-24
Add verbose option

jkutner:
This PR adds the `--verbose` or `-v` option to `pack` commands. In doing so, it changes the default log level from `Debug` to `Info`, and adjusts the marker used for a number of calls that were `Debug/Debugf` but should be `Info/Infof`.

In particular, we're changing the marker used for the lifecycle output from `Debug` to `Info` in order to support more verbose levels. This creates a better mapping between the output of the two processes (previously, the `Info` logs from lifecycle were being logged as `Debug` in `pack`).


jkutner:
The acceptance tests seem like they need a lifecycle update:

```
--- FAIL: TestAcceptance/acceptance_suite/p_current_cb_current_lc_current/rebase/--publish/--run-image/uses_provided_run_image (1.27s)
    acceptance_test.go:702: Expected nil: failed to execute command: [/tmp/pack.acceptance.binary.004242971/pack build --no-color localhost:32770/some-org/yaunlufpon -p testdata/mock_app --builder localhost:32770/some-org/wgbvbutuwd --run-image localhost:32770/run-before/pwknyulxey --no-pull], exit status 1, ===> DETECTING
        [detector] flag provided but not defined: -log-level
        [detector] Usage of /lifecycle/detector:
        [detector]   -app string
        [detector]     	path to app directory (default "/workspace")
        [detector]   -buildpacks string
        [detector]     	path to buildpacks directory (default "/cnb/buildpacks")
        [detector]   -group string
        [detector]     	path to group.toml (default "./group.toml")
        [detector]   -order string
        [detector]     	path to order.toml (default "/cnb/order.toml")
        [detector]   -plan string
        [detector]     	path to plan.toml (default "./plan.toml")
        [detector]   -platform string
        [detector]     	path to platform directory (default "/platform")
        [detector]   -version
        [detector]     	show version
        ERROR: failed with status code: 2
```

jkutner:
I think this might need a lifecycle release in order to update [builder/lifecycle.go](https://github.com/buildpack/pack/blob/master/builder/lifecycle.go#L22) and get the acceptance tests working.

build/phases.go
```diff
@@ -17,6 +17,7 @@ func (l *Lifecycle) Detect(ctx context.Context) error {
 	detect, err := l.NewPhase(
 		"detector",
 		WithArgs(
+			"-log-level", l.determineLogLevel(),
```
jromero: Based on the results of the acceptance tests we probably want to make this flag optionally set based on lifecycle version to ease the transition (backwards compatibility).

jkutner: oof, good call

jkutner: Fixed in 1f07183496bafccb05050df48a90ad3571d23a4b

jkutner:
@jromero @ekcasey addressed your comments, and CI is passing now.

PR #320, opened 2019-09-25
Colors will now work on Windows. 

murphybytes:
Signed-off-by: John Murphy <johnmurphy@salesforce.com>

This change addresses the problems with console output on Windows, in short, colors now work on both Windows and Posix platforms. Another issue that is fixed is that writers that receive output from shell scripts running on build packs now intercept that output and interpret the color information correctly on Windows as well as Posix platforms.  

In order to get things to work I had to abandon `fatih/color` and replace it with a refactored color package `heroku/color`  The original package is archived and has some fundamental flaws such as race conditions so a refactor made more sense than a fork for the original.  The changes and the rationale for doing them are outlined [here](https://github.com/heroku/color/blob/master/README.md).

Edit - I've included a fix for the image pull update issue referenced below by @ekcasey.


ekcasey:
@murphybytes I haven't gotten a chance to review the code yet, but I built from this ref and tried it out real quick. The image pull output looks borked. The lines are no longer being overwritten:

Screen Shot 2019-09-26 at 9 44 47 AM

ekcasey:
Apologies @murphybytes, it appears the docker pull output was actually already broken on `master`. Apologies for blaming your PR!

However, I do want to investigate and fix this regression before we merge more changes to the CLI output.

internal/logging/logging_test.go
```diff
@@ -34,254 +34,86 @@ func newTestLogger(stdout, stderr io.Writer) *logWithWriters {
 	return &lw
 }
 
+func mockStd() (*color.Console, func() string) {
```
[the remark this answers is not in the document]
murphybytes: Sure.  `color.NewConsole` is designed to wrap stdout or stderr and as such takes `*os.File` as an argument.  This is a consequence of one of the libraries that the color package depends on taking an *os.File as an argument.  Anyway, I use `os.Pipe` to get input and output `*os.File` and pass back a function that is called to retrieve the bytes written to the input file.

builder/lifecycle_test.go
```diff
@@ -18,7 +18,8 @@ import (
 )
 
 func TestLifecycle(t *testing.T) {
-	color.NoColor = true
+	color.Disable(true)
+	defer func() { color.Disable(false) }()
```
ameyer-pivotal: Probably don't need to re-enable it right? Unless I haven't looked further down yet and this modifies some local configuration or something.

image/fetcher.go
```diff
@@ -87,7 +86,15 @@ func (f *Fetcher) pullImage(ctx context.Context, imageID string) error {
 		return err
 	}
 	writer := logging.GetInfoWriter(f.logger)
-	termFd, isTerm := term.GetFdInfo(writer)
+	type descriptor interface {
+		Fd() uintptr
+	}
+	var termFd uintptr
+	var isTerm bool
```
ameyer-pivotal: Minor: Maybe `isTerm := false` for clarity?

style/style.go
```diff
@@ -3,18 +3,18 @@ package style
...
 }
 
 var Symbol = func(format string, a ...interface{}) string {
-	if color.NoColor {
-		format = fmt.Sprintf("'%s'", format)
+	if color.Enabled() {
+		return Key(format, a...)
 	}
-	return Key(format, a...)
+	return fmt.Sprintf("'%s'", Key(format, a...))
```
ameyer-pivotal: This should not include a call to `Key()`. I suppose color might just get ignored here, but the intention is clearer without the call

builder/lifecycle_test.go
murphybytes: The color is enabled by default, so when I disable it I want to re-enable it when I'm done testing to restore it to its default state.  Is it needed strictly speaking, maybe not, but in keeping with principle of least surprise I'd like the color library to be in it's default state for any other tests that might use it.

image/fetcher.go
murphybytes: IMHO it's good practice to use zero values for this kind of thing.

builder/lifecycle_test.go
ameyer-pivotal: I see your point

PR #328, opened 2019-10-03
Add a --network flag to the pack cli 

### LINKED ISSUE #309: Allow attaching pack to network for build and run
For local builds, sometimes it is useful to connect to local container networks in order to access stuff on separate networks. As an example, some buildpacks may create database connections - ruby seems to do this when compiling assets... - that may be sitting on a non-default network.

ForestEckhardt:
This flag will allow you to specify the network mode used during detect and build. This is a done to fulfill the request in the following issue: https://github.com/buildpack/pack/issues/309

Co-authored-by: Ryan Moran <rmoran@pivotal.io>

Resolves #309


build.go
```diff
@@ -125,6 +126,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		HTTPProxy:  proxyConfig.HTTPProxy,
 		HTTPSProxy: proxyConfig.HTTPSProxy,
 		NoProxy:    proxyConfig.NoProxy,
+		Network:    opts.Network,
```
ekcasey: I am not sure this belongs in `ProxyConfig`. Can we add a `ContainerConfig` field to `LifecycleOptions` with a `Network` field. This would give us a nice way to extend to include future `docker run` type flags

build/exports_test.go
```diff
@@ -0,0 +1,7 @@
+package build
+
+import "github.com/docker/docker/api/types/container"
+
+func GetPhaseHostConfig(phase *Phase) *container.HostConfig {
```
ekcasey: I don't love adding a getter to the package's exported interface just for testing. I would rather move the phase tests directly into the `build` package so we can test the unexported field if adding another test endpoint to https://github.com/buildpack/pack/blob/master/build/testdata/fake-lifecycle/phase.go is too annoying (the design of these tests could be simpler).

ForestEckhardt:
We extracted the Network setting into it's own object refactored the testing to avoid using a getter on the phases.

jromero:
Resolves #309

Issue #329, opened 2019-10-03
Order suggested builders alphabetically


jromero:
Based on conversation in the working group meeting. We have agreed to organize the output of suggested builders and stacks based on an alphabetical order.

Relates to #311

PR #330, opened 2019-10-04
Improve stack/builders suggestions

### LINKED ISSUE #311: Update `suggest-stacks` entries
We've determined that the cnb bionic stack `io.buildpacks.stacks.bionic` is to be used solely for learning purposes (as a sample and for tutorials) only. Thereby, it should not be suggested via the command `suggest-stacks`.

Related story: #156

dgageot:
+ Order stack/builders suggestions alphabetically
 + Move every stack under the `Stacks maintained by the community` category
 + Move code portions next to where they are used
 + Retrieve the builder descriptions concurrently
 + Simplify the code as much as I could

Fixes #311 and #329

### DIFF
--- a/commands/suggest_builders.go
+++ b/commands/suggest_builders.go
@@ -39,10 +39,9 @@ func SuggestBuilders(logger logging.Logger, client PackClient) *cobra.Command {
 		Use:   "suggest-builders",
 		Short: "Display list of recommended builders",
 		Args:  cobra.NoArgs,
-		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
+		Run: func(*cobra.Command, []string) {
 			suggestBuilders(logger, client)
-			return nil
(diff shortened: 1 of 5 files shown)


commands/suggest_builders.go
```diff
@@ -1,22 +1,97 @@
...
+
+func suggestBuilders(logger logging.Logger, client PackClient) {
+	sort.Slice(suggestedBuilders, func(i, j int) bool { return suggestedBuilders[i].Image < suggestedBuilders[j].Image })
+
+	logger.Info("Suggested builders:")
+
+	// Fetch descriptions concurrently.
+	descriptions := make([]string, len(suggestedBuilders))
+
+	var wg sync.WaitGroup
```
jromero: 👍 on the concurrency-based optimization.

PR #337, opened 2019-10-08
added debugging info to phase

### LINKED ISSUE #335: pack build should log lifecycle arguments
It's difficult to debug differences between running pack build locally and a remote build that calls the lifecycle binaries individually. Adding additional debugging info when calling lifecycle steps would be extremely helpful.

For example, the run function could print out the binaries, arguments, and environment variables it uses:

https://github.com/buildpack/pack/blob/master///build/phase.go#L164

josephlewis42:
Fixes #335

### DIFF
--- a/build/phase.go
+++ b/build/phase.go
@@ -161,6 +161,15 @@ func (p *Phase) Run(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to copy files to '%s' container", p.name)
 	}
 
+	p.logger.Debugf("Running container: %v", p.ctr.ID)
+	if ctrConf := p.ctrConf; ctrConf != nil {
+		p.logger.Debugf("  Cmd: %v", ctrConf.Cmd)
+		p.logger.Debugf("  Env: %v", ctrConf.Env)
+		p.logger.Debugf("  User: %s", ctrConf.User)
+		p.logger.Debugf("  Image: %s", ctrConf.Image)
+		p.logger.Debugf("  Labels: %v", ctrConf.Labels)
+	}
+
 	return container.Run(
 		ctx,
 		p.docker,


build/phase.go
```diff
@@ -161,6 +161,15 @@ func (p *Phase) Run(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to copy files to '%s' container", p.name)
 	}
 
+	p.logger.Debugf("Running container: %v", p.ctr.ID)
+	if ctrConf := p.ctrConf; ctrConf != nil {
+		p.logger.Debugf("  Cmd: %v", ctrConf.Cmd)
+		p.logger.Debugf("  Env: %v", ctrConf.Env)
```
ekcasey: I think including the `Env` here might be misleading. When users pass environment variables to `pack build` with the `--env` flag they actually get written to disk at `/platform/env` as described in the spec https://github.com/buildpack/spec/blob/master/buildpack.md#provided-by-the-platform.

I think seeing an empty Env array after providing an `--env` flag is likely to confuse users even when everything is working as expected.

build/phase.go
```diff
@@ -161,6 +161,15 @@ func (p *Phase) Run(ctx context.Context) error {
 		return errors.Wrapf(err, "failed to copy files to '%s' container", p.name)
 	}
 
+	p.logger.Debugf("Running container: %v", p.ctr.ID)
```
ekcasey: I don't see the value in providing the container ID w/o providing a way to prevent the intermediate containers from being removed after running.

ameyer-pivotal: I think making the cmd the top-level output (i.e. the output without the prepended spaces), rather than the container ID, would be a good alternative, as the cmd describes which phase is being run.

build/phase.go
ameyer-pivotal: @ekcasey While I agree that it could be confusing, wouldn't this be useful information?

Alternatively, could this information be logged in lifecycle instead of `pack`? That way we could read the actual env vars, which should include the image-configured vars PLUS the `--env` vars (correct me if I'm wrong @ekcasey). One issue with that might be that there's no way (currently, I believe) for `pack` to tell lifecycle when to be verbose, so if it were logged in lifecycle, it might always show up.

ameyer-pivotal:
Requested changes, mostly related to the container ID being output. For now, maybe it's best to keep the "Env" section (even though it may lead to some confusion). Perhaps we just call it "Env (Image)" instead of "Env" for now to disambiguate a bit.

zmackie:
@josephlewis42 Checking in on this. What can the maintainers do to help it move along?

josephlewis42:
@zmackie I'll close this one and let the maintainers decide among yourselves the best way to do debugging in here since it seems like there's not a general consensus.

Issue #239, opened 2019-07-11
`pack create-package` from buildpack blobs


ekcasey:
This story should be done on a branch so the community can experiment with it, but not merged until https://github.com/buildpack/rfcs/pull/11 is finalized

## Motivations

Please see https://github.com/buildpack/rfcs/pull/11

## `pack create-package`
**Given** I have the following `package.toml` file
```
[default]
id = "id.buildpack.one"
version = "0.0.1"

[[blobs]]
uri = "https://example.com/buildpack1.tgz"

[[stacks]]
id = "some.stack.id"
```
**And** `https://example.com/buildpack1.tgz` points to an archive with the following entries
```
buildpack.toml
bin
bin/detect
bin/build
```
where `buildpack.toml` has the following contents
```
[buildpack]
id = "id.buildpack.one"
version = "0.0.1"

[[stacks]]
id = "some.stack.id"
```

**When** I run `pack create-package -p <path-to-package.toml> example.com/some/package:tag`
**Then** An image with the tag `example.com/some/package:tag` should be exported to the docker daemon
**And** the package should have a label with key `io.buildpacks.buildpackage.metadata` and the following value
```
{
  "id": "id.buildpack.one",
  "version": "0.0.1",
  "stacks": [
    {
      "id": "some.stack.id",
    }
  ]
}
```
**And** a layer should be created containing `id.buildpack.one` with the following files
```
/cnb/buildpacks/id.buildpack.one/0.0.1/buildpack.toml
/cnb/buildpacks/id.buildpack.one/0.0.1/bin/
/cnb/buildpacks/id.buildpack.one/0.0.1/bin/detect
/cnb/buildpacks/id.buildpack.one/0.0.1/bin/build
```

### --publish
**When** I run `pack create-package -p package.toml example.com/some/package:tag --publish`
**Then** It behaves exactly as above but exports the package to the `example.com` registry

### Notes
* `uri` field should support all `uri` type currently supported by `pack create-builder`
* should use download cache
* multiple buildpack blobs and multiple stacks should be supported

jromero:
@ekcasey  Based on the [spec](https://github.com/buildpack/spec/blob/master/distribution.md#buildpackage) the label should be `io.buildpacks.buildpackage.metadata`

ameyer-pivotal:
This passes acceptance.

Issue #179, opened 2019-05-09
Validate mixins during `create-builder` and `build`


ekcasey:
### Adds buildpack mixins to builder layers label

**Given** I have a `buildpack.toml` with the following section:
```
[[stacks]]
  id = "some.stack.id"
  mixins = ["mixinA", "mixinB"]
```
**When** I run `pack create-builder` against the buildpack and stack `some.stack.id`
**Then** The builder contains the buildpack mixins in its `io.buildpacks.buildpack.layers` label:
```
{
  "some.buildpack.id": {
    "some.buildpack.version": {
      "stacks": [{
        "id": "some.stack.id",
        "mixins": ["mixinA", "mixinB"]
      }],
      ...
    }
  },
  ...
}
```

> **NOTE**: Order buildpacks will not have mixins (as they have no stacks), so they can be skipped over

----

#### Mixins satisfied
**Given** I have a build image with a `io.buildpacks.stack.mixins` label: 
```
[ "mixinA", "build:mixinB", "mixinC" ]
```
**And** I have a buildpack that requires mixins: `mixinA`, `build:mixinB`, `run:mixinD`
**When** I run `pack create-builder` against the build image and buildpack
**Then** the command succeeds

> **NOTE**: `run:`-prefixed mixins do not need to be validated during `create-builder` (e.g. `run:mixinD` above).

----

#### Mixins unsatisfied
**Given** I have a build image `some/build` with a `io.buildpacks.stack.mixins` label: 
```
[ "mixinA", "build:mixinB", "mixinC" ]
```
**And** I have a buildpack that requires mixins: `mixinX`, `mixinY`, `build:mixinB`, `run:mixinD`
**When** I run `pack create-builder` against the build image and buildpack
**Then** the command fails with error similar to:
```
Error: invalid build image 'some/build': buildpack 'some.buildpack.id@some.buildpack.version' requires missing mixin(s): mixinX, mixinY
```

> **NOTE**: `run:`-prefixed mixins do not need to be validated during `create-builder` (e.g. `run:mixinD` above).

#### Mixins match
**Given** I have a builder image with a `io.buildpacks.stack.mixins` label:
```
["mixinA", "build:mixinB"]
```
**And** the builder's corresponding run image has a `io.buildpacks.stack.mixins` label:
```
["mixinA", "run:mixinC"]
```
**When** I run `pack build` against the builder
**Then** the command succeeds

----

#### Missing one or more common mixins
[...]

PR #341, opened 2019-10-09
Validate package creation

### LINKED ISSUE #240: `pack create-package` validates default ID and version
[PLACEHOLDER]

**Given** a `package.toml` with the following contents
```
[default]
id = "id.buildpack.two"
version = "0.0.2"

[[blobs]]
uri = "https://example.com/buildpack1.tgz"

[[stacks]]
id = "some.stack.id"
```
**Where** `id.buildpack.two` is not a buildpack defined in the `buildpack.toml` in the given blob
**Then** I receive a helpful error message

jromero:
The validations here are based on the [spec](https://github.com/buildpack/spec/blob/master/distribution.md#buildpackage) with the omission of mixins as that is currently incomplete (see #179)

- Resolves #240
- Relates to #239

### DIFF
--- a/buildpackage/builder.go
+++ b/buildpackage/builder.go
@@ -1,0 +1,135 @@
+package buildpackage
+
+import (
+	"io/ioutil"
+	"os"
+
+	"github.com/buildpack/imgutil"
+	"github.com/pkg/errors"
+
+	"github.com/buildpack/pack/dist"
+	"github.com/buildpack/pack/style"
+)
+
+type ImageFactory interface {
+	NewImage(repoName string, local bool) (imgutil.Image, error)
+}
+
+type PackageBuilder struct {
+	defaultBpInfo dist.BuildpackInfo
+	buildpacks    []dist.Buildpack
+	stacks        []dist.Stack
+	imageFactory  ImageFactory
+}
+
+func NewBuilder(imageFactory ImageFactory) *PackageBuilder {
+	return &PackageBuilder{
+		imageFactory: imageFactory,
+	}
+}
+
+func (p *PackageBuilder) SetDefaultBuildpack(bpInfo dist.BuildpackInfo) {
+	p.defaultBpInfo = bpInfo
+}
+
+func (p *PackageBuilder) AddBuildpack(buildpack dist.Buildpack) {
+	p.buildpacks = append(p.buildpacks, buildpack)
+}
+
+func (p *PackageBuilder) AddStack(stack dist.Stack) {
+	p.stacks = append(p.stacks, stack)
+}
+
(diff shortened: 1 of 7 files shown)


buildpackage/builder.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+type PackageBuilder struct {
+	defaultBpInfo dist.BuildpackInfo
+	buildpacks    []dist.Buildpack
+	stacks        []dist.Stack
+	imageFactory  ImageFactory
+}
+
+func NewBuilder(imageFactory ImageFactory) *PackageBuilder {
```
ameyer-pivotal: I hate how overloaded the term "builder" is in our code, but I suppose we can't get away from that at the moment...

buildpackage/builder.go
```diff
@@ -0,0 +1,138 @@
...
+	defer os.RemoveAll(tmpDir)
+
+	for _, bp := range p.buildpacks {
+		bpLayerTar, err := dist.BuildpackLayer(tmpDir, 0, 0, bp)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := image.AddLayer(bpLayerTar); err != nil {
+			return nil, errors.Wrapf(err, "adding layer tar for buildpack %s", style.Symbol(bp.Descriptor().Info.ID+"@"+bp.Descriptor().Info.Version))
```
ameyer-pivotal: Can now use `.Info.FullName()`

buildpackage/builder.go
```diff
@@ -0,0 +1,138 @@
...
+	if !bpExists(bps, defBp) {
+		return errors.Errorf("selected default %s is not present",
+			style.Symbol(defBp.ID+"@"+defBp.Version),
+		)
+	}
+
+	return nil
+}
+
+func validateStacks(bps []dist.Buildpack, stacks []dist.Stack) error {
```
ameyer-pivotal: This is probably a larger question, but I wonder why we even need to declare stacks in `package.toml`. They seem to be computable on the fly as the intersection of all stack sets from each buildpack. cc: @sclevine, @ekcasey

image/factory.go
```diff
@@ -0,0 +1,27 @@
+package image
+
+import (
+	"github.com/buildpack/imgutil"
+	"github.com/docker/docker/client"
+	"github.com/google/go-containerregistry/pkg/authn"
+)
+
+type DefaultImageFactory struct {
```
ameyer-pivotal: Wondering if the word "Default" is appropriate for this type... What about it makes it "default-ey"? 

Side note: I see the convenience of the `NewImage` method, but aside from that, is this really buying us much?

PR #350, opened 2019-10-15
Add more linters

dgageot:
Also restore the check on imports being organised
in 3 groups.


jromero:
Thanks for the contribution! Keep it up!

PR #351, opened 2019-10-15
Add layer digest and layer diff id to builder layers label

### LINKED ISSUE #251: Builder Metadata Format
**Given** I have a builder.toml with buildpacks `buildpack.id@1.2.4` and `other.id@3.4.5`
**When** I create a builder
**Then** the builder contains the following labels:

-  `io.buildpacks.buildpack.layers` (`version`s are not optional):
```json
{
  "buildpack.id": {
    "1.2.4": {
      "layerDigest": "sha256:8abc...",
      "order": [
        {"group": [{"id": "other.id", "version": "3.4.5"}]}
      ]
    }
  },
  "other.id": {
    "3.4.5": {
      "layerDigest": "sha256:9acc..."
    }
  }
}
```
- `io.buildpacks.buildpack.order` (builder-only, `version`s are optional):
```json
[
   {"group": [{"id": "buildpack.id", "version": "1.2.3"}]},
   {"group": [{"id": "other.id"}]}
]
```

- `io.buildpacks.builder.metadata` with current metadata contents (for backwards compatibility)

**Notes**
- Moved buildpackage requirements to #289

jromero:
Resolves #251


dist/layers.go
```diff
@@ -109,3 +113,44 @@ func embedBuildpackTar(tw *tar.Writer, uid, gid int, bp Buildpack, baseTarDir st
...
+	defer fh.Close()
+
+	// h <----------------- +
+	//                      |
+	// zh <- zw (gunzip) <- + <-- fh
+
+	h := sha256.New()
+	zh := sha256.New()
+
+	zw, err := gzip.NewWriterLevel(zh, gzip.DefaultCompression)
```
ekcasey: Will `gzip.DefaultCompression` get the correct digest when we are working with remote images given that ggcr uses `gzip.BestSpeed`?

jromero: Yeah... just looked around `imgutil` and found that we are [using the `tarball.Layer`](https://github.com/buildpack/imgutil/blob/master/remote/remote.go#L308) which would give us a different digest based on what we've found. The problem from what I've found is that it all depends on the client that compresses and pushes the image to the registry (`pack build --publish` vs `docker push` vs any other tool). This means that relying on this information might be a bit too brittle for our use case.

FWIW: I've got a [PR up on GGRC](https://github.com/google/go-containerregistry/pull/574) which would allow us to make it consistent with the `docker` client.

jromero: FYI: created issue in imgutil (https://github.com/buildpack/imgutil/issues/15) to track necessary compression change.

Issue #349, opened 2019-10-11
Add `packages` support to `create-package`


jromero:
## Motivation

See [RFC-7](https://github.com/buildpack/rfcs/blob/master/text/0007-spec-distribution.md#buildpack-developer)

### AC1. `create-package` with package in `package.toml`

**Given** I have the following `package.toml` file
```
[default]
id = "buildpack.1"
version = "0.0.1"

[[buildpacks]]
uri = "https://example.com/buildpack1.tgz"

[[packages]]
ref = "package/bp2-bp3"

[[stacks]]
id = "some.stack.id"
```
**And** `https://example.com/buildpack1.tgz` points to an archive with the following entries
```
buildpack.toml
bin
bin/detect
bin/build
```
where `buildpack.toml` has the following contents
```
[buildpack]
id = "buildpack.1"
version = "0.0.1"

[[stacks]]
id = "some.stack.id"
```
**And** there is an image `package/bp2-bp3` created by `pack create-package` with the following `package.toml` contents:

```
[default]
id = "buildpack-2"
version = "0.0.1"

[[buildpacks]]
uri = "buildpack-2-v0.0.1.tgz"

[[buildpacks]]
uri = "buildpack-3-v0.0.1.tgz"

[[stacks]]
id = "some.stack.id"
```
**When** I run `pack create-package -p <path-to-package.toml> example.com/some/package:tag`
**Then** An image with the tag `example.com/some/package:tag` should be exported to the docker daemon
**And** the package should have a label with key `io.buildpacks.buildpackage.metadata` and the following value
```
{
  "id": "buildpack.1",
  "version": "0.0.1",
  "stacks": [
    {
      "id": "some.stack.id",
    }
  ]
}
```
**And** a layer should be created containing `buildpack.1` with the following files:
```
/cnb/buildpacks/buildpack.1/0.0.1/buildpack.toml
/cnb/buildpacks/buildpack.1/0.0.1/bin/
/cnb/buildpacks/buildpack.1/0.0.1/bin/detect
/cnb/buildpacks/buildpack.1/0.0.1/bin/build
```
**And** a layer should be created containing `buildpack.2` with the following files:
```
/cnb/buildpacks/buildpack.2/0.0.1/buildpack.toml
/cnb/buildpacks/buildpack.2/0.0.1/bin/
/cnb/buildpacks/buildpack.2/0.0.1/bin/detect
/cnb/buildpacks/buildpack.2/0.0.1/bin/build
```
[...]

Issue #348, opened 2019-10-11
`pack build --buildpack` accepts a buildpackage as an argument


jromero:
If you have an image for a buildpackage on a registry, that can be passed in the `--buildpack` flag during `pack build`

#### Acceptance criteria:
```gherkin
Given I have access to a  buildpackage on a registry or your local docker daemon
When you run `pack build MY_APP --buildpack <BUILDPACKAGE:tag>`
Then the build executes using the packaged buildpacks using the current default builder

Given I have access to a  buildpackage on a registry or your local docker daemon
When you run `pack build MY_APP --buildpack <BUILDPACKAGE:tag> --buildpack <BUILDPACKAGE:tag>`
Then the build executes using the packaged buildpacks follow the existing semantics


Given I have access to a  buildpackage on a registry or your local docker daemon
When you run `pack build MY_APP --buildpack <BUILDPACKAGE@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855>`
Then the build executes using the packaged buildpacks follow the existing semantics
```

### Note:
- This issue does not include support for `.cnb` files.
- This issue only initially tackles building from unambigous buildpackages, IE `repo/image:tag`. Later story will support disambiguating.

Issue #347, opened 2019-10-11
Add `packages` support to `create-builder`


jromero:
## Motivation

See [RFC-7](https://github.com/buildpack/rfcs/blob/master/text/0007-spec-distribution.md#buildpack-developer)

### AC1. `create-builder` - package image

**Given** there is a `builder.toml` with the following contents:
```
[buildpacks]]
uri = "buildpack-1-v0.0.1.tgz"

[[packages]]
ref = "package/bp2-bp3:bionic"

[stack]
id = "io.buildpacks.samples.stacks.bionic"
run-image = "cnbs/sample-stack-run:bionic"
build-image = "cnbs/sample-stack-build:bionic"
```
**And** there is an image `package/bp2-bp3:bionic` created by `pack create-package` with the following `package.toml` contents:

```
[default]
id = "buildpack-2"
version = "0.0.1"

[[buildpacks]]
uri = "buildpack-2-v0.0.1.tgz"

[[buildpacks]]
uri = "buildpack-3-v0.0.1.tgz"

[[stacks]]
id = "io.buildpacks.samples.stacks.bionic"
```
**And** we run `pack create-builder foo-builder --builder-config builder.toml`
**When** we run `inspect-builder`
**Then** it should output:
```
...
Buildpacks:
  ID                    VERSION
  buildpack-1           0.0.1
  buildpack-2           0.0.1
  buildpack-3           0.0.1
...
```

### AC2. `create-builder` - nested package image

**Given** there is a `builder.toml` with the following contents:
```
[buildpacks]]
uri = "buildpack-1-v0.0.1.tgz"

[[packages]]
ref = "package/bp2-bp3-bp4-bp5:bionic"

[stack]
id = "io.buildpacks.samples.stacks.bionic"
run-image = "cnbs/sample-stack-run:bionic"
build-image = "cnbs/sample-stack-build:bionic"
```
**And** there is an image `package/bp4-bp5:bionic` created by `pack create-package` with the following `package.toml` contents:

```
[default]
id = "buildpack-4"
version = "0.0.1"

[[blobs]]
uri = "buildpack-4-v0.0.1.tgz"

[[blobs]]
uri = "buildpack-5-v0.0.1.tgz"

[[stacks]]
id = "io.buildpacks.samples.stacks.bionic"
```
**And** there is an image `package/bp2-bp3-bp4-bp5:bionic` created by `pack create-package` with the following `package.toml` contents:

```
[default]
id = "buildpack-2"
[...]

PR #353, opened 2019-10-16
Hide create-package command

ameyer-pivotal:
We are hiding this command as it's not ready for prime-time without additional follow up issues (#240, #347, #348, #349) yet some early adopters might want to play with it.


commands/create_package.go
```diff
@@ -52,6 +52,9 @@ func CreatePackage(logger logging.Logger, client PackClient) *cobra.Command {
 	cmd.MarkFlagRequired("package-config")
 	cmd.Flags().BoolVar(&flags.Publish, "publish", false, "Publish to registry")
 	AddHelpFlag(cmd, "create-package")
+	
+	cmd.Hidden = true
```
ameyer-pivotal: This will also have the effect of disabling the acceptance tests, since it will no longer show up in the help output. I'm okay with that if everyone else is.

PR #354, opened 2019-10-16
Make download URL and progress output consistent

### LINKED ISSUE #355: Download progress shown without a download URL
pack is showing the download progress, even when the URL being downloaded from is not shown.

This should be fixed such that progress and URL are both shown ~only when the `--verbose` flag is provided.~ always, regardless of log level

ameyer-pivotal:
Always show download URL and progress

- Also tidied other output

Resolves #355

### DIFF
--- a/build.go
+++ b/build.go
@@ -258,8 +258,6 @@ func (c *Client) processBuildpacks(ctx context.Context, buildpacks []string) ([]
 				return nil, dist.OrderEntry{}, err
 			}
 
-			c.logger.Debugf("fetching buildpack from %s", style.Symbol(bp))
-
 			blob, err := c.downloader.Download(ctx, bp)
 			if err != nil {
 				return nil, dist.OrderEntry{}, errors.Wrapf(err, "downloading buildpack from %s", style.Symbol(bp))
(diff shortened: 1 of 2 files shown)


ekcasey:
@ameyer-pivotal I'm not sure I agree we should hide progress in the default case. When the download is slow it's hard to know whether the command is working. I can imagine users wanting download progress w/o wanting debug level output.

ameyer-pivotal:
> @ameyer-pivotal I'm not sure I agree we should hide progress in the default case. When the download is slow it's hard to know whether the command is working. I can imagine users wanting download progress w/o wanting debug level output.

@ekcasey That makes sense. I had considered that too. Do you vote we always show this output then?

build.go
```diff
@@ -258,8 +258,6 @@ func (c *Client) processBuildpacks(ctx context.Context, buildpacks []string) ([]
 				return nil, dist.OrderEntry{}, err
 			}
 
-			c.logger.Debugf("fetching buildpack from %s", style.Symbol(bp))
```
ameyer-pivotal: Downloader shows this info already

PR #360, opened 2019-10-21
Use diffID instead of digest in builder metadata

ameyer-pivotal:


cache/image_cache.go
```diff
@@ -16,7 +16,7 @@ type ImageCache struct {
 }
 
 func NewImageCache(imageRef name.Reference, dockerClient *client.Client) *ImageCache {
```
ekcasey: We removed support for image caches in pack. We should probably by removing these functions instead of updating them.

PR #375, opened 2019-11-06
Move most packages into internal

ameyer-pivotal:
Moves most packages into internal, as a first step toward reorganizing our packages.

Phase 2, which will be a later PR, will be to minimize/simplify the packages within `internal`.


client.go
```diff
@@ -37,21 +38,21 @@ func WithLogger(l logging.Logger) ClientOption {
...
-func WithFetcher(f ImageFetcher) ClientOption {
+func WithFetcher(f internal.ImageFetcher) ClientOption {
 	return func(c *Client) {
 		c.imageFetcher = f
 	}
 }
 
 // WithDownloader supply your own downloader.
-func WithDownloader(d Downloader) ClientOption {
+func WithDownloader(d internal.Downloader) ClientOption {
```
jromero: A general rule should be that anything _used_ by a public API cannot reference an internal component. This is an example of where that is violated. Another example would be returning an internal object/interface from a public method.

ameyer-pivotal: I agree, and missed this. I also missed that the package here ended up as `internal`, which I believe is not recommended.

ameyer-pivotal: Fixed.

create_builder.go
```diff
@@ -8,15 +8,16 @@ import (
...
+	"github.com/buildpack/pack/internal/builder"
+	"github.com/buildpack/pack/internal/dist"
+	"github.com/buildpack/pack/internal/image"
+	"github.com/buildpack/pack/internal/style"
 )
 
 type CreateBuilderOptions struct {
 	BuilderName   string
-	BuilderConfig builder.Config
+	BuilderConfig builder2.Config
```
jromero: :| Not a fan of `builder2`. Feel like if we are going to be running into this issue more frequently we should have a standard way of differentiating them.

ameyer-pivotal: Agreed. That's definitely not intentional.

internal/buildpackage/buildpackage.go
```diff
@@ -0,0 +1,18 @@
+package buildpackage
+
+import (
+	dist2 "github.com/buildpack/pack/internal/dist"
```
jromero: Why is this `dist2`?

internal/commands/inspect_builder.go
```diff
@@ -12,12 +12,12 @@ import (
...
 
 	"github.com/buildpack/pack"
-	"github.com/buildpack/pack/api"
-	"github.com/buildpack/pack/builder"
-	"github.com/buildpack/pack/config"
-	"github.com/buildpack/pack/dist"
+	"github.com/buildpack/pack/internal/api"
+	"github.com/buildpack/pack/internal/builder"
+	"github.com/buildpack/pack/internal/config"
+	dist2 "github.com/buildpack/pack/internal/dist"
```
jromero: `dist2`?

interfaces.go
```diff
@@ -1,30 +1,7 @@
 package pack
```
jromero: The remnants here look out of place.

client.go
```diff
@@ -1,20 +1,37 @@
...
-	"github.com/buildpack/pack/config"
-	"github.com/buildpack/pack/image"
+	"github.com/buildpack/pack/internal/blob"
+	"github.com/buildpack/pack/internal/build"
+	"github.com/buildpack/pack/internal/config"
+	"github.com/buildpack/pack/internal/image"
 	"github.com/buildpack/pack/logging"
 )
 
+type ImageFetcher interface {
```
jromero: These are missing their mock generation comments.

internal/buildpackage/buildpackage.go
ameyer-pivotal: Something automatic by the IDE. Will fix. Good catch!

internal/commands/inspect_builder.go
ameyer-pivotal: Same as other comment, will fix.

client.go
ameyer-pivotal: They're in `interfaces.go`. Happy to fix, though this was an issue before this PR

interfaces.go
ameyer-pivotal: See response in `client.go`.

ameyer-pivotal: Agreed it's sort of weird to have them here, and I'm happy to disperse them to their homes.

ameyer-pivotal: I think these were grouped here for convenience, but now that we have `make generate`, there's no need.

PR #377, opened 2019-11-08
Typed Images

ameyer-pivotal:
_WIP: acceptance tests pass (needs unit test work)_

This PR adds the concept of typed images. Essentially, a typed image is a lightweight abstraction of a specific kind of image (e.g. `stack.Image`, `stack.BuildImage`, `stack.RunImage`, `builder.Image`, `buildpackage.Image`), which wraps an underlying "raw image" (currently an `imgutil.Image`).

Typed images provide specific methods that are relevant for the given image type. For instance, any `stack.Image` knows how to read the stack ID and mixins off of the image, via `StackID()` and `Mixins()` methods, respectively.

Each new interface interface relates to others in the same way that the underlying images do (due to having labels, layers, etc. in common), allowing for intuitive use in code. For instance:

- A `stack.BuildImage` is a `stack.Image`
- A `stack.RunImage` is a `stack.Image`
- A `builder.Image` is a `stack.BuildImage`

### What about Builders? (In the design pattern sense)
Our code contains two types that follow the "builder" design pattern: `builder.Builder` and `buildpackage.Builder`. It should be clarified that a typed image is *not* responsible for the same behavior that a builder is (though much of their current work can be moved into the typed image instead). The typed images are about convenient domain-centered image manipulation (somewhat "dumb" data access objects), while the builders are responsible for orchestrating which manipulations happen, ensuring the validity of such manipulations, and for saving the final result.

### Overall gains
- Simplifies the current builders (again, "builder" in the design pattern sense) due to separation of concerns
- Creates "strongly-typed" images in a world of generic `imgutil.Image`s (analogous to using enums instead of strings in a coding project)
- Clarifies the responsibilities of each domain (buildpackages, builders, stacks)
- Decoupling will eventually allow easier removal of `imgutil` when we migrate to GGCR (there will be far fewer places to swap out `imgutil.Image` with its replacement)

### Considerations/Plans
These can be done as part of this PR or as separate PRs which evolve the typed images:
- Add a typed image for the app image (`app.Image` perhaps)
- Move more manipulation logic into typed images (e.g. add an `AddBuildpack()` method to `builder.Image` and `buildpackage.Image`)
- Fix unit tests (will be part of this PR)


image/image.go
```diff
@@ -0,0 +1,68 @@
...
+	"encoding/json"
+
+	"github.com/buildpack/imgutil"
+	"github.com/pkg/errors"
+
+	"github.com/buildpack/pack/style"
+)
+
+type Image interface {
+	Raw() imgutil.Image
```
jromero: This seems like a code smell. Seems like we should be able to use **embedding** (analogous to delegate pattern in OOP) to effectively adhere to the `imgutil.Image` interface?

For example:
```go
type Image interface {
	imgutil.Image
}

type image struct {
	imgutil.Image
}
```

Ref: https://yourbasic.org/golang/inheritance-object-oriented/

jromero: Seems like you are using embedding in other places like `stack.Image`. Is there a limitation that I'm not seeing here?

builder/builder.go
```diff
@@ -123,22 +98,23 @@ func constructBuilder(img imgutil.Image, newName string, metadata Metadata) (*Bu
 		platformAPIVersion = metadata.Lifecycle.API.PlatformVersion
 	}
 
+	// TODO: When metadata groups are deprecated, use `builderImage.Order()` instead
```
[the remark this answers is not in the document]
ameyer-pivotal: Great idea!

image/image.go
ameyer-pivotal: I suppose you could, but I wanted to limit how much these wrappers can do at their top-level interface, to keep them simple and focused. `Raw()` just gives a way to do what you need to in a pinch (when the top-level interface doesn't suffice).

My hope is that eventually, `Raw()` will go away completely, once the pertinent manipulations are exposed as appropriate methods on each wrapper type.

ameyer-pivotal: Removal of `Raw()` is something we might even be able to pull off in this PR

jromero: Although I agree with the goals of the future I think that the fact that `Raw` is currently there would mean that it would break the API anyway in the future yet if we embed the `imgutil.Image` we get a much cleaner API for how things currently exist.

ameyer-pivotal: Keep in mind PR #375. I wouldn't think these images would be part of the exported API, at least not initially.

jromero:
As discussed, closing for now.

PR #380, opened 2019-11-11
Update lifecycle

jromero:
* Update lifecycle.
    * Most adaptations around `BOM` metadata and passing credential information to the lifecycle.
* Inherits resolution to an issue with [differing compression configuration generating differing layer digest](https://github.com/buildpack/pack/issues/251#issuecomment-543388120).


testhelpers/testhelpers.go
```diff
@@ -228,31 +232,72 @@ func Eventually(t *testing.T, test func() bool, every time.Duration, timeout tim
...
-func CreateImageOnLocal(t *testing.T, dockerCli *client.Client, repoName, dockerFile string) {
+func CreateImage(t *testing.T, dockerCli *client.Client, repoName, dockerFile string) {
 	t.Helper()
-	ctx := context.Background()
-
 	buildContext, err := archive.CreateSingleFileTarReader("Dockerfile", dockerFile)
 	AssertNil(t, err)
 
-	res, err := dockerCli.ImageBuild(ctx, buildContext, dockertypes.ImageBuildOptions{
+	err = checkImageBuild(dockerCli.ImageBuild(context.Background(), buildContext, dockertypes.ImageBuildOptions{
```
simonjjones: I'm not sure the approach of one-lining the return values of `ImageBuild` into the args for `checkImageBuild` is the most readable. Given that we then have to handle the potential `error` from `ImageBuild` in the check method anyway, would it be clearer to assign the returned values, perform the error checking and then call `checkImageBuild` with just the response?

jromero: Personally, I think it's more readable (once you recognize that the language feature exists) but I don't have a strong opinion. Will split into two assertions where used.

ekcasey: Checking the error first separately feel more idiomatic to me.

PR #384, opened 2019-11-14
Support combined exporter/cacher and reversed analyzer/restorer.

### LINKED ISSUE #372: Add support for Platform API 0.2
## Motivation
An [upcoming change to the lifecycle](https://github.com/buildpack/lifecycle/pull/205) will break the current Platform API. This will effectively raise the Platform API version from 0.1 to 0.2. In order to mitigate, the breaking change we'd like to release `pack` with compatibility for both Platform API versions _before_ the change is released. 

### Platform API 0.1
**Given**  I have a builder `builder/platform-01` with a lifecycle v0.5.0 with descriptor:
```
[api]
 platform = "0.1"

...
```
**When** I run `pack build app/platform-01 --builder builder/platform-01 -p <app_dir>`
**Then** build is successful
**And** I can run the application via `docker run -it --rm app/platform-01`

### Platform API 0.2
**Given**  I have a builder `builder/platform-02` with a lifecycle based on [PR](https://github.com/buildpack/lifecycle/pull/205) with descriptor:
```
[api]
 platform = "0.2"

...
```
**When** I run `pack build app/platform-02 --builder builder/platform-02 -p <app_dir>`
**Then** build is successful
**And** I can run the application via `docker run -it --rm app/platform-02`

lukasberger:
* Update Platform API Version to 0.2
* Extend the incompatible platform version check to include a minimum-supported version.
* When encountering lifecycle with platform API version 0.2 and above:
  * Do not check that cacher exists when creating a new builder.
  * Switch analyzer and restorer build phases.
  * Remove cacher build phases.
  * Change -path cache flag name to -cache-dir.
* Related lifecycle PRs:
  * https://github.com/buildpack/lifecycle/pull/205
  * https://github.com/buildpack/lifecycle/pull/207

Resolves #372


lukasberger:
This is an attempt to add "expocacher" support for pack while maintaining compatibility with the current platform API version. Any suggestions on how this can be made clearer appreciated!

internal/build/build.go
```diff
@@ -10,27 +10,33 @@ import (
...
 	"github.com/buildpack/pack/internal/cache"
 	"github.com/buildpack/pack/internal/style"
 	"github.com/buildpack/pack/logging"
 )
 
-// PlatformAPIVersion is the current Platform API Version supported by this version of pack.
-const PlatformAPIVersion = "0.1"
+const (
+	// MinPlatformAPIVersion is the minimum Platform API Version supported by this version of pack.
+	MinPlatformAPIVersion = "0.1"
```
jromero: `Min` here seems to misconstrued how platform versions are supported. At least for `pre-stable`, each version is independent (not semver). [Here's a little more detail](https://github.com/buildpack/spec/blob/master/buildpack.md#buildpacktoml-toml) although only mentioned for Buildpack API it also should apply to Platform API version.

A better representation could be:
```
var SupportedPlatformAPIVersions := ["0.1", "0.2"]
```

jromero:
Resolves #372

build.go
```diff
@@ -121,10 +121,12 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		lcPlatformAPIVersion = descriptor.API.PlatformVersion
 	}
 
-	if !api.MustParse(build.PlatformAPIVersion).SupportsVersion(lcPlatformAPIVersion) {
+	if lcPlatformAPIVersion.Compare(api.MustParse(build.MinPlatformAPIVersion)) == -1 ||
+		lcPlatformAPIVersion.Compare(api.MustParse(build.PlatformAPIVersion)) == 1 {
 		return errors.Errorf(
-			"pack %s (Platform API version %s) is incompatible with builder %s (Platform API version %s)",
+			"pack %s (Platform API versions %s and %s) is incompatible with builder %s (Platform API version %s)",
```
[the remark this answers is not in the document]
lukasberger: I felt the same way. I like your suggestion, updated.

internal/build/build.go
lukasberger: Good point, updated.

PR #385, opened 2019-11-15
Refactor/stringset

ameyer-pivotal:
Adds a `stringset` package to simplify some set-related logic, especially for mixins

- Also added a few small auxiliary refactors in tests, etc.


internal/stack/mixins.go
```diff
@@ -5,67 +5,52 @@ import (
...
-func mixinSet(mixins []string, imageName string, run bool) (map[string]interface{}, error) {
-	set := map[string]interface{}{}
-	invalidPrefix := "run"
-	excludePrefix := "build"
-	if run {
-		invalidPrefix = "build"
-		excludePrefix = "run"
-	}
-
+// TODO: test
```
jromero: TODO

PR #389, opened 2019-11-20
Remove fake logger

jromero:
Reason: Other implementations satisfy the purpose of the fake logger. It's thereby
simply more code to maintain without added value.


micahyoung (approved):
Looks good. Approved.

aliamish:
I agree with this. If the existing implementations already cover the purpose of the fake logger, keeping another implementation would add unnecessary code to maintain.

PR #396, opened 2019-11-25
Change `pack version` to format `<semver>+<gitsha1, first 7 chars>`

### LINKED ISSUE #390: Update pack version format to semver
Based on changes made for https://github.com/buildpack/lifecycle/issues/143, we'd like to have the pack version being displayed as semver.

### Release Candidates
**Given** I have built a `pack` binary using `VERSION=0.4.0 make build`
**When** I run `./pack version`
**Then** The output should report the given version and git sha as `<VERSION>+<first 7 chars from git commit>`

Example:
```
$ ./out/pack version
0.6.0+f49a44e
```

### Local Development
**Given** I have built a `pack` binary locally using `make build`
**When** I run `./pack version`
**Then** The output should report the version as `0.0.0+<first 7 chars from git commit>`

Example:
```
$ ./out/pack version
0.0.0+f49a44e
```

micahyoung:
- `make build` in a git repo at sha abc1234xyz... will yield `0.0.0+abc1234`
- `PACK_VERSION=1.2.3 make build` in a git repo at sha abc1234xyz... will yield `1.2.3+abc1234`
- `PACK_VERSION=1.2.3 PACK_GITSHA1=abc1234xyz make build` will always yield `1.2.3+abc1234xyz`
- When not in git repo or when `PACK_GITSHA1= make build`, version will be `<semver>` only
- Implement via Makefile for the same reason `PACK_VERSION` is set in Makefile - it is not intrinsic to pack codebase, rather comes from the version control system of the codebase and also should be overridable
- Complex Makefile conditional/strip syntax is meant to handle messy whitespace if git fails or is not present (see `ifeq` section): https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html)

Resolves #390

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,18 @@
 GOCMD?=go
 GOENV=CGO_ENABLED=0
 GOFLAGS?=-mod=vendor
-PACK_VERSION?=dev-$(shell date +%Y-%m-%d-%H:%M:%S)
+PACK_VERSION?=0.0.0
+PACK_GITSHA1?=$(shell git rev-parse --short=7 HEAD)
 PACK_BIN?=pack
 PACKAGE_BASE=github.com/buildpack/pack
 PACKAGES:=$(shell $(GOCMD) list ./... | grep -v /testdata/)
 SRC:=$(shell find . -type f -name '*.go' -not -path "*/vendor/*")
 ARCHIVE_NAME=pack-$(PACK_VERSION)
+
+# append git sha to version, if not-empty
+ifneq ($(strip ${PACK_GITSHA1}),)
+PACK_VERSION:=${PACK_VERSION}+${PACK_GITSHA1}
+endif
 
 export GOFLAGS:=$(GOFLAGS)
 


jromero:
Hyphen denotes pre-release versions. Plus sign denotes metadata. The git sha should be metadata hence should use a plus sign.

micahyoung:
Ugh, swing and a miss... Fixed now.

micahyoung:
@jromero Also, looking deeper into this, lifecycle uses an [non-overridable variable called `SCM_COMMIT`](https://github.com/buildpack/lifecycle/blob/c9f1056b97126211a68f20dbe2accaff9be1411b/Makefile#L9). Would we want to use that syntax here instead and/or otherwise change pack and lifecycle Makefiles to be more consistent?

ameyer-pivotal:
> @jromero Also, looking deeper into this, lifecycle uses an [non-overridable variable called `SCM_COMMIT`](https://github.com/buildpack/lifecycle/blob/c9f1056b97126211a68f20dbe2accaff9be1411b/Makefile#L9). Would we want to use that syntax here instead and/or otherwise change pack and lifecycle Makefiles to be more consistent?

@jromero @micahyoung Agree that we should be consistent. I tend to lean toward the way lifecycle does it. I don't see a use case for allowing the var to be overridable.

Makefile
```diff
@@ -1,13 +1,19 @@
...
-PACK_VERSION?=dev-$(shell date +%Y-%m-%d-%H:%M:%S)
+PACK_VERSION?=0.0.0
+PACK_GITSHA1?=$(shell git rev-parse --short=7 HEAD)
 PACK_BIN?=pack
 PACKAGE_BASE=github.com/buildpack/pack
 PACKAGES:=$(shell $(GOCMD) list ./... | grep -v /testdata/)
 SRC:=$(shell find . -type f -name '*.go' -not -path "*/vendor/*")
 ARCHIVE_NAME=pack-$(PACK_VERSION)
 
+# append git sha to version, if not-empty
```
ameyer-pivotal: Do you need to explicitly skip this if the `PACK_VERSION` was already overridden (perhaps by someone making a local version and wanting to give it a special value)? Otherwise it seems this might stomp on it regardless.

Makefile
```diff
@@ -1,13 +1,19 @@
 GOCMD?=go
 GOENV=CGO_ENABLED=0
 GOFLAGS?=-mod=vendor
-PACK_VERSION?=dev-$(shell date +%Y-%m-%d-%H:%M:%S)
+PACK_VERSION?=0.0.0
+PACK_GITSHA1?=$(shell git rev-parse --short=7 HEAD)
```
ameyer-pivotal: Should be consistent with lifecycle repo, per comments

natalieparellano:
We've adjusted our strategy slightly, this is now captured in #572 and #574 building on the work in this PR. Therefore, I think this one could be closed.

PR #397, opened 2019-11-26
Include processes in inspect-image output when available

simonjjones:
When using lifecycle version 0.6.0 or greater to build an image, the output of `pack inspect-image` will now include a list of processes defined by a buildpack. The output will indicate which process, if any, is the default.

Calls to `pack.Client`'s `InspectImage` command will include an additional `Processes` element of `ProcessDetails` including a `DefaultProcess` if there is one, as well as a slice of `OtherProcesses`.

### DIFF
--- a/internal/commands/inspect_image.go
+++ b/internal/commands/inspect_image.go
@@ -3,11 +3,15 @@ package commands
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
+	"strings"
 	"text/tabwriter"
 	"text/template"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
+	"github.com/buildpack/lifecycle"
 
 	"github.com/buildpack/pack"
 	"github.com/buildpack/pack/internal/config"
@@ -76,6 +80,7 @@ func logBOM(remote *pack.ImageInfo, local *pack.ImageInfo, logger logging.Logger
 func logDetails(remote *pack.ImageInfo, local *pack.ImageInfo, cfg config.Config, logger logging.Logger) {
 	imgTpl := template.Must(template.New("runImages").Parse(runImagesTemplate))
 	imgTpl = template.Must(imgTpl.New("buildpacks").Parse(buildpacksTemplate))
+	imgTpl = template.Must(imgTpl.New("processes").Parse(processesTemplate))
 	imgTpl = template.Must(imgTpl.New("image").Parse(imageTemplate))
 	remoteOutput, err := inspectImageOutput(remote, cfg, imgTpl)
 	if err != nil {
@@ -91,6 +96,13 @@ func logDetails(remote *pack.ImageInfo, local *pack.ImageInfo, cfg config.Config
 	}
 }
 
+type process struct {
+	PType   string
+	Shell   string
+	Command string
+	Args    string
+}
+
 func inspectImageOutput(
 	info *pack.ImageInfo,
 	cfg config.Config,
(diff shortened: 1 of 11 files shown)


simonjjones:
Resolves https://github.com/buildpack/pack/issues/370

internal/commands/inspect_image.go
```diff
@@ -101,20 +113,56 @@ func inspectImageOutput(
 	}
 	var buf bytes.Buffer
 	localMirrors := getLocalMirrors(info.Stack.RunImage.Image, cfg)
+	processes := displayProcesses(info.Processes)
 	tw := tabwriter.NewWriter(&buf, 0, 0, 8, ' ', 0)
 	defer tw.Flush()
 	if err := tpl.Execute(tw, &struct {
```
ameyer-pivotal: This anonymous struct could be a map and it might shorten this code a bit.

internal/commands/inspect_image.go
```diff
@@ -91,6 +96,13 @@ func logDetails(remote *pack.ImageInfo, local *pack.ImageInfo, cfg config.Config
 	}
 }
 
+type process struct {
+	PType   string
```
ameyer-pivotal: Why not just `Type` here? `lifecycle.Process` already does it.

internal/commands/inspect_image.go
```diff
@@ -101,20 +113,56 @@ func inspectImageOutput(
...
 		info,
 		localMirrors,
+		processes,
 	}); err != nil {
 		return "", err
 	}
 	return buf.String(), nil
 }
 
+func displayProcess(p lifecycle.Process, d bool) process {
```
ameyer-pivotal: This function name makes it sound like it should be printing (or generating a string). I guess it depends on whether you read "display" as a verb or an adjective. Maybe something like `renderProcess`? Though you might be able to make the same argument with that name (it's the best I could come up with on the spot).

Also, I feel like primitive parameters should have more descriptive names (the `lifecycle.Process` param is fine, but the meaning of the `bool` is hard to infer).

internal/commands/inspect_image_test.go
```diff
@@ -215,6 +247,35 @@ func testInspectImageCommand(t *testing.T, when spec.G, it spec.S) {
   test.bp.two.local\s+2.0.0`)
 				})
 
+				it("displays process info for local and remote", func() {
+					h.AssertNil(t, command.Execute())
+					h.AssertContainsMatch(t, remoteOutput(outBuf), `Processes:
+  TYPE\s+SHELL\s+COMMAND\s+ARGS
```
ameyer-pivotal: I think we should use regexes in test assertions sparingly. If there's any way to use a raw string (or perhaps a fixture), I think that route will be easier to maintain going forward.

internal/commands/inspect_image.go
simonjjones: type is a reserved word in golang

internal/commands/inspect_image_test.go
simonjjones: Due to the tabwriter, the amount of whitespace is unpredictable. We could get there with trial and error, but it would create a relatively fragile test in that if any of the inputs changed the header will also change. We'd also be essentially testing behaviour of the tabwriter. This way, we just care that their separated

internal/commands/inspect_image.go
ameyer-pivotal: I had assumed that was the reason, but I saw that `lifecycle.Process` has a `Type` field.

internal/commands/inspect_image_test.go
ameyer-pivotal: ✅

PR #399, opened 2019-11-29
[WIP] Generate Markdown documentation

Zanadar:
**Context**:
I was looking for common documentation of `pack` commands to point somebody on our slack to and I couldn't find them online. I realized that cobra can generate markdown of helptext for use in documentation. Hugo uses to document its commands online (and then publishes them online with ...Hugo...). Anyway I thought this could be a nice way to get generated documentation for pack commands online that we could include on `buildpacks.io` (or wherever).

I also factored this code such that we open up the possibility of using the other generators that `cobra` provides: man pages and bash-completion (although the latter looks like its already done).
All this is totally scratch, WIP code, but just wanted to get the idea out there.

https://github.com/buildpack/pack/pull/399/commits/6e5657839843efd9a9eff1f45a0ba0add3e15806 has examples of the markdown that's generated.


jromero:
Really cool feature! Here are a couple initial thoughts.

1. This could be hosted under a `pack` section in [References](https://buildpacks.io/docs/reference/)
2. We probably don't want the generated docs to live in the repo. We had similar conversations about the [docs](https://github.com/buildpack/docs) (website) and we ultimately landed on using CI to generate the files. Same would be applied here as they'll likely be living on the website itself.

Zanadar:
Glad you're open to it. Agree with 1 and 2. Just added the MD here as an example. We can add  headers, etc. that let us do custom generation of markdown for the docs in cobra.

Zanadar:
Closing for now. Will followup later.

PR #406, opened 2019-12-06
inspect-builder throws error with missing remote and local builder

### LINKED ISSUE #386: pack inspect-builder <non-existent-tag> does not error
No error when I try to inspect a non-existant or unreachable builder:
```
± zm |master ✓| → pack inspect-builder cloudfoundry/cnb:0.0.40-bionic
Inspecting builder: cloudfoundry/cnb:0.0.40-bionic

REMOTE:
(not present)

LOCAL:
(not present)

```

ktpv:

### DIFF
--- a/internal/commands/inspect_builder.go
+++ b/internal/commands/inspect_builder.go
@@ -36,28 +36,33 @@ func InspectBuilder(logger logging.Logger, cfg config.Config, client PackClient)
 				imageName = args[0]
 			}
 
+			presentRemote, remoteOutput, remoteWarnings, remoteErr := inspectBuilderOutput(client, cfg, imageName, false)
+			presentLocal, localOutput, localWarnings, localErr := inspectBuilderOutput(client, cfg, imageName, true)
+
+			if !presentRemote && !presentLocal {
+				return errors.New(fmt.Sprintf("Unable to find builder '%s' locally or remotely.\n", imageName))
+			}
+
 			if imageName == cfg.DefaultBuilder {
 				logger.Infof("Inspecting default builder: %s\n", style.Symbol(imageName))
 			} else {
 				logger.Infof("Inspecting builder: %s\n", style.Symbol(imageName))
 			}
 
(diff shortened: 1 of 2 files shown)


internal/commands/inspect_builder.go
```diff
@@ -36,28 +36,33 @@ func InspectBuilder(logger logging.Logger, cfg config.Config, client PackClient)
 				imageName = args[0]
 			}
 
+			remoteOutput, remoteWarnings, remoteErr := inspectBuilderOutput(client, cfg, imageName, false)
+			localOutput, localWarnings, localErr := inspectBuilderOutput(client, cfg, imageName, true)
+
+			if remoteOutput == "(not present)" && localOutput == "(not present)" {
```
jromero: I would feel more comfortable with a string comparison if it was a constant used inside the `inspectBuilderOutput` function and in this statement.

Alternatively, an additional return value similar to how maps returns a boolean value (`ok`) if it was found would work here.

No strong opinion either way but something should be changed to not just do arbitrary freeform string comparisons IMO.

internal/commands/inspect_builder_test.go
```diff
@@ -60,14 +60,12 @@ func testInspectBuilderCommand(t *testing.T, when spec.G, it spec.S) {
 
 	when("#Get", func() {
 		when("image cannot be found", func() {
-			it("logs 'Not present'", func() {
+			it("logs 'ERROR Unable'", func() {
```
jromero: This should read `it("errors when no image is found")`.

jromero: We also need at least one test (preferably two) where we test when only one of the two is returned and ensure it doesn't error.

ktpv: Done and done after reorganizing test data.

internal/commands/inspect_builder.go
ktpv: Did the 2nd option and added additional return value 'present'

PR #408, opened 2019-12-09
removes dangling images in acceptance tests

### LINKED ISSUE #402: Pack acceptance tests leak docker images
When running acceptance tests on a workstation we see two additional images leaked per run of the test suite:

```
$  LIFECYCLE_PATH=/Users/pivotal/Downloads/lifecycle-v0.6.0-build.7+linux.x86-64.tgz make acceptance
...
PASS | github.com/buildpack/pack/acceptance 291.178s
$  docker images | wc -l
     154
$  LIFECYCLE_PATH=/Users/pivotal/Downloads/lifecycle-v0.6.0-build.7+linux.x86-64.tgz make acceptance
...
PASS | github.com/buildpack/pack/acceptance 275.289s
$  docker images | wc -l
     156
```

We should track down which images are being leaked and make sure we're cleaning them up

ktpv:
fixes #402

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -256,7 +256,7 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 				})
 
 				it.After(func() {
-					dockerCli.ImageRemove(context.TODO(), repoName, dockertypes.ImageRemoveOptions{Force: true, PruneChildren: true})
+					h.DockerRmi(dockerCli, repoName)
 					ref, err := name.ParseReference(repoName, name.WeakValidation)
 					h.AssertNil(t, err)
 					cacheImage := cache.NewImageCache(ref, dockerCli)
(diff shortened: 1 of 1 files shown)


acceptance/acceptance_test.go
```diff
@@ -1131,6 +1132,7 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 
 			packageName := createPackageLocally(aggregatePackageToml)
 			assertImageExistsLocally(packageName)
+			defer h.DockerRmi(dockerCli, nestedPackageName, packageName)
```
jromero: We should break this into two `defer` statements. One for each respective images. The precaution is for the fact that this `defer` will not register if any of the assertions fail. We would want to `defer` the clean up as soon as the image is created.

ktpv: Changes made.

acceptance/acceptance_test.go
```diff
@@ -1131,6 +1132,9 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 
 			packageName := createPackageLocally(aggregatePackageToml)
 			assertImageExistsLocally(packageName)
+
+			defer h.DockerRmi(dockerCli, nestedPackageName)
```
jromero: Minor: Sorry if I wasn't clear enough. 

This line should be right after line 1127, where the `nestedPackageName` was created.

The line below should be right after line 1133, where `packageName` was created.

This would ensure that defer is registered as early as possible and execute even if other assertions fail.

ktpv: Not a problem.  I found additional images like the one above and am in the process of fixing them.

PR #415, opened 2019-12-13
inspect-image only deserializes fields it requires

ekcasey:
* Avoids breaking change due to slice layer metadata


ekcasey:
Without this PR changes from https://github.com/buildpacks/lifecycle/pull/210 will break `pack inspect-image`. @elbandito and decided to deserialize just a subset of fields so the code will be compatible with both metadata schemas.

PR #420, opened 2019-12-16
Update project.toml format

### LINKED ISSUE #409: Update package.toml format
Based on outcome of #376, we are changing the structure of the project.toml format.

Additionally, stacks should be computed as the intersection of all stacks declared in the included buildpacks. There MUST be at least one compatible stack.

jromero:
- Compute supported stacks based on buildpack and it's dependencies

Resolves #409


internal/buildpackage/builder_test.go
```diff
@@ -52,107 +52,109 @@ func testPackageBuilder(t *testing.T, when spec.G, it spec.S) {
...
+					}, 0644)
+					h.AssertNil(t, err)
 
-			when("no stacks are set", func() {
-				it("returns error", func() {
-					_, err := subject.Save(fakePackageImage.Name(), false)
-					h.AssertError(t, err, "must specify at least one supported stack")
+					subject.SetBuildpack(buildpack)
+					_, err = subject.Save("some/package", false)
+					h.AssertError(t, err, "buildpack 'bp.1.id@bp.1.version' must at least support one stack")
```
ameyer-pivotal: Slight suggestion: "must support at least one stack"

internal/stack/aggregate.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// 	stacksA = [{ID: "stack1", mixins: ["build:mixinA"]}}]
+// 	stacksB = [{ID: "stack2", mixins: ["mixinA", "run:mixinB"]}}]
+// 	result = []
+//
+func Aggregate(stacksA []dist.Stack, stacksB []dist.Stack) []dist.Stack {
+	set := map[string][]string{}
+
+	for _, s := range stacksA {
+		set[s.ID] = s.Mixins
```
ameyer-pivotal: This wouldn't allow two stack entries with the same ID in `stacksA`. For instance:

```
stacksA = [
  {ID: "stack1", mixins: ["build:mixinA", "mixinB", "run:mixinC"]}},
  {ID: "stack1", mixins: ["build:mixinX", "mixinY", "run:mixinZ"]}}
]
```

The above is legal, right @sclevine?

internal/stack/aggregate.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// 	stacksA = [{ID: "stack1", mixins: ["build:mixinA"]}}, {ID: "stack2", mixins: ["mixinA"]}}]
+// 	stacksB = [{ID: "stack2", mixins: ["mixinA", "run:mixinB"]}}]
+// 	result = [{ID: "stack2", mixins: ["mixinA", "run:mixinB"]}}]
+//
+// 	stacksA = [{ID: "stack1", mixins: ["build:mixinA"]}}]
+// 	stacksB = [{ID: "stack2", mixins: ["mixinA", "run:mixinB"]}}]
+// 	result = []
+//
+func Aggregate(stacksA []dist.Stack, stacksB []dist.Stack) []dist.Stack {
```
ameyer-pivotal: Maybe rename to something like `MergeCompatibleStacks`? I wonder if it would also be good to rename the variables to shed light on what is compatible with what. For instance:
`stacksA` => `compatStacks`
`stacksB` => `buildpackStacks` or `newStacks`

or something similar

ameyer-pivotal: Does this need to be exported?

sclevine: This is used in `internal/buildpackage`, so I think it does need to be exported.

Naming wise, the intent is clear to me personally, especially given that the package is already called `stack`. `stack.MergeCompatibleStacks` could be considered a naming stutter.

internal/stack/aggregate.go
sclevine: It isn't legal for a buildpack to specify the same stack ID twice with multiple sets of mixins. So I think this is okay.

ameyer-pivotal: OK, sounds good!

internal/stack/aggregate.go
ameyer-pivotal: That makes sense. I missed the package name. Perhaps `MergeCompatible` then?

ameyer-pivotal: Also, what I meant by my export comment was that I think this is buildpackage-specific logic, and it should be in that package (and unexported).

internal/stack/aggregate.go
```diff
@@ -0,0 +1,55 @@
+package stack
+
+import (
+	"sort"
+
+	"github.com/buildpacks/pack/internal/dist"
+	"github.com/buildpacks/pack/internal/stringset"
+)
+
+// Aggregate aggregates two set of stacks into a merged set of stacks where compatibility exists.
```
ameyer-pivotal: Maybe:
```
... determines the allowable set of stacks that a combination of buildpacks may run on, given each buildpack's set of stacks. Compatibility between the two sets of buildpack stacks is defined by the following rules:
  
  1) The stack must be supported by both buildpacks. That is, any resulting stack ID must appear in both input sets.

  2) For each supported stack ID, all required mixins for all buildpacks must be provided by the result. That is, mixins for the stack ID in both input sets are unioned.

Examples:
...
```

PR #421, opened 2019-12-17
Add support for project.toml

jkutner:
This PR adds initial support for the Project Descriptor defined in [Project Descriptor RFC](https://github.com/buildpacks/rfcs/pull/32)

The `project.toml` file path is specified via the `--descriptor` flag (defaults to `project.toml`).  The project file is then parsed to support the following features of the `project.toml` RFC:

* **[project]** for specifying `name`, `id`, and `version`
* **[project.licenses]** for specifying license `type` or `uri` information
* **[[build.buildpacks]]** to specify the buildpacks that the platform should use on the repo.
* **[[build.env]]** for setting environment variables at build time

The plan is to provide additional features in later PR(s), such as the following:
* **[build.include]** A list of files to include in the build (while excluding everything else
* **[build.exclude]** A list of files to exclude from the build (while including everything else)

### DIFF
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -1,8 +1,10 @@
 package commands
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -10,21 +12,24 @@ import (
 
 	"github.com/buildpacks/pack"
 	"github.com/buildpacks/pack/internal/config"
+	"github.com/buildpacks/pack/internal/paths"
+	"github.com/buildpacks/pack/internal/project"
 	"github.com/buildpacks/pack/internal/style"
 	"github.com/buildpacks/pack/logging"
 )
 
 type BuildFlags struct {
-	AppPath    string
-	Builder    string
-	RunImage   string
-	Env        []string
-	EnvFiles   []string
-	Publish    bool
-	NoPull     bool
-	ClearCache bool
-	Buildpacks []string
-	Network    string
+	AppPath        string
+	Builder        string
+	RunImage       string
+	Env            []string
+	EnvFiles       []string
+	Publish        bool
+	NoPull         bool
+	ClearCache     bool
+	Buildpacks     []string
+	Network        string
+	DescriptorPath string
 }
 
 func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cobra.Command {
@@ -41,10 +46,39 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 				suggestSettingBuilder(logger, packClient)
 				return MakeSoftError()
 			}
-			env, err := parseEnv(flags.EnvFiles, flags.Env)
+
+			descriptor, actualDescriptorPath, err := parseProjectToml(flags.AppPath, flags.DescriptorPath)
 			if err != nil {
 				return err
 			}
+			if actualDescriptorPath != "" {
+				logger.Debugf("Using project descriptor located at '%s'", actualDescriptorPath)
+			}
+
+			env, err := parseEnv(descriptor, flags.EnvFiles, flags.Env)
+			if err != nil {
+				return err
+			}
+
+			buildpacks := flags.Buildpacks
+			if len(buildpacks) == 0 {
+				buildpacks = []string{}
+				projectDescriptorDir := filepath.Dir(actualDescriptorPath)
+				for _, bp := range descriptor.Build.Buildpacks {
+					if len(bp.URI) == 0 {
+						// there are several places through out the pack code where the "id@version" format is used.
+						// we should probably central this, but it's not clear where it belongs
+						buildpacks = append(buildpacks, fmt.Sprintf("%s@%s", bp.ID, bp.Version))
+					} else {
+						uri, err := paths.ToAbsolute(bp.URI, projectDescriptorDir)
+						if err != nil {
+							return err
+						}
+						buildpacks = append(buildpacks, uri)
+					}
+				}
+			}
+
 			if err := packClient.Build(ctx, pack.BuildOptions{
 				AppPath:           flags.AppPath,
 				Builder:           flags.Builder,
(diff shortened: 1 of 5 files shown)


internal/commands/build.go
```diff
@@ -41,10 +45,31 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
...
 			if err != nil {
 				return err
 			}
+
+			buildpacks := flags.Buildpacks
+			if len(buildpacks) == 0 {
+				buildpacks = []string{}
+				for _, bp := range descriptor.Build.Buildpacks {
+					if len(bp.Uri) == 0 {
+						// there are several places through out the pack code where the "id@version" format is used.
```
jromero: We have this as part of a [`BuildpackInfo#FullName()`](https://github.com/buildpacks/pack/blob/master/internal/dist/buildpack.go#L46)

jkutner: That's is where it constructs the string, but there isn't a place were it gets parsed. It didn't look like creating a `BuildpackInfo` in each place was straight forward so i decided not to deal with it in this PR (hence leaving the comment).

internal/project/project_test.go
```diff
@@ -0,0 +1,204 @@
...
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/BurntSushi/toml"
+)
+
+func TestDecodeSimple(t *testing.T) {
```
jromero: Think we should keep tests consistent in using `spec`.

elbandito: @jromero remind me, are we pretty much using `spec` across the board for all CNB sponsored repos?

elbandito: @jromero I went ahead and updated the test suite to be `spec` based.  Let me know if I missed anything else.  Thanks.

zmackie:
@jkutner Looks like this needs a rebase

elbandito:
@zmackie I went ahead and performed the rebase.  Is there any other outstanding items?

internal/commands/build.go
jromero: Theoretically, `project.Buildpack` could embed `dist.BuildpackInfo` and thereby inherit that functionality but I can see how that's not optimal since it would require this package to have a dependency on `dist`. There are additional benefits that come from it as well like ensuring there is only a single location for the definition of what "Buildpack Info" is. Additionally, we use this pattern in multiple places. For example:

https://github.com/buildpacks/pack/blob/589d9975ff9321aa4b9dae66f412708752b259be/internal/builder/metadata.go#L22-L25

jkutner: I don't have any strong opinions on the solution, but i do think it should be done in a separate line of work (hence the `TODO`).

jkutner: Oh, looks like I didn't actually put `TODO`, but you get the idea. It's not the goal of this PR to fix the `BuildpackInfo`.

internal/commands/build.go
```diff
@@ -128,3 +157,21 @@ func addEnvVar(env map[string]string, item string) map[string]string {
...
+	}
+
+	if _, err := os.Stat(fullDescriptorPath); !os.IsNotExist(err) {
+		return project.ReadProjectDescriptor(fullDescriptorPath)
+	}
+	return project.Descriptor{}, nil
+}
+
+func parseProjectToml(appPath string, descriptorPath string) (project.Descriptor, error) {
+	return readProjectDescriptor(appPath, descriptorPath)
```
jromero: This indirection seems unnecessary. The functionality of `readProjectDescriptor` could just live here.

elbandito: Good point!

jromero:
@jkutner / @elbandito 

I've approved it for the changes that are currently PR'd but it seems like the overall features of `project.toml` aren't implemented. eg. actually excluding or including app dir contents and adding labels.

What's the plan for those features? Can we add a description to this PR as to what's implemented and what's left?

jkutner:
@jromero this is the first step in implementing the [spec extension](https://github.com/buildpacks/spec/pull/70), and we'll introduce the rest piece by piece. I think exclude/include is the only missing feature that requires some behavior. We'll work on that in a separate PR.

elbandito:
@jromero I updated the PR description to be more explicit on the intentions of this PR, and what features remain for subsequent PRs.  Let me know if there's anything else.

jkutner:
> buildpack.uri is relative to working dir

I'm not sure I agree that it should be relative to `project.toml`. If you're using `-p` you're changing the context that the it runs in. This is the same as if you passed `--buildpack`, no? Why would we change or be different from that?

elbandito:
> **No error if file is not found**

Fixed

> **No indication error is related to project.toml**

Fixed

jromero:
> I'm not sure I agree that it should be relative to project.toml. If you're using -p you're changing the context that the it runs in. This is the same as if you passed --buildpack, no? Why would we change or be different from that?

The example I have in mind which works well thus far is `create-builder` and `build`.

```
pack build ... --buildpack <this-path-is-relative-to-pwd>
```

```
pack create-builder ... --builder-config <this-path-is-relative-to-pwd>
```

`builder.toml`:
```
[[buildpackls]]
uri = "<relative-path-to-this-file>"
```
See https://buildpacks.io/docs/reference/builder-config/

`package.toml`:
```
...
[[dependencies]]
uri = "<relative-path-to-this-file>"
```

jkutner:
@jromero i'm not sure i agree that what happens in those examples is preferable. 

What if you want to use a `project.toml` that sits outside of the repo (so you could decouple how an an app built from the app itself)?

I'm not going to die on this hill, but I think in this case (and probably the cases you describe) the value passed to the option should work the same way whether it's a CLI option or codified in a descriptor file.

elbandito:
@jromero I added changes for making the BPs `uri` paths relative to the location of the `project.toml` and a `debug()` log indicating to the user that a project descriptor file was found.

internal/commands/build.go
```diff
@@ -128,3 +166,20 @@ func addEnvVar(env map[string]string, item string) map[string]string {
 	}
 	return env
 }
+
+func parseProjectToml(descriptorPath string) (project.Descriptor, string, error) {
+	actualDescriptorPath := descriptorPath
+	if descriptorPath == "" {
+		actualDescriptorPath = "project.toml"
```
jromero: I suspect this needs to be relative to the app path.

elbandito: I must have removed it during a refactor.

PR #425, opened 2019-12-18
Fix: support meta-buildpack in packages

### LINKED ISSUE #289: Buildpackage layers label format
### Based on work from #251

**Given** I have buildpack `bp.one@1.2.3` with `buildpack.toml`:
```
[buildpack]
  id = "bp.one"
  version = "1.2.3"

[[order]]
[[order.group]]
  id = "bp.nested"
  version = "2.3.4"
```
**When** I create a buildpackage with the following configuration:
```
[default]
id = "bp.one"
version = "1.2.3"

[[buildpacks]]
uri = "https://example.com/bp.one_1.2.3.tgz"

[[packages]]
ref = "package/bp.nested_2.3.4"

[[stacks]]
id = "some.stack.id"
```
**Then** the buildpackage contains the following labels:
-  `io.buildpacks.buildpack.layers`:
```json
{
  "bp.one": {
    "1.2.3": {
      "layerDiffID": "8abc...",
      "order": [
        {"group": [{"id": "bp.nested", "version": "2.3.4"}]}
      ]
    }
  },
  "bp.nested": {
    "2.3.4": {
      "layerDiffID": "9acc..."
    }
  }
}
```

jromero:
Resolves #289

### DIFF
--- a/internal/buildpackage/builder.go
+++ b/internal/buildpackage/builder.go
@@ -59,6 +59,10 @@ func (p *PackageBuilder) Save(repoName string, publish bool) (imgutil.Image, err
 		}
 	}
 
+	if len(stacks) == 0 {
+		return nil, errors.Errorf("no compatible stacks among provided buildpacks")
+	}
+
 	image, err := p.imageFactory.NewImage(repoName, !publish)
 	if err != nil {
 		return nil, errors.Wrapf(err, "creating image")
--- a/internal/buildpackage/builder.go
+++ b/internal/buildpackage/builder.go
@@ -42,22 +42,20 @@ func (p *PackageBuilder) Save(repoName string, publish bool) (imgutil.Image, err
 	}
 
 	stacks := p.buildpack.Descriptor().Stacks
-	if len(stacks) == 0 {
-		return nil, errors.Errorf(
-			"buildpack %s must support at least one stack",
-			style.Symbol(p.buildpack.Descriptor().Info.FullName()),
-		)
-	}
-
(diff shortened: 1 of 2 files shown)


internal/buildpackage/builder.go
```diff
@@ -42,22 +42,20 @@ func (p *PackageBuilder) Save(repoName string, publish bool) (imgutil.Image, err
 	}
 
 	stacks := p.buildpack.Descriptor().Stacks
-	if len(stacks) == 0 {
```
jromero: Removed this since it's not possible to construct an invalid buildpack in our current code flow:

See https://github.com/buildpacks/pack/blob/master/internal/dist/buildpack.go#L205

internal/buildpackage/builder.go
```diff
@@ -42,22 +42,20 @@ func (p *PackageBuilder) Save(repoName string, publish bool) (imgutil.Image, err
...
+			stacks = stack.MergeCompatible(stacks, bpd.Stacks)
+			if len(stacks) == 0 {
+				return nil, errors.Errorf(
+					"buildpack %s does not support any stacks from %s",
+					style.Symbol(p.buildpack.Descriptor().Info.FullName()),
+					style.Symbol(bpd.Info.FullName()),
+				)
+			}
 		}
 	}
```
ameyer-pivotal: I think we're still missing a check for `len(stacks) == 0` here. Though unlikely, there's a chance that stacks could be empty at this point.

PR #434, opened 2019-12-24
Display error when no lifecycle descriptor is provided 

### LINKED ISSUE #267: Display error when no lifecycle descriptor is provided
As part of RFCs [#10](https://github.com/buildpack/rfcs/blob/master/text/0010-api-versions.md) and [#11](https://github.com/buildpack/rfcs/blob/master/text/0011-lifecycle-descriptor.md), it is imperative that we have the lifecycle descriptor available to determine compatibility between lifecycle and provided buildpacks.

Relates to #260 

#### Fail when descriptor not found
**Given**  I have a lifecycle archive at path `/path/to/lifecycle.tgz` WITHOUT a `lifecycle.toml` file at the root of the archive
**When** I run `pack create-builder -b builder.toml` and the `lifecycle.uri` field in the`builder.toml` points to the given lifecycle
**Then** `pack` should fail with exit code not equal to 0
**And** `pack` should output `ERROR: provided lifecycle is missing lifecycle descriptor.`

ktpv:
fixes #267 

### DIFF
--- a/internal/builder/lifecycle_test.go
+++ b/internal/builder/lifecycle_test.go
@@ -46,12 +46,9 @@ func testLifecycle(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		when("there is no descriptor file", func() {
-			it("assumes Platform API version 0.1", func() {
-				lifecycle, err := builder.NewLifecycle(&fakeEmptyBlob{})
-				h.AssertNil(t, err)
-				h.AssertEq(t, lifecycle.Descriptor().Info.Version.String(), "0.3.0")
-				h.AssertEq(t, lifecycle.Descriptor().API.PlatformVersion.String(), "0.1")
-				h.AssertEq(t, lifecycle.Descriptor().API.BuildpackVersion.String(), "0.1")
+			it("throws an error ", func() {
+				_, err := builder.NewLifecycle(&fakeEmptyBlob{})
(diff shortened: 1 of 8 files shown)


internal/builder/lifecycle_test.go
```diff
@@ -46,12 +44,9 @@ func testLifecycle(t *testing.T, when spec.G, it spec.S) {
...
 
 		when("there is no descriptor file", func() {
-			it("assumes Platform API version 0.1", func() {
-				lifecycle, err := builder.NewLifecycle(&fakeEmptyBlob{})
-				h.AssertNil(t, err)
-				h.AssertEq(t, lifecycle.Descriptor().Info.Version.String(), "0.3.0")
-				h.AssertEq(t, lifecycle.Descriptor().API.PlatformVersion.String(), "0.1")
-				h.AssertEq(t, lifecycle.Descriptor().API.BuildpackVersion.String(), "0.1")
+			it("throws an error ", func() {
+				_, err := builder.NewLifecycle(blob.NewBlob(filepath.Join("testdata", "lifecycle-bad")))
```
jromero: Is there a reason why the filesystem was used instead of the empty blob? We are moving towards reducing the overhead of maintaining fixtures in certain areas where the same effect could be achieved via inversion of control. The method `builder.NewLifecycle` shouldn't care if it's a file system based blob or an in-memory one.

PR #440, opened 2020-01-06
Create issue templates for bugs and features

Zanadar:
Creates a standard template for bug reports and feature requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,36 @@
+---
+name: Bug
+about: Bug report
+title: ''
+labels: status/triage, type/bug
+assignees: ''
+
+---
+### Summary
+<!--- Please provide a general summary of the issue. -->
+
+
+---
+
+### Reproduction
+
+##### Steps
+<!--- What steps should be taken to reproduce the issue? -->
+
+1.
+2.
+3.
+
+##### Expected
+<!--- What did you expect to happen -->
+
+
+---
+
+### Environment
+
+##### pack info
+<!--- Run `pack report` and copy output here. -->
+
+##### docker info
+<!--- Run `docker info` and copy output here. -->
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,8 @@
+---
+name: Bug
+about: Bug report
+title: ''
+labels: 'needs-triage'
```
jromero: Could we set the following labels: `status/triage, type/bug`

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,8 @@
...
+assignees: ''
+
+---
```
jromero: We should add request for more info here:

Things like:

```
### Summary
<!--- Provide a general summary of the issue in the Title above. -->


---

### Reproduction

##### Steps
<!--- What steps should be taken to reproduce the issue? -->

1.
2.
3.

##### Expected
<!--- Tell us what should happen. -->


---

### Environment

##### pack info
<!--- Run `pack report` and copy output here. -->

##### docker info
<!--- Run `docker info` and copy output here. -->

```

Zanadar:
Don't know if this should be an RFC or not @sclevine still kind of figuring out what does and doesn't need an RFC.

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,22 @@
+---
+name: Feature request
+about: Suggest a new feture or an improvement to existing functionality
+title: ''
+labels: type/enhancement, status/triage
+assignees: ''
+
+---
+
+**Description**
```
jromero: Nitpick: Might want to use headers `####` to be more visually separating given that might use bold text in their responses.

Zanadar: Good call to make them consistent.

Issue #449, opened 2020-01-11
Remove pack run


Zanadar:
Now that https://github.com/buildpacks/rfcs/pull/36 is merged, we should remove the `pack run` command.

michiboo:
Hi Can I have a go at this issue?

Zanadar:
@michiboo That'd be great!

Zanadar:
Accepting:
```
± zm |master U:1 ✗| → ./out/pack run
Error: unknown command "run" for "pack"
Run 'pack --help' for usage.
```

PR #451, opened 2020-01-13
remove command run from client program

michiboo:
This pull request is to address #449


jromero:
:wave:  @michiboo 

Nice to see this PR! Seems like the acceptance test are failing. You can verify this via `make acceptance` if you are on Mac or Linux.

Also, we'd probably want to get rid of the dead code associated with the Run command. That means `pack/Client#Run` and everything used within that operation.

michiboo:
@jromero  Hi, I think I fixed the issue.

PR #459, opened 2020-01-16
Fix create-builder with multiple buildpacks of same id

### LINKED ISSUE #430: pack create-builder should de-dup buildpacks list
We recently ran into issues creating a builder packaged with CNBs that reference the same dependency. Here's a snippet of our builder.toml that is causing this issue:

```
[[buildpacks]]
id = "org.cloudfoundry.node-engine"
uri = "/tmp/build/59a8880f/sources/dotnet-core-cnb/org.cloudfoundry.node-engine"

[[buildpacks]]
id = "org.cloudfoundry.node-engine"
uri = "/tmp/build/59a8880f/sources/nodejs-cnb/org.cloudfoundry.node-engine"
```

When we ran `pack create-builder localhost:5000/cloudfoundry/cnb:bionic --builder-config builder.toml --publish` we see the following error:

```
ERROR: failed to register layer: Error processing tar file(duplicates of file paths not supported): 
```

**Expected Behavior**
- If a builder toml has multiple entries with the same id that both reference the same version of a CNB, pack should de-dup these entries, and `pack create-builder` succeeds
- If a builder.toml has multiple entries with the same id that both reference different versions of the CNB, pack adds both versions separately

natalieparellano:
Resolves #430 

- Remove comptability layer (except for lifecycle symlink)
- Remove comptability metadata from builder creation
- Introduce COMPILE_PACK_WITH_VERSION env variable for running acceptance tests
- Add pack_previous_fixtures_overrides directory to enable overriding fixtures used
  for previous pack versions

Signed-off-by: Natalie Arellano <narellano@pivotal.io>


Makefile
```diff
@@ -48,7 +48,7 @@ unit: format lint
 
 acceptance: format lint
 	@echo "> Running acceptance tests..."
-	$(GOCMD) test -v -count=1 -parallel=1 -timeout=0 -tags=acceptance ./acceptance
+	COMPILE_PACK_WITH_VERSION=$(or ${COMPILE_PACK_WITH_VERSION}, 0.0.0) $(GOCMD) test -v -count=1 -parallel=1 -timeout=0 -tags=acceptance ./acceptance
```
jromero: What's the purpose of setting a version on a short lived pack binary? It doesn't seem to be used for anything. Is it primarily aesthetic?

acceptance/acceptance_test.go
```diff
@@ -137,11 +142,18 @@ func TestAcceptance(t *testing.T) {
 		h.AssertNil(t, err)
 	}
 
+	tmpPreviousPackFixturesPath, err := ioutil.TempDir("", "previous-pack-fixtures")
+	h.AssertNil(t, err)
+	defer os.RemoveAll(tmpPreviousPackFixturesPath)
+
+	h.RecursiveCopy(t, previousPackFixturesPath, tmpPreviousPackFixturesPath)
+	h.RecursiveCopy(t, filepath.Join("testdata", "pack_previous_fixtures_overrides"), tmpPreviousPackFixturesPath)
```
jromero: I _think_ I understand why this is needed. Maybe it's worth creating a `README.md` in that directory that entails exactly _why_ you would put stuff in that directory.

Also, would it not suffer from the same problems we originally had with "pack_previous" in which you had to update the repo after a release because "current" was (de)moted to "previous"?

ameyer-pivotal: Would vote on punting on the docs so we can get this out. Thoughts @jromero?

It won't suffer from the same problems if it just has the version-related files like it has now. But true, it could suffer in general. Hopefully these are the only type of files we'd need. We'll keep an eye on this.

Makefile
ameyer-pivotal: Saw your later comment. Yes, it's needed for version-specific fixtures to be found.

natalieparellano: Also because the run combination for `make acceptance` is current-current-current the version-specific fixtures are not relevant, hence `0.0.0` is an okay value - but for `make acceptance-all` it seemed prudent to require either `COMPILE_PACK_WITH_VERSION` or `PACK_PATH` to be set to avoid confusing test failures when `pack version` is incorrect.

PR #485, opened 2020-01-31
Enable interception via shell per lifecycle phase execution

### LINKED ISSUE #62: Hijack feature: shell into a buildpack container to look at the state of things
One of the neatest features of concourse CI is the `fly hijack` command. It would be useful to copy that somehow in `pack` so that app developers can get better visibility into what happens inside the build containers. E.g. 

```
$ pack build my/app --hijack org.cloudfoundry.openjdk:build` 
workspace/app$ # poke around and check the state of the container
workspace/app$ ^D
... build continues
```

I get a shell entry point into the "build" phase container of the "org.cloudfoundry.openjdk" buildpack. As a developer I can poke around in there and make sure all the pieces of the app and the preceding buildpacks are in place and have their proper values. When I leave the shell the build continues.

jromero:
##### Known Issues:
- stdin not captured properly per phase preventing subsequent phase shell from accepting initial input
    - **workaround**: type `ls` first to ensure shell can receive input

##### TODO:
- [x] Create Flag: `--intercept`
- [ ] Window Resizing

### Output

```
┌─[javier.romero][jromero-air][±][hack/intercept ↑1 ↓1 ✓][~/dev/pack]
└─▪ go run cmd/pack/main.go build my-app -B cnbs/sample-builder:bionic -p ../samples/apps/ruby-bundler/ --verbose
Pulling image index.docker.io/cnbs/sample-builder:bionic
bionic: Pulling from cnbs/sample-builder
Digest: sha256:f81f1c46d1164d3e7870724c64514e4edfaa021a1cf8144827e1e645da940ae8
Status: Image is up to date for cnbs/sample-builder:bionic
Selected run image cnbs/sample-stack-run:bionic
Pulling image cnbs/sample-stack-run:bionic
bionic: Pulling from cnbs/sample-stack-run
Digest: sha256:bbccf4244b2646e8ad69f167b5f6c4240679cc7164b6ef795accd33d573bda95
Status: Image is up to date for cnbs/sample-stack-run:bionic
Using build cache volume pack-cache-cbd6ecf04229.build
===> DETECTING
Intercepting...
-----------
To continue to the next phase type: exit
To manually run the phase type:
/cnb/lifecycle/detector -log-level debug -app /workspace -platform /platform
-----------
$ 
```

### Related Issues

Resolves #62


ekcasey:
@jromero if we move using `lifecycle/all` (buildpacks/rfcs#46) in the near future (which we should because because it saves us approximately 6s per build) this phase container based implementation might become moot?

Unless we choose run separate containers in the `--intercept` case, specifically for this purpose.

jromero:
@ekcasey yeah, I thought about that as I came back to try to finish this off. My thought was that whatever that may look like we could find a solution for it. For example, your proposal or we can simply list the `/all` command as the thing they should execute plus a way to list out individual commands. There are possibly other solutions here as well. I just didn't want this to collect too much dust. IMO, the need will always be there for a developer (or bp author) to better be able to debug the app building process.

ekcasey:
@jromero I think it's also worth noting that the original feature request in the referenced issue is to hijack between the execution of individual buildpacks during the `builder` phase, not between phases. Not that I can't imagine the latter also being useful, but perhaps this could benefit from a little user research (cc @z)

jromero:
This is now stale. Closing.

Issue #453, opened 2020-01-14
Expose create-package command


zmackie:
*Blocked by*: #348, #441 

To prepare for the release of buildpackages in pack, we need to expose the `create-package`.

```
Given: I run `pack` or `pack --help`
Then: I see `create-package NAME` as a documented command
```

PR #497, opened 2020-02-14
Update help for --buildpack flag

ameyer-pivotal:

### DIFF
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -114,7 +114,7 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
 	cmd.Flags().StringArrayVar(&buildFlags.EnvFiles, "env-file", []string{}, "Build-time environment variables file\nOne variable per line, of the form 'VAR=VALUE' or 'VAR'\nWhen using latter value-less form, value will be taken from current\n  environment at the time this command is executed")
 	cmd.Flags().BoolVar(&buildFlags.NoPull, "no-pull", false, "Skip pulling builder and run images before use")
 	cmd.Flags().BoolVar(&buildFlags.ClearCache, "clear-cache", false, "Clear image's associated cache before building")
-	cmd.Flags().StringSliceVarP(&buildFlags.Buildpacks, "buildpack", "b", nil, "Buildpack reference in the form of '<buildpack>@<version>',\n  path to a buildpack directory (not supported on Windows), or\n  path/URL to a buildpack .tar or .tgz file"+multiValueHelp("buildpack"))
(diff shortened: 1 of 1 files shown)


internal/commands/build.go
```diff
@@ -114,7 +114,7 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
 	cmd.Flags().StringArrayVar(&buildFlags.EnvFiles, "env-file", []string{}, "Build-time environment variables file\nOne variable per line, of the form 'VAR=VALUE' or 'VAR'\nWhen using latter value-less form, value will be taken from current\n  environment at the time this command is executed")
 	cmd.Flags().BoolVar(&buildFlags.NoPull, "no-pull", false, "Skip pulling builder and run images before use")
 	cmd.Flags().BoolVar(&buildFlags.ClearCache, "clear-cache", false, "Clear image's associated cache before building")
-	cmd.Flags().StringSliceVarP(&buildFlags.Buildpacks, "buildpack", "b", nil, "Buildpack reference in the form of '<buildpack>@<version>',\n  path to a buildpack directory (not supported on Windows), or\n  path/URL to a buildpack .tar or .tgz file"+multiValueHelp("buildpack"))
+	cmd.Flags().StringSliceVarP(&buildFlags.Buildpacks, "buildpack", "b", nil, "Buildpack reference in the form of '<buildpack>@<version>',\n  path to a buildpack directory (not supported on Windows),\n  path/URL to a buildpack .tar or .tgz file, or\n  the name of a packaged buildpack image"+multiValueHelp("buildpack"))
```
zmackie:
```suggestion
	cmd.Flags().StringSliceVarP(&buildFlags.Buildpacks, "buildpack", "b", nil, "Buildpack reference in the form of '<buildpack>@<version>',\n  path to a buildpack directory (not supported on Windows),\n  path/URL to a buildpack .tar or .tgz file, or\n  the name of a packaged buildpack image (eg. gcr.io/cncf-buildpacks-ci/buildpackages:nodejs-cnb)"+multiValueHelp("buildpack"))
```

zmackie: maybe an example here?

ameyer-pivotal: @zmackie A couple of thoughts:

- There is a long-form help string that we can set for more detailed information. We don't currently use the long help strings, but we could start to (I'd recommend not in this PR). This short-form string is supposed to be more "at a glance."
- We don't provide examples for all the other image-related flags in other commands. That doesn't mean we shouldn't. Maybe we should, though it might be better for the long-form help text.

ameyer-pivotal:
Going to roll this into #453

PR #502, opened 2020-02-18
Migrate tests from travis to github actions

### LINKED ISSUE #500: Migrate Travis tests to use GitHub Actions
Eventually, we'd like our whole build pipeline to be run by GitHub Actions. This will help reduce the number of third-party integrations, easing community engagement. It will also enable more streamlined integration with our codebase (PRs, releases, compatibility matrix tests, etc.). As an initial step in this direction, we'd like to remove our dependency on Travis CI.

- [ ] Migrate current Travis tests to use GitHub Actions
- [ ] Enable these tests for PRs and commits to `master` (but not all branches)

jromero:
These changes migrate from travis to GitHub actions. They provide the same level of functionality plus the addition of storing build artifacts.

- Skips docker dependent tests on `darwin` and `windows` (follow up work necessary)
    - `darwin`: we have no runner yet
    - `windows`: some tests need to be gated
- Skips code format/lint on `windows` (follow up work might be necessary)
    - formatting is showing differences
- Uploads built artifacts

Resolves #500 

image


.travis.yml
```diff
@@ -1,24 +0,0 @@
...
-        - NO_DOCKER=true
-    - os: windows
-      script: go test -mod=vendor -parallel=1 -v ./...
-      after_success: go build -mod=vendor -o pack ./cmd/pack
-      env:
-        - NO_DOCKER=true
-
-branches:
-  only:
-    - master
```
simonjjones: Is there any value in using a bimodal testing approach while we iterate towards Github Actions being the only way?

jromero: Can't think of a single value keeping Travis around would provide. You?

.github/workflows/build.yml
```diff
@@ -0,0 +1,55 @@
...
+        if: matrix.os != 'windows'
+        run: make verify
+      - name: Test
+        run: make test
+      - name: Build
+        run: make build
+      - uses: actions/upload-artifact@v1
+        with:
+          name: pack-${{ matrix.os }}
+          path: out/${{ env.PACK_BIN }}
```
zmackie: Love it!

Makefile
```diff
@@ -42,15 +42,15 @@ lint: install-golangci-lint
...
 	@echo "> Running unit/integration tests..."
 	$(GOCMD) test -v -count=1 -parallel=1 -timeout=0 ./...
 
-acceptance: format lint
+acceptance:
 	@echo "> Running acceptance tests..."
 	COMPILE_PACK_WITH_VERSION=$(or ${COMPILE_PACK_WITH_VERSION}, 0.0.0) $(GOCMD) test -v -count=1 -parallel=1 -timeout=0 -tags=acceptance ./acceptance
 
-acceptance-all: format lint
+acceptance-all:
```
zmackie: I think its a good change to remove these as a pre-req for the test runs.

Issue #493, opened 2020-02-10
Run Containers with mirrored TTY


nebhale:
Currently the build container is always started without a TTY.  Given that most applications use the existence of a TTY to determine whether to display color, any sub-process run within that container assumes that it is not in a TTY and does not display color even though a user might want it to.

This problem is even more insidious as parts of the buildpacks ecosystem have had to do various work arounds to get any color output.  For example, [`libcfbuildpack` always enables](https://github.com/cloudfoundry/libcfbuildpack/blob/master/logger/logger.go#L46) color regardless of the user's setting.  The `lifecycle` always outputs in color and then that color is stripped ([sometimes incompletely](https://github.com/buildpacks/pack/issues/123#issuecomment-583096038)) by `pack`.  In both of these cases (`libcfbuildpack` and `lifecycle`) the underlying color library would have properly enabled and disabled itself based on the existence of a TTY in the container that it ran it. Being able to depend on this removes a bad piece of code from `libcfbuildpack` and a _lot_ of code in `pack`.

`pack` should start its underlying containers with a TTY if the following are true.

1. `pack` is running in a TTY
1. `--no-color` has **not** been passed

It is a reasonable expectation that if a user has chosen to run `pack` in a TTY they expect color from `pack`, from the `lifecycle`, from buildpacks, and from things the buildpacks run.  It is also a reasonable expectation that if a user has chosen to run `pack` in a non-TTY or passed `--no-color` they expect no-color from `pack`, from the `lifecycle`, from buildpacks, and from things the buildpacks run.

By using a TTY as the color indicator we integrating into a vast existing ecosystem of tools that already use this heuristic.

jromero:
Related to https://github.com/buildpacks/pack/issues/486, yet this issue takes it much further by providing this information to the `lifecycle`.

natalieparellano:
Took a look at this today - in some sense starting the container with a TTY may be as simple as adding `Tty: true` [here](https://github.com/buildpacks/pack/blob/20305671ee25047ae437a534f89f6831a7176204/internal/build/phase.go#L36) and switching to use `io.Copy` [here](https://github.com/buildpacks/pack/blob/20305671ee25047ae437a534f89f6831a7176204/internal/container/run.go#L32) (see convo [here](https://github.com/moby/moby/issues/40336) on why that's needed).

[...]

PR #514, opened 2020-02-27
Mirror tty when in tty and not no-color passed

zmackie:
- Fixes https://github.com/buildpacks/pack/issues/493

Signed-off-by: Zander Mackie <zmackie@gmail.com>

### DIFF
--- a/internal/build/phase.go
+++ b/internal/build/phase.go
@@ -37,6 +37,14 @@ func (l *Lifecycle) NewPhase(name string, ops ...func(*Phase) (*Phase, error)) (
 		Image:  l.builder.Name(),
 		Labels: map[string]string{"author": "pack"},
 	}
+
+	if l.TTY {
+		ctrConf.Tty = true
+		ctrConf.AttachStdout = true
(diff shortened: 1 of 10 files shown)


internal/build/phase.go
```diff
@@ -172,8 +180,9 @@ func (p *Phase) Run(ctx context.Context) error {
 		ctx,
 		p.docker,
 		p.ctr.ID,
-		logging.NewPrefixWriter(logging.GetWriterForLevel(p.logger, logging.InfoLevel), p.name),
-		logging.NewPrefixWriter(logging.GetWriterForLevel(p.logger, logging.ErrorLevel), p.name),
+		logging.GetWriterForLevel(p.logger, logging.InfoLevel),
+		logging.GetWriterForLevel(p.logger, logging.ErrorLevel),
```
natalieparellano: Ah, so no more prefixing of log lines from lifecycle binaries? I recall somewhere that this was discussed, not sure where we landed on it.

zmackie:
I may be wrong, but I believe it’s necessary for the demux to work
properly.

PR #520, opened 2020-03-03
Update dependencies

jromero:
* Update `pack` dependencies
* Update `tools` dependencies


simonjjones:
This looks good to me, seems like a lot of the `x/sys` transitive deps have been deduplicated by ARCH which explains some of the discrepancy between lines added and removed.

PR #532, opened 2020-03-09
Unit test phases

### LINKED ISSUE #488: Phases should be unit-testable
### Description
See [internal/build/phases.go](https://github.com/buildpacks/pack/blob/1a6e03a48c9444629c355bf83562d73206698f66/internal/build/phases.go)

This code is not unit-tested, therefore we may not be verifying that the correct arguments are passed to each phase binary.

We already shipped a bug (see https://github.com/buildpacks/pack/issues/476) as a result of not verifying the arguments.

### Proposed solution
The code should be unit-tested. See [WIP branch](https://github.com/buildpacks/pack/tree/wip-di-phasemanager) with a proposed implementation.

At a high level:
* Declare new interface, `PhaseManager`
* `PhaseManager` is responsible for running each phase
* Fake `PhaseManager` in the tests
* Helper methods (such as `WithArgs`) live on `PhaseManager` so that we can assert that the helpers are called with correct arguments in the tests

Do nothing. Risk shipping more bugs.

Go with another implementation / approach.

### Additional context
`Detect` is already half-implemented on the branch. It would be great to get some sort of feedback that this is the approach we want to take, before tackling the other phases.

cc @simonjjones

natalieparellano:
Fixes #488

### DIFF
--- a/internal/build/phase_config_provider.go
+++ b/internal/build/phase_config_provider.go
@@ -1,0 +1,110 @@
+package build
+
+import (
+	"fmt"
+
+	"github.com/docker/docker/api/types/container"
+)
+
+type PhaseConfigProviderOperation func(*PhaseConfigProvider)
+
+type PhaseConfigProvider struct {
+	ctrConf  *container.Config
+	hostConf *container.HostConfig
+	name     string
+}
+
+func NewPhaseConfigProvider(name string, lifecycle *Lifecycle, ops ...PhaseConfigProviderOperation) *PhaseConfigProvider {
+	provider := &PhaseConfigProvider{
+		ctrConf:  new(container.Config),
+		hostConf: new(container.HostConfig),
+		name:     name,
+	}
+
+	provider.ctrConf.Cmd = []string{"/cnb/lifecycle/" + name}
+	provider.ctrConf.Image = lifecycle.builder.Name()
+	provider.ctrConf.Labels = map[string]string{"author": "pack"}
+
+	ops = append(ops,
+		WithLifecycleProxy(lifecycle),
+		WithBinds([]string{
+			fmt.Sprintf("%s:%s", lifecycle.LayersVolume, layersDir),
+			fmt.Sprintf("%s:%s", lifecycle.AppVolume, appDir),
+		}...),
+	)
+
+	for _, op := range ops {
+		op(provider)
+	}
+
+	return provider
+}
+
+func (p *PhaseConfigProvider) ContainerConfig() *container.Config {
+	return p.ctrConf
+}
+
+func (p *PhaseConfigProvider) HostConfig() *container.HostConfig {
+	return p.hostConf
+}
+
+func (p *PhaseConfigProvider) Name() string {
+	return p.name
+}
+
+func WithArgs(args ...string) PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.ctrConf.Cmd = append(provider.ctrConf.Cmd, args...)
+	}
+}
+
+func WithBinds(binds ...string) PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.hostConf.Binds = append(provider.hostConf.Binds, binds...)
+	}
+}
+
+func WithDaemonAccess() PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.ctrConf.User = "root"
+		provider.hostConf.Binds = append(provider.hostConf.Binds, "/var/run/docker.sock:/var/run/docker.sock")
+	}
+}
+
+func WithLifecycleProxy(lifecycle *Lifecycle) PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		if lifecycle.httpProxy != "" {
+			provider.ctrConf.Env = append(provider.ctrConf.Env, "HTTP_PROXY="+lifecycle.httpProxy)
+			provider.ctrConf.Env = append(provider.ctrConf.Env, "http_proxy="+lifecycle.httpProxy)
+		}
+
+		if lifecycle.httpsProxy != "" {
+			provider.ctrConf.Env = append(provider.ctrConf.Env, "HTTPS_PROXY="+lifecycle.httpsProxy)
+			provider.ctrConf.Env = append(provider.ctrConf.Env, "https_proxy="+lifecycle.httpsProxy)
+		}
+
+		if lifecycle.noProxy != "" {
+			provider.ctrConf.Env = append(provider.ctrConf.Env, "NO_PROXY="+lifecycle.noProxy)
+			provider.ctrConf.Env = append(provider.ctrConf.Env, "no_proxy="+lifecycle.noProxy)
+		}
+	}
+}
+
+func WithNetwork(networkMode string) PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.hostConf.NetworkMode = container.NetworkMode(networkMode)
+	}
+}
+
+func WithRegistryAccess(authConfig string) PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.ctrConf.Env = append(provider.ctrConf.Env, fmt.Sprintf(`CNB_REGISTRY_AUTH=%s`, authConfig))
+		provider.hostConf.NetworkMode = container.NetworkMode("host")
+	}
+}
+
+func WithRoot() PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.ctrConf.User = "root"
+	}
+}
--- a/internal/build/phase_factory.go
+++ b/internal/build/phase_factory.go
@@ -1,0 +1,23 @@
+package build
+
+type DefaultPhaseFactory struct {
+	lifecycle *Lifecycle
+}
+
+func NewDefaultPhaseFactory(lifecycle *Lifecycle) *DefaultPhaseFactory {
+	return &DefaultPhaseFactory{lifecycle: lifecycle}
+}
+
+func (m *DefaultPhaseFactory) New(provider *PhaseConfigProvider) RunnerCleaner {
+	return &Phase{
+		ctrConf:  provider.ContainerConfig(),
+		hostConf: provider.HostConfig(),
+		name:     provider.Name(),
+		docker:   m.lifecycle.docker,
+		logger:   m.lifecycle.logger,
+		uid:      m.lifecycle.builder.UID,
+		gid:      m.lifecycle.builder.GID,
+		appPath:  m.lifecycle.appPath,
+		appOnce:  m.lifecycle.appOnce,
+	}
+}
--- a/internal/build/phase_test.go
+++ b/internal/build/phase_test.go
@@ -14,8 +14,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/buildpacks/pack/internal/archive"
-
 	"github.com/buildpacks/imgutil/local"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
@@ -23,12 +21,15 @@ import (
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
 
+	"github.com/buildpacks/pack/internal/archive"
 	"github.com/buildpacks/pack/internal/build"
 	"github.com/buildpacks/pack/internal/builder"
 	ilogging "github.com/buildpacks/pack/internal/logging"
 	"github.com/buildpacks/pack/logging"
 	h "github.com/buildpacks/pack/testhelpers"
 )
+
+const phaseName = "phase"
 
 var (
 	repoName  string
@@ -58,7 +59,8 @@ func TestPhase(t *testing.T) {
 
 func testPhase(t *testing.T, when spec.G, it spec.S) {
 	var (
-		subject        *build.Lifecycle
+		lifecycle      *build.Lifecycle
+		phaseFactory   *build.DefaultPhaseFactory
 		outBuf, errBuf bytes.Buffer
 		docker         client.CommonAPIClient
 		logger         logging.Logger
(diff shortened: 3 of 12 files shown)


internal/build/fakes/fake_phase_manager.go
```diff
@@ -0,0 +1,104 @@
+package fakes
+
+import "github.com/buildpacks/pack/internal/build"
+
+type FakePhaseManager struct {
```
jromero: This looks like a lot of boilerplate. Is this auto-generated? If not, I wouldn't look forward to maintaining this...

internal/build/phases.go
```diff
@@ -15,17 +15,32 @@ const (
 	platformDir    = "/platform"
 )
 
-func (l *Lifecycle) Detect(ctx context.Context, networkMode string, volumes []string) error {
-	detect, err := l.NewPhase(
+type RunnerCleaner interface {
```
jromero: I get why this name was chosen but it's still slightly odd. Could it not simply be `Runner`?

internal/build/phases.go
```diff
@@ -15,17 +15,32 @@ const (
...
+	Cleanup() error
+	Run(ctx context.Context) error
+}
+
+type PhaseManager interface {
+	New(name string, ops ...PhaseOperation) (RunnerCleaner, error)
+	WithArgs(args ...string) PhaseOperation
```
jromero: Ops shouldn't need to be created by a service class. This shouldn't be necessary and if I understand the PR it's only here for the purpose of testing the operations. If that true, I don't think this is ultimately the right use of this pattern.

Some alternatives could be a builder pattern or state verification on the end result by inspecting what mutations a concrete operation yields.

internal/build/phase_manager.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"fmt"
+
+	"github.com/buildpacks/lifecycle/auth"
+	dcontainer "github.com/docker/docker/api/types/container"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/pkg/errors"
+)
+
+type ConcretePhaseManager struct {
```
jromero: Suggested alternative name: `DefaultPhaseManager`.

internal/build/phases.go
```diff
@@ -15,17 +15,32 @@ const (
...
 )
 
-func (l *Lifecycle) Detect(ctx context.Context, networkMode string, volumes []string) error {
-	detect, err := l.NewPhase(
+type RunnerCleaner interface {
+	Cleanup() error
+	Run(ctx context.Context) error
+}
+
+type PhaseManager interface {
```
jromero: I like the idea of something that creates a `Phase` but not sure manager is the right pattern to use here. See other comments. In my opinion, this should be more like a `Factory` (or `Builder` if you want some DSL and tracking capabilities for testing).

internal/build/fakes/fake_phase_manager.go
simonjjones: It's not autogenerated, which is a purposeful choice to reduce our set of dependencies. An auto-generated version of this file would be many multiples in length as it would be required to solve for the generic problem of multiple types of response & assertions, whereas with this hand-rolled fake we can optimize for the expected use cases (with guards against unexpected use)

internal/build/phases.go
simonjjones: The go idiom of an interface called `Runner` would be to expect it to consist of a single method called `Run`. The compound equivalent is fairly common throughout the std library for interfaces of two or even three methods. It could possibly be shortened to something like `RunCleaner` which would be consistent with interface names like `ReadCloser`

jromero: I did keep that in mind. The difference here being that it's not simply an interface for the composition of generic interfaces. It's an interface for a service type. There is no value to add `Cleaner` in my opinion given that it's a known and required function for the overall interface.

For example, [`Signal`](https://golang.org/pkg/os/#Signal) is not called `SignalStringer`. Another example, [`Listener`](https://golang.org/pkg/net/#Listener) is not called `ListenerCloser`.

I don't have a strong opinion, just a preference.

internal/build/fakes/fake_phase_manager.go
jromero: > reduce our set of dependencies

what type of dependencies are referring to ?

> auto-generated version of this file would be many multiples in length

But given that it would be auto-generated, do we care?

> we can optimize for the expected use cases

Not entire sure what optimize here means. If it's size or performance, it's not really applicable since it's only used within the testing scope. If it's usage, I'm not convinced that using something else wouldn't allow you to yield at the same results.

natalieparellano: We could use something like [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) to generate these fakes. Would this be acceptable?

internal/build/phases.go
natalieparellano: I do not see how "state verification on the end result" would be possible given that the fields of `Phase` are unexported.

Could you elaborate on what a builder pattern might look like? I could see us creating something like a "phase configuration" that we pass to the phase manager and verify that the configuration is correct.

internal/build/phases.go
natalieparellano: Agree the wording isn't quite right. We can come back to this when we settle on the scope of the interface.

internal/build/phases.go
jromero: There are two common builder patterns. Which one to pick is more or less per use case (for example if you wanted to pass around `BuildConfig`) and opinion:

```
// Single service constructs phase
PhaseBuilder()
 .withRoot(...)
 .withBind(..)
 .build() => Phase()

// A builder service and it's config
PhaseBuilder.build(BuildConfig().withRoot().withBind()) => Phase()
```

ameyer-pivotal: A couple of thoughts:
- Could we just add getters for the unexported fields and avoid the larger refactor? I'd never be a fan of exporting just for the sake of testing, but adding read-only getters seems OK to me.

- If you didn't move the "opts" to their own methods, and instead left them as arguments on the `Lifecycle.NewPhase()` method, think you can still use `gomock` to inspect them. After all, functions are still first-class citizens. Did you try that and it didn't work?

ameyer-pivotal: Update: I tested out bullet 2, and `gomock` either has a bug or doesn't allow something like this:

```go
mockLifecycle.EXPECT().NewPhase("build", []pack.LifecycleOp{build.WithArgs("x", "y", "z")})
mockLifecycle.NewPhase("build", build.WithArgs("x", "y", "z"))
```

It's strange because it __should__ work, per the test output, since functions are just pointers, but it still fails:

```
Expected call at /Users/ameyer/workspace/pack/internal/build/some_test.go:30 doesn't match the argument at index 1.
                Got: [0x14ac630]
                Want: is equal to [0x14ac630]
```

First, I tried moving the `WithArgs()` call to a stored variable, to ensure the reference/pointer was exactly the same, but there was no difference in the output above.

ameyer-pivotal: (Also note, I added `NewPhase` to the interface to allow this, and added the `LifecycleOp` type alias for convenience)

natalieparellano: We stubbed out a possible implementation that we are somewhat happy with: https://github.com/buildpacks/pack/tree/di-phasemanager-phaseconfigprovider

The key concepts being here: https://github.com/buildpacks/pack/blob/e83f703002967a06ec82c5d51b1031ec0a230fea/internal/build/phases.go#L45

Thoughts?

natalieparellano: >```
> // Single service constructs phase
> PhaseBuilder()
>  .withRoot(...)
>  .withBind(..)
>  .build() => Phase()

I see how this would work but I don't see how this is testable, given the limitations discussed above. Am I missing something?

>```
> // A builder service and it's config
> PhaseBuilder.build(BuildConfig().withRoot().withBind()) => Phase()

This is approaching the implementation we are currently recommending. By inspecting the config object, we can build confidence that the phase is configured correctly.

jromero: 👍 Looks good. Achieves "state verification on the end result" in a cleaner way which I'm really happy with.

natalieparellano:
Pushed some changes just now. Still todo:

* actually use phase config provider when creating a phase
* resolve outstanding review items
* answer code todos, in particular:
  * decide if config provider needs an interface
  * test verbose logging
  * fix weird registry auth expectation
  * decide if docker container import needs to be aliased where it's used

internal/build/fakes/fake_phase_manager.go
natalieparellano: With the latest changes, the fakes are a lot simpler and shouldn't need to change as often.

internal/build/phases.go
natalieparellano: I'm going to leave this as is for now, as the "cleanup" role of this interface seems essential to its function (given that we defer cleanup to ensure that it happens).

internal/build/phases.go
natalieparellano: Changing to `Factory` (we have enough builders in this code base!)

internal/build/phases.go
```diff
@@ -30,13 +41,24 @@ func (l *Lifecycle) Detect(ctx context.Context, networkMode string, volumes []st
...
 }
 
-func (l *Lifecycle) Restore(ctx context.Context, cacheName string) error {
-	restore, err := l.NewPhase(
-		"restorer",
+func (l *Lifecycle) Restore(ctx context.Context, cacheName string, phaseFactory PhaseFactory) error {
+	phaseName := "restorer"
+
+	configProvider, err := NewPhaseConfigProvider(
+		phaseName,
```
natalieparellano: This is a little awkward; maybe the config provider can hold onto the name and expose it to the factory?

internal/build/phases_test.go
```diff
@@ -0,0 +1,505 @@
...
+			it("configures the phase with registry access", func() {
+				lifecycle := fakeLifecycle(t, false)
+				fakePhaseFactory := fakes.NewFakePhaseFactory()
+				expectedRepos := []string{"some-repo-name"}
+
+				err := lifecycle.Analyze(context.Background(), expectedRepos[0], "test", true, false, fakePhaseFactory)
+				h.AssertNil(t, err)
+
+				configProvider := fakePhaseFactory.NewCalledWithProvider
+				h.AssertIncludeAllExpectedPatterns(t, configProvider.ContainerConfig().Env, []string{"CNB_REGISTRY_AUTH={}"})
```
natalieparellano: Is this expectation okay? That is indeed the value that is set in the environment, but I am surprised it does not reference the repo name that was passed in.

natalieparellano:
@jromero @ameyer-pivotal @simonjjones think this is ready for another round.

internal/build/phase_factory.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (m *DefaultPhaseFactory) New(name string, provider *PhaseConfigProvider) (RunnerCleaner, error) {
+	err := provider.Update(
+		WithLifecycle(m.lifecycle),
+		WithBinds([]string{
+			fmt.Sprintf("%s:%s", m.lifecycle.LayersVolume, layersDir),
+			fmt.Sprintf("%s:%s", m.lifecycle.AppVolume, appDir),
+		}...),
+	)
```
natalieparellano: Upon reflection I think this belongs in the constructor for the provider.

natalieparellano: I am also going to break up `WithLifecycle` as right now it is doing both "image" and "proxy" things.

natalieparellano:
Updated with the latest feedback.

internal/build/phase_test.go
```diff
@@ -70,66 +73,87 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
...
 	it.After(func() {
-		h.AssertNil(t, subject.Cleanup())
+		h.AssertNil(t, lifecycle.Cleanup())
 	})
 
 	when("Phase", func() {
 		when("#Run", func() {
 			it("runs the subject phase on the builder image", func() {
-				phase, err := subject.NewPhase("phase")
+				configProvider, err := build.NewPhaseConfigProvider(phaseName, lifecycle)
```
simonjjones: If we were using a fake implementation of our interface here, we could assert that the relevant config provider methods were in fact called, which would become the test of [our factory implementation](https://github.com/buildpacks/pack/blob/b321dd4c4643c933eca842444c2871a39d33a4f0/internal/build/phase_factory.go#L13-L15)

internal/build/phase_test.go
```diff
@@ -139,17 +163,18 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
...
-					assertAppModTimePreserved(t, subject, &outBuf, &errBuf)
+					assertAppModTimePreserved(t, lifecycle, phaseFactory, &outBuf, &errBuf)
 				})
 			})
 
 			when("app is a zip", func() {
 				it("preserves original mod times", func() {
 					var err error
-					subject, err = CreateFakeLifecycle(filepath.Join("testdata", "fake-app.zip"), docker, logger)
+					lifecycle, err = CreateFakeLifecycle(filepath.Join("testdata", "fake-app.zip"), docker, logger, repoName)
```
simonjjones: I've long held a belief that there's a go idiom that the logger should be the final argument passed to a method, but I can't currently find any evidence to back that up. It would feel more consistent with other parts of our codebase.

internal/build/phase_config_provider.go
```diff
@@ -0,0 +1,136 @@
...
+		provider.hostConf.NetworkMode = container.NetworkMode(networkMode)
+		return nil
+	}
+}
+
+func WithRegistryAccess(repos ...string) PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) error {
+		authConfig, err := auth.BuildEnvVar(authn.DefaultKeychain, repos...)
+		if err != nil {
+			return err
```
simonjjones: Our operation method signature is required to return an error for every type of operation, where this is the only place in our seven current methods that we could actually return an error. If we could find a way to move this logic out of this operation we could simplify every other operation implementation.

internal/build/phase_test.go
```diff
@@ -218,7 +242,9 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
 
 			when("#WithArgs", func() {
 				it("runs the subject phase with args", func() {
-					phase, err := subject.NewPhase("phase", build.WithArgs("some", "args"))
+					configProvider, err := build.NewPhaseConfigProvider(phaseName, lifecycle, build.WithArgs("some", "args"))
```
simonjjones: Here we are transitively testing the implementation of `NewPhaseConfigProvider` through the implementation of our `phaseFactory` and the implementation of this phases `Run` method, relying on log output. We should look to add explicit tests to our `PhaseConfigProvider` that specifically test the implementation of our `With...` commands.

We can then hopefully drop a lot of the `#With...` contexts from this test suite.

internal/build/phases.go
```diff
@@ -30,13 +39,15 @@ func (l *Lifecycle) Detect(ctx context.Context, networkMode string, volumes []st
...
 	defer detect.Cleanup()
 	return detect.Run(ctx)
 }
 
-func (l *Lifecycle) Restore(ctx context.Context, cacheName string) error {
-	restore, err := l.NewPhase(
-		"restorer",
+func (l *Lifecycle) Restore(ctx context.Context, cacheName string, phaseFactory PhaseFactory) error {
+	configProvider, err := NewPhaseConfigProvider(
+		"restorer", l,
```
simonjjones: We should probably have `l` on a separate line, even though that might seem like a waste of space, I think visually when the rest of the args are on separate lines it's easy to notice there are two on this line.

testhelpers/arg_patterns.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func sliceEmpty(slice interface{}) (bool, error) {
+	switch reflect.TypeOf(slice).Kind() {
+	case reflect.Slice:
+		return reflect.ValueOf(slice).Len() == 0, nil
+	default:
+		return true, fmt.Errorf("invoked with non slice actual: %v", slice)
+	}
+}
```
simonjjones: I know I added these methods, but I'm wondering if they need to be as flexible now. Do we ever pass these anything other than a slice of string? If we don't need to be generic as to what sort of slice gets passed we shouldn't.

internal/build/phases_test.go
```diff
@@ -0,0 +1,516 @@
+package build_test
```
simonjjones: I wonder if it would be worthwhile to separate out each phase into it's own file and keep a separate corresponding test file? The only real argument for it is that this file has now become quite large.

internal/build/phase_test.go
jromero: My $0.02, I would think it should be near the beginning because it's a **common** argument, similar to context in the standard lib.

As a visual example:

```go
func foo1(context, logger, arg0, arg1, arg2)
func foo2(context, logger, arg0, arg1)
func bar(context, arg0, arg1)
func fizz(arg0, arg1)
```

vs 


```go
func foo1(context, arg0, arg1, arg2, logger)
func foo2(context, arg0, arg1, logger)
func bar(context, arg0, arg1)
func fizz(arg0, arg1)
```

internal/build/phase_test.go
natalieparellano: Discussed with @simonjjones - we can imagine future work to refactor `phase` and `phase.Run()` to be more unit testable, so that we don't have to spin up a real docker container in order to assert that it was configured correctly. However we believe that work could be done outside of this PR.

internal/build/phase_test.go
natalieparellano: The closest sister method to `CreateFakeLifecycle` is `build.NewLifecycle(docker, logger)` so I would like to keep `docker, logger` as either the first two or the last two arguments. Going with the first two because ¯\_(ツ)_/¯

internal/build/phase_config_provider.go
natalieparellano: This seems reasonable and remove a bunch of error checking from `phases.go`.

internal/build/phase_test.go
natalieparellano: Adding tests for phase config provider. As mentioned in reply to other comment, we'll save a larger refactor of the test boundaries for `phase` for the future.

testhelpers/arg_patterns.go
natalieparellano: You are right, we only ever pass string slices. Making this change.

internal/build/phases_test.go
natalieparellano: Fair enough!

natalieparellano: On trying this out, I think it might actually be better to leave everything in the same file. GoLand needs the top level `TestPhases(t *testing.T)` to run the tests with Ctrl+Shift+R. If we give each phase its own file and top level `TestXXXPhase(t *testing.T)`, we have to do the same clunky setup for each phase (as described here: https://github.com/buildpacks/pack/issues/541). I'd rather leave the setup as-is and break up `phases_test.go` when we tackle https://github.com/buildpacks/pack/issues/541.

natalieparellano:
@jromero @simonjjones I think this is ready for another round.

jromero:
Nothing changes from a user's perspective.

PR #535, opened 2020-03-12
Validate contents of builder.toml

### LINKED ISSUE #510: Validate contents of builder.toml
While working on https://github.com/buildpacks/pack/issues/503 we found that `builder.toml` is not being validated like `package.toml`. We should validate `builder.toml`.

### Description
`pack` should validate `builder.toml` during builder creation

#### AC1. Tables

**Given** `builder.toml` contains an array of tables that is not expected:
```toml
[[buildpacks]]
uri = "noop-buildpack.tgz"

[[ordre]] # Notice: this is misspelled
    [[order.group]]
    id = "some-buildpack-id"
    version = "some-buildpack-version"
```
**When** I run `pack create-builder builder-image --builder-config builder.toml`
**Then** it should error with the following message:
```
Unknown configuration element 'ordre' in 'builder.toml'
```

#### AC2. Keys

**Given** `builder.toml` contains a key that is not expected:
```toml
[[buildpacks]]
# Notice: this should be `uri`
url = "noop-buildpack.tgz"

[[buildpacks]]
image = "some/package-dep"
```
**When** I run `pack create-builder builder-image --builder-config builder.toml`
**Then** it should error with the following message:
```
Unknown configuration element 'url' in 'builder.toml'
```

supra08:
Fixes #510

This PR contains modified config reader for `builder.toml` which is used to verify missing keys and values for the file.
It also contains additional tests for that.
@jromero @natalieparellano please review.

Signed-off-by: Supratik Das <rick.das08@gmail.com>

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -208,8 +208,10 @@ lifecycle:
 func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packPath, packCreateBuilderPath, configDir, lifecyclePath string, lifecycleDescriptor builder.LifecycleDescriptor) {
 
 	var (
-		bpDir    = buildpacksDir(*lifecycleDescriptor.API.BuildpackVersion)
-		packHome string
+		bpDir      = buildpacksDir(*lifecycleDescriptor.API.BuildpackVersion)
+		packHome   string
+		packVer    string
+		packSemver *semver.Version
 	)
 
 	// subjectPack creates a pack `exec.Cmd` based on the current configuration
@@ -221,6 +223,9 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 		var err error
 		packHome, err = ioutil.TempDir("", "buildpack.pack.home.")
 		h.AssertNil(t, err)
+		packVer, err = packVersion(packPath)
+		h.AssertNil(t, err)
+		packSemver = semver.MustParse(strings.TrimPrefix(strings.Split(packVer, " ")[0], "v"))
 	})
 
 	it.After(func() {
@@ -534,14 +539,31 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 								"--volume", fmt.Sprintf("%s:%s", tempVolume, "/my-volume-mount-target"),
 							))
 
-							packVer, err := packVersion(packPath)
-							h.AssertNil(t, err)
-							packSemver := semver.MustParse(strings.TrimPrefix(strings.Split(packVer, " ")[0], "v"))
-
 							if packSemver.GreaterThan(semver.MustParse("0.9.0")) || packSemver.Equal(semver.MustParse("0.0.0")) {
 								h.AssertContains(t, output, "Detect: Reading file '/platform/my-volume-mount-target/some-file': some-string")
 							}
 							h.AssertContains(t, output, "Build: Reading file '/platform/my-volume-mount-target/some-file': some-string")
+						})
+					})
+
+					when("--default-process", func() {
+						it("sets the default process from those in the process list", func() {
+							h.SkipIf(t, !packSupports(packPath, "build --default-process"), "--default-process flag is not supported")
+
+							h.SkipIf(t,
+								lifecycleDescriptor.Info.Version.LessThan(semver.MustParse("0.7.0")),
+								"skipping default process. Lifecycle does not support it",
+							)
+
+							h.Run(t, subjectPack(
+								"build", repoName,
+								"--default-process", "hello",
+								"-p", filepath.Join("testdata", "mock_app"),
+								"--buildpack", filepath.Join("testdata", "mock_buildpacks", "0.2", "simple-layers-buildpack"),
+							))
+
+							assertMockAppLogs(t, repoName, "hello world")
+
 						})
 					})
 
@@ -1679,6 +1701,23 @@ func assertMockAppRunsWithOutput(t *testing.T, repoName string, expectedOutputs 
 	assertMockAppResponseContains(t, launchPort, 10*time.Second, expectedOutputs...)
 }
 
+func assertMockAppLogs(t *testing.T, repoName string, expectedOutputs ...string) {
+	t.Helper()
+	containerName := "test-" + h.RandString(10)
+	ctr, err := dockerCli.ContainerCreate(context.Background(), &container.Config{
+		Image: repoName,
+	}, nil, nil, containerName)
+	h.AssertNil(t, err)
+
+	var b bytes.Buffer
+	err = h.RunContainer(context.Background(), dockerCli, ctr.ID, &b, &b)
+	h.AssertNil(t, err)
+
+	for _, expectedOutput := range expectedOutputs {
+		h.AssertContains(t, b.String(), expectedOutput)
+	}
+}
+
 func assertMockAppResponseContains(t *testing.T, launchPort string, timeout time.Duration, expectedOutputs ...string) {
 	t.Helper()
 	resp := waitForResponse(t, launchPort, timeout)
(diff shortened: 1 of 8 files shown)


zmackie:
@supra08 Yay! Thanks a ton for this PR. We'll take a look soon.

builder/config.go
```diff
@@ -114,13 +115,49 @@ func getWarningsForObsoleteFields(reader io.Reader) ([]string, error) {
...
+		if len(errorKeys) > 1 {
+			pluralizedElement += "s"
+		}
+
+		return Config{}, errors.Errorf("unknown configuration %s %s in %s",
+			pluralizedElement,
+			strings.Join(errorKeys, ", "),
+			style.Symbol(path),
+		)
+	}
```
natalieparellano: This code is duplicated in buildpackage/config_reader.go - could it be extracted into some sort of helper method?

builder/config.go
```diff
@@ -114,13 +115,49 @@ func getWarningsForObsoleteFields(reader io.Reader) ([]string, error) {
...
+
+	fmt.Println(builderConfig.Buildpacks.Buildpacks())
```
natalieparellano: We should remove this (I'm guessing this was added during debugging)

builder/config.go
```diff
@@ -114,13 +115,49 @@ func getWarningsForObsoleteFields(reader io.Reader) ([]string, error) {
...
 	for i, bp := range builderConfig.Buildpacks.Buildpacks() {
+		if bp.URI == "" {
+			return Config{}, errors.Errorf("missing %s configuration", style.Symbol("buildpack.uri"))
+		}
+
```
natalieparellano: This check is actually unnecessary since the call to `.Buildpacks()` scopes the list to bp's with non-empty URI

builder/config.go
supra08: Yeah, working on it.

builder/config.go
supra08: got it!

supra08:
I need some help with the failing `Acceptance Test`. It passes in my local build but fails here showing `invalid_builder.toml` not found. This was the file to verify the `builder.toml` in acceptance tests.

acceptance/acceptance_test.go
```diff
@@ -278,6 +283,19 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
...
+					h.CopyFile(t, filepath.Join(packFixturesDir, "invalid_builder.toml"), filepath.Join(tmpDir, "invalid_builder.toml"))
+
+					_, err := h.RunE(subjectPack("create-builder", "some-builder:build", "--builder-config", filepath.Join(tmpDir, "invalid_builder.toml")))
+					h.AssertError(t, err, "failed to execute command:")
+				})
+			})
+
+			it.After(func() {
+				h.AssertNil(t, os.RemoveAll(tmpDir))
+			})
```
natalieparellano: We should move this up so that it is immediately after the `it.Before`

builder/config_reader_test.go
```diff
@@ -86,6 +86,7 @@ func testConfig(t *testing.T, when spec.G, it spec.S) {
 
 				it("returns warnings", func() {
 					_, warns, err := builder.ReadConfig(builderConfigPath)
+					// h.AssertError(t, err, "parse contents of")
```
natalieparellano: We should remove this comment

builder/config_reader.go
```diff
@@ -95,31 +96,53 @@ func ReadConfig(path string) (config Config, warnings []string, err error) {
...
+		errorKeys := config.ParseUndecodedKeys(undecodedKeys)
+
+		pluralizedElement := "element"
+		if len(errorKeys) > 1 {
+			pluralizedElement += "s"
+		}
+
+		if obsoleteConfigs > 0 {
+			return builderConfig, nil
+		}
```
natalieparellano: If there are other invalid elements in addition to `groups` in the builder.toml, this logic would result in us not failing as we intend.

builder/config_reader.go
```diff
@@ -95,31 +96,53 @@ func ReadConfig(path string) (config Config, warnings []string, err error) {
...
+	return warnings, len(obsoleteConfig.Groups), nil
 }
 
 // parseConfig reads a builder configuration from reader and resolves relative buildpack paths using `relativeToDir`
-func parseConfig(reader io.Reader, relativeToDir string) (Config, error) {
-	var builderConfig Config
-	if _, err := toml.DecodeReader(reader, &builderConfig); err != nil {
+func parseConfig(reader io.Reader, relativeToDir, path string, obsoleteConfigs int) (Config, error) {
+	builderConfig := Config{}
+	fmt.Println(obsoleteConfigs)
```
natalieparellano: We should remove this print statement

acceptance/acceptance_test.go
```diff
@@ -278,6 +283,19 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 				builderName = value
 			})
 
+			when("builder.toml is invalid", func() {
+				it("displays an error", func() {
+					h.CopyFile(t, filepath.Join(packFixturesDir, "invalid_builder.toml"), filepath.Join(tmpDir, "invalid_builder.toml"))
```
natalieparellano: This test is failing in our combo test suites because we wouldn't expect this behavior to exist in previous versions of pack. We can skip this test using `h.SkipIf()` and some code similar to https://github.com/buildpacks/pack/blob/007ab0cb7d61d8c999efee282c82953b56bb5b80/acceptance/acceptance_test.go#L536-L540.

builder/config_reader.go
natalieparellano: Actually, we've open https://github.com/buildpacks/pack/issues/538 to remove the warning completely and just error when `groups` (or any other invalid config) is provided. Given that, I think we can accept this implementation in this pull request, and address obsolete configs in the other issue.

builder/config_reader_test.go
```diff
@@ -108,6 +109,33 @@ func testConfig(t *testing.T, when spec.G, it spec.S) {
...
+				it.Before(func() {
+					h.AssertNil(t, ioutil.WriteFile(builderConfigPath, []byte(`
+[buildpack]
+url = "noop-buildpack.tgz"
+`), 0666))
+				})
+
+				it("returns errors", func() {
+					_, _, err := builder.ReadConfig(builderConfigPath)
+					h.AssertError(t, err, "parse contents of")
```
jromero: Could we make these error assertions more explicit to what we expect to be stated?

builder/config_reader_test.go
```diff
@@ -108,6 +109,33 @@ func testConfig(t *testing.T, when spec.G, it spec.S) {
...
+				it.Before(func() {
+					h.AssertNil(t, ioutil.WriteFile(builderConfigPath, []byte(`
+[buidlpack]
+uri = "noop-buildpack.tgz"
+`), 0666))
+				})
+
+				it("returns errors", func() {
+					_, _, err := builder.ReadConfig(builderConfigPath)
+					h.AssertError(t, err, "parse contents of")
```
jromero: Could we make these error assertions more explicit to what we expect to be stated?

supra08: sure!

builder/config_reader_test.go
supra08: yeah!

supra08:
Alright @jromero I make those changes suggested as soon as possible!

supra08:
Hi @jromero @natalieparellano I've updated the tests. But now Darwin build is failing (cannot install Go in the container). And the acceptance combo is failing again, bringing about a lot of new errors which didn't arose in a local test.
Any solution for this?

natalieparellano:
@supra08 thanks for updating your PR! It looks like the Darwin failure was just a network flake. I'm sorry you're hitting more errors with the combo tests - I'll take a more careful look tomorrow to see if I can understand what is going on. It's possible we could merge your PR as-is and make the needed changes on our end.

supra08:
Thanks @natalieparellano! I'd be glad to work on fixing the tests if something can be done.

builder/config_reader_test.go
```diff
@@ -84,11 +84,10 @@ func testConfig(t *testing.T, when spec.G, it spec.S) {
 `), 0666))
 				})
 
-				it("returns warnings", func() {
+				it("returns error when obsolete 'groups' field is used", func() {
```
natalieparellano: Technically with this change we could remove this function call (and its definition): https://github.com/buildpacks/pack/pull/535/files#diff-9e3eaa38298a8257cec0740ad5e24fd9R79 - but we can also save this for #538.

builder/config_reader_test.go
natalieparellano: I think we could be even more explicit - e.g., instead of `parse contents of` we could say `unknown configuration elements buildpacks.url` since that is also in the error message.

builder/config_reader_test.go
natalieparellano: Same comment as above but we could reference `unknown configuration elements buidlpack`.

natalieparellano:
Hi @supra08 - nice work! It seems the combo failure for "pack previous" is due to an incorrect skip (the semver comparison is a bit off). This led me to investigate the other combo scenarios and I found that the error check in the test is too generic - for "pack current" we indeed get a failure but it is due to missing stack, not incorrect keys in builder.toml. I sent you a PR that should address this, let me know what you think.

Left a couple other comments but otherwise this is looking good!

builder/config_reader_test.go
supra08: I am a bit unsure about this. If we just remove the function call then its fine and we can address #538 with this too. But if we save this for issue #538, then what would we leave here? Means after the buildpack validations, we can either remove the test or identify it as an error in the test. 
@natalieparellano

supra08: I guess it would be better if this function is removed for now. Your opinions?

PR #550, opened 2020-03-30
Feature/524 download acceptance assets

### LINKED ISSUE #524: Acceptance tests should attempt to download pack & lifecycle pre-requisites when not specified by environment variable
### Description
When running `make acceptance-all`, there are a variety of pre-requisites that are expected to be present on my local system, with their locations specified by environment variable. This presents a challenging environment for contributors hoping to validate that changes continue to be compatible with previous versions of pack and lifecycle, as well as builders created with a previous version of pack. The current feedback loop requires either understanding this complex set of requirements, or creating a PR in order to trigger these tests in a [GitHub Action](https://github.com/buildpacks/pack/actions?query=workflow%3Acompatibility)

### Proposed solution
It is possible to replicate the logic used defined in our [GitHub Action workflow](https://github.com/buildpacks/pack/blob/master/.github/workflows/compatibility.yml) to identify which versions of each pre-requisite should be used, and then download these to a local system before continuing with our acceptance suite.

This proposal is to add functionality to our acceptance test suite that will download and consume any pre-requisites that have not been specified from the GitHub locations currently used in our workflows. These downloads should be communicated to the contributor very clearly, with instructions as to how to more explicitly configure them if preferred. The assets should be downloaded to a temporary folder and cleaned up after each run.

If the pre-requisites are locally specified by environment variable, there should be no change to the existing flow.

- Improved documentation for running combinatorial acceptance tests locally in the [Contributing document](https://github.com/buildpacks/pack/blob/master/CONTRIBUTING.md)
- Earlier, more explicit failure in acceptance tests, with a clear explanation of which configuration properties and guidelines as to how to manually download the required assets.

### Additional context
- Our current workflow for identifying the latest lifecycle asset involves consuming the [GitHub Actions Workflows API](https://developer.github.com/v3/actions/workflows/) which appears to require an authenticated user session. This may mean we require a GitHub authentication token as part of our test configuration. It's not clear whether this is strictly a better solution than the existing setup of having contributors manually configure this, so we may be better off falling back on improved documentation & failure messaging. Getting the latest lifecycle is an existing pre-requisite for running `make acceptance` so contributors should be somewhat familiar with the process.

natalieparellano:
Fixes #524


acceptance/acceptance_test.go
```diff
@@ -1291,110 +1269,6 @@ func testAcceptance(t *testing.T, when spec.G, it spec.S, packFixturesDir, packP
 	})
 }
 
-type runCombo struct {
```
natalieparellano: Moved run combo related things to their own file.

acceptance/acceptance_test.go
```diff
@@ -1839,60 +1713,3 @@ func taskKey(prefix string, args ...string) string {
 	}
 	return fmt.Sprintf("%s-%s", prefix, hex.EncodeToString(hash.Sum(nil)))
 }
-
-type SuiteManager struct {
```
natalieparellano: Moved suite manager related things to their own file.

acceptance/run_combination.go
```diff
@@ -0,0 +1,121 @@
...
+	lifecycleDescriptor          builder.LifecycleDescriptor
+}
+
+func (c *runCombo) String() string {
+	return fmt.Sprintf("p_%s cb_%s lc_%s", c.Pack, c.PackCreateBuilder, c.Lifecycle)
+}
+
+func getRunCombinations() ([]runCombo, error) {
+	combos := []runCombo{
+		{Pack: "current", PackCreateBuilder: "current", Lifecycle: "default"}, // TODO: the current behavior for
```
natalieparellano: We should discuss and align on this. I think the code as-is preserves the current behavior.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+
+func NewGithubAssetFetcher() (*GithubAssetFetcher, error) {
+	packHome, err := config.PackHome()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting pack home")
+	}
+
+	return &GithubAssetFetcher{
+		ctx:          context.TODO(),
+		logger:       ilogging.NewLogWithWriters(os.Stdout, os.Stderr), // TODO: not sure if this is what we really want
```
natalieparellano: Logging with testing is a little bit awkward. I'm using `fmt.Printf` in places to get the output to show while the test is running. Otherwise it would print when the test finishes and only if there's a failure. I'm not sure what we want.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+		return value, nil
+	}
+	return "", nil
+}
+
+func (f *GithubAssetFetcher) loadCacheManifest() (assetCache, error) {
+	if _, err := os.Stat(filepath.Join(f.cacheDir, assetCacheManifest)); os.IsNotExist(err) {
+		return assetCache{}, nil
+	}
+	// TODO: invalidate cache manifest if more than X old
```
natalieparellano: How old? One hour?

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+	packHome, err := config.PackHome()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting pack home")
+	}
+
+	return &GithubAssetFetcher{
+		ctx:          context.TODO(),
+		logger:       ilogging.NewLogWithWriters(os.Stdout, os.Stderr), // TODO: not sure if this is what we really want
+		githubClient: github.NewClient(nil),
+		cacheDir:     filepath.Join(packHome, assetCacheDir),
```
natalieparellano: We'll keep all cached assets and a manifest of entries in a sub-directory inside pack home.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+	}, nil
+}
+
+func (f *GithubAssetFetcher) FetchReleaseAsset(owner, repo, version string, expr *regexp.Regexp, extract bool) (string, error) {
```
natalieparellano: `regexp` is something like `pack-v\d+.\d+.\d+-macos`

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+			desiredAsset = asset
+			break
+		}
+	}
+	if desiredAsset == nil {
+		return "", fmt.Errorf("could not find asset matching expression %s", expr.String())
+	}
+
+	var returnPath string
+	if extract && strings.Contains(*desiredAsset.Name, ".tgz") {
```
natalieparellano: This is a bit awkward. I am doing this in one big `if/else` block so that we can write once to the cache before returning.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+		cache[owner][repo].Sources[version] = destDir
+	})
+	if err != nil {
+		f.logger.Warn(errors.Wrap(err, "failed to write cache").Error())
+	}
+	return destDir, nil
+}
+
+func (f *GithubAssetFetcher) FetchReleaseVersion(owner, repo string, n int) (string, error) {
+	if version, _ := f.cachedVersion(owner, repo, n); version != "" {
```
natalieparellano: Cache the version as well to avoid rate limiting. In theory if we blow away the cache manifest but the same release versions exist, we just need to query the API once to get the versions and can re-use the assets that would still be in the cache directory.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		return errors.Wrap(err, "marshaling cache content")
+	}
+	content = append(content, "\n"...)
+
+	return ioutil.WriteFile(filepath.Join(f.cacheDir, assetCacheManifest), content, 0644)
+}
+
+func (f *GithubAssetFetcher) downloadAndSave(assetURI, destPath string) error {
+	downloader := blob.NewDownloader(f.logger, f.cacheDir)
```
natalieparellano: Re-use the blob downloader.

acceptance/input_paths_manager.go
```diff
@@ -0,0 +1,183 @@
...
+			return err
+		}
+
+		fmt.Printf("using %s for lifecycle path\n", assetPath)
+		m.lifecyclePath = assetPath
+	}
+	if c.Lifecycle == "previous" && m.previousLifecyclePath == "" {
+		fmt.Printf("run combination %+v requires %s to be set\n", c, style.Symbol(envPreviousLifecyclePath))
+
+		version, err := githubAssetFetcher.FetchReleaseVersion("buildpacks", "lifecycle", -1) // TODO: this is crude. We should use semver to ensure it's the latest patch of the previous minor.
```
natalieparellano: TBD

acceptance/input_paths_manager.go
```diff
@@ -0,0 +1,183 @@
...
+			return err
+		}
+
+		sourceDir, err := githubAssetFetcher.FetchReleaseSource("buildpacks", "pack", version)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("using %s for previous pack fixtures path\n", sourceDir)
+		m.previousPackFixturesPath = sourceDir // TODO: fix; this is the top level directory - need to traverse directory
```
natalieparellano: TBD - this is complicated by the fact that the source has a top-level directory that is versioned by commit, e.g. `buildpacks-pack-d42c384`. In theory there would only ever be one directory so it's probably fine.

natalieparellano:
This is not ready to be merged but potentially could use some high-level feedback on the approach as I'm ironing out the issues.

natalieparellano:
Our linter is not happy. It thinks code is unused that is actually used :/

natalieparellano:
I'm able to run the combo suite with just `COMPILE_PACK_WITH_VERSION=0.0.0 make acceptance-all` - yay!

I wonder if (when we're happy with this) we could switch the GitHub actions workflow to this way of running things, so that it's implicitly tested.

jromero:
> I wonder if (when we're happy with this) we could switch the GitHub actions workflow to this way of running things, so that it's implicitly tested.

I'd like that generally. The only drawback would be running not running each combination in parallel would greatly increase testing time. Plus, the overall output/logs might be harder to consume. Things to consider.

natalieparellano:
> > I wonder if (when we're happy with this) we could switch the GitHub actions workflow to this way of running things, so that it's implicitly tested.
> 
> I'd like that generally. The only drawback would be running not running each combination in parallel would greatly increase testing time. Plus, the overall output/logs might be harder to consume. Things to consider.

We could have 4 different workflows with different values for `ACCEPTANCE_SUITE_CONFIG`. I'm also realizing that we'd want to add authentication to avoid getting rate limited. But I think it's doable and potentially worth it.

natalieparellano:
I am not sure why the linter is unhappy - maybe we're running into this bug? https://github.com/golangci/golangci-lint/issues/791 ...the code it thinks is unused is actually used!

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,462 @@
...
+	err = f.writeCacheManifest(owner, repo, func(cache assetCache) {
+		cache[owner][repo].Sources[version] = destDir
+	})
+	if err != nil {
+		f.logger.Warn(errors.Wrap(err, "writing cache").Error())
+	}
+	return destDir, nil
+}
+
+func (f *GithubAssetFetcher) FetchReleaseVersion(owner, repo string, n int) (string, error) {
```
jromero, line 174: What is `n` here? Maybe godocs would help.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,462 @@
...
+}
+
+func (f *GithubAssetFetcher) loadCacheManifest() (assetCache, error) {
+	cacheManifest, err := os.Stat(filepath.Join(f.cacheDir, assetCacheManifest))
+	if os.IsNotExist(err) {
+		return assetCache{}, nil
+	}
+
+	// invalidate cache manifest that is too old
+	if time.Since(cacheManifest.ModTime()) > 1*time.Hour {
```
jromero, line 277: Can we make `1*time.Hour` a const so it's easier to identify?

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,462 @@
...
+	for i, release := range releases {
+		version, err := semver.NewVersion(*release.TagName)
+		if err != nil {
+			return "", errors.Wrap(err, "parsing semver")
+		}
+		versions[i] = version
+	}
+	sort.Sort(semver.Collection(versions))
+
+	latestVersion := versions[len(versions)-1]
```
jromero, line 197: Shouldn't happen, but there doesn't seem to be a check if there are 0 releases.

acceptance/github_asset_fetcher.go
```diff
@@ -0,0 +1,462 @@
...
+		}
+	}
+
+	op(cache)
+
+	content, err := json.Marshal(cache)
+	if err != nil {
+		return errors.Wrap(err, "marshaling cache manifest content")
+	}
+	content = append(content, "\n"...)
```
jromero, line 319: Why are we appending a new line?

acceptance/input_paths_manager.go
```diff
@@ -0,0 +1,194 @@
...
+		sourceDir, err := githubAssetFetcher.FetchReleaseSource("buildpacks", "pack", version)
+		if err != nil {
+			return errors.Wrap(err, "fetching release source")
+		}
+
+		fis, err := ioutil.ReadDir(sourceDir)
+		if err != nil {
+			return errors.Wrapf(err, "reading directory %s", sourceDir)
+		}
+		innerDir := fis[0].Name()
```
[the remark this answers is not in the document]
natalieparellano: GitHub source tarballs have a top-level directory whose name includes the current commit sha. We know enough about the release to calculate this directory name, I was just being lazy.

acceptance/acceptance_test.go
```diff
@@ -116,43 +115,22 @@ func TestAcceptance(t *testing.T) {
...
 		lifecycleDescriptor, err = extractLifecycleDescriptor(lifecyclePath)
 		if err != nil {
 			t.Fatal(err)
 		}
 	}
 
-	previousLifecycleDescriptor := lifecycleDescriptor
-	previousLifecyclePath := os.Getenv(envPreviousLifecyclePath)
+	previousLifecyclePath := inputPathsManager.previousLifecyclePath
+	var previousLifecycleDescriptor builder.LifecycleDescriptor
```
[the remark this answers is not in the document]
natalieparellano: I saw this in the code and it confused me why we'd ever want to use `previousLifecycleDescriptor` when `previousLifecyclePath` is empty. I could put it back.

acceptance/run_combination.go
```diff
@@ -0,0 +1,123 @@
...
+	return parseSuiteConfig(suiteConfig)
+}
+
+func parseSuiteConfig(config string) ([]runCombo, error) {
+	var cfgs []runCombo
+	if err := json.Unmarshal([]byte(config), &cfgs); err != nil {
+		return nil, errors.Wrap(err, "parse config")
+	}
+
+	validate := func(jsonKey, value string) error {
```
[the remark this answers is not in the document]
natalieparellano: We only ever call this validation when `ACCEPTANCE_SUITE_CONFIG` is set.

acceptance/github_asset_fetcher.go
natalieparellano: I think this is an artifact of my debugging - will take this out.

acceptance/github_asset_fetcher.go
natalieparellano: I will add this check.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
jromero: It could have very well been unnecessary. Just curious. I'm sure that if all the tests are passing then this shouldn't be necessary.

acceptance/input_paths_manager.go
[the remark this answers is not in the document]
natalieparellano: I left a comment for now. If you prefer we can derive the directory name, but I don't know if that's any less fragile.

natalieparellano:
@jromero I think this is ready for another round.

natalieparellano:
@jromero updated GitHub actions workflow to download assets using the new code.

Issue #548, opened 2020-03-24
Use exporter -run-image flag instead of -image (platform API 0.3+)


ekcasey:
### Description
When platform API >=0.3 use the `-run-image` flag on the `exporter` instead of the deprecated `-image` flag.

### Additional context
> which image?

For clarity, `-image` has been deprecated in favor of `-run-image` on the `exporter` as of Platform API 0.3
- https://github.com/buildpacks/lifecycle/releases/tag/v0.7.0

jromero:
Resolved by #560

Issue #541, opened 2020-03-17
Lifecycle should rely on a builder interface


natalieparellano:
### Description
As in here: https://github.com/buildpacks/pack/blob/007ab0cb7d61d8c999efee282c82953b56bb5b80/internal/build/build.go#L25

It's frustrating that we have to `docker build` an image in our tests in order to be able to call things like `lifecycle.builder.Name()`.

### Proposed solution
Lifecycle should rely on a builder interface. Should make our tests faster and easier to write.

### Describe alternatives you've considered
Do nothing.

jromero:
Not sure I follow the connection between a `docker` binary execution and the underlying implementation. Could you elaborate? What does `docker build` have to do with whether the `lifecycle` uses a concrete object or interface?

natalieparellano:
This was closed in #560

PR #560, opened 2020-04-03
Pass -run-image to exporter if platform API >= 0.3

natalieparellano:
- Refactor: lifecycle relies on builder interface, so that builder
  can be faked in tests i.e., we don't have to build a docker image
  to test phases.


natalieparellano:
Fixes #548 and also #541. I decided to tackle the refactor in this PR because otherwise the work for #548 wouldn't be tested (or so complicated to setup properly that the work would've been more than the refactor).

internal/build/phases_test.go
```diff
@@ -1,77 +1,36 @@
...
-	phasesRepoName string
-)
-
 func TestPhases(t *testing.T) {
 	rand.Seed(time.Now().UTC().UnixNano())
 
 	color.Disable(true)
 	defer color.Disable(false)
 
-	h.RequireDocker(t)
```
jromero, line 36: 💯

internal/build/phases_test.go
```diff
@@ -1,77 +1,36 @@
...
-
-	it.Before(func() {
-		var err error
-		tmpDir, err = ioutil.TempDir("", "phases-test")
-		h.AssertNil(t, err)
-
-		dockerConfigDir, err := ioutil.TempDir("", "empty-docker-config-dir")
-		h.AssertNil(t, err)
-
-		h.AssertNil(t, os.Setenv("DOCKER_CONFIG", dockerConfigDir))
```
jromero, line 64: This is necessary to ensure the docker configuration is not contaminated for the tests. For example, if you have logged in to docker, it will have configuration that is not expected in these tests.

natalieparellano: I believe you but I don't understand - how do these tests even interact with docker?

jromero: You don't believe me... * nods head *. 😄 

I'm curious myself. Let me follow the crumbs...

1. https://github.com/buildpacks/pack/blob/55b940fbf29d0b4159b2ba02c2964b493722fb5a/internal/build/phases_test.go#L179
2. https://github.com/buildpacks/pack/blob/55b940fbf29d0b4159b2ba02c2964b493722fb5a/internal/build/phases.go#L71
3. https://github.com/buildpacks/pack/blob/55b940fbf29d0b4159b2ba02c2964b493722fb5a/internal/build/phases.go#L91
4. https://github.com/google/go-containerregistry/blob/3d03ed9b1ca2ad5d78d43832e8e46adc31d2b961/pkg/authn/keychain.go#L51
5. https://github.com/google/go-containerregistry/blob/3d03ed9b1ca2ad5d78d43832e8e46adc31d2b961/pkg/authn/keychain.go#L62

As suspected, it's GGCR that does a lookup on the EnvVar based on the keychain provided. We could make that something that we inject via DI.

natalieparellano: Ohhh okay thank you, this makes sense. I am going to put this back for now, DI sounds like something we want to do but maybe out of scope for this PR.

jromero:
Nice to see some improvement of the tests.

PR #574, opened 2020-04-09
Feature/version sevmer gitsha1 update

### LINKED ISSUE #572: pack version should be proper semver
### Description
Today, the version that we use to compile pack is not a proper semver:
```
$  pack version
v0.9.0 (git sha: d42c384a39f367588f2653f2a99702db910e5ad7)
```

### Proposed solution
It should follow the format `0.9.0+build-XXX.sha-asdasdf`

We could start from the work in this PR: https://github.com/buildpacks/pack/pull/396

When running `make build` in GitHub actions, we should compile pack with a meaningful semver if the branch being tested is a release branch. The desired major / minor / patch can be inferred from the branch name. When it's time to cut a release, we can ship the latest artifacts built from automated checks on the release branch.

If the branch being tested is not a release branch, we can compile pack with `0.0.0+build-XXX.sha-asdasdf`.

Do nothing.

### Additional context
https://semver.org/
> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.

### Questions
Do we want to remove the leading `v`? This is removed for now.
Do we need the build number? How do we keep track of it? Using [GITHUB_RUN_NUMBER](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables).
Is this the right format? `0.9.0+build-XXX.sha-asdasdf` See discussion below.
Do we need to warn end-users about this change? (What if there's tooling that parses the output of `pack version`?)

natalieparellano:
This PR supersedes #396 

Fixes #572 

Pending alignment on the format of the semver.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,8 @@ jobs:
         run: make test
       - name: Build
         run: make build
+        env:
+          - PACK_BUILD: ${{ GITHUB_RUN_NUMBER }}
       - uses: actions/upload-artifact@v1
         with:
           name: pack-${{ matrix.os }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,11 @@ jobs:
         run: make verify
       - name: Test
         run: make test
+      - name: Derive pack version from branch name
+        run: |
+          [[ $GITHUB_REF =~ ^refs\/heads\/release/(.*)$ ]] && version=${BASH_REMATCH[1]} || version=0.0.0
+          echo "::set-env name=PACK_VERSION::$version"
+        shell: bash
       - name: Build
         run: make build
         env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Build
         run: make build
         env:
-          - PACK_BUILD: ${{ GITHUB_RUN_NUMBER }}
+          PACK_BUILD: ${{ github.run_number }}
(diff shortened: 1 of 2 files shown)


natalieparellano:
Flipping back to `master` so that GHA workflows will run.

natalieparellano:
This isn't ready to merge - it's not deriving the version from the branch yet.

natalieparellano:
This is "done" in the sense that it **should** work and I personally am happy with the metadata format (though the format could easily be changed).

On the latest `build` workflow run, the output artifact has the following version:
```
$  ~/Downloads/pack version
0.0.0+b-235.g-293db8f
```

Note the base branch name of `release/0.99.0`. If we merge this in we should see an artifact with version `0.99.0+b-<some-build>.g-<some-sha>` generated for the merge commit. We could then cherry-pick this onto `release/0.10.0`.

jromero:
# ✅ Accepted

### `pack --version`

Downloaded from attached CI artifacts.

```
./Downloads/pack --version
0.0.0+git-e66c5c6.build-239
```

jromero:
Thanks for the contribution!

It's nice to see the improvements.

Issue #368, opened 2019-10-24
Proposal: Feature flags


jromero:
As we strive for a time-based release cadence instead of feature-based releases there are times where it doesn't make sense to have a partially implemented feature (eg. packages https://github.com/buildpack/pack/pull/353) or features that aren't yet intended for GA. Although these feature are not "ready" certain individual might want to start experimenting with said features. Here are a few proposals to standardize the way in which we provide additional _incomplete_ or _experimental_ features.

#### Option 1 - go build [ldflags](https://golang.org/cmd/link/)

[Setting variables at build time](https://blog.alexellis.io/inject-build-time-vars-golang/) could allow us to completely gate code/features that are incomplete.

`go build -ldflags "-X 'github.com/buildpack/pack/feature.FeatureA=true'" ...`

#### Option 1 - ~/.pack/config.toml individual features

A list of experimental features in the configuration.

**Example:**

```toml
# Enable experimental features (false by default)
[experimental]
packages = true
slices = false
```

#### Option 2 - ~/.pack/config.toml single configuration

A single configuration element enables all experimental features.

**Example:**

```toml
# Enable experimental features (false by default)
experimental = true
```

simonjjones:
The real question we should answer is whether we want build time or runtime feature flagging, they could each be implemented in a number of ways. I think the answer is possibly both, they solve different problems.

While we are still working on features or groups of features in the main codebase I think it makes sense to continue to ship binaries that have no way of consuming these features. This can be achieved by a build time feature flag that simply doesn't wire up the feature(s) in question. Anyone who builds the project for themselves would be able to use these features, and it's more likely that they'd understand any risks associated with doing so.

Runtime feature flagging makes sense in the context of shipping experimental features; ones where we have a reasonable amount of confidence, but perhaps aren't yet feature complete or we're not ready to call them GA.

jromero:
@simonjjones Completely agree with most, if not all, points in you comment. I think the main purpose of this issue should be to define the standard mechanisms to do so instead of implementing something last minute and before we know it we have X number of ways to do something similar.

With that said, you bring up a good point about probably needing both build time and runtime enablement. I don't know of many more options for build time enablement but could think of at least another variant to runtime enablement. Instead of individually calling out each individual feature and potentially creating a stale list of experimental features we could set a single flag that enables all experimental features. Similar to the [docker experiment config setting](https://github.com/docker/docker-ce/blob/master/components/cli/experimental/README.md#use-docker-experimental).

[...]

PR #576, opened 2020-04-14
Adds experimental flagging

zmackie:
- Addresses #368

This global flag should support future development of runtime experimental feature, per the discussion in #368 .

`pack report` provides an example of how to consume that flag, in addition to reporting it's state:

```bash
± zm |feature/experimental → origin S:4 U:1 ✗| → ./out/pack report
Pack:
  Version:  0.0.0+git-27f8c87
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.7.2

Config:
  default-builder-image = "heroku/buildpacks:18"
  experimental = true

  [[run-images]]
    image = "cnbs/test"
    mirrors = ["one", "two"]

  [[run-images]]
    image = "cnbs/test2"
    mirrors = ["one", "two"]
```

### DIFF
--- a/cmd/pack/main.go
+++ b/cmd/pack/main.go
@@ -28,7 +28,6 @@ func main() {
 		logger.Error(err.Error())
 		os.Exit(1)
 	}
-
 	rootCmd := &cobra.Command{
 		Use: "pack",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
--- a/internal/commands/report_test.go
+++ b/internal/commands/report_test.go
@@ -40,7 +40,10 @@ func testReportCommand(t *testing.T, when spec.G, it spec.S) {
 		packHome, err = ioutil.TempDir("", "")
 		h.AssertNil(t, err)
 		packConfigPath = filepath.Join(packHome, "config.toml")
-		h.AssertNil(t, ioutil.WriteFile(packConfigPath, []byte(`default-builder-image = "some/image"`), 0666))
+		h.AssertNil(t, ioutil.WriteFile(packConfigPath, []byte(`
+default-builder-image = "some/image"
+experimental = true
+`), 0666))
 		packMissingHome, err = ioutil.TempDir("", "")
 		h.AssertNil(t, err)
 	})
@@ -63,6 +66,7 @@ func testReportCommand(t *testing.T, when spec.G, it spec.S) {
 			it("presents output", func() {
 				h.AssertNil(t, command.Execute())
 				h.AssertContains(t, outBuf.String(), `default-builder-image = "some/image"`)
+				h.AssertContains(t, outBuf.String(), `experimental = true`)
 			})
 		})
 		when("config.toml is not present", func() {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -11,6 +11,7 @@ import (
 type Config struct {
 	RunImages      []RunImage `toml:"run-images"`
 	DefaultBuilder string     `toml:"default-builder-image,omitempty"`
+	Experimental   bool       `toml:"experimental,omitempty"`
 }
 
 type RunImage struct {
--- a/.github/ISSUE_TEMPLATE/feature.md
+++ b/.github/ISSUE_TEMPLATE/feature.md
@@ -9,7 +9,7 @@ assignees: ''
 
 ### Description
 <!-- A concise description of what problem the feature solves and why solving it matters.
-Ex. My shoelaces won't stay tied and I keep tripping [...] -->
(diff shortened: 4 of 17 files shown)


internal/commands/report.go
```diff
@@ -38,12 +40,13 @@ func Report(logger logging.Logger) *cobra.Command {
...
 }
 
-func generateOutput(writer io.Writer) error {
+func generateOutput(writer io.Writer, experimental bool) error {
 	tpl := template.Must(template.New("").Parse(`Pack:
   Version:  {{ .Version }}
   OS/Arch:  {{ .OS }}/{{ .Arch }}
 
 Default Lifecycle Version:  {{ .DefaultLifecycleVersion }}
+Experiment Features Enables: {{ .Experimental  }}
```
natalieparellano, line 49: Could this be `Experimental Features Enabled`?

cmd/pack/main.go
```diff
@@ -45,14 +45,19 @@ func main() {
...
 			}
 
 			packClient = initClient(logger)
+			packClient.Experimental(experimental)
 		},
 	}
 
 	rootCmd.PersistentFlags().Bool("no-color", false, "Disable color output")
 	rootCmd.PersistentFlags().Bool("timestamps", false, "Enable timestamps in output")
+	rootCmd.PersistentFlags().Bool("experimental", false, "Enable experimental features")
```
jromero, line 60: I'm curious what the expected experience for this type of implementation. The initial proposal might not be clear enough but the initial intent would be that it's a persisted state similar to the [docker CLI](https://docs.docker.com/engine/reference/commandline/cli/#experimental-features).

internal/commands/report_test.go
```diff
@@ -22,6 +24,13 @@ func TestReportCommand(t *testing.T) {
 	spec.Run(t, "Commands", testReportCommand, spec.Random(), spec.Report(report.Terminal{}))
 }
 
+type fakeClient struct {
```
natalieparellano, line 27: Do we want to make any test assertions around the contents of `pack report` w.r.t. experimental being enabled or disabled?

cmd/pack/main.go
```diff
@@ -45,14 +45,19 @@ func main() {
 				if flag, err := fs.GetBool("timestamps"); err == nil {
 					logger.WantTime(flag)
 				}
+				if flag, err := fs.GetBool("experimental"); err == nil {
+					experimental = flag
+				}
 			}
 
 			packClient = initClient(logger)
+			packClient.Experimental(experimental)
```
natalieparellano, line 54: I don't see that this is persisted anywhere... would the user need to pass `--experimental` every time? This seems contradictory to the idea that we could see the current configuration by running `pack report`.

internal/commands/report_test.go
```diff
@@ -30,17 +39,19 @@ func testReportCommand(t *testing.T, when spec.G, it spec.S) {
...
 		logger = ilogging.NewLogWithWriters(&outBuf, &outBuf)
-		command = commands.Report(logger)
+		client = &fakeClient{}
+		command = commands.Report(logger, client)
 
 		packHome, err = ioutil.TempDir("", "")
 		h.AssertNil(t, err)
 		packConfigPath = filepath.Join(packHome, "config.toml")
-		h.AssertNil(t, ioutil.WriteFile(packConfigPath, []byte(`default-builder-image = "some/image"`), 0666))
+		h.AssertNil(t, ioutil.WriteFile(packConfigPath, []byte(`default-builder-image = "some/image"`), 0666)) // TODO we should probably restore the existing value of this
```
natalieparellano, line 54: This is a horrible bug. We should add an issue for this. I think we are trying to adhere to only adding TODOs in the code when there is a corresponding issue.

natalieparellano: Actually the bug I was thinking of is coming from set_default_builder_test.go. I just put up a [PR](#581) for this, it has been bothering me for awhile.

I believe this line of code is okay because `packConfigPath` is residing in a temporary directory.

zmackie: will clean up

internal/config/config.go
```diff
@@ -11,6 +11,7 @@ import (
 type Config struct {
 	RunImages      []RunImage `toml:"run-images"`
 	DefaultBuilder string     `toml:"default-builder-image,omitempty"`
+	Experimental   bool       `toml:"experimental,omitempty"`
```
natalieparellano, line 14: We should test this, maybe along the lines of https://github.com/buildpacks/pack/blob/11e3a677ce38ee1166c96546171aa0b9877dbb37/internal/config/config_test.go#L47.

PR #577, opened 2020-04-14
Add release job to build workflow

jromero:
This new job prepares a draft release when any new commits are made to the release branch (`release/**`).

Additional features:
- Automatically generate `pack --help` in release notes
- Automatically detect default lifecycle version

This feature has been tested on a fork with the exception of Windows artifacts due to the current configuration needs. Here is an example release produced: https://github.com/jromero/pack/releases/tag/v0.91.0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,7 +167,7 @@ jobs:
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_path: ./pack-darwin.tgz
-          asset_name: pack-v${{ env.PACK_VERSION }}-darwin.tgz
+          asset_name: pack-v${{ env.PACK_VERSION }}-macos.tgz
           asset_content_type: application/gzip
       - name: Upload Release Asset - linux
         uses: actions/upload-release-asset@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,11 @@ jobs:
       NO_DOCKER: ${{ matrix.no_docker }}
     steps:
       - uses: actions/checkout@v2
+      - name: Derive pack version from branch name
+        run: |
+          [[ $GITHUB_REF =~ ^refs\/heads\/release/(.*)$ ]] && version=${BASH_REMATCH[1]} || version=0.0.0
+          echo "::set-env name=PACK_VERSION::$version"
+        shell: bash
       - name: Set up go
         uses: actions/setup-go@v2-beta
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -49,16 +54,134 @@ jobs:
...
+        run: |
+          LIFECYCLE_VERSION=$(./pack-linux/pack report | grep 'Default Lifecycle Version:' | grep -o '[^ ]*$')
+          echo "::set-output name=version::$LIFECYCLE_VERSION"
+      - name: Extract pack help
+        id: pack_help
+        run: |
+          PACK_HELP=$(./pack-linux/pack --help)
+          PACK_HELP="${PACK_HELP//'%'/'%25'}"
+          PACK_HELP="${PACK_HELP//$'\n'/'%0A'}"
+          PACK_HELP="${PACK_HELP//$'\r'/'%0D'}"
```
jromero, line 108: Should have probably commented this out but this all has to do with multiline outputs. 

See https://github.community/t5/GitHub-Actions/set-output-Truncates-Multiline-Strings/m-p/38372/highlight/true#M3322.

.github/workflows/build.yml
```diff
@@ -49,16 +54,134 @@ jobs:
...
+        id: create_release
+        uses: actions/create-release@latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: v${{ env.PACK_VERSION }}
+          release_name: pack v${{ env.PACK_VERSION }}
+          draft: true
+          prerelease: false
+          body: |
```
natalieparellano, line 120: This is so exciting! I love that we are templating this... then it will be consistent :)

natalieparellano:
I am sure you've tested this on your fork but just to call it out, the expected behavior if there's more than one commit on the release branch is that the existing draft release gets deleted and replaced?

.github/workflows/build.yml
```diff
@@ -183,5 +185,5 @@ jobs:
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_path: ./pack-windows.zip
-          asset_name: pack-${{ env.PACK_VERSION }}-windows.tgz
+          asset_name: pack-v${{ env.PACK_VERSION }}-windows.tgz
```
natalieparellano, line 188: I believe we want to use the `zip` extension.

jromero:
> I am sure you've tested this on your fork but just to call it out, the expected behavior if there's more than one commit on the release branch is that the existing draft release gets deleted and replaced?

The behaviour would be that there would be multiple draft releases available for maintainers to comb through and publish (or delete).

PR #581, opened 2020-04-15
Preserve pack config when running units

### LINKED ISSUE #580: pack units contaminate pack config
### Summary
Running `make unit` sets the default builder to `some/image`. This could be confusing to users of pack who are also working on pack.

---

### Reproduction
```
$  rm ~/.pack/config.toml && make unit
$  cat ~/.pack/config.toml
default-builder-image = "some/image"
```

##### Expected behavior
I expect that `~/.pack/config.toml` is untouched.

---

natalieparellano:
Fixes #580.

### DIFF
--- a/internal/commands/set_default_builder_test.go
+++ b/internal/commands/set_default_builder_test.go
@@ -3,6 +3,8 @@ package commands_test
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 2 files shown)


testhelpers/testhelpers.go
```diff
@@ -580,3 +580,18 @@ func writeTAR(t *testing.T, srcDir, tarDir string, mode int64, w io.Writer) {
 	err := archive.WriteDirToTar(tw, srcDir, tarDir, 0, 0, mode, true, nil)
 	AssertNil(t, err)
 }
+
+func CreateAndAssignTempPackHome(t *testing.T) (string, error) {
```
natalieparellano, line 584: On second thought, this helper is probably unnecessary...

internal/commands/set_default_builder_test.go
```diff
@@ -33,17 +34,24 @@ func testSetDefaultBuilderCommand(t *testing.T, when spec.G, it spec.S) {
...
 		logger = ilogging.NewLogWithWriters(&outBuf, &outBuf)
 		command = commands.SetDefaultBuilder(logger, config.Config{}, mockClient)
+
+		tempPackHome, err = h.CreateAndAssignTempPackHome(t)
+		h.AssertNil(t, err)
 	})
 
 	it.After(func() {
 		mockController.Finish()
+		h.AssertNil(t, os.RemoveAll(tempPackHome))
```
natalieparellano, line 54: We should also unset PACK_HOME

testhelpers/testhelpers.go
jromero: The downside to this helper is that is tries to hide the need to know that `PACK_HOME`  should be set but you need to remember to unset it (or even better set it to the previously set value). I can see the value of a helper struct more than a single method. One that lets you register and unregister. That said, it's value is relatively small.

PR #585, opened 2020-04-17
Move tests that don't require specific builders to their own suite

natalieparellano:
The current setup requires complicated workarounds when the output of
specific commands changes.

For example, we require the pack_previous_fixtures_overrides directory
and files such as inspect_X.Y.Z_builder_output.txt to change the desired
expectation when a previous version of pack is used to either
create a builder or inspect a builder.

This is cumbersome. We can reduce the number of tests that require
these workarounds by only applying the compatibility matrix for tests
that truly require a compatibility check. For example, we do not need to
re-evaluate the output of `pack suggest-builders` from a previous version
of pack.

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -142,6 +142,20 @@ func TestAcceptance(t *testing.T) {
 		}
 	}
 
+	suiteManager = &SuiteManager{out: t.Logf}
+	suite := spec.New("acceptance suite", spec.Report(report.Terminal{}))
+
+	suite("p_current", func(t *testing.T, when spec.G, it spec.S) {
+		testWithoutSpecificBuilderRequirement(
+			t,
+			when,
+			it,
+			currentPackFixturesDir,
+			packPath,
+			api.MustParse(builder.DefaultBuildpackAPIVersion),
+		)
+	}, spec.Report(report.Terminal{}))
+
 	resolvedCombos, err := resolveRunCombinations(
 		combos,
 		packPath,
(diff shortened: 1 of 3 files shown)


acceptance/testdata/pack_previous_fixtures_overrides/inspect_X.Y.Z_builder_output.txt
```diff
@@ -0,0 +1 @@
+Files like these represent the expected output of calling `pack inspect-builder` on a builder created with pack vX.Y.Z
```
natalieparellano, line 1: Added this example because typically we'd want to use versioned overrides with this directory. I think that was the root of the latest problem (no version).

acceptance/acceptance_test.go
```diff
@@ -142,6 +142,20 @@ func TestAcceptance(t *testing.T) {
 		}
 	}
 
+	suiteManager = &SuiteManager{out: t.Logf}
+	suite := spec.New("acceptance suite", spec.Report(report.Terminal{}))
+
+	suite("p_current", func(t *testing.T, when spec.G, it spec.S) {
```
jromero, line 148: Correct me if I'm wrong but it would appear that these would always execute regardless of ACCEPTANCE_SUITE_CONFIG. For example, in cases where we want to test (pack=previous, cb_pack=\*, lifecycle=\*) it would run the standard suite of tests against the current version of pack even though it's not what we requested.

natalieparellano: Ah, this is true... I'll add a check here. For the value of `ACCEPTANCE_SUITE_CONFIG` that you describe, I believe the desired behavior is that we skip these tests completely, since they would have run and passed when the previous version of pack was released.

jromero: As an alternative the check can be made within the loop.

natalieparellano: Added a check that `combos` has at least one element with `Pack` equal to "current".

natalieparellano:
> Need to update it a tad to only run `pack=current` acceptance when pack is set to `current`.

I think this is fixed in the latest update.

PR #589, opened 2020-04-20
restore doesn't require daemon access

ekcasey:
removes unnecessary mount of `docker.sock` into restore container


zmackie:
> I downgraded to 2.2.0.4 and yes, it just works :slightly_smiling_face: Your hunch about this being a os/docker-for-mac version issue is correct!

https://buildpacks.slack.com/archives/CD61YAG69/p1587409048142900 for context

Issue #538, opened 2020-03-16
builder.toml that contains `groups` should error, not warn


natalieparellano:
### Description
If builder.toml contains `groups`, we are currently warning the user: https://github.com/buildpacks/pack/pull/263. Now that we are validating that builder.toml contains only the expected tables and keys, we should error instead of warning.

### Proposed solution
Don't treat `groups` as a special case for invalid builder config.

**AC**

Given a builder.toml that contains `groups` as a table name
When I run `pack create-builder my-builder -B builder.toml`
Then it should error with a helpful message (see #510)

natalieparellano:
I believe this has already been done as part of #535. But there is some unused code relating to the old behavior that still needs to be removed.

natalieparellano:
This is closed in #592!

PR #592, opened 2020-04-21
Support process substitution for config files

retrohacker:
When building, you may want to make some things dynamic. When using `bash`/`make` I like to use `envsubst` for this and do process substitution. For example:

```toml
# Buildpacks to include in builder
[[buildpacks]]
id = "com.my.buildpacks.nodejs"
version = "0.0.1"
uri = "${PWD}/buildpacks/nodejs"

# Order used for detection
[[order]]
[[order.group]]
id = "com.my.buildpacks.nodejs"

# Stack that will be used by the builder
[stack]
id = "com.my.buildpacks.baseos.bionic"
run-image = "${REGISTRY}/baseos-cnb-test/run:bionic"
build-image = "${REGISTRY}/baseos-cnb-test/build:bionic"
```

And my `Makefile`:

```
#!make
SHELL=/bin/bash
REGISTRY?=my.personal.registry.net:7002
PREFIX?=$(REGISTRY)/baseos-cnb-test

build:
  pack create-builder $(PREFIX)/builder:bionic --builder-config <(cat builders/bionic/builder.toml | envsubst) --no-pull --publish $(PACK_FLAGS)
```

Since `pack` currently uses `io.Seek` on the file to scan for warnings first and then parse the TOML second, it can't read from file descriptors that point to pipes.

This PR updates `pack` to scan for warnings and read in the config in a single pass.

### DIFF
--- a/builder/config_reader.go
+++ b/builder/config_reader.go
@@ -43,7 +43,6 @@ type Config struct {
 	Order       dist.Order          `toml:"order"`
 	Stack       StackConfig         `toml:"stack"`
 	Lifecycle   LifecycleConfig     `toml:"lifecycle"`
-	groups      []interface{}
 }
 
 type BuildpackConfig struct {
@@ -123,10 +122,6 @@ func parseConfig(reader io.Reader, relativeToDir, path string) (Config, []string
 			return Config{}, warnings, errors.Wrap(err, "transforming lifecycle URI")
 		}
 		builderConfig.Lifecycle.URI = uri
-	}
-
-	if len(builderConfig.groups) > 0 {
-		warnings = append(warnings, fmt.Sprintf("%s field is obsolete in favor of %s", style.Symbol("groups"), style.Symbol("order")))
 	}
 
 	return builderConfig, warnings, nil
--- a/builder/config_reader.go
+++ b/builder/config_reader.go
@@ -76,7 +76,7 @@ func ReadConfig(path string) (config Config, warnings []string, err error) {
 	}
 	defer file.Close()
 
-	config, warnings, err = parseConfig(file, builderDir, path)
+	config, err = parseConfig(file, builderDir, path)
 	if err != nil {
 		return Config{}, nil, errors.Wrapf(err, "parse contents of '%s'", path)
 	}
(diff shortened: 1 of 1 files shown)


natalieparellano:
@retrohacker thanks for this PR! It's nice to see these improvements.

As it turns out, `getWarningsForObsoleteFields` is no longer needed as the presence of `groups` in the config will cause `parseConfig` to return an error. We have #538 to get rid of this check, but didn't get to it yet and I'm afraid that's making your work more complicated than it needs to be. I think though that we can still accept this PR as-is, unless you'd like to change it.

hone:
@retrohacker thanks for the PR! can add DCO to your commits?

```
git commit --amend --signoff
```

retrohacker:
@natalieparellano thanks for the quick response!

https://github.com/buildpacks/pack/issues/538#issuecomment-614841188

Happy to make the change. It looks like the tests cover this, so if I remove the relevant blocks and tests pass I _think_ this is good to go?

@hone oh hai :heart: sure thing!

retrohacker:
I updated this PR to remove `groups` from the toml parser config.

builder/config_reader.go
```diff
@@ -76,15 +76,7 @@ func ReadConfig(path string) (config Config, warnings []string, err error) {
 	}
 	defer file.Close()
 
-	warnings, err = getWarningsForObsoleteFields(file)
-	if err != nil {
-		return Config{}, nil, errors.Wrapf(err, "check warnings for file '%s'", path)
-	}
-	if _, err := file.Seek(0, io.SeekStart); err != nil {
-		return Config{}, nil, errors.Wrap(err, "reset config file pointer")
-	}
```
natalieparellano, line 85: I think in the end all we need is this deletion (and the removal of unused `getWarningsForObsoleteFields`).

natalieparellano:
> I updated this PR to remove `groups` from the toml parser config.

@retrohacker thank you for updating the PR! Since we never set `warnings` in `parseConfig`, we could probably remove that as well.

retrohacker:
Ah, right :facepalm: thanks :-) updated.

natalieparellano:
This looks great! Adding my user acceptance below. @jromero I think this is good to go.

Modified builder.toml with variable interpolation:
```
$  head builder.toml
# Buildpacks to include in builder
[[buildpacks]]
id = "buildpacks-io-samples/java-maven"
version = "0.0.1"
uri = "${PWD}/buildpacks/java-maven"

[[buildpacks]]
id = "buildpacks-io-samples/kotlin-gradle"
version = "0.0.1"
uri = "${PWD}/buildpacks/kotlin-gradle"
```

Released pack errors when provided this config:
```
$  pack create-builder test-builder --builder-config <(cat ./builder.toml | envsubst)
ERROR: invalid builder toml: reset config file pointer: seek /dev/fd/63: illegal seek
```

Modified pack runs successfully:
```
$  ~/workspace/pack/out/pack create-builder test-builder --builder-config <(cat ./builder.toml | envsubst)
alpine: Pulling from cnbs/sample-stack-build
...
Successfully created builder image test-builder
```

jromero:
Thank you @retrohacker for the contribution!

Issue #523, opened 2020-03-04
use `/cnb/lifecycle/creator` to improve performance


ekcasey:
### Description
Given a builder with platform API >= 0.3. `pack build` should execute the lifecycle in a single container to improve performance

### Proposed solution
When Platform API >= 0.3 `pack` should execute `/cnb/lifecycle/creator` rather than executing each phase individually.

### Additional context
RFC: https://github.com/buildpacks/rfcs/blob/master/text/0026-lifecycle-all.md
Lifecycle PR: buildpacks/lifecycle#243

NOTE: Should only apply when `--trust-builder` flag is provided. Further acceptance to be done in #566 to verify that `creator` is used when a builder is statefully trusted with `pack trust-builder`.

jromero:
👍 This would simplify our build process significantly.

https://github.com/buildpacks/pack/blob/master/internal/build/build.go#L67-L111

ameyer-pivotal:
Blocked currently by PR #532

yaelharel:
Hi, I'm going to start looking at this issue.

yaelharel:
This issues depends on issue #566 and on introducing a new flag: `--trust-builder`.
Since I'm working on it while #566 is still open, I think that we need to create the `--trust-builder` flag and move the A/Cs that relate to the `pack trust-builder` command to be part of #566.

natalieparellano:
Thanks @yaelharel for the clarification. I updated the description here and in #566 to reflect the updated A/C.

ekcasey:
@yaelharel @natalieparellano I think this issue has interactions with trusted builders but could be implemented independently. Right now all builders are trusted. We wouldn't be making matters worse by migrating all builds to `creator`. The interactions only come in once we make a distinction between untrusted/trusted builders. If trust/untrust lands first, `creator` should only apply used w/ trusted builders. If this lands first, that the trusted builder feature would be responsible for returning to container phases when a builder is untrusted.

yaelharel:
Discussion about performance testing can be found in the following slack thread:
https://buildpacks.slack.com/archives/CD61YAG69/p1588004835159100

PR #595, opened 2020-04-25
Execute the lifecycle in a single container

yaelharel:
- Add a `trust-builder` flag to `pack build` (default is false).
- Call lifecycle.Create if platform API is at least 0.3 and the builder is trusted.
- Add `networkMode` to Restore, Analyze and Export phases.

Fixes issue #523

Signed-off-by: Yael Harel <yharel@vmware.com>


acceptance/acceptance_test.go
```diff
@@ -812,6 +811,75 @@ func testAcceptance(
...
+									semver.MustParse(lifecycleDescriptor.API.PlatformVersion.String()).LessThan(semver.MustParse("0.3")),
+									"ignore trust-builder (and the creator). Lifecycle does not support it",
+								)
+								h.SkipIf(t,
+									!packSupports(packPath, "build --trust-builder"), "trust-builder not supported",
+								)
+							})
+
+							when("the builder is trusted", func() {
+								it("calls the creator", func() {
```
natalieparellano, line 851: I wonder how far we want to take this - we're asserting that we've called the creator but we don't test that a build completes successfully. It might be nice to see the simplest happy path `pack build` at a minimum. But if something were to go wrong it might be because some combination of user inputs was not correctly translated into arguments passed to the creator - which is presumably why we have phases_test.go, but perhaps acceptance tests would give us additional security that would be helpful. Thinking about how the current acceptance tests could be extended to use creator vs. not-creator (such as with a [table test](https://github.com/buildpacks/pack/blob/6bccf6c9cb0b3f5be910d1a748bb4ca8a4393bb5/internal/buildpackage/builder_test.go#L66-L78)) boggles the mind a little bit and I am not sure that we want to go down that path - I am just raising the question.

natalieparellano: I was thinking about this a little bit more, given that we're moving toward everything using the creator (in the short term) we could change test expectations that use the binary name (such as [this one](https://github.com/buildpacks/pack/blob/0501b34a9be65832f2b8a60cbcffbb4d1342b333/acceptance/acceptance_test.go#L683)) to be regexes. We'll still need to answer the question of trusted vs. untrusted builders and the test matrix that might introduce at some point, but it's maybe not as immediate now.

acceptance/acceptance_test.go
```diff
@@ -812,6 +811,75 @@ func testAcceptance(
 						})
 					})
 
+					when("--trust-builder", func() {
+						when("platform API version is up to 0.2", func() {
+							it.Before(func() {
+								h.SkipIf(t,
+									semver.MustParse(lifecycleDescriptor.API.PlatformVersion.String()).GreaterThan(semver.MustParse("0.2")),
+									"skipping trust-builder (and the creator). Lifecycle does not support it",
```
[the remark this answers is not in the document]
yaelharel: You're right. I'll change it.

internal/build/build.go
```diff
@@ -93,35 +96,40 @@ func (l *Lifecycle) Execute(ctx context.Context, opts LifecycleOptions) error {
...
+		}
 
-	l.logger.Info(style.Step("EXPORTING"))
-	if err := l.Export(ctx, opts.Image.Name(), opts.RunImage, opts.Publish, launchCache.Name(), buildCache.Name(), phaseFactory); err != nil {
-		return err
+		l.logger.Info(style.Step("EXPORTING"))
+		if err := l.Export(ctx, opts.Image.Name(), opts.RunImage, opts.Publish, launchCache.Name(), buildCache.Name(), opts.Network, phaseFactory); err != nil {
+			return err
+		}
+	} else {
```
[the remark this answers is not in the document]
yaelharel: I copied the style from that was previously here (as you can see [here](https://github.com/buildpacks/pack/blob/dd793414d1094f42b2b0671d0a9171091614f13c/internal/build/build.go)).
But I think you're right and there is no need for the else statement.
Will change it.

internal/build/phases.go
```diff
@@ -29,6 +29,66 @@ type PhaseFactory interface {
...
+
+	args := []string{
+		"-run-image", runImage,
+		repoName,
+	}
+
+	if clearCache {
+		args = append([]string{"-skip-restore"}, args...)
+	} else {
+		args = append([]string{"-cache-dir", cacheDir}, args...)
```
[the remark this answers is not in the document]
yaelharel: As far as I understand, `restore` is being called only when the `clearCache` flag is set to true (as can be seen [here](https://github.com/buildpacks/pack/blob/dd793414d1094f42b2b0671d0a9171091614f13c/internal/build/build.go#L107-L112)).
I think it makes sense since the restore phase is responsible on restoring any previously-cached dependencies but we can skip it.
I think it's similar with the export phase but I'm not sure and I don't see a proof in the code.
Anyway I thought that the `creator` may or may not use the cache.
Maybe @ekcasey or @jromero can clarify that.

internal/build/phases.go
```diff
@@ -29,6 +29,66 @@ type PhaseFactory interface {
 	New(provider *PhaseConfigProvider) RunnerCleaner
 }
 
+func (l *Lifecycle) Create(ctx context.Context, publish, clearCache bool, runImage, launchCacheName, cacheName, repoName, networkMode string, phaseFactory PhaseFactory) error {
+	var configProvider *PhaseConfigProvider
+	var binds []string
+
+	args := []string{
```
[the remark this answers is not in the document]
yaelharel: You're right. We're not passing the flags you mentioned since we're using their default values.
I don't mind changing the other phases but I'm not sure it should be part of this PR.
What do you think?

internal/build/phases_test.go
```diff
@@ -47,6 +48,126 @@ func testPhases(t *testing.T, when spec.G, it spec.S) {
...
+
+				err := lifecycle.Create(context.Background(), true, true, "test", "test", "test", "test", expectedNetworkMode, fakePhaseFactory)
+				h.AssertNil(t, err)
+
+				configProvider := fakePhaseFactory.NewCalledWithProvider
+				h.AssertEq(t, configProvider.HostConfig().NetworkMode, container.NetworkMode(expectedNetworkMode))
+			})
+		})
+
+		when("publish is false and clear cache is false", func() {
```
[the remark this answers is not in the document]
yaelharel: Please correct me if I'm wrong but I think you want to test it similarly to how we test `Analyze` (there are 5 cases: "clear cache", "clear cache is false", "publish", "publish is false" and some general tests).
I agree that it's possible that `publish` will be true and `clearCache` will be false and vice versa but I wanted to reduce the number of cases that we test. I know that it's important to cover all of the scenarios but sometimes it's ok to leave some scenarios without coverage since the other scenarios are good enough. Maybe this isn't the case and it's better to split the cases.
What do you think?

internal/build/phases.go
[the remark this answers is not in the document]
natalieparellano: Ah, that makes sense about restore. I think we'll want to confirm about export though.

internal/build/phases.go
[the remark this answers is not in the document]
natalieparellano: I can see how this would be its own unit of work. If this is something that we want to do, we should probably create an issue for it so that it doesn't get lost.

internal/build/phases_test.go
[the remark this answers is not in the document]
yaelharel: @natalieparellano , following our discussion and since it seems from the tests that there is some relation between the value of `publish` and `clearCache`, I'll split the code to 5 cases as appears in `Analyze`.

internal/build/phases.go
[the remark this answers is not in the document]
yaelharel: I wrote it to myself. Thanks!

acceptance/acceptance_test.go
yaelharel: Following the discussion in [this slack thread](https://buildpacks.slack.com/archives/CSFE49VGR/p1588091913054100?thread_ts=1587752732.046400&cid=CSFE49VGR), I'll make this changes after removing the `--trusted-builder` flag and updating the acceptance-tests accordingly.
Thanks for the idea!

internal/build/phases_test.go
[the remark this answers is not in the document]
yaelharel: I've just uploaded the changes.
Please notice that the general cases don't fully cover all of the cases.
For example, I don't test that the `log-level`, `run-image` and `expectedRepo` are configured as expected when `publish` and `clearCache` are true.
I think that the current coverage is enough but we can have another discussion if you think differently.

acceptance/acceptance_test.go
```diff
@@ -678,44 +678,59 @@ func testAcceptance(
 						t.Log("app is runnable")
 						assertMockAppRunsWithOutput(t, repoName, "Launch Dep Contents", "Cached Dep Contents")
 
+						version, err := packVersion(packPath)
+						h.AssertNil(t, err)
+						packSemver = semver.MustParse(strings.TrimPrefix(strings.Split(version, " ")[0], "v"))
+						packSupport := packSemver.GreaterThan(semver.MustParse("0.10.0")) || packSemver.Equal(semver.MustParse("0.0.0"))
```
natalieparellano, line 684: Could we change this variable name to `packSupportsCreator`? We have a function called `packSupports` that is called with the name of a command, so this would read similarly.

acceptance/acceptance_test.go
```diff
@@ -678,44 +678,59 @@ func testAcceptance(
...
+
 						t.Log("restores the cache")
-						if lifecycleDescriptor.Info.Version.LessThan(semver.MustParse("0.6.0")) {
-							h.AssertContainsMatch(t, output, `(?i)\[restorer] restoring cached layer 'simple/layers:cached-launch-layer'`)
-							h.AssertContainsMatch(t, output, `(?i)\[analyzer] using cached launch layer 'simple/layers:cached-launch-layer'`)
-						} else {
+						if semver.MustParse(lifecycleDescriptor.API.PlatformVersion.String()).LessThan(semver.MustParse("0.3")) || !packSupport {
```
natalieparellano, line 687: I wonder if there would be value in introducing a new variable such as `lifecycleSupportsCreator` to make this more readable.

acceptance/acceptance_test.go
```diff
@@ -678,44 +678,59 @@ func testAcceptance(
...
+							h.AssertContainsMatch(t, output, `(?i)\[exporter] reusing layer 'simple/layers:cached-launch-layer'`)
 						} else {
-							h.AssertContainsMatch(t, output, `(?i)\[exporter] Reusing cache layer 'simple/layers:cached-launch-layer'`)
+							h.AssertContainsMatch(t, output, `(?i)\[creator] reusing layer 'simple/layers:cached-launch-layer'`)
 						}
 
-						t.Log("rebuild with --clear-cache")
-						output = h.Run(t, subjectPack("build", repoName, "-p", appPath, "--clear-cache"))
-						h.AssertContains(t, output, fmt.Sprintf("Successfully built image '%s'", repoName))
+						// TODO: check if we need the following test - no cacher exists
```
natalieparellano, line 702: I'm not too familiar with this code but it looks like we might need three cases here? One for platform api version 0.2 that uses `cacher`, one that uses `exporter`, and one that uses `creator`? The `cacher` one might be too old at this point, maybe?

acceptance/acceptance_test.go
```diff
@@ -847,31 +889,33 @@ func testAcceptance(
...
+						when("--default-process", func() {
+							it("sets the default process from those in the process list", func() {
+								h.SkipIf(t, !packSupports(packPath, "build --default-process"), "--default-process flag is not supported")
 
-							h.SkipIf(t,
-								lifecycleDescriptor.Info.Version.LessThan(semver.MustParse("0.7.0")),
-								"skipping default process. Lifecycle does not support it",
-							)
+								h.SkipIf(t,
+									semver.MustParse(lifecycleDescriptor.API.PlatformVersion.String()).GreaterThan(semver.MustParse("0.2")),
```
natalieparellano, line 905: I believe what you've provided is equivalent to the old skip criteria. I think we've been leaning toward using the platform API version vs. the lifecycle version so this seems preferable to me.

acceptance/acceptance_test.go
```diff
@@ -678,44 +678,59 @@ func testAcceptance(
...
 						t.Log("restores the cache")
-						if lifecycleDescriptor.Info.Version.LessThan(semver.MustParse("0.6.0")) {
-							h.AssertContainsMatch(t, output, `(?i)\[restorer] restoring cached layer 'simple/layers:cached-launch-layer'`)
-							h.AssertContainsMatch(t, output, `(?i)\[analyzer] using cached launch layer 'simple/layers:cached-launch-layer'`)
-						} else {
+						if semver.MustParse(lifecycleDescriptor.API.PlatformVersion.String()).LessThan(semver.MustParse("0.3")) || !packSupport {
 							h.AssertContainsMatch(t, output, `(?i)\[restorer] Restoring data for "simple/layers:cached-launch-layer" from cache`)
 							h.AssertContainsMatch(t, output, `(?i)\[analyzer] Restoring metadata for "simple/layers:cached-launch-layer" from app image`)
+						} else {
+							h.AssertContainsMatch(t, output, `(?i)\[creator] Restoring data for "simple/layers:cached-launch-layer" from cache`)
```
natalieparellano, line 691: Our reliance on the format of log messages is a bummer. One day maybe we'll have structured data and this problem will go away...

acceptance/acceptance_test.go
yaelharel: When running `make acceptance-all` on the main branch, we don't get to the `cacher` in any configuration so I deleted this part because I guess it's too old (kept only the `exporter` and the `creator`.

acceptance/acceptance_test.go
yaelharel: This test is failing currently and this is why I comment it out.
I guess my skip condition isn't right.
I'll try to figure out what should be the right skip condition.

acceptance/acceptance_test.go
yaelharel: I totally agree that we should change it!
On the other hand, I don't think it should be part of this issue.

acceptance/acceptance_test.go
yaelharel: Good idea! Thanks!

acceptance/acceptance_test.go
yaelharel: Sure! Will do.

internal/build/phases.go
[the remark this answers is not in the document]
yaelharel: Please see: https://github.com/buildpacks/pack/issues/610

internal/build/phases.go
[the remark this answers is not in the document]
yaelharel: So you were right!
We need to pass the `-cache-dir` even when `clearCache` is set to true.
Otherwise, the `cacheStore` will be `nil` [here](https://github.com/buildpacks/lifecycle/blob/3dde35fd5d4f1a3a3412eac7b819fa597d597cd9/cmd/lifecycle/exporter.go#L215) and therefore we won't get to "Adding cache layer" [here](https://github.com/buildpacks/lifecycle/blob/3dde35fd5d4f1a3a3412eac7b819fa597d597cd9/exporter.go#L321).
Please see [this commit](https://github.com/buildpacks/pack/pull/595/commits/b6961bf7c99560fcc5dd0a9deb1ed607dd876aa9) that fixes this issue.
Thanks!

internal/build/phases_test.go
```diff
@@ -47,6 +48,166 @@ func testPhases(t *testing.T, when spec.G, it spec.S) {
...
+				err := verboseLifecycle.Create(context.Background(), false, false, "test", "test", "test", "test", "test", fakePhaseFactory)
+				h.AssertNil(t, err)
+
+				configProvider := fakePhaseFactory.NewCalledWithProvider
+				h.AssertEq(t, configProvider.Name(), "creator")
+				h.AssertIncludeAllExpectedPatterns(t,
+					configProvider.ContainerConfig().Cmd,
+					[]string{"-cache-dir", "/cache"},
+				)
+				h.AssertSliceContains(t, configProvider.HostConfig().Binds, "/var/run/docker.sock:/var/run/docker.sock")
```
natalieparellano, line 126: Just being nit-picky here, I don't think the docker socket being mounted has anything to do with clear cache - it is a side-effect of passing `false` for publish. We can probably remove this line.

internal/build/phases_test.go
```diff
@@ -47,6 +48,166 @@ func testPhases(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertNil(t, err)
+
+				configProvider := fakePhaseFactory.NewCalledWithProvider
+				h.AssertEq(t, configProvider.Name(), "creator")
+				h.AssertIncludeAllExpectedPatterns(t,
+					configProvider.ContainerConfig().Cmd,
+					[]string{"-daemon"},
+					[]string{"-launch-cache", "/launch-cache"},
+				)
+				h.AssertSliceContains(t, configProvider.HostConfig().Binds, "/var/run/docker.sock:/var/run/docker.sock")
```
natalieparellano, line 182: This and line 194 are superfluous (I think) to line 206.

natalieparellano:
@yaelharel this is looking great! I ran through a very basic user acceptance, output below. Note the confusing warning about default process, I think it's because [this line](https://github.com/buildpacks/pack/blob/cb1ef794c4f7e56bbfdb433c051e44a93d4bc115/internal/build/phases.go#L168) doesn't include a check that the default process is specified, but that's unrelated to this PR.

@jromero I think this is ready for your review. 

```
$  ~/workspace/pack/out/pack build test-creator -B cloudfoundry/cnb:bionic
bionic: Pulling from cloudfoundry/cnb
...
Status: Image is up to date for gcr.io/paketo-buildpacks/run:base-cnb
===> CREATING
Warning: You specified a default process type but that is not supported by this version of the lifecycle
[creator] ---> DETECTING
[creator] 6 of 15 buildpacks participating
...
[creator] ---> ANALYZING
[creator] Restoring metadata for "paketo-buildpacks/bellsoft-liberica:class-counter" from app image
...
[creator] ---> RESTORING
[creator] Restoring data for "paketo-buildpacks/bellsoft-liberica:jdk" from cache
...
[creator] ---> BUILDING
[creator] 
[creator] Paketo BellSoft Liberica Buildpack 2.5.3
...
[creator] ---> EXPORTING
[creator] Reusing layer 'launcher'
...
[creator] Reusing 1/1 app layer(s)
[creator] Adding layer 'config'
[creator] *** Images (375a31961170):
[creator]       index.docker.io/library/test-creator:latest
[creator] Reusing cache layer 'paketo-buildpacks/bellsoft-liberica:jdk'
...
Successfully built image test-creator
```

internal/build/phases_test.go
yaelharel: You're right. I'll remove it.

internal/build/phases_test.go
yaelharel: You're right. Removed it.

PR #600, opened 2020-05-01
Add tests for archive.IsZip()

dfreilich:
* Additionally, update the vendor directory, which seems to have been updated. 

Q: Should we use automation to update the vendor dir in the future? Is there typically?


jromero:
> Q: Should we use automation to update the vendor dir in the future? Is there typically?

1. imgutil and lifecycle have moved to "un-vendored".
2. imgutil has started using dependabot ([see PRs](https://github.com/buildpacks/imgutil/issues?q=label%3Adependencies+is%3Aclosed)).

Maybe we want to follow suit.

PR #602, opened 2020-05-01
Publish should also support default process

### LINKED ISSUE #599: Default process type is not set when running against a registry
### Summary
Currently, the default process type is only set if publish is false: https://github.com/buildpacks/pack/blob/82549aa7c8980ed7a42b0c8947c0a37d6d1edade/internal/build/phases.go#L191 (note the return on line 185)

We should set the default process type for both the registry and the daemon case.

---

##### Steps

Making some assumptions here...
1. `pack build <some-image> --default-process <some-process-type> --publish`
2. `pack inspect-image` - observe that the default process is not set

##### Current behavior
It is not set.

##### Expected behavior
It should be set.

zmackie:
Fixes #599

### DIFF
--- a/internal/build/phases.go
+++ b/internal/build/phases.go
@@ -7,8 +7,6 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/buildpacks/lifecycle/auth"
 	"github.com/google/go-containerregistry/pkg/authn"
-
-	"github.com/buildpacks/pack/internal/api"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/build/phases.go
```diff
@@ -183,31 +187,22 @@ func (l *Lifecycle) newExport(repoName, runImage string, publish bool, launchCac
 		)
 
 		return phaseFactory.New(configProvider), nil
-	}
+	} else {
```
natalieparellano, line 190: I can see why you added this... this `else` would've likely alerted us all to the initial problem. But now the linter is unhappy :/

natalieparellano:
Besides the failing linter these changes look good to me. Awesome that this is now tested :)

jromero:
@zmackie any follow up on this?

zmackie:
@jromero Should be ready to go. lint fixed.

jromero:
@zmackie seems to still be failing :/

You can try running `make prepare-for-pr` locally.

zmackie:
> @zmackie seems to still be failing :/
> 
> You can try running `make prepare-for-pr` locally.

Should be GTG according to `make prepare-for-pr` (although I think that `make` command is not working properly see:
```
-----------------
NOTICE: There are some files that have not been committed.
-----------------

On branch fix/default-proc-publish
Your branch is up to date with 'origin/fix/default-proc-publish'.

nothing to commit, working tree clean

-----------------
NOTICE: There are some files that have not been committed.
-----------------

```

PR #613, opened 2020-05-08
Refactor create builder

dfreilich:
I aimed to split a lot of the functionality into separate methods and make it significantly clearer, while hopefully increasing test coverage as well

### DIFF
--- a/create_builder.go
+++ b/create_builder.go
@@ -29,6 +29,26 @@ type CreateBuilderOptions struct {
 
 // CreateBuilder allows users to create a builder
 func (c *Client) CreateBuilder(ctx context.Context, opts CreateBuilderOptions) error {
+	if err := c.validateConfig(ctx, opts); err != nil {
+		return err
+	}
+
+	builder, err := c.createBaseBuilder(ctx, opts)
+	if err != nil {
+		return errors.Wrap(err, "failed to create builder")
+	}
+
+	if c.addBuildpacksToBuilder(ctx, opts, builder); err != nil {
+		return errors.Wrap(err, "failed to add buildpacks to builder")
+	}
+
+	builder.SetOrder(opts.Config.Order)
+	builder.SetStack(opts.Config.Stack)
+
+	return builder.Save(c.logger)
+}
+
+func (c Client) validateConfig(ctx context.Context, opts CreateBuilderOptions) error {
 	if err := validateBuilderConfig(opts.Config); err != nil {
 		return errors.Wrap(err, "invalid builder config")
 	}
@@ -37,25 +57,7 @@ func (c *Client) CreateBuilder(ctx context.Context, opts CreateBuilderOptions) e
 		return err
 	}
 
-	baseImage, err := c.imageFetcher.Fetch(ctx, opts.Config.Stack.BuildImage, !opts.Publish, !opts.NoPull)
-	if err != nil {
-		return errors.Wrap(err, "fetch build image")
-	}
-
-	builder, err := c.createBaseBuilder(ctx, opts, baseImage)
-	if err != nil {
-		return errors.Wrap(err, "failed to create builder")
-	}
-
-	err = c.addBuildpacksToBuilder(ctx, opts, builder)
-	if err != nil {
-		return errors.Wrap(err, "failed to add buildpacks to builder")
-	}
-
-	builder.SetOrder(opts.Config.Order)
-	builder.SetStack(opts.Config.Stack)
-
-	return builder.Save(c.logger)
+	return nil
 }
 
 func validateBuilderConfig(conf pubbldr.Config) error {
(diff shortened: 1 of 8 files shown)


create_builder.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/buildpacks/pack/internal/style"
 )
 
+// CreateBuilderOptions are options passed into CreateBuilder
```
zmackie, line 21: Love the godocs here.

pack.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/buildpacks/pack/internal/style"
 )
 
-func extractPackagedBuildpacks(ctx context.Context, pkgImageRef string, fetcher ImageFetcher, publish, noPull bool) (mainBP dist.Buildpack, depBPs []dist.Buildpack, err error) {
+func extractPackagedBuildpacksFromImage(ctx context.Context, pkgImageRef string, fetcher ImageFetcher, publish, noPull bool) (mainBP dist.Buildpack, depBPs []dist.Buildpack, err error) {
```
zmackie, line 13: [not blocking] function name is long, IMHO and the parameters tell you its from an image so that's redudant.

create_builder.go
```diff
@@ -40,16 +42,94 @@ func (c *Client) CreateBuilder(ctx context.Context, opts CreateBuilderOptions) e
...
+	}
+
+	if conf.Stack.RunImage == "" {
+		return errors.New("stack.run-image is required")
+	}
+
+	return nil
+}
+
+func (c Client) validateRunImageConfig(ctx context.Context, opts CreateBuilderOptions) error {
```
zmackie, line 79: we're mixing pointer and value receivers here. This should also be grouped with other `*Client` methods.

```suggestion
func (c *Client) validateRunImageConfig(ctx context.Context, opts CreateBuilderOptions) error {
```

create_builder.go
```diff
@@ -40,16 +42,94 @@ func (c *Client) CreateBuilder(ctx context.Context, opts CreateBuilderOptions) e
...
+		return errors.Wrap(err, "failed to add buildpacks to builder")
+	}
+
+	builder.SetOrder(opts.Config.Order)
+	builder.SetStack(opts.Config.Stack)
+
+	return builder.Save(c.logger)
+}
+
+func validateBuilderConfig(conf pubbldr.Config) error {
```
zmackie, line 63: Move below `*Client` methods.

zmackie: [not blocking] Also, this could be public and tested even though its internal? It could also be a method of `Config` ie `Config.Validate() error`?

create_builder.go
```diff
@@ -40,16 +42,94 @@ func (c *Client) CreateBuilder(ctx context.Context, opts CreateBuilderOptions) e
...
+				style.Symbol(stackID),
+				style.Symbol(img.Name()),
+			)
+		}
+	}
+
+	return nil
+}
+
+func (c Client) createBaseBuilder(ctx context.Context, opts CreateBuilderOptions, baseImage imgutil.Image) (*builder.Builder, error) {
```
zmackie:
```suggestion
func (c *Client) createBaseBuilder(ctx context.Context, opts CreateBuilderOptions, baseImage imgutil.Image) (*builder.Builder, error) {
```

create_builder.go
```diff
@@ -26,30 +27,100 @@ type CreateBuilderOptions struct {
...
+	if c.addBuildpacksToBuilder(ctx, opts, builder); err != nil {
+		return errors.Wrap(err, "failed to add buildpacks to builder")
+	}
+
+	builder.SetOrder(opts.Config.Order)
+	builder.SetStack(opts.Config.Stack)
+
+	return builder.Save(c.logger)
+}
+
```
natalieparellano, line 50: This is awesome! I love how clearly this reads now.

pack.go
dfreilich: Fair point! I was going to make a similar method to extractPackagedBuildpacksFromFile, which was why I renamed it, but I don't think I'll do that in the end

create_builder.go
dfreilich: Why should this be a pointer receiver? This isn't changing anything on the client, it's just validating that the information from it?

dfreilich: Never mind, it's much better, given what could be the size of the client struct. Will change!

create_builder.go
dfreilich: I moved it to be a method of the `StackConfig`. In terms of design, though, would it be better to have it there, or not, given that that's adding funcs to data objects?

jromero: @dfreilich I would be in favor of not adding logic to data types but everything else in this PR is a great improvement. This is not something that should block this PR and we can change it later.

PR #614, opened 2020-05-08
Add Arch Linux pack-cli and pack-cli-bin test

jromero:
- Update build task in Makefile for best practices
- Update release notes to include Arch Linux packages


.github/workflows/build.yml
```diff
@@ -142,24 +142,38 @@ jobs:
...
+            
+            On Linux you can use the one of the following methods of installation.
+            
+            ##### Arch Linux
+            
+            - [pack-cli](https://aur.archlinux.org/packages/pack-cli/)
+            - [pack-cli-bin](https://aur.archlinux.org/packages/pack-cli-bin/)
             
-            On Linux you can use the following command:
+            ##### Command
```
zmackie, line 162: If I'm reading this right there are two paths here: arch and others (which I take to mean "other distros").
If that's right, I think the command should be nested under "Others".

jromero: If we put command under others I would feel like it would be inconsistent since it's not applicable to Windows. Hope is to add other distros here soon.

PR #628, opened 2020-05-15
Add test for Version command

dfreilich:

### DIFF
--- a/internal/commands/version_test.go
+++ b/internal/commands/version_test.go
@@ -1,0 +1,37 @@
+package commands_test
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


internal/commands/version_test.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func testVersionCommand(t *testing.T, when spec.G, it spec.S) {
+	var (
+		command *cobra.Command
+		outBuf  bytes.Buffer
+	)
+
+	it.Before(func() {
+		command = commands.Version(logging.NewLogWithWriters(&outBuf, &outBuf), cmd.Version)
```
jromero, line 28: Ideally we should hardcode an expected value here instead of using a build time constant.

PR #638, opened 2020-05-19
Configure codecov

jromero:
- Makes codecov wait for 4 (number of parallel unit tests executed) reports before notifying/posting coverage
- Allow for coverage fudging up to 1% via threshold

Also:

- Don't upload windows artifacts twice.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,21 +15,25 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos, linux, windows-lcow, windows-wcow]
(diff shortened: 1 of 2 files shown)


jromero:
The degradation in coverage appears to be related to how Windows coverage is reported based on what the report states. 


For example, [config_reader.go](https://codecov.io/gh/buildpacks/pack/pull/638/src/buildpackage/config_reader.go) on Windows reports coverage change on `}` braces. Likely related to [this](https://docs.codecov.io/docs/fixing-reports#go) which appears to not be happening on Windows.

image

PR #640, opened 2020-05-20
Add acceptance test for trust-builder

natalieparellano:
This ensures the command is exported.


natalieparellano:
There appears to be a flake in the package buildpack tests that we are encountering here.

PR #648, opened 2020-05-22
Fix PREVIOUS_PACK_PATH when there is a draft release

natalieparellano:


natalieparellano:
Umm I think something went wrong... this PR has more commits than I intended :/

PR #649, opened 2020-05-22
Fix PREVIOUS_PACK_PATH when there is a draft release

natalieparellano:

### DIFF
--- a/acceptance/github_asset_fetcher.go
+++ b/acceptance/github_asset_fetcher.go
@@ -196,12 +196,23 @@ func (f *GithubAssetFetcher) FetchReleaseVersion(owner, repo string, n int) (str
 	}
 
 	// get all release versions
-	releases, _, err := f.githubClient.Repositories.ListReleases(f.ctx, owner, repo, nil)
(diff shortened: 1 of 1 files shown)


acceptance/github_asset_fetcher.go
```diff
@@ -196,14 +196,25 @@ func (f *GithubAssetFetcher) FetchReleaseVersion(owner, repo string, n int) (str
...
 	}
-	if len(releases) == 0 {
+	if len(rawReleases) == 0 {
 		return "", fmt.Errorf("no releases found for %s/%s", owner, repo)
 	}
 
+	// exclude drafts
+	var releases []*github.RepositoryRelease
+	for _, release := range rawReleases {
+		if !*release.Draft {
```
jromero, line 210: Would this be better written as `release.Draft != nil` if that's the same thing?

I'm having trouble mapping `!*`.

dfreilich: @natalieparellano To unblock a PR, I pushed that change to this branch.

jromero: @dfreilich thanks for attempting the change. I saw that the change didn't yield the results that we wanted. I'm taking it in as is. We can decipher it later. :D

PR #644, opened 2020-05-21
Fail when using an untrusted builder and no lifecycle image available

natalieparellano:
Also consider lifecycle image supported when version is 0.6.1 (this version
of the lifecycle does not require the user to be valid in the image, so a
lifecycle image is possible).

Signed-off-by: Natalie Arellano <narellano@vmware.com>

PR #653, opened 2020-05-25
Update log output for creator scenario

### LINKED ISSUE #641: Remove [creator] eye catcher
### Description
With the recent update to use a single container for all phases, there is now a single `[creator]` eye-catcher on every single line.  This makes the output redundant and should be removed.

### Additional context
Screen Shot 2020-05-20 at 11 56 33

jromero:
Resolves #641 

#### Note
The log output will look a little bland when using creator with lifecycle 0.7.5 due to https://github.com/buildpacks/lifecycle/issues/299 but should look much better for the following lifecycle version.

### Individual Phases (lifecycle 0.7.5)
Screen Shot 2020-05-25 at 11 18 24 AM

### Creator (lifecycle 0.7.5)
Screen Shot 2020-05-25 at 11 17 44 AM

### Creator (lifecycle master - see https://github.com/buildpacks/lifecycle/pull/300)
Screen Shot 2020-05-25 at 11 30 08 AM

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -739,27 +739,14 @@ func testAcceptance(
 						assertMockAppRunsWithOutput(t, repoName, "Launch Dep Contents", "Cached Dep Contents")
 
 						t.Log("restores the cache")
-						if creatorSupported {
-							h.AssertContainsMatch(t, output, `(?i)\[creator] Restoring data for "simple/layers:cached-launch-layer" from cache`)
-							h.AssertContainsMatch(t, output, `(?i)\[creator] Restoring metadata for "simple/layers:cached-launch-layer" from app image`)
-						} else {
-							h.AssertContainsMatch(t, output, `(?i)\[restorer] Restoring data for "simple/layers:cached-launch-layer" from cache`)
-							h.AssertContainsMatch(t, output, `(?i)\[analyzer] Restoring metadata for "simple/layers:cached-launch-layer" from app image`)
-						}
+						h.AssertContainsMatch(t, output, `(?i)Restoring data for "simple/layers:cached-launch-layer" from cache`)
+						h.AssertContainsMatch(t, output, `(?i)Restoring metadata for "simple/layers:cached-launch-layer" from app image`)
 
 						t.Log("exporter reuses unchanged layers")
-						if creatorSupported {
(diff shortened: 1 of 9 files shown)


acceptance/acceptance_test.go
```diff
@@ -739,27 +739,14 @@ func testAcceptance(
...
 
 						t.Log("restores the cache")
-						if creatorSupported {
-							h.AssertContainsMatch(t, output, `(?i)\[creator] Restoring data for "simple/layers:cached-launch-layer" from cache`)
-							h.AssertContainsMatch(t, output, `(?i)\[creator] Restoring metadata for "simple/layers:cached-launch-layer" from app image`)
-						} else {
-							h.AssertContainsMatch(t, output, `(?i)\[restorer] Restoring data for "simple/layers:cached-launch-layer" from cache`)
-							h.AssertContainsMatch(t, output, `(?i)\[analyzer] Restoring metadata for "simple/layers:cached-launch-layer" from app image`)
-						}
+						h.AssertContainsMatch(t, output, `(?i)Restoring data for "simple/layers:cached-launch-layer" from cache`)
```
natalieparellano, line 742: I love this as it's much cleaner. I worry that we're losing the ability to assert that the creator (vs. not creator) was actually run. This is unit tested pretty well, so maybe it's okay. But for example the assertions here, I wouldn't want to change: https://github.com/buildpacks/pack/pull/644/files#diff-38e2f40dfce97bbc5beadcb903ffe700R693-R696.

acceptance/acceptance_test.go
```diff
@@ -1349,18 +1309,13 @@ func testAcceptance(
...
+								err := cmd.Wait()
+								h.AssertNotNil(t, err)
 
 								if creatorSupported {
-									go terminateAtStep(t, cmd, &buf, "[creator]")
-									err := cmd.Wait()
-									h.AssertNotNil(t, err)
-
-									h.AssertContains(t, buf.String(), "[creator]")
+									h.AssertNotContains(t, buf.String(), "[detector]")
```
natalieparellano, line 1317: Ah well, to my previous comment, this is nice.

acceptance/acceptance_test.go
jromero: Yeah, I went through the same thought process but as you noted, we already have a specific acceptance set for that.

nebhale:
@jromero I'm reviewing this right now.

jromero:
Thank you @nebhale for the review!

PR #659, opened 2020-05-27
--no-pull also applies to the lifecycle image

### LINKED ISSUE #658: --no-pull should apply to the lifecycle image
`pack build --no-pull --builder not-trusted-builder` should not pull the lifecycle image. It is currently always pulled, with or without `--no-pull` flag.

When `--no-pull` is set and the lifecycle image is not available locally, the build should error out.

dgageot:
Fixes #658

### DIFF
--- a/build.go
+++ b/build.go
@@ -181,7 +181,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 				ctx,
 				fmt.Sprintf("%s:%s", lifecycleImageRepo, lifecycleVersion.String()),
 				true,
-				true,
+				!opts.NoPull,
 			)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


natalieparellano:
Thanks for catching this! I wonder if we could add a test for this - trying this out on master, adding the following lines to build_test.go at line ~1490 (under `uses the local builder and run images without updating`) produces a failure:

```
args = fakeImageFetcher.FetchCalls[fakeLifecycleImage.Name()]
					h.AssertEq(t, args.Daemon, true)
					h.AssertEq(t, args.Pull, false)
```

jromero:
Given the severity of the issue and the trivial change I would like to get this in without blocking it on additional tests.

dgageot:
I'll add a test in another PR. Thanks to both of you!

PR #661, opened 2020-05-27
Let's be pedantic with import groups

dgageot:


dfreilich:
Is this something that `go-format`/`make format` should have done? Should we update it?

jromero:
> Is this something that `go-format`/`make format` should have done? Should we update it?

Yes, should. We've tweaked it a few times and have gotten mixed results. It would be worth revisiting.

Issue #261, opened 2019-08-16
Add created-by metadata to builder


ekcasey:
## Motivation

`pack` has changed it's builder creation process over time. It would be nice for users to know which version of `pack` created a given builder was so that they:

1. know what metadata to expect on the builder
1. know the organizational structure of the image FS (Example: lifecycle binary paths)
1. know whether a builder is compatible to run with another version of `pack`.
1. can provide this information in bug reports involving a custom builder
1. can satisfy their curiosity

### `pack create-builder`

**Given** I have `pack` version `1.2.3`
**When** I run `pack create-builder <builder-image>`
**Then** The `<builder-image>` will have the following information it it's `io.buildpacks.builder.metadata` label
```json
{... "createdBy": {"name": "Pack CLI", "version": "1.2.3"}}}
```

### `pack inspect-builder`
**Given** a builder `<builder-image>` with `createdBy` metadata in the above format
**When** I run `pack inspect-builder <builder-image>`
**Then** I will see the following information in the output
```
$pack inspect-builder <builder-image>
...
Created By: 
  Name: Pack CLI
  Version: 1.2.3
```

**Given** a builder `<builder-image>` w/o `createdBy` in it's metadata
**When** I run `pack inspect-builder <builder-image>`
**Then** `Created By:` will not appear in the output

ameyer-pivotal:
Accepted.

PR #663, opened 2020-05-27
Make command creation accessible by other tools

dfreilich:
To support https://github.com/buildpacks/docs/issues/119, and adding to the work of #399 , this pulls the creation of the pack cobra command into a separately accessible func. 

This will make it easier to generate docs for the command, based on this: https://github.com/spf13/cobra/blob/master/doc/md_docs.md#generate-markdown-docs-for-the-entire-command-tree, so that we can simply do:
```
pack := cli.NewPackCommand(logger)
err := doc.GenMarkdownTree(pack, "./")
if err != nil {
		log.Fatal(err)
}
```


yaelharel:
Hi @dfreilich, your changes look good to me but since I'm new to this project, I guess it'll be better if someone else will review them as well.

cli/pack_command.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/buildpacks/pack/buildpackage"
+	"github.com/buildpacks/pack/cmd"
+	"github.com/buildpacks/pack/internal/commands"
+	"github.com/buildpacks/pack/internal/config"
+	clilogger "github.com/buildpacks/pack/internal/logging"
+	"github.com/buildpacks/pack/logging"
+)
+
+var (
+	packClient pack.Client
```
jromero, line 18: This might be carried over from the move to this new file but what's the purpose of setting the package level variable? It doesn't appear to be used.

cli/pack_command.go
```diff
@@ -0,0 +1,110 @@
+package cli
```
jromero, line 1: I like the overall extraction of the command building to this new function. I am curious as to why `cli` and not `cmd`. It would appear as though the convention is more towards `cmd` and consumers would still be able to achieve the same.

dfreilich: We already have a cmd package, which seemingly functions just to provide `cmd.Version`. That is then used in a bunch of places, so we can't this function in the same package as the `cmd.Version` var, because of import cycles. I can move Version to another package, but I was trying to minimize what I touched inside the code. What do you think?

cli/pack_command.go
dfreilich: Hmm. I took that from the existing code, but it looks like it was created originally to allow for initializing it in PreRun. I can try without doing that though.

dfreilich:
I think so. It anyways makes it the startup a bit cleaner, and it makes us able to generate docs wherever we want to.

cli/pack_command.go
jromero: > Perhaps I'll move the Version to the pack package (so it would be pack.Version, which anyways sees clearer), and move this into cmd...

Version is a build time value, it doesn't make sense in `pack` package as that is the scope for the pack library.

I reviewed where `cmd.Version` is getting used and your statement is correct. Yet, I feel like the _right_ approach would be to propagate the version to the `Client` from `cmd` so that it's properly set in all cases, allows for configuration (other tools using pack as a library), and solves the cyclic dependency.

I'd have to investigate if there any assumptions being made about the `Client` name and version but I believe it was solely introduced for traceability/compliance.

jromero: > Version is a build time value that applies to the cli only, it doesn't make sense in pack package as that is the scope for the pack library.

We can discuss whether this makes sense to change and possible hardcode the value instead of providing it in at build time in which case it would be accurate for both cli and lib use cases.

jromero: Found the original feature request: https://github.com/buildpacks/pack/issues/261

jromero: TLDR: Moving it to the top-level `pack` package makes sense and instead of using some external build-time value we should hardcode it (can be done later).

> 1. know what metadata to expect on the builder
> 1. know the organizational structure of the image FS (Example: lifecycle binary paths)
> 1. know whether a builder is compatible to run with another version of pack.
> 1. can provide this information in bug reports involving a custom builder
> 1. can satisfy their curiosity

Based on this information:

1. Builders should be an extension spec instead of relying on pack version (@ekcasey 😄)
2. We should hardcode the value so that the pack version is also applied when used as a lib.

Sorry for the long windedanswerr @dfreilich.

dfreilich: Interesting. I know that @sclevine had been a fan of build-time versions, so I wonder what he thinks.

dfreilich: But, to clarify, you think I should move the `Version` to the `pack` package, and then move the `NewPackCommand(...)` to `cmd`?

cli/pack_command.go
dfreilich: It was added in here, when we first created pack.Client: https://github.com/buildpacks/pack/commit/237719e6220f4ced26a1670ac3bfb1a944bee8c7#diff-d9d0491c5f9a7fee85ac4da6dc4f2058. It works without it though.

cli/pack_command.go
dfreilich: @jromero I assumed that was what you meant, and went ahead with it (I also had to inject the version into the builders Version method, in order to prevent cycles). Let me know what you think!

internal/builder/builder.go
```diff
@@ -259,7 +258,7 @@ func (b *Builder) SetStack(stackConfig builder.StackConfig) {
 }
 
 // Save saves the builder
-func (b *Builder) Save(logger logging.Logger) error {
+func (b *Builder) Save(logger logging.Logger, version string) error {
```
jromero, line 261: With the caveat that the architecture of `Builder` has been problematic and due for some refactor, I think at minimum to make this change acceptable we should change it from `..., version string)` to `..., creator CreatorMetadata)` so that it's up to the caller to provide the entire set of information for `CreatedBy`.

PR #667, opened 2020-05-28
Send repository dispatch to docs on new release

dfreilich:
* Helps with buildpacks/docs#115
* Helps with buildpacks/docs#150


dfreilich:
In order for the [docs](github.com/buildpacks/docs) to know when a new version is released (and thereby update the install pack links ([here](https://buildpacks.io/docs/install-pack/#linux)) to the right link), the simplest method is for pack to send a repository dispatch to docs.

dfreilich:
@jromero I think that when we work with `release:published`, it won't trigger for a draft being created. [This](https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release) says that, as does [this stackoverflow](https://stackoverflow.com/questions/59319281/github-action-different-between-release-created-and-published)

.github/workflows/delivery-release-dispatch.yml
```diff
@@ -0,0 +1,18 @@
...
+    types:
+      - published
+
+jobs:
+  release-dispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch to Docs
+        uses: peter-evans/repository-dispatch@v1
+        if
```
jromero, line 14: This looks incomplete.

dfreilich: That's... a disturbing mistake I left in. Took it out!

dfreilich: From my understanding, though, this should be the only step needed. There's no need to actually load the repository, all we really need to do is be triggered at the appropriate time and send a dispatch to `buildpacks/docs`

PR #673, opened 2020-06-04
Rewrite of acceptance test suite

simonjjones:
This PR includes a number of very broad changes to our existing acceptance test suite, but doesn't include any functional changes to what is being tested.

## Draft PR Purpose
The purpose of this PR is primarily to solicit feedback on the possible direction of these changes. While it can be merged as is, it may prove simpler to break out into smaller PRs to provide simpler reviews. As part of any review it may be useful to highlight which abstractions would be considered higher priority for you.

## Overview
The guiding principles of these changes are to make acceptance tests:

- Easier to write and change
- Easier to read and understand
- Separable into smaller files divided by function

To achieve this I have attempted to abstract as much boilerplate and repeated, preparation and functionality into clearly named objects implementing well defined interfaces.

The implementations of these interfaces are immutable to avoid any unwanted changes between test cases. For the sake of efficiency we do share consistent stacks, builders and packages between tests as we did previously using the suitemanager's `RunOnce*` methods.

We can divide the work into the following areas, which could make good candidates for separation into smaller PRs

- Tidy up of input parsing, run combination generation, pre-requisite fetching and asset management throughout the suite
- Assertion Managers provide assertion specific helpers instantiated with names such as `assert`, `assertOutput` & `assertImage` that provide very context specific assertions such as `ReportsSuccessfulRebase()`
- Pack and lifecycle as service objects instantiated with details of which binaries and fixtures they use (current/previous/default) and providing a variety of helper methods to execute binaries or provide information about these components
- Manager abstractions for stacks, buildpacks, builders and packages provide methods to safely ensure that default versions of pre-requisites exist during a test run, and where useful can be used to create additional versions for use in specific test cases

## Drawbacks
In order to provide repeatable functionality in as simple a way as possible to test cases, it has proven necessary to move and increase the complexity of the objects that implement this functionality on behalf of the test cases. This will cause new _concepts_ to have a higher barrier to writing their first acceptance tests while maintaining the existing patterns.

## Reviewing
If you have bandwidth to review the whole thing, then great, however realistically I understand this is a huge change. I would advise starting with `acceptance_test.go` as this is the area we should be most familiar with. If we can validate which abstractions consumed within the main tests are or aren't useful, we can then go on to validate their actual implementations. I'm available to talk about any questions/concerns/etc.


acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
 
 	when("set-default-builder", func() {
 		it("sets the default-stack-id in ~/.pack/config.toml", func() {
-			output := h.Run(t, subjectPack("set-default-builder", "gcr.io/paketo-buildpacks/builder:base"))
-			h.AssertContains(t, output, "Builder 'gcr.io/paketo-buildpacks/builder:base' is now the default builder")
+			builderName := "gcr.io/paketo-buildpacks/builder:base"
+
+			output := pack.SuccessfulRunWithOutput("set-default-builder", builderName)
+
+			assert.NewOutputAssertionManager(output).ReportsSettingDefaultBuilder(builderName)
```
dfreilich, line 154: nit: I would add in an assertion that it exists in the `config.toml` as well, for completeness

acceptance/managers/buildpack_manager.go
```diff
@@ -0,0 +1,42 @@
...
+	buildpackAPI := api.MustParse(builder.DefaultBuildpackAPIVersion).String()
+
+	return BuildpackManager{
+		testObject: t,
+		assert:     assert,
+		source:     filepath.Join("testdata", "mock_buildpacks", buildpackAPI),
+	}
+}
+
+func (b BuildpackManager) PlaceBuildpacksInDir(destination string, buildpacks ...components.TestBuildpack) {
```
dfreilich, line 32: I felt like this name was a bit odd; I would suggest something more like `CopyBuildpacksToDir`

acceptance/components/pack_executor.go
```diff
@@ -0,0 +1,392 @@
...
+
+func (e *PackExecutor) RunWithCombinedOutput(name string, args ...string) (string, error) {
+	e.testObject.Helper()
+
+	output, err := e.cmd(name, args...).CombinedOutput()
+
+	return string(output), err
+}
+
+func (e *PackExecutor) SuccessfulRunWithOutput(name string, args ...string) string {
```
dfreilich, line 96: I know that this is a common pattern, but when I first read it, I thought that the args provided were what we were expecting the output to be, and not the args to pack, so I was a bit confused.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+				assert.Contains(output, fmt.Sprintf("Successfully created package '%s'", packageName))
+				defer containerManager.RemoveImages(packageName)
+
+				assert.ImageExistsLocally(packageName)
 			})
-		})
 
-		when("--format image", func() {
-			it("creates the package", func() {
+			it("creates the package as a local image", func() {
```
dfreilich, line 222: This test doesn't seem appropriate for the test, given that it's testing more that package-buildpack can work with a package buildpack as one of the dependencies.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
 				it("publishes image to registry", func() {
-					nestedPackageName := packageBuildpackRemotely(filepath.Join(tmpDir, "package.toml"))
-					defer h.DockerRmi(dockerCli, nestedPackageName)
-					aggregatePackageToml := generateAggregatePackageToml("simple-layers-parent-buildpack.tgz", nestedPackageName)
+					nestedPackageName := registryConfig.RepoName(newPackageName())
+					assert.PackagePublished(
+						nestedPackageName,
+						pack.PackageBuildpack(nestedPackageName, simplePackageConfigPath, "--publish"),
+					)
+					defer containerManager.RemoveImages(nestedPackageName)
```
dfreilich, line 254:
```suggestion
assert.ImageExistsOnlyInRegistry(nestedPackageName, registryConfig)
```

acceptance/managers/container_manager.go
```diff
@@ -0,0 +1,278 @@
...
+func (c TestContainerManager) HostOS() string {
+	c.testObject.Helper()
+
+	daemonInfo, err := c.dockerCli.Info(context.Background())
+	c.assert.Nil(err)
+
+	return daemonInfo.OSType
+}
+
+func (c TestContainerManager) CleanupTaskForImageByName(imageName string) func() {
```
dfreilich, line 52: I found the pattern of calling this and then `defer`ing the returned func a bit unnecessary; it could have just defered a CleanupImageByName func which did all of this.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
 		it("displays suggested builders", func() {
-			output := h.Run(t, subjectPack("suggest-builders"))
-
-			h.AssertContains(t, output, "Suggested builders:")
-			h.AssertMatch(t, output, `Google:\s+'gcr.io/buildpacks/builder'`)
-			h.AssertMatch(t, output, `Heroku:\s+'heroku/buildpacks:18'`)
-			h.AssertMatch(t, output, `Paketo Buildpacks:\s+'gcr.io/paketo-buildpacks/builder:base'`)
-			h.AssertMatch(t, output, `Paketo Buildpacks:\s+'gcr.io/paketo-buildpacks/builder:full-cf'`)
-			h.AssertMatch(t, output, `Paketo Buildpacks:\s+'gcr.io/paketo-buildpacks/builder:tiny'`)
+			output := pack.SuccessfulRunWithOutput("suggest-builders")
```
natalieparellano, line 130: When I see `SuccessfulRunWithOutput` my first thought is that "suggest-builders" is the output we're expecting to receive. I wonder if we could just call this `SuccessfulRun`.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
 
 		it.Before(func() {
 			h.SkipIf(t,
-				!packSupports(packPath, "package-buildpack"),
+				!pack.Supports("package-buildpack"),
 				"pack does not support 'package-buildpack'",
 			)
 
-			h.SkipIf(t, dockerHostOS() == "windows", "These tests are not yet compatible with Windows-based containers")
+			containerManager = managers.NewTestContainerManager(t, assert, dockerCli)
```
natalieparellano, line 183: Curious why we're instantiating `containerManager` here instead of passing it in at the suite level?

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
-				defer h.DockerRmi(dockerCli, nestedPackageName)
-				assertImageExistsLocally(nestedPackageName)
-
-				t.Log("package w/ buildpacks and packages")
-				aggregatePackageToml := generateAggregatePackageToml("simple-layers-parent-buildpack.tgz", nestedPackageName)
-				packageName := packageBuildpackLocally(aggregatePackageToml)
-				defer h.DockerRmi(dockerCli, packageName)
-				assertImageExistsLocally(packageName)
+				nestedPackageName := newPackageName()
+				assert.PackageCreated(
```
natalieparellano, line 225: I believe `pack.PackageBuildpack` is the "thing" we're testing here, but it's a little hard to glean this from having it occur within the call to `assert.PackageCreated`. I wonder if this could be made clearer.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+					nestedPackageName,
+					pack.PackageBuildpack(nestedPackageName, simplePackageConfigPath),
+				)
+				defer containerManager.RemoveImages(nestedPackageName)
+
+				assert.ImageExistsLocally(nestedPackageName)
+
+				aggregatePackageConfigFile := pack.AggregatePackageFixture(
```
natalieparellano, line 233: At first glance it's a little unclear to me what `pack.AggregatePackageFixture` is doing... I had to click around a bit and am still not 100% sure. It's a little strange to me that `PackExecutor` is doing test-type things (such as filling in templates) and "real" things (such as calling `pack package-buildpack`). I wonder if separating those concerns could make things a bit easier to follow.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+
+			cacheCleanupRequired = true
+		})
+
+		it.After(func() {
+			if cacheCleanupRequired {
+				h.DockerRmi(dockerCli, appImageName)
+				ref, err := name.ParseReference(appImageName, name.WeakValidation)
+				assert.Nil(err)
+				cacheImage := cache.NewImageCache(ref, dockerCli)
```
natalieparellano, line 560: Maybe it's worth adding a comment explaining what we are doing here? It seems a little strange to be creating a new thing in the test cleanup, but I assume that is so we can call `Clear` on it.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+			t.Log("rebuild")
+			output = pack.SuccessfulRunWithOutput("build", appImageName, "-p", appPath)
+			cleanupRebuiltAppImageTask := containerManager.CleanupTaskForImageByName(appImageName)
+			defer cleanupRebuiltAppImageTask()
+
+			assertOutput = assert.NewOutputAssertionManager(output)
+			assertOutput.ReportsSuccessfulImageBuild(appImageName)
+			assertOutput.ReportsSelectingRunImageMirrorFromLocalConfig(localRunImageMirror)
+
+			assertMixedOutput := assert.NewMixedComponentOutputAssertionManager(output, mixedComponentMessenger)
```
natalieparellano, line 609: The "mixed output" part is a bit hard to follow.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+			appPath := filepath.Join("testdata", "mock_app.zip")
+			output := pack.SuccessfulRunWithOutput("build", appImageName, "-p", appPath)
+
+			assert.NewOutputAssertionManager(output).ReportsSuccessfulImageBuild(appImageName)
+
+			appImageCleanupTask := containerManager.CleanupTaskForImageByName(appImageName)
+			defer appImageCleanupTask()
+		})
+
+		when("builder is untrusted", func() {
```
natalieparellano, line 663: There is some stuff around "lifecycle image" that appears to be missing here (see https://github.com/buildpacks/pack/pull/673/files#diff-38e2f40dfce97bbc5beadcb903ffe700L723) unless it's captured elsewhere.

acceptance/acceptance_test.go
```diff
@@ -5,221 +5,91 @@ package acceptance
...
-	return len(p), nil
-}
-
 func TestAcceptance(t *testing.T) {
 	var err error
 
 	h.RequireDocker(t)
 	rand.Seed(time.Now().UTC().UnixNano())
 
 	dockerCli, err = client.NewClientWithOpts(client.FromEnv, client.WithVersion("1.38"))
```
yaelharel, line 42: Don't we prefer to create a constant variable for the version (`1.38`)?
In this way we can have all of the constants in one place and if something is changed, we don't need to search for them in all of the file.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+
+				output, err := createBuilderPack.RunWithCombinedOutput(
+					"create-builder",
+					"--no-color",
+					registry.RepoName(fmt.Sprintf("test/builder-%s", h.RandString(8))),
+					"-b", builderConfigPath,
+				)
+
+				assert.NotNil(err)
+				assert.NewOutputAssertionManager(output).ReportsWindowsContainersExperimental()
```
yaelharel, line 514: Can you please explain why we need this function? Why can't we do `assert.Contains(output, "Windows containers support is currently experimental")`?

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+			testBuilder := builderManager.EnsureComboBuilderExists()
+			pack.SuccessfulRun("set-default-builder", testBuilder.Name())
+			if pack.Supports("trust-builder") {
+				pack.SuccessfulRun("trust-builder", testBuilder.Name())
+			}
+
+			appPath := filepath.Join("testdata", "mock_app")
+			output := pack.SuccessfulRunWithOutput("build", appImageName, "-p", appPath)
+			cleanupAppImageTask := containerManager.CleanupTaskForImageByName(appImageName)
+			defer cleanupAppImageTask()
```
yaelharel, line 580: Is there a difference between running the cleanup as part of `it.After...` and setting it under `defer`?
Sometimes you choose the first option and sometimes the second one.
I'm not sure what is better.
You'll probably less likely forget to add the `defer` but I think the code is clearer when all of the "after commands" appear together.

yaelharel: On second thought, it's clear, in this case, why you have the `defer` and not the `it.After` but my question is not specifically to this case.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+				cacheImage := cache.NewImageCache(ref, dockerCli)
+				buildCacheVolume := cache.NewVolumeCache(ref, "build", dockerCli)
+				launchCacheVolume := cache.NewVolumeCache(ref, "launch", dockerCli)
+				cacheImage.Clear(context.TODO())
+				buildCacheVolume.Clear(context.TODO())
+				launchCacheVolume.Clear(context.TODO())
+			}
+		})
+
+		it("creates a runnable, rebuildable image on daemon from app dir", func() {
```
yaelharel, line 569: This `it` is really long. What do you think about putting each "step" in a different function?

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
 
-			when("experimental is enabled", func() {
+			when("the network mode is not provided", func() {
+				it("reports that build and detect are online", func() {
+					stackManager.EnsureDefaultStackCreated()
+					testBuilder := builderManager.EnsureComboBuilderExists()
+					pack.SetDefaultTrustedBuilder(testBuilder.Name())
 
-				var packCreateBuilderHome string
+					buildpackManager.PlaceBuildpacksInDir(tmpDir, components.InternetCapableBuildpack)
```
yaelharel, line 708: I think you can move these lines to `it.Before`

yaelharel: Moreover, the first three lines repeat many times in the tests. Maybe it's worth creating a function for them.

acceptance/acceptance_test.go
simonjjones: I intentionally haven't added any tests or functionality in this PR, my goal was to present the exact set of tests that exist on main. We should add this test, but probably as a different PR

acceptance/managers/buildpack_manager.go
simonjjones: The trouble with this is there are other ways of preparing buildpacks than copying. Adding this interface is intended to simplify testing all the different ways buildpacks can be prepared. I agree it's a bit weird though, the directory that's passed in is ignored for package buildpacks, and may not make sense for other buildpacks. Because the directories are generally made after the buildpacks are instantiated it makes it kind of hard to get it to the buildpack concrete implementation at the time it's needed while keeping the objects immutable. Happy to explore alternative patterns, or even an alternative name, but I don't think it should be `Copy`

acceptance/components/pack_executor.go
simonjjones: What might have made it easier to understand the intent and keep the brevity?

The interface this provides gives us 
```
output := pack.SuccessfulRunWithOutput("build", "app-name", "-B", "builder", ...)
```

from the perspective of the tests consuming the interface most things I can think of become longer
```
output := pack.CommandWithArgs("build", "app-name", "-B", "builder", ...).RunSuccessfullyReturningOutput()
```
would be more explicit.

Would renaming this to `SuccessfulRunReturningOutput()` be clearer? It adds another few characters to an already long command name.

acceptance/acceptance_test.go
simonjjones: I'm not sure I understand the concern? Do you think the test shouldn't exist? Have I misunderstood the test and renamed it incorrectly? I believe this test exists in the current suite, so I wouldn't want to remove it as part of this PR, but happy to do so in another PR if it's not appropriate.

acceptance/acceptance_test.go
simonjjones: We don't currently test this, so again, in service of the goal of achieving a like for like rewrite I'd prefer to add this in another PR.

acceptance/managers/container_manager.go
simonjjones: This is tricky, and I think I may have introduced some leakage through inconsistent application of `defer`ing the returned method.

What I _think_ is happening, is that some of our `pack` invocations will create copies of images and in doing so rename the existing image. This task effectively transforms that name into an ID which is then encapsulated in the `defer`red function and _should_ ensure we clean up the original images as well as the copied ones.

I'm not certain that's what's happening though, but this approach does mirror what's happening in our existing tests.

I think I have a couple of areas to investigate out of this:
* Are my assumptions around image renaming accurate, or have I simply inherited existing poor patterns
* If this is necessary, how can we accurately communicate the intent of returning a deferrable function rather than the more obvious approach of deferring the `CleanupImageByName`
* Again, if this is necessary, how can ensure that people adding new tests don't follow the *obvious* and on the face of it correct path (which I have done in this PR many times)
* Do we need to bake in stronger guarantees around our cleanup into our test suite?

.github/workflows/compatibility.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         pack_kind: [current, previous]
         create_builder_kind: [current, previous]
-        lifecycle_kind: [current, previous]
+        lifecycle_kind: [default, previous]
```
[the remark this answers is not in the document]
simonjjones: The underlying implementation of what `default` means here is the same as what `current` is in the existing tests. I wasn't aware that there's actually a distinction. I believed that our intent was to not test the actual `current` of lifecycle because that could break and cause PRs to be blocked for reasons outside of contributor's control.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
-	// subjectPack creates a pack `exec.Cmd` based on the current configuration
-	subjectPack := func(name string, args ...string) *exec.Cmd {
-		return packCmd(packHome, packPath, name, args...)
-	}
-
 	it.Before(func() {
-		var err error
-		packHome, err = ioutil.TempDir("", "buildpack.pack.home.")
-		h.AssertNil(t, err)
+		pack = managers.NewPackExecutor(t, assetManager, managers.Current, registryConfig, assert)
```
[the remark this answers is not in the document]
simonjjones: Yes, they actually have slightly different signatures, mainly around `components` not wanting to be interested in things like `ComboValue` and `AssetManager`. The `NewPackExecutor` in `managers` does some small transformations that feel more _managery_. The `NewPackExecutor` in `components` is a far simpler instantiation method.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
+			h.SkipIf(t, containerManager.HostOS() == "windows", "These tests are not yet compatible with Windows-based containers")
 
 			var err error
 			tmpDir, err = ioutil.TempDir("", "package-buildpack-tests")
-			h.AssertNil(t, err)
+			assert.Nil(err)
 
-			h.CopyFile(t, filepath.Join(packFixturesDir, "package.toml"), filepath.Join(tmpDir, "package.toml"))
+			simplePackageConfigPath = filepath.Join(tmpDir, "package.toml")
+			h.CopyFile(t, pack.FixtureMustExist("package.toml"), simplePackageConfigPath)
```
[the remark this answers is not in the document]
simonjjones: Yes, these locations are defined by the version of pack that's instantiated, along with the location of the executable so it felt right to tie the ideas together.

acceptance/acceptance_test.go
```diff
@@ -229,1905 +99,1417 @@ func testWithoutSpecificBuilderRequirement(
...
 			)
-			h.AssertNil(t, err)
 		})
 
 		it.After(func() {
-			h.AssertNil(t, os.RemoveAll(tmpDir))
+			assert.Nil(os.RemoveAll(tmpDir))
 		})
 
-		packageBuildpackLocally := func(absConfigPath string) string {
```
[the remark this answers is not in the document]
simonjjones: There are two differences between the existing `packageBuildpackLocally` and `packageBuildpackRemotely`:
* The `--publish` flag is passed when packaging remotely
* The output asserted against will include the word `created` locally and `published` remotely.

The implementation of this now uses `pack.PackageBuildpack` and the tests themselves decide whether to pass in the `--publish` flag, actually removing a level of abstraction/obfuscation.

Our assertion manager includes `PackageCreated` and `PackagePublished` to address the distinction in outputs.

acceptance/acceptance_test.go
```diff
@@ -5,221 +5,91 @@ package acceptance
...
-	defaultCompilePackVersion = "0.0.0"
-	defaultPlatformAPIVersion = "0.3"
-)
-
 var (
-	dockerCli      client.CommonAPIClient
-	registryConfig *h.TestRegistryConfig
-	suiteManager   *SuiteManager
+	dockerCli    *client.Client
+	suiteManager *managers.SuiteManager
```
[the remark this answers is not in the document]
simonjjones: It probably doesn't, but I had to draw the line somewhere

acceptance/acceptance_test.go
simonjjones: @dfreilich had a similar concern, so this is obviously a problem. I added some commentary to his [comment](https://github.com/buildpacks/pack/pull/673#discussion_r442904946).

I really want to capture the intent that there is a return value, and it is the output. I wouldn't expect `SuccessfulRun` to return anything.

acceptance/acceptance_test.go
simonjjones: We don't need it for every test, so when focusing on tests that don't use containers we shouldn't instantiate this.

The test object we have in our test examples is different to the one we have before `spec` invokes our test commands, so if you instantiate anything with `t` in `TestAcceptance` and then pass it in to `testAcceptance` via `suite` it won't include the same `t` that is also passed into `testAcceptance`. This leads to inconsistent output and won't include the tracing of which test case is being executed.

acceptance/acceptance_test.go
simonjjones: Yes, I think I'll move `PackageCreated` and `PackagePublished` to the OutputAssertionManager anyway, and separate these lines to bring consistency with other tests.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
yaelharel: Isn't it better to change the name? I think it's a bit confusing the way it's right now.

acceptance/acceptance_test.go
simonjjones: I would like to figure out how we can keep the concepts of the source of fixtures tightly coupled to the implementation of pack we're using. If we separate these, I can see a situation where it becomes confusing which fixture source we should be using for which version of pack.

There is, however, clearly a mixture of concerns here so there's some more thought required to balance tightly coupling execution and fixture source with separation of command execution.

acceptance/acceptance_test.go
simonjjones: I agree this is a confusing pattern, but it's what we have today, so I'd opt to keep it out of scope for this PR.

acceptance/acceptance_test.go
simonjjones: Yes! It's a complicated concept, basically the expectations are based on both the `pack` and `lifecycle` versions we're using. I'd love to work out how to make that clearer. Any thoughts on a better name? Or perhaps the limited number of times this is used make it a candidate for being explicit and repeating this complexity in our test cases?

acceptance/acceptance_test.go
simonjjones: Good catch, I'll add that back

acceptance/acceptance_test.go
yaelharel: I thought that since you're trying to make the code clearer in this PR, it's ok to add a comment in this case, but I don't have a strong opinion on that.
If you decide not to add it, maybe it's better to create an issue and to specify unclear code that we want to explain.

acceptance/acceptance_test.go
simonjjones: Probably, but it's kind of not an interesting input to us. Pack doesn't care about using version 1.38, it just needs a docker client that works. It would be a change across a number of files outside of the acceptance test suite, so I wouldn't consider it part of the scope of this PR.

acceptance/acceptance_test.go
simonjjones: It's not so much about the testObject. `assert` has its own testObject, and `assert.Contains` also calls `Helper()`.

It's more of philosophical choice to separate the concerns of *the message exists* and *the message is exactly this*. From the perspective of our tests, I think the important thing is that the message is shown, not exactly what it is. The OutputAssertionManager becomes the single source of truth for what the message should be.

Practically it means fewer changes if we ever decide to change the wording in our output.

acceptance/acceptance_test.go
simonjjones: @dfreilich had concerns around a similar area, I [commented](https://github.com/buildpacks/pack/pull/673#discussion_r442912727) on his question.

That doesn't cover the `After()` vs `defer` discussion though. If we perform an action in an `After` it will always run, regardless of how much of the test was executed, which could cause further failures and pollute our test output. With a `defer` we can tightly couple to the stage of the test that instantiates an object. We also potentially don't have to share a context wide state - I'd rather avoid an additional nested context in which we'd define some variables and an `After` if I can.

That said, `defer` does come with downsides, failures that occurs in a `defer` are unreliably reported to the tester.

I think the differences throughout the code are likely to be a combination of different circumstances feeling more or less appropriate, influence from the existing test suite and how I felt as they were written. I don't know if any of them a right or wrong.

acceptance/acceptance_test.go
simonjjones: Yes, it tests a resource intensive process that we would rather not repeat multiple times. I agree, and if this was a millisecond operation that's how I would right the tests. Unfortunately `pack build` is a multi-second operation even with the simplest of apps.

Focussing on the scope of this PR, I have intended to replicate the tests in the existing suite as closely as possible, so breaking this test up into smaller tests isn't something I'd like to include, but I'd be happy to revisit in the future.

acceptance/acceptance_test.go
simonjjones: This is another of those philosophical decisions. Technically it would be very simple to move these repeated lines into a couple of `Before` functions.

The decision is based on my internalized idea of things I consider **interesting** enough to become part of each test. Each of these things is an important and (potentially) state changing pre-requisite for each of the tests as viewed in isolation.

They're not interesting enough to want to be explicit about how each of them are created, that sort of thing is the **subject** of the test. I do think they're interesting enough that it's worth indicating that these things are happening, and are necessary for the subject to succeed.

Practically this will make it much easier to separate the test suite into multiple files, which is a goal we've discussed previously.

I'd welcome further discussion around this, because it is different to how a lot of tests are written. I'm also thinking about how to externalize my thoughts around the different levels of interesting that exist in a test suite, and how we can organize them most effectively.

acceptance/acceptance_test.go
yaelharel: Thank you @simonjjones but I didn't fully understand.
What is the difference between this case and line 479 (`assert.Contains(output, "invalid builder toml")`)?
And why does it mean fewer changes if we ever decide to change the wording in the output? It looks like it'll be either make the changes in the `it` block or in the `ReportsWindowsContainersExperimental` function.
I understand that I'm missing something but I'm not sure what is it.

acceptance/acceptance_test.go
yaelharel: Thank you for your detailed answer. It totally makes sense!

acceptance/acceptance_test.go
yaelharel: Sounds good.
As I wrote in another comment, since we get some good ideas from everyone, I think it's worth creating an issue and specify all of them.

acceptance/acceptance_test.go
yaelharel: Thank you for your detailed answer!
From my point of view, although these lines are important and interesting, I think that duplicate code is really bad.
If you want to change the behavior or you find a bug, you need to change it in multiple places. Moreover, contributors that add new tests might forget to add these lines.
Anywary, I totally agree that think it's worth having further discussion around this.

acceptance/acceptance_test.go
simonjjones: I don't think you've missed anything @yaelharel. There's no real difference between this and `assert.Contains(output, "invalid builder toml")`.  I do think both approaches are around about as clear, one small difference that may have prompted the difference is the line length of the message.

I didn't consider 100% consistency to actually be a goal. Even if it were achieved in this PR it would be hard to enforce going forward. Both approaches have validity, and people will have different preferences going forward.

> And why does it mean fewer changes if we ever decide to change the wording in the output?

This applies if the message is reused in multiple tests. That isn't the case with this particular message and probably never will be. A lot of the other messages in `OutputAssertionManager` are used repeatedly. Taking `IncludesGoogleBuilder` as an example, this output crops up on a number of occasions - as part of `suggested-builders`, `build` without a configured builder and `list-trusted-builders`. If Google decide to change the name of their builder those three tests don't really care what it changes to, but they will likely care that we're still showing the Google builder whatever it's called.

acceptance/components/pack_executor.go
dfreilich: Hmm. I know the `...WithOutput` is a pattern, but I would find it clearer to just have `SuccesfulRun("build"...)` just return the output by default, and someone who's not interested could just `_ =` it.

acceptance/acceptance_test.go
dfreilich: I think that I was confused by the naming of the test, because this and the previous one (creates the package as image) both seem to be testing that it can create a package as a local image. Instead, the differentiation between this test and the previous test is that PackageBuildpack can itself consume a packaged buildpack image as one of its dependency buildpacks.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
dfreilich: Agreed, if I was creating a new acceptance tests I would be confused about which to use.

acceptance/acceptance_test.go
dfreilich: Overall, I'm a fan of the use of the `OutputAssertionManager`. I feel like it really helped equalize the level of abstraction we were dealing with in the test; specific wording may be important to codify, but it was nice to see it pulled out of the tests themselves.

acceptance/managers/container_manager.go
dfreilich: > This is tricky, and I think I may have introduced some leakage

**With PR**:
With PR

**Pre PR**:
Pre PR

There definitely does seem to be some leakage from the tests, though there was some minor leakage from before the refactor as well (I would expect the `registry` to be cleaned up as well)

acceptance/acceptance_test.go
yaelharel: @simonjjones, thanks again for the explanation!
I think it's worth being consistent, at least in this commit, since it'll increase the chances that the next PR will keep the same approach.
Maybe it's only me, but as you can understand from my comments, it confused me to see the two approaches in a commit that its purpose is to organize the code.

jromero:
Hi @simonjjones, could you provide some guidance on how to proceed with this draft PR. I know we've picked it apart into smaller PRs. Is there much left to be incorporated?

dfreilich:
I spent a bit of time looking into this, and it seems that there still is some work present in this PR that isn't represented in our current acceptance tests. Specifically, some things that I'm going to break out into future PRs is:
* `ContainerManager` &rarr; To manage and remove images. This can also be used to create the `buildRunImage` in the `rebase` test
* Move container assertions to `assertionManager`, and create an `ImageAssertionManager`. This would also contain all the logic for testing the different permutations of `inspect-image --output`
* Move registry assertions (`AppExistsInCatalog`, `ImageExistsInRegistry`) to `assertionManager` (`assert`)
* `StackManager` &rarr; To handle stack creation and cleanup
* `PackageManager` &rarr; To generate package configuration, and perhaps the creation of buildpack packages
* `BuilderManager` &rarr; To generate builder configuration and required artifacts, create the builders, and handle cleaning up
* `Components Application Manager` &rarr; To take care of any adjustments to application we need to take care. Examples are the `--descriptor` tests

In addition, some other things we should clean up in our current acceptance tests are:
* Making the buildpack for the `--descriptor` test follow our other patterns of having a manager that takes care of copying all of the relevant files (@dwillist if you want to take this, that would be 🤙 )
* Clean up any assertions in our test that explicitly test for wording

This would end up removing virtually all of the functions from our `acceptance_test.go` file, and make it significantly cleaner to read. 

I think this should hold until #1068 is pulled in, given that that may cause merge conflicts otherwise.

jromero:
The history of this PR will always live on but this PR is now divergent enough that it should be closed. Thank you @simonjjones for your efforts.

PR #680, opened 2020-06-09
Added basic scaffolding of publish-buildpack command

jkutner:
Implementation of https://github.com/buildpacks/rfcs/pull/75


dfreilich:
Could we also add a test for the command with this PR, and not just the client func?

PR #693, opened 2020-06-15
Add list-trusted-builders command

### LINKED ISSUE #567: pack should provide an easy way to list trusted builders
### Description

We should provide an easy way to see which builders are currently trusted. #566 assumes that we can do this with `pack report`, but maybe there is a better way.

### Proposed solution
AC 1
When I run `pack list-trusted-builders`
Then I should see the default builders (from `pack suggest-builders`)

AC 2
When I run `pack trust-builder <some-builder>`
And I run `pack list-trusted-builders`
Then I should see `<some-builder>` in the output

simonjjones:
## Summary
Provide a way for pack users to see which builders are trusted

## Output
```
$ pack list-trusted-builders
Trusted Builders:
  gcr.io/buildpacks/builder
  heroku/buildpacks:18
  gcr.io/paketo-buildpacks/builder:base
  gcr.io/paketo-buildpacks/builder:full-cf
  gcr.io/paketo-buildpacks/builder:tiny

$ pack trust-builder simons-new-builder
Builder simons-new-builder is now trusted

$  ./out/pack list-trusted-builders
Trusted Builders:
  gcr.io/buildpacks/builder
  heroku/buildpacks:18
  gcr.io/paketo-buildpacks/builder:base
  gcr.io/paketo-buildpacks/builder:full-cf
  gcr.io/paketo-buildpacks/builder:tiny
  simons-new-builder
```

## Documentation

    - [ ] Yes, see #___
    - [x] No (I don't think so, it doesn't look like `trust-builder` is documented, so if we want to we should document both

Resolves #567

### DIFF
--- a/internal/commands/list_trusted_builders.go
+++ b/internal/commands/list_trusted_builders.go
@@ -27,6 +27,6 @@ func ListTrustedBuilders(logger logging.Logger, cfg config.Config) *cobra.Comman
 		}),
 	}
 
-	AddHelpFlag(cmd, "trust-builder")
+	AddHelpFlag(cmd, "list-trusted-builders")
 	return cmd
 }
--- a/internal/commands/list_trusted_builders.go
+++ b/internal/commands/list_trusted_builders.go
@@ -1,6 +1,8 @@
 package commands
 
 import (
+	"sort"
+
 	"github.com/spf13/cobra"
 
 	"github.com/buildpacks/pack/internal/config"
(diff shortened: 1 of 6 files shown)


jromero:
Without thinking into the implementation, I would have expected the output to be alphabetically ordered based on the output (image name).

internal/commands/list_trusted_builders.go
```diff
@@ -0,0 +1,32 @@
...
+
+			for _, builder := range cfg.TrustedBuilders {
+				logger.Infof("  %s", builder.Name)
+			}
+
+			return nil
+		}),
+	}
+
+	AddHelpFlag(cmd, "trust-builder")
```
dfreilich, line 30:
```suggestion
	AddHelpFlag(cmd, "list-trusted-builders")
```

simonjjones: Is this line actually necessary at all? Even with the existing line the output works as expected

```
$  ./out/pack help list-trusted-builders
List Trusted Builders.

Show the builders that are either trusted by default or have been explicitly trusted locally using `trust-builder`

Usage:
  pack list-trusted-builders [flags]
...
```

and it doesn't break `trust-builder` either

```
$  ./out/pack help trust-builder
Trust builder.

When building with this builder, all lifecycle phases will be run in a single container using the builder image.

Usage:
  pack trust-builder <builder-name> [flags]
```

jromero: Yes, at least historically, IIRC it's purpose is to override the help flag message and capitalizing it to be consistent.

The output you presented omitted the flags section, that's where it would be apparent.

simonjjones: Got it, updated

simonjjones:
> Without thinking into the implementation, I would have expected the output to be alphabetically ordered based on the output (image name).

> Could we update this PR to be alphabetically ordered

Done

```
$  pack list-trusted-builders
Trusted Builders:
  gcr.io/buildpacks/builder
  gcr.io/paketo-buildpacks/builder:base
  gcr.io/paketo-buildpacks/builder:full-cf
  gcr.io/paketo-buildpacks/builder:tiny
  heroku/buildpacks:18
  simons-new-builder
```

PR #698, opened 2020-06-17
Rename master branch to main

jromero:


yaelharel:
Since I've never renamed `master` before, I'm wondering whether contributors that have a fork of `pack` and/or have a copy of this repository locally will need to do anything after this change.
I guess there will be a few commands that everyone will need to run locally (like deleting the old branch). Maybe it's worth adding these steps to the PR's description.

jromero:
> Since I've never renamed master before, I'm wondering whether contributors that have a fork of pack and/or have a copy of this repository locally will need to do anything after this change.
I guess there will be a few commands that everyone will need to run locally (like deleting the old branch). Maybe it's worth adding these steps to the PR's description.

Not sure PR description is the best place to put these instructions. Let's chat on slack about this and what's the best we can do.

yaelharel:
You're right. It's probably not the right place.

PR #699, opened 2020-06-18
Add untrust-builder command

### LINKED ISSUE #670: pack should provide a way to set a builder as untrusted
### Description
In #631 the concept of trusted builders was introduced. We would like to provide a way to set a builder as untrusted.

### Proposed solution
**AC 1**
When I run `pack trust-builder <some-builder>`
And I run `pack untrust-builder <some-builder>`
And I run `pack list-trusted-builders`
Then I should not see `<some-builder>` in the output

**AC 2**
When I run `pack trust-builder <some-builder>`
And I run `pack untrust-builder <some-builder>`
And I run `pack inspect-builder <some-builder>`
Then I should see something indicating that the builder is NOT trusted

### Additional context
This issue is blocked on #567 and #669 in order to make the AC will easier.
If we prefer not to block this issue on the issues above, we can change the AC to check that `pack` will run the `lifecycle` in five separate phases when running `pack build my-app --publish -B <some-builder>` (`lifecycle image` should be available, otherwise the command will fail).
For more information, see [this blog post](https://medium.com/buildpacks/faster-more-secure-builds-with-pack-0-11-0-4d0c633ca619).

simonjjones:

## Summary
`pack untrust-builder <builder-name>` will remove the named builder from `pack`'s local configuration for trusted builders, set with `pack trust-builder <builder-name>`

#### AC1 from #670
```
$  pack trust-builder simons-new-builder
Builder simons-new-builder is now trusted
$  pack list-trusted-builders
Trusted Builders:
  gcr.io/buildpacks/builder
  gcr.io/paketo-buildpacks/builder:base
  gcr.io/paketo-buildpacks/builder:full-cf
  gcr.io/paketo-buildpacks/builder:tiny
  heroku/buildpacks:18
  simons-new-builder
$  pack untrust-builder simons-new-builder
Builder simons-new-builder is no longer trusted
$  pack list-trusted-builders
Trusted Builders:
  gcr.io/buildpacks/builder
  gcr.io/paketo-buildpacks/builder:base
  gcr.io/paketo-buildpacks/builder:full-cf
  gcr.io/paketo-buildpacks/builder:tiny
  heroku/buildpacks:18
```

#### AC2 from #670 - builder must exist
```
$  pack trust-builder simons-new-builder
Builder simons-new-builder is now trusted
$  pack inspect-builder simons-new-builder
Inspecting builder: simons-new-builder
...
Created By:
  Name: Pack CLI
  Version: 0.0.0+git-79d84ba

Trusted: Yes

Stack:
  ID: pack.test.stack
...
$  pack untrust-builder simons-new-builder
Builder simons-new-builder is no longer trusted
$  pack inspect-builder simons-new-builder
Inspecting builder: simons-new-builder
...
Created By:
  Name: Pack CLI
  Version: 0.0.0+git-79d84ba

Trusted: No

Stack:
  ID: pack.test.stack
...
```

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related
Resolves #670

### DIFF
--- a/internal/commands/untrust_builder.go
+++ b/internal/commands/untrust_builder.go
@@ -14,13 +14,8 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 		Use:   "untrust-builder <builder-name>",
 		Short: "Stop trusting builder",
 		Long:  "Stop trusting builder.\n\nWhen building with this builder, all lifecycle phases will be no longer be run in a single container using the builder image.",
-		Args:  cobra.MaximumNArgs(1),
+		Args:  cobra.ExactArgs(1),
 		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
-			if len(args) < 1 || args[0] == "" {
-				logger.Infof("Usage:\n\t%s\n", cmd.UseLine())
-				return nil
-			}
-
 			builderName := args[0]
 			existingBuilders := cfg.TrustedBuilders
 			cfg.TrustedBuilders = []config.TrustedBuilder{}
--- a/internal/commands/untrust_builder.go
+++ b/internal/commands/untrust_builder.go
@@ -1,0 +1,56 @@
+package commands
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/buildpacks/pack/internal/config"
+	"github.com/buildpacks/pack/internal/style"
+	"github.com/buildpacks/pack/logging"
+)
+
+func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "untrust-builder <builder-name>",
+		Short: "Stop trusting builder",
+		Long:  "Stop trusting builder.\n\nWhen building with this builder, all lifecycle phases will be no longer be run in a single container using the builder image.",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 || args[0] == "" {
+				logger.Infof("Usage:\n\t%s\n", cmd.UseLine())
+				return nil
+			}
+
+			builderName := args[0]
+			existingBuilders := cfg.TrustedBuilders
+			cfg.TrustedBuilders = []config.TrustedBuilder{}
+			for _, builder := range existingBuilders {
+				if builder.Name == builderName {
+					continue
+				}
+
+				cfg.TrustedBuilders = append(cfg.TrustedBuilders, builder)
+			}
+
+			if len(existingBuilders) == len(cfg.TrustedBuilders) {
+				logger.Infof("Builder %s wasn't trusted", style.Symbol(builderName))
+				return nil
+			}
+
+			configPath, err := config.DefaultConfigPath()
+			if err != nil {
+				return errors.Wrap(err, "getting config path")
+			}
+			err = config.Write(cfg, configPath)
+			if err != nil {
+				return errors.Wrap(err, "writing config file")
+			}
+
+			logger.Infof("Builder %s is no longer trusted", style.Symbol(builderName))
+			return nil
+		}),
+	}
+
+	AddHelpFlag(cmd, "untrust-builder")
+	return cmd
+}
--- a/internal/commands/untrust_builder_test.go
+++ b/internal/commands/untrust_builder_test.go
@@ -59,7 +59,10 @@ func testUntrustBuilderCommand(t *testing.T, when spec.G, it spec.S) {
 				cfg := configManager.configWithTrustedBuilders()
 				command := commands.UntrustBuilder(logger, cfg)
 				command.SetArgs([]string{})
-				h.AssertNil(t, command.Execute())
+				command.SetOut(&outBuf)
+
+				err := command.Execute()
+				h.AssertError(t, err, "accepts 1 arg(s), received 0")
 				h.AssertContains(t, outBuf.String(), "Usage:")
 			})
 		})
--- a/internal/commands/untrust_builder_test.go
+++ b/internal/commands/untrust_builder_test.go
@@ -1,0 +1,148 @@
+package commands_test
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/buildpacks/pack/internal/style"
+
+	"github.com/heroku/color"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+	"github.com/spf13/cobra"
+
+	"github.com/buildpacks/pack/internal/commands"
+	"github.com/buildpacks/pack/internal/config"
+	ilogging "github.com/buildpacks/pack/internal/logging"
+	"github.com/buildpacks/pack/logging"
+	h "github.com/buildpacks/pack/testhelpers"
+)
+
+func TestUntrustBuilder(t *testing.T) {
+	color.Disable(true)
+	defer color.Disable(false)
+	spec.Run(t, "Commands", testUntrustBuilderCommand, spec.Random(), spec.Report(report.Terminal{}))
+}
+
+func testUntrustBuilderCommand(t *testing.T, when spec.G, it spec.S) {
+	var (
+		logger       logging.Logger
+		outBuf       bytes.Buffer
+		tempPackHome string
+		configPath   string
+	)
+
+	it.Before(func() {
+		var err error
+
+		logger = ilogging.NewLogWithWriters(&outBuf, &outBuf)
+
+		tempPackHome, err = ioutil.TempDir("", "pack-home")
+		h.AssertNil(t, err)
+		h.AssertNil(t, os.Setenv("PACK_HOME", tempPackHome))
+
+		configPath = filepath.Join(tempPackHome, "config.toml")
+	})
+
+	it.After(func() {
+		h.AssertNil(t, os.Unsetenv("PACK_HOME"))
+		h.AssertNil(t, os.RemoveAll(tempPackHome))
+	})
+
+	when("#UntrustBuilder", func() {
+		when("no builder is provided", func() {
+			it("prints usage", func() {
+				command := untrustBuilderCommandWithTrustedBuilders(t, logger, configPath)
+				command.SetArgs([]string{})
+				h.AssertNil(t, command.Execute())
+				h.AssertContains(t, outBuf.String(), "Usage:")
+			})
+		})
+
+		when("builder is already trusted", func() {
+			it("removes builder from the config", func() {
+				builderName := "some-builder"
+
+				command := untrustBuilderCommandWithTrustedBuilders(t, logger, configPath, builderName)
+				command.SetArgs([]string{builderName})
+
+				h.AssertNil(t, command.Execute())
+
+				b, err := ioutil.ReadFile(configPath)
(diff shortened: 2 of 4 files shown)


internal/commands/untrust_builder.go
```diff
@@ -0,0 +1,56 @@
...
+			builderName := args[0]
+			existingBuilders := cfg.TrustedBuilders
+			cfg.TrustedBuilders = []config.TrustedBuilder{}
+			for _, builder := range existingBuilders {
+				if builder.Name == builderName {
+					continue
+				}
+
+				cfg.TrustedBuilders = append(cfg.TrustedBuilders, builder)
+			}
```
yaelharel, line 33: Can we change `cfg.TrustedBuilders` to be a hashset?
If we will do so, we won't need to go over all of the trusted builders in order to search for the builder that we want to untrust. Instead, we'll be able to do it instantly.
Maybe it's a bit of an overkill if we think there shouldn't be too many trusted builder but I don't think it should be too complicated to make this change.
If not a hashset, we can use a list (or other data structure) so we'll be able to stop the for loop when we find the builder that we want to untrust, instead of going over all of the existing builders.

internal/commands/untrust_builder_test.go
```diff
@@ -0,0 +1,148 @@
...
+		h.AssertNil(t, os.RemoveAll(tempPackHome))
+	})
+
+	when("#UntrustBuilder", func() {
+		when("no builder is provided", func() {
+			it("prints usage", func() {
+				command := untrustBuilderCommandWithTrustedBuilders(t, logger, configPath)
+				command.SetArgs([]string{})
+				h.AssertNil(t, command.Execute())
+				h.AssertContains(t, outBuf.String(), "Usage:")
```
yaelharel, line 62: I think it's better to check that all of the "usage" message is written correctly.

internal/commands/untrust_builder.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd := &cobra.Command{
+		Use:   "untrust-builder <builder-name>",
+		Short: "Stop trusting builder",
+		Long:  "Stop trusting builder.\n\nWhen building with this builder, all lifecycle phases will be no longer be run in a single container using the builder image.",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 || args[0] == "" {
+				logger.Infof("Usage:\n\t%s\n", cmd.UseLine())
+				return nil
+			}
```
jromero, line 22: This validation should be something that cobra can take care of for us if we were to set the `Args` params with proper requirements.

internal/commands/untrust_builder.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/buildpacks/pack/internal/style"
+	"github.com/buildpacks/pack/logging"
+)
+
+func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "untrust-builder <builder-name>",
+		Short: "Stop trusting builder",
+		Long:  "Stop trusting builder.\n\nWhen building with this builder, all lifecycle phases will be no longer be run in a single container using the builder image.",
+		Args:  cobra.MaximumNArgs(1),
```
jromero, line 17: I believe there is an option for "exact" number of Args.

internal/commands/untrust_builder_test.go
```diff
@@ -0,0 +1,148 @@
...
+				h.AssertContains(t,
+					outBuf.String(),
+					fmt.Sprintf("Builder %s wasn't trusted", style.Symbol(neverTrustedBuilder)),
+				)
+			})
+		})
+	})
+}
+
+//nolint:whitespace // A leading line of whitespace is left after a method declaration with multi-line arguments
```
jromero, line 130: Not sure I understand why this is required. Could you elaborate?

internal/commands/untrust_builder_test.go
```diff
@@ -0,0 +1,148 @@
...
+func untrustBuilderCommandWithTrustedBuilders(
```
jromero, line 131: This function seems unnecessary if we were to instead just have a function "newConfigWithTrustedBuilders":

Then usages would be two lines but would be more understandable IMO:

```go
cfg := newConfigWithTrustedBuilders(builders)
command := commands.UntrustBuilder(logger, cfg)
```

internal/commands/untrust_builder.go
simonjjones: The concern I'd have with the hashset approach is that this data structure is a representation of the config file. We'd lose all of the toml encoding & decoding magic we get for free right now. I don't imagine this list becoming unmanageable and while iterating through the whole list is not the most efficient approach, it's also unlikely to be noticeably slow.

internal/commands/untrust_builder_test.go
simonjjones: The literal requirement for the comment is that linting fails without it.

The choice to leave the whitespace is based on feedback from previous teams that once you have split method arguments on to multiple lines, leaving a new line helps to provide a visual distinction between method definition and implementation.

Is the question whether we should include the comment, or whether we should include the whitespace?

internal/commands/untrust_builder.go
simonjjones: I expect that's true, this implementation is consistent with `trust-builder` so we should update that too.

internal/commands/untrust_builder_test.go
simonjjones: I'm not sure that's true as doing so would be testing `cobra`'s implementation. In this case I feel confident that cobra will output appropriate usage instructions, but I don't care what it is.

This is also consistent with how we test `trust-builder` and none of the other commands test the usage output at all. That alone is not a reason not to do the right thing, but in this case I feel like this is good enough.

internal/commands/untrust_builder.go
simonjjones: The small side effect to using `cobra.ExactArgs(1)` is that the command will exit with non-zero status, which is probably fine, but does lead to a small change in functionality.

simonjjones: A bigger side effect is that cobra writes the usage to stdout rather than return an error that we can write to our logger. I think that makes the approach unsuitable for us unfortunately.

internal/commands/untrust_builder_test.go
simonjjones: Agreed, updated. Used a small manager struct to avoid repeatedly passing around `t` & `configPath` that aren't really relevant to the test cases.

internal/commands/untrust_builder_test.go
simonjjones: The cleaner implementation means this resolves itself

internal/commands/untrust_builder.go
yaelharel: I agree that it's ok for now since the trusted builder array is probably not that big, but I always prefer to be as proficient as possible so we won't have any problem with the performance in the future.
Can you please elaborate on what is the magic you're talking about and what we would need to do in order to change the trusted builder data structure?

simonjjones: The "magic" is just [`toml.Decode`](https://github.com/buildpacks/pack/blob/main/internal/config/config.go#L50) and [`toml.Encode`](https://github.com/buildpacks/pack/blob/main/internal/config/config.go#L68) and the human readable format of the resultant toml file. I'm not sure how these would handle a `map[TrustedBuilder]interface{}` I'm not even sure if toml is capable of expressing that pattern.

yaelharel: Interesting. I'm wondering if it's possible to do it with a list or a hashset.
Of course it doesn't have to be part of this PR.

internal/commands/untrust_builder.go
jromero: > A bigger side effect is that cobra writes the usage to stdout rather than return an error that we can write to our logger. I think that makes the approach unsuitable for us unfortunately.

This seems like a side consequence that we might want to explore further. I wouldn't use it as an argument not to use cobra's built-in validation. I would say it's stranger to have `< 1` and `> 1` output via two different streams than for both to use the same mechanism and thereby go to the same stream.

internal/commands/untrust_builder_test.go
```diff
@@ -0,0 +1,152 @@
...
+				h.AssertContains(t,
+					outBuf.String(),
+					fmt.Sprintf("Builder %s wasn't trusted", style.Symbol(neverTrustedBuilder)),
+				)
+			})
+		})
+	})
+}
+
+type configManager struct {
```
jromero, line 129: Not particularly fond of a object to simply persist only two params. Seems like overkill for this particular case where a single function taking the arguments would be easier to understand and within context of it's usage.

internal/commands/untrust_builder.go
jromero: > I expect that's true, this implementation is consistent with trust-builder so we should update that too.

That's interesting, I'd be curious as to how that came about or if it was simply an oversight. Nonetheless, I agree that they should be consistent.

internal/commands/untrust_builder_test.go
simonjjones: > Not particularly fond of a object to simply persist only two params

That's unfortunate. I'm super fond of anything that means I don't have to keep passing `t` into everything...

jromero: I'm not trying to be particularly picky here as I can see this going in as-is. (The main blocker atm is [this](https://github.com/buildpacks/pack/pull/699/files#r442964759)). The consequences with this helper class is its added indirection albeit minor in this context. I'm purely making a statement that it seems like it's buying very little while adding some overhead in code and indirection.

internal/commands/untrust_builder.go
simonjjones: > This seems like a side consequence that we might want to explore further.

[Possible resolution to the inconsistent output streams](https://github.com/buildpacks/pack/pull/701#issuecomment-647617183).

> Nonetheless, I agree that they should be consistent.

It's [fairly](https://github.com/buildpacks/pack/blob/main/internal/commands/create_builder.go#L39) [inconsistent](https://github.com/buildpacks/pack/blob/main/internal/commands/inspect_builder.go#L27-L30) [right](https://github.com/buildpacks/pack/blob/main/internal/commands/inspect_image.go#L30) [now](https://github.com/buildpacks/pack/blob/main/internal/commands/package_buildpack.go#L42), [and](https://github.com/buildpacks/pack/blob/main/internal/commands/rebase.go#L17) [almost](https://github.com/buildpacks/pack/blob/main/internal/commands/trust_builder.go#L19-L22) [exclusively](https://github.com/buildpacks/pack/blob/main/internal/commands/set_default_builder.go#L21-L24) [untested](https://github.com/buildpacks/pack/blob/main/internal/commands/set_run_image_mirrors.go#L19).

dfreilich: > A bigger side effect is that cobra writes the usage to stdout rather than return an error that we can write to our logger. I think that makes the approach unsuitable for us unfortunately.

Just to clarify - this means that when a person does the command with invalid args, cobra will print usage to Stdout? Why would you expect it to go to Stderr?

simonjjones: > Why would you expect it to go to Stderr?

The command I've encountered wrote to Stdout, however cobra also uses Stderr. I would rather not do a full analysis of Cobra to determine if we'll ever care about Stderr in our use cases. I'd also not like to rely on Cobra's implementation never changing as I could see a justification that this output does belong on Stderr and Cobra could potentially change to that in the future.

PR #701, opened 2020-06-19
Move logger creation into PackCmd creation to allow for external use

dfreilich:
* Add short description for docs

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -36,6 +36,14 @@ type Logger interface {
 	IsVerbose() bool
 }
 
+// ConfigurableLogger defines behavior required by the Pack cobra.Command
+type ConfigurableLogger interface {
+	Logger
+	WantTime(f bool)
+	WantQuiet(f bool)
+	WantVerbose(f bool)
+}
+
 // WithSelectableWriter is an optional interface for loggers that want to support a separate writer per log level.
 type WithSelectableWriter interface {
 	WriterForLevel(level Level) io.Writer
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -36,14 +36,6 @@ type Logger interface {
 	IsVerbose() bool
 }
 
-// ConfigurableLogger defines behavior required by the Pack cobra.Command
-type ConfigurableLogger interface {
-	Logger
-	WantTime(f bool)
-	WantQuiet(f bool)
-	WantVerbose(f bool)
-}
-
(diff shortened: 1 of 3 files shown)


jromero:
Not sure I understand the purpose of this change. There are logger alternatives. Why are we removing a dependency injection point?

dfreilich:
We reference a number of internal methods as part of the creation of commands - for instance, `WantQuiet`, `WantVerbose`, or `WantTime`. We could add those to our external logger, and I'm happy to do that, I was just going for the smallest change.

dfreilich:
@jromero Thanks for the feedback, definitely made sense, and I reevaluated. I didn't want to change the `Logger` interface, given that it's a public API, and is used in creating the `Client` (so that would've broken anyone who uses it), so I thought my best bet was to add it as a new interface. Let me know your thoughts!

logging/logging.go
```diff
@@ -36,6 +36,14 @@ type Logger interface {
 	IsVerbose() bool
 }
 
+// ConfigurableLogger defines behavior required by the Pack cobra.Command
+type ConfigurableLogger interface {
```
jromero, line 40: TL;DR: Move to `cmd` package.

This interface makes sense but going with `go` idioms this should be defined in the package it is used. This specially makes sense to move given the fact that there is a comment about where it's being used.

dfreilich: Done.

simonjjones:
I actually think there could be value in moving the logger creation inside `NewPackCommand`, but passing in the writer for stdout & stderr.

The reason for this is that we can't propagate our `ConfigurableLogger` to the cobra command, so while the existing approach will appropriately redirect any of our outputs, anything emitted directly from `cobra` will default to `os.Stdout` & `os.Stderr`.

If we used `NewPackCommand(stdout, stderr io.Writer) (*cobra.Command, err)` we could then add `rootCmd.SetOut(stdout)` & `rootCmd.SetErr(stderr)` to ensure everything uses consistent writers.

This is relevant to the change [requested](https://github.com/buildpacks/pack/pull/699/files#r442964759) on #699 where we intend to use cobra's built in validation.

dfreilich:
@jromero What do you think, re: 

> I actually think there could be value in moving the logger creation inside `NewPackCommand`, but passing in the writer for stdout & stderr.
>

dfreilich:
I tried to settle both, by using the `logging.GetWriterForLevel` func to get the appropriate `out` and `err` to set.

PR #705, opened 2020-06-19
Reorder Funcs/Tests in Registry

dfreilich:
## Summary

This PR clarifies the funcs and tests in the Registry package, without doing any restructuring. There is room for further refactoring, but that would be best left for another PR. 

Additionally, added GoDocs to the public funcs. 

## Documentation

    - [ ] Yes, see #___
    - [x] No


internal/registry/registry_cache.go
```diff
@@ -63,54 +62,73 @@ func NewRegistryCache(logger logging.Logger, home, registryURL string) (Cache, e
...
+			highestVersion := entry.Buildpacks[0]
+			if len(entry.Buildpacks) > 1 {
+				for _, bp := range entry.Buildpacks[1:] {
+					if semver.Compare(fmt.Sprintf("v%s", bp.Version), fmt.Sprintf("v%s", highestVersion.Version)) > 0 {
+						highestVersion = bp
+					}
+				}
+			}
+			return highestVersion, highestVersion.Validate()
+		}
```
simonjjones, line 93: This gets pretty deeply nested, so I would be tempted to move this in to a few helper methods or a service.

Could `Entry` provide us a `BuildpackForVersion(version string)`?

I feel like this would only return information about the data structure rather than perform another service, so I treat is as a single concern. If we wanted to strongly maintain a separation of concerns could we add a `RegistryCacheManager` or something similar to provide this method?

Within that method could we use some private helpers such as `highestVersion()`? Could we use tail recursion to iterate through our list rather than a range?

internal/registry/registry_cache.go
```diff
@@ -63,54 +62,73 @@ func NewRegistryCache(logger logging.Logger, home, registryURL string) (Cache, e
...
+
+		for _, bpIndex := range entry.Buildpacks {
+			if bpIndex.Version == version {
+				return bpIndex, bpIndex.Validate()
+			}
+		}
+		return Buildpack{}, fmt.Errorf("could not find version for buildpack: %s", bp)
```
simonjjones, line 100:
```suggestion
		return Buildpack{}, fmt.Errorf("finding version for buildpack: %s", bp)
```

internal/registry/registry_cache.go
```diff
@@ -63,54 +62,73 @@ func NewRegistryCache(logger logging.Logger, home, registryURL string) (Cache, e
...
 }
 
-func (r *Cache) validateCache() error {
-	r.logger.Debugf("Validating registry cache for %s/%s", r.url.Host, r.url.Path)
+// Refresh local Registry Cache
+func (r *Cache) Refresh() error {
+	r.logger.Debugf("Refreshing registry cache for %s/%s", r.url.Host, r.url.Path)
+
+	if err := r.Initialize(); err != nil {
+		return errors.Wrapf(err, "could not initialize (%s)", r.Root)
```
simonjjones, line 111:
```suggestion
		return errors.Wrapf(err, "initializing (%s)", r.Root)
```

internal/registry/registry_cache.go
```diff
@@ -63,54 +62,73 @@ func NewRegistryCache(logger logging.Logger, home, registryURL string) (Cache, e
...
+	}
 
 	repository, err := git.PlainOpen(r.Root)
 	if err != nil {
-		return errors.Wrap(err, "could not open registry cache")
+		return errors.Wrapf(err, "could not open (%s)", r.Root)
```
simonjjones, line 116:
```suggestion
		return errors.Wrapf(err, "opening (%s)", r.Root)
```

internal/registry/registry_cache.go
```diff
@@ -63,54 +62,73 @@ func NewRegistryCache(logger logging.Logger, home, registryURL string) (Cache, e
...
 	}
 
-	remotes, err := repository.Remotes()
+	w, err := repository.Worktree()
 	if err != nil {
-		return errors.Wrap(err, "could not access registry cache")
+		return errors.Wrapf(err, "could not read (%s)", r.Root)
```
simonjjones, line 121:
```suggestion
		return errors.Wrapf(err, "reading (%s)", r.Root)
```

internal/registry/registry_cache.go
```diff
@@ -136,28 +154,48 @@ func (r *Cache) Initialize() error {
...
 		return err
 	}
 
-	repository, err := git.PlainOpen(r.Root)
+	repository, err := git.PlainClone(root, false, &git.CloneOptions{
+		URL: r.url.String(),
+	})
 	if err != nil {
-		return errors.Wrapf(err, "could not open (%s)", r.Root)
+		return errors.Wrap(err, "could not clone remote registry")
```
simonjjones, line 169:
```suggestion
		return errors.Wrap(err, "cloning remote registry")
```

internal/registry/registry_cache.go
```diff
@@ -136,28 +154,48 @@ func (r *Cache) Initialize() error {
...
-		return nil
+	return os.Rename(w.Filesystem.Root(), r.Root)
+}
+
+func (r *Cache) validateCache() error {
+	r.logger.Debugf("Validating registry cache for %s/%s", r.url.Host, r.url.Path)
+
+	repository, err := git.PlainOpen(r.Root)
+	if err != nil {
+		return errors.Wrap(err, "could not open registry cache")
```
simonjjones, line 185:
```suggestion
		return errors.Wrap(err, "opening registry cache")
```

internal/registry/registry_cache.go
```diff
@@ -136,28 +154,48 @@ func (r *Cache) Initialize() error {
...
 	}
-	return err
+
+	remotes, err := repository.Remotes()
+	if err != nil {
+		return errors.Wrap(err, "could not access registry cache")
```
simonjjones, line 190:
```suggestion
		return errors.Wrap(err, "accessing registry cache")
```

internal/registry/registry_cache_test.go
```diff
@@ -0,0 +1,227 @@
...
+						Name:  "John Doe",
+						Email: "john@doe.org",
+						When:  time.Now(),
+					},
+				})
+				h.AssertNil(t, err)
+
+				_, err = r.CommitObject(commit)
+				h.AssertNil(t, err)
+			})
```
simonjjones, line 180: This is a long `Before` method that applies to a single test. I would prefer to find a way to have this setup take place in the test case. I suspect the case for not doing so is how long it is. I would look to externalize a lot of this into a clearly name helper method (or two, or three) that can be triggered within the test case.

internal/registry/registry_cache_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+			when("url is empty string", func() {
+				it("fails to clone cache", func() {
+					registryCache.url = &url.URL{}
+
+					err = registryCache.Initialize()
+					h.AssertError(t, err, "could not clone remote registry")
+				})
+			})
+		})
```
simonjjones, line 225: There's quite a lot of nesting happening here, none of these cases have any other state than `registryCache.Root = ""` so could that be moved in to the parent's `Before` (or is it even the default returned by `NewRegistryCache`?)

Does "url is an empty string" need to be a sub-context of "root is empty string"? On the surface they don't seem to be related. My guess is the check for an empty URL occurs before the check for an empty root so these tests will pass. Does that order need to be guaranteed in this way?

internal/registry/registry_cache_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+		when("root is empty string", func() {
+			it.Before(func() {
+				registryCache.Root = ""
+			})
+
+			it("fails to create registry cache", func() {
+				err = registryCache.Initialize()
+				h.AssertError(t, err, "could not create registry cache")
```
simonjjones, line 214:
```suggestion
				h.AssertError(t, err, "creating registry cache")
```

internal/registry/registry_cache_test.go
```diff
@@ -0,0 +1,227 @@
...
+			})
+
+			when("url is empty string", func() {
+				it("fails to clone cache", func() {
+					registryCache.url = &url.URL{}
+
+					err = registryCache.Initialize()
+					h.AssertError(t, err, "could not clone remote registry")
```
simonjjones, line 222:
```suggestion
					h.AssertError(t, err, "cloning remote registry")
```

internal/registry/registry_cache_test.go
dfreilich: These cases were contrived to try and force some of the error cases, and would definitely not be standard. The nesting of url and root both being empty was actually relevant, because when the Root is empty, it mostly passes (it fetches the cache to a new `Tmpdir`), it just fails on the `os.Rename` at the end of the `createCache` func (although, thinking about it, we could easily make it more resilient and not fail in that case, as unlikely as it is). However, when the `url` is empty, then it fails to even fetch the cache in the first place.

internal/registry/registry_cache_test.go
dfreilich: I definitely hear that. I simplified it somewhat. The middle chunk is just creating a new commit, which is only done in this test (right now). I'm not sure whether the one use warrants creating an external func for it, what do you think?

internal/registry/registry_cache.go
```diff
@@ -24,32 +23,32 @@ const defaultRegistryURL = "https://github.com/buildpacks/registry-index"
...
+// NewDefaultRegistryCache creates a new registry cache with default options
+func NewDefaultRegistryCache(logger logging.Logger, home string) (Cache, error) {
+	return NewRegistryCache(logger, home, defaultRegistryURL)
 }
 
+// NewRegistryCache creates a new registry cache
 func NewRegistryCache(logger logging.Logger, home, registryURL string) (Cache, error) {
 	if _, err := os.Stat(home); err != nil {
-		return Cache{}, err
+		return Cache{}, errors.Wrapf(err, "failed to find home %s", home)
```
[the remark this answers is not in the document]
dfreilich: This was a great point. Fixed them in f8b31447f43b88f9ea11df11c0166e51fdc1c8c4

internal/registry/registry_cache.go
dfreilich: This I didn't change; this was a generated error, so I figured that using the `could not find` syntax would be a bit clearer.

internal/registry/registry_cache.go
dfreilich: I definitely hear that. I was trying to avoid a larger scale refactor in this PR, but I can definitely move this into a separate func (perhaps not on `Entry`, so as to keep it off data structures, though perhaps I should also move validation off `Buildpacks`...)

Issue #606, opened 2020-05-05
Acceptance tests could be clearer, easier to separate and simpler to add to


simonjjones:
### Description
The primary acceptance test file has grown to over 2000 lines and is proving hard to reason about, especially for people who are new to working with it. The tests are supported by a range of methods, state and service objects that are variously declared throughout the codebase and instantiated anywhere from the package level to individual tests.

### Proposed solution
I propose a refactor of the acceptance test suite to accommodate the scale it has grown to, in doing so I suggest the following principles:

- As much as reasonable, *prepare*, *act* and *assert* within the test body.
  - Avoid significant, scenario specific activities obscured in nested `Before` trees.
  - Each step can be abstracted by helper service methods, or stateless package methods. The methods should be very descriptively named.
- Use immutable service objects to pass around state, and state related test methods.
  - Consider the test object `*testing.T` to be state associated with these objects, rather than repeatedly pass to stateless methods.
  - Package (static) methods should ideally only be used when all of the inputs are outcomes of test preparation or action.
- Use service objects to encapsulate our dependencies, such as test stacks and buildpacks.
  - Include more contextual state about these dependencies without reusing our implementation.
  - Allow more flexibility around how these are generated.
- Declare shared state and service objects at the deepest level of test to be used by the tests that require them.
- Identify missing, or no longer required test cases

### Additional context
- Work out how to deliver this incrementally?

dfreilich:
I think it's fair to say that this issue has been relatively resolved, with the epic based on #673.

Issue #593, opened 2020-04-22
Add code coverage to acceptance tests


jromero:
### Description
https://github.com/buildpacks/pack/pull/591 adds code coverage for unit/integration tests. In order, to truly know the full coverage of our tests we need to add coverage to acceptance tests.

### Proposed solution

The proposed solution would apply to `pack=current, create builder pack=current, lifecycle=default` when pack itself is built by the acceptance tests. A possible solution could be that instead of it being built to binary, a wrapper/interface could be created that does ["magic"](https://blog.cloudflare.com/go-coverage-with-external-tests/) when calling operations to `pack`. 

I propose this "hack" because it appears that [`go run -cover` or `go build -cover` don't exist](https://github.com/golang/go/issues/16537). The implementation should be a tidy as possible limiting the scope of the "hack" to a very specific section of code.

PR #708, opened 2020-06-23
Acceptance Config & Asset setup into immutable managers

simonjjones:

## Summary
Following on from discussion in #673 this is the first group of changes intended to improve the readability, maintainability and user experience of working with our acceptance test suite.

This changes create `AssetManager`, `InputConfigurationManager` & `RunCombination` objects which handle the task of turning our input configuration environment variables into a local development environment ready to execute our acceptance tests.

The logic of which assets are downloaded & configured remains identical to the existing test suite, this rearrangement serves the goal of decluttering the main `acceptance_test.go` file.

## Documentation
    - [ ] Yes, see #___
    - [x] No

## Related
Relates to #606 - doesn't resolve it alone


acceptance/managers/input_configuration_manager.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	envPackPath                 = "PACK_PATH"
```
dfreilich, line 14: The only ones of these I see used is GITHUB_TOKEN and ACCEPTANCE_SUITE_CONFIG. Where are the rest used? Can we take them out?

acceptance/managers/input_configuration_manager.go
```diff
@@ -0,0 +1,91 @@
...
+func NewInputConfigurationManager() (InputConfigurationManager, error) {
+	packPath := os.Getenv(envPackPath)
+	previousPackPath := os.Getenv(envPreviousPackPath)
+	previousPackFixturesPath := os.Getenv(envPreviousPackFixturesPath)
+	lifecyclePath := os.Getenv(envLifecyclePath)
+	previousLifecyclePath := os.Getenv(envPreviousLifecyclePath)
+	compilePackWithVersion := os.Getenv(envCompilePackWithVersion)
+	githubToken := os.Getenv(envGitHubToken)
+
+	err := absolutizePaths(&packPath, &previousPackPath, &previousPackFixturesPath, &lifecyclePath, &previousLifecyclePath)
```
dfreilich, line 44: Didn't realize absolutize is a word! https://www.merriam-webster.com/dictionary/absolutize

acceptance/managers/run_combination.go
```diff
@@ -0,0 +1,189 @@
...
+	default:
+		return ""
+	}
+}
+
+var defaultRunCombo = []*RunCombo{
+	{Pack: Current, PackCreateBuilder: Current, Lifecycle: DefaultKind},
+}
+
+type RunCombo struct {
```
dfreilich, line 39: For clarity, perhaps we could have the `RunCombo` typedef first, then `defaultRunCombo`, then `ComboValue`? Or do you want to connect `RunCombo` with its `func`s?

acceptance/managers/run_combination.go
```diff
@@ -0,0 +1,189 @@
...
+	Pack              ComboValue `json:"pack"`
+	PackCreateBuilder ComboValue `json:"pack_create_builder"`
+	Lifecycle         ComboValue `json:"lifecycle"`
+}
+
+func (c *RunCombo) UnmarshalJSON(b []byte) error {
```
dfreilich, line 45: Right now, this seems to be unused. Did you mean to use it in `NewInputConfigurationManager`?

dfreilich:
A note for other reviewers, mentioned by @natalieparellano here: https://github.com/buildpacks/pack/pull/673#pullrequestreview-430878948

> One thing I'll note for other reviewers, in reading through the code, it is necessary to configure GoLand according to this page: https://www.jetbrains.com/help/go/configuring-build-constraints-and-vendoring.html with acceptance as a custom tag in order to be able to click through to function definitions. Maybe we could document this somewhere for others newly approaching this code, to save a bit of confusion.

acceptance/variables/variables.go
```diff
@@ -0,0 +1,6 @@
+// +build acceptance
+// +build !windows
+
+package variables
+
+const PackBinaryName = "pack"
```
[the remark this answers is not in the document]
simonjjones: Why do we need a separate name for linux & darwin?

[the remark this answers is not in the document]
dfreilich: We wouldn't, I just thought we may want to get rid of `variables.go` if it only has one `const`

acceptance/managers/input_configuration_manager.go
simonjjones: Just be glad I didn't try and colonise https://en.wiktionary.org/wiki/absolutise

acceptance/managers/run_combination.go
simonjjones: This is used in `NewInputConfigurationManager` - this causes `RunCombo` to implement [`json.Unmarshaler`](https://golang.org/pkg/encoding/json/#Unmarshaler) and will be called internally by `json.Unmarshal`

.github/workflows/compatibility.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         pack_kind: [current, previous]
         create_builder_kind: [current, previous]
-        lifecycle_kind: [current, previous]
+        lifecycle_kind: [default, previous]
```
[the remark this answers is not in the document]
simonjjones: I believe (but happy to be shown incorrectly) that this change mainly represents the reality of what we're doing today.

The difference between `pack` & `lifecycle` was intentional. As this test suite is for `pack` it's important that we test the current code. It's not intended to test `main` of `lifecycle` because that repo offers fewer guarantees and could cause us to block changes in this repo unnecessarily.

The use of previous, and it's different meanings, also hasn't changed in this PR. The change here is to call out that we're using the `default` lifecycle for the latest `pack` code, rather than `current` which was also behaving differently between `pack` and `lifecycle`

acceptance/variables/variables.go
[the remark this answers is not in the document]
simonjjones: What if there were [two](https://github.com/buildpacks/pack/pull/673/files#diff-fc54a12719f47ed2d6f8b2b77eb50fefR8-R10)?

acceptance/managers/asset_manager.go
```diff
@@ -0,0 +1,313 @@
...
+	previousPackPath            string
+	previousPackFixturesPath    string
+	lifecyclePath               string
+	lifecycleDescriptor         builder.LifecycleDescriptor
+	previousLifecyclePath       string
+	previousLifecycleDescriptor builder.LifecycleDescriptor
+	testObject                  *testing.T
+}
+
+func ConvergedAssetManager(t *testing.T, inputConfig InputConfigurationManager) AssetManager {
```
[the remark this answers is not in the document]
simonjjones: Is it OK for something that looks like one of our standard constructors to do **so much**?

The `Converged` in this case is intended to represent the idea that we will use pre-requisites either from our environment, or we'll download them for you. We'll determine based on your `InputConfiguration`, and the state of your environment, which things we'll use locally, and which we'll download for you. That's a lot of things, I wouldn't expect to happen in a standard constructor.

The multiple points do exist. The `inputConfig` simply helps us determine which points to use.

acceptance/managers/asset_manager.go
```diff
@@ -0,0 +1,313 @@
...
+
+	packTmpDir, err := ioutil.TempDir("", "pack.acceptance.binary.")
+	h.AssertNil(b.testObject, err)
+
+	packPath := filepath.Join(packTmpDir, variables.PackBinaryName)
+
+	cwd, err := os.Getwd()
+	h.AssertNil(b.testObject, err)
+
+	cmd := exec.Command("go", "build",
```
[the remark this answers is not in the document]
simonjjones: Ultimately we'll probably stop using `build` at all as part of [including acceptance tests in code coverage](https://github.com/buildpacks/pack/issues/593).

acceptance/managers/input_configuration_manager.go
simonjjones: These are used in [`NewInputConfigurationManager`](https://github.com/buildpacks/pack/pull/708/files#diff-638d0c574e1c03af8d170c505421b931R36-R42) unless I'm missing your point?

acceptance/managers/run_combination.go
simonjjones: That was the goal. Happy to move it above `defaultRunCombo` but I'd probably rather keep all of the `ComboValue` bits together.

.github/workflows/compatibility.yml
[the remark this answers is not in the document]
dfreilich: 100%, my earlier comment was inaccurate. I just found it slightly confusing (when it came down to the `Fetcher`, where `previous` pack `relativeVersion` is still 0), but it makes sense.

acceptance/managers/input_configuration_manager.go
dfreilich: What I had meant is, are those actually set by any environments/users such that the values in `NewInputConfigurationManager` will ever not be `""`? It's great that we offer this flexibility, I just didn't see anywhere that actually set that environment variable, though I easily could have missed it.

acceptance/managers/run_combination.go
dfreilich: Ahh ok, that makes sense, I hadn't looked into the `Unmarshaler` type.

acceptance/managers/input_configuration_manager.go
simonjjones: Yeah, I think it's pretty important to maintain our older contract (prior to automatically downloading missing pre-requisites these variables were required). Auto-downloading should be considered a helper and a fallback to defining these variables.

acceptance/managers/asset_manager.go
```diff
@@ -0,0 +1,313 @@
...
+			b.inputConfig.combinations,
+			style.Symbol(envLifecyclePath),
+		)
+
+		lifecyclePath = b.downloadLifecycle(0)
+
+		b.testObject.Logf("using %s for default lifecycle path", lifecyclePath)
+	}
+
+	lifecycle, err := builder.NewLifecycle(blob.NewBlob(lifecyclePath))
```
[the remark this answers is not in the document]
simonjjones: Can you elaborate what you been by recreate it every time?

The `descriptor` is a member element of the `lifecycle`, it's instantiated once during `NewLifecycle`. The response of this method includes the descriptor separately to help with clarity when assigning variables which will instantiate our `AssetManager`. This method is only called once during construction of that `AssetManager`, so I can't see any duplicated creation

.github/workflows/compatibility.yml
natalieparellano, line 19: I still sort of wonder about this change. Right now, `current` means "latest released"; `previous` means "latest patch of previous minor"; `default` means "whatever pack thinks is the default". It's possible these could all be different. I don't know what makes the most sense to automatically test. Also we should probably document these definitions somewhere (wherever we land).

acceptance/managers/asset_manager.go
```diff
@@ -0,0 +1,313 @@
...
+	if inputConfig.combinations.requiresPreviousPack() {
+		convergedPreviousPackPath = assetBuilder.ensurePreviousPack()
+		convergedPreviousPackFixturesPath = assetBuilder.ensurePreviousPackFixtures()
+
+		h.RecursiveCopy(t, previousPackFixturesOverridesDir, convergedPreviousPackFixturesPath)
+	}
+
+	if inputConfig.combinations.requiresDefaultLifecycle() {
+		convergedDefaultLifecyclePath, convergedDefaultLifecycleDescriptor = assetBuilder.ensureDefaultLifecycle()
+	}
```
natalieparellano, line 73: Are we doing away with the current definition of `current` lifecycle?

acceptance/managers/asset_manager.go
```diff
@@ -0,0 +1,313 @@
...
+	lifecyclePath := b.inputConfig.lifecyclePath
+
+	if lifecyclePath == "" {
+		b.testObject.Logf(
+			"run combinations %+v require %s to be set",
+			b.inputConfig.combinations,
+			style.Symbol(envLifecyclePath),
+		)
+
+		lifecyclePath = b.downloadLifecycle(0)
```
natalieparellano, line 221: I believe the use of "default" in this PR actually means "current" (by our current definition) - because of this line here. `0` results in us downloading the `n - 0` previous minor version. I question if we want to do away with "default" as it's currently defined (`builder.DefaultLifecycleVersion`). Or maybe since what we are doing is actually downloading the "latest" lifecycle maybe we should call it `latest`.

acceptance/managers/github_asset_fetcher.go
```diff
@@ -353,7 +374,9 @@ func (f *GithubAssetFetcher) writeCacheManifest(owner, repo string, op func(cach
 }
 
 func (f *GithubAssetFetcher) downloadAndSave(assetURI, destPath string) error {
-	downloader := blob.NewDownloader(f.logger, f.cacheDir)
+	f.testObject.Helper()
+
+	downloader := blob.NewDownloader(logging.New(&testWriter{t: f.testObject}), f.cacheDir)
```
natalieparellano, line 379: Nice!

acceptance/managers/input_configuration_manager.go
```diff
@@ -0,0 +1,91 @@
...
+		githubToken:              githubToken,
+		combinations:             combos,
+	}, nil
+}
+
+func (i InputConfigurationManager) Combinations() ComboSet {
+	return i.combinations
+}
+
+func absolutizePaths(paths ...*string) error {
```
natalieparellano, line 76: TIL "absolutize" is a real word. As an alternative, perhaps `resolveAbsolutePaths` would be clearer.

.github/workflows/compatibility.yml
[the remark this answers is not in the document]
natalieparellano: Ah, I missed this convo - some of this concern is duplicated in my other comments. The way `default` lifecycle works today is SO convoluted, thank goodness we are changing it. But as I understand it, when `default` is set we won't try to fill in the lifecycle path, resulting in a blank string being passed here: https://github.com/buildpacks/pack/blob/972ed1ef930df623bd8a0bdb9351519745b8f41f/acceptance/acceptance_test.go#L197. Therefore, ultimately, pack will download the lifecycle, and it will download the one that is specified in `builder.DefaultLifecycleVersion`. 

I think it would be nice if `default` meant `builder.DefaultLifecycleVersion`. If we could enable that through the inputs manager, even better.

acceptance/managers/input_configuration_manager.go
natalieparellano: Haha @dfreilich great minds think alike: https://github.com/buildpacks/pack/pull/708#discussion_r445182334 :)

acceptance/managers/asset_manager.go
```diff
@@ -0,0 +1,313 @@
...
+		packPath = a.previousPackPath
+		packFixturesPaths = a.previousPackFixturesPath
+	default:
+		a.testObject.Fatalf("pack kind must be current or previous, was %s", kind)
+	}
+
+	return packPath, packFixturesPaths
+}
+
+func (a AssetManager) Lifecycle(kind ComboValue) (lifecyclePath string, lifecycleDescriptor builder.LifecycleDescriptor) {
```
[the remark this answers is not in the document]
simonjjones: `LifecycleDescriptor` isn't a path unfortunately. Both of the elements returned in `Pack` are paths. Can you think of a better name to describe return both a path & a descriptor object?

acceptance/acceptance_test.go
```diff
@@ -56,15 +50,6 @@ var (
...
-type testWriter struct {
-	t *testing.T
-}
-
-func (w *testWriter) Write(p []byte) (n int, err error) {
-	w.t.Logf(string(p))
-	return len(p), nil
-}
-
 func TestAcceptance(t *testing.T) {
```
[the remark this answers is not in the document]
simonjjones: Not only did the logic not change, but the code didn't. The verification is that there are no changes in the diff.

.github/workflows/compatibility.yml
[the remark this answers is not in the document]
simonjjones: OK, so what I think we're describing here is that there could be a difference between `current` & `default`.

`current` should pick the latest GH release from the lifecycle repo. This accurately describes the existing implementation, as well as the like for like refactor that exists in this PR. Therefore `default` is objectively wrong in this case.

`default` would describe using `builder.DefaultLifecycleVersion` from the codebase being tested to determine which version to test. This is not currently being tested, although `builder.DefaultLifecycleVersion` _may_ correlate with the latest GH release for lifecycle repo, it may also correlate with `previous` or be a different version entirely.

As the purpose of this PR is refactoring, without changing testing functionality, I propose that I revert to using `current` & `previous` for lifecycle, removing all references to `default`. We should then explore separately what would be involved with adding the `default` concept for lifecycle to the suite.

How does that sound @natalieparellano @dfreilich @yaelharel ?

simonjjones: Responding in a [consistent location](https://github.com/buildpacks/pack/pull/708#discussion_r445808594)

acceptance/managers/asset_manager.go
simonjjones: Responding in a [consistent location](https://github.com/buildpacks/pack/pull/708#discussion_r445808594)

simonjjones: Although... That said, I do like `latest`

.github/workflows/compatibility.yml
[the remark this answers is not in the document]
simonjjones: Although, @natalieparellano's [suggestion](https://github.com/buildpacks/pack/pull/708#discussion_r445180541) of `latest` is really cool, and more accurate too.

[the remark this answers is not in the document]
yaelharel: I agree that @natalieparellano's suggestion is better, but as @simonjjones mentioned, since the purpose of this PR is only refactoring, I think we should revert to using `current` and `previous` and create a new PR for making the change to `latest`.

acceptance/managers/asset_manager.go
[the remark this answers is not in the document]
yaelharel: You're right. I missed the fact that there are two return values in the `Lifecycle` image (and that the function name is `PackPaths` and not `PackPath` :) ).
What do you think about the name `LifecyclePathAndDesciptor`?
Although it's pretty long, I think it's clearer than just `Lifecycle`.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
yaelharel: I'll add a comment next to some lines that aren't clear to me.

acceptance/managers/asset_manager.go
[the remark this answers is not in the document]
simonjjones: I wouldn't really like such a long method name. The return arguments are named in this method signature which should help alleviate any confusion about what's getting returned.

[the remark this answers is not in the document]
yaelharel: Personally, I have no problem with long method names (even much longer than `LifecyclePathAndDescriptor`). I think it's better to have a long and clear name.
You're right that the return arguments help but I would also change the method name to be more specific. If, theoretically, there would be another function in this file that returns some arguments that relate to `Lifecycle`, how would you change this function name?
Usually, I don't like to prepare code just in case we'll need to make some changes, but in this case I think that the method name is too vague.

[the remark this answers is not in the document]
simonjjones: If there were another lifecycle related thing I'd probably separate this into multiple methods. So I think I should just do that if this isn't clear enough. I'll replace `Lifecycle()` with `LifecyclePath` & `LifecycleDescriptor`

.github/workflows/compatibility.yml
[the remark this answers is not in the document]
natalieparellano: I'd be happy with keeping `current` vs. `latest` terminology for this PR, though I am glad you like my suggestion :)

I do think that `default` is currently being run in GHA [here](https://github.com/buildpacks/pack/blob/016be5a61a24b8ee0a9ef1f7a2c8292c95683828/.github/workflows/build.yml#L66) - because when you run acceptance without setting `LIFECYCLE_PATH` or `ACCEPTANCE_SUITE_CONFIG`, that's the code path we go down.

acceptance/managers/asset_manager.go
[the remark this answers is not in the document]
natalieparellano: I agree with @yaelharel `Lifecycle` is too overloaded in our codebase :)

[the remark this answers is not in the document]
yaelharel: Thank you @simonjjones. Sounds like a good idea!

acceptance/managers/run_combination.go
```diff
@@ -0,0 +1,186 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func (combos ComboSet) requiresDefaultLifecycle() bool {
+	return combos.requiresLifecycleKind(Current)
+}
```
natalieparellano, line 172: This mixture of "default" and "current" terminology still worries me.

yaelharel: According to https://github.com/buildpacks/pack/pull/708#discussion_r445808594, I think that @simonjjones is planning to revert to using current & previous for lifecycle (removing all references to default).

simonjjones: This was an oversight, I had thought I'd already reverted the behavior, but I'd missed a number of method names, should be resolved with latest push.

natalieparellano:
@simonjjones this is great. I confirmed with the latest changes that `make acceptance` and `make acceptance-all` function the same as on `main`.

PR #709, opened 2020-06-24
Release Docker Image

### LINKED ISSUE #468: Docker Image with Pack
### Description
As part of the release process, create & publish a Docker image that contains pack, like:
https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/pack

To use pack from systems like GitHub Actions, Cloud Build, etc it is useful to have an image with pack in it so that builder files can be written like:
https://github.com/jamesward/hello-php/blob/master/cloudbuild.yaml

dfreilich:
## Summary

This PR adds a Github action to release a `pack` Docker image to `cnbs/pack` on a new release. It currently creates three tags:
* `buildpacksio/pack:0.11.2`
* `buildpacksio/pack:latest`

I ran it, and created the images at `dfreilich/pack:...`
 
## Output

`$ act -j deliver-docker -s DOCKER_USERNAME=dfreilich -s DOCKER_PASSWORD=<PWD> -e .github/event.json`
image

## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

Resolves #468

### DIFF
--- a/.github/workflows/delivery-docker.yml
+++ b/.github/workflows/delivery-docker.yml
@@ -1,0 +1,25 @@
+name: delivery / docker
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  deliver-docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Determine version
+        run: echo "::set-env name=PACK_VERSION::`echo ${{ github.event.release.tag_name }} | cut -d "v" -f2`"
+        shell: bash
+      - name: Build and Push Image
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: cnbs/pack
+          tags: latest, ${{ env.PACK_VERSION }}
+          tag_with_sha: true
+          path: .github/workflows/delivery/docker
+          build_args: PACK_VERSION=${{ env.PACK_VERSION }}
--- a/.github/workflows/delivery-docker.yml
+++ b/.github/workflows/delivery-docker.yml
@@ -11,15 +11,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Determine version
-        run: echo "::set-env name=PACK_VERSION::`echo ${{ github.event.release.tag_name }} | cut -d "v" -f2`"
+        id: version
         shell: bash
+        run: |
+          echo "::set-output name=VERSION::$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)"
       - name: Build and Push Image
         uses: docker/build-push-action@v1
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           repository: cnbs/pack
-          tags: latest, ${{ env.PACK_VERSION }}
+          tags: latest, ${{ steps.version.outputs.VERSION }}
           tag_with_sha: true
           path: .github/workflows/delivery/docker
-          build_args: PACK_VERSION=${{ env.PACK_VERSION }}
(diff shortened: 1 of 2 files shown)


.github/workflows/delivery-docker.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  deliver-docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Determine version
+        id: version
+        shell: bash
+        run: |
+          echo "::set-output name=VERSION::$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)"
```
dfreilich, line 17: I tried extensively to use `set-env`, and then consume it later in the file, but it didn't seem to work as an input to a task, regardless of what I tried. 

[This](https://github.community/t/how-to-pass-environment-variable-to-an-input/16277/2) seems to say that it should be done with output, though the solution is a bit dated. If you can get it working with `set-env`, let me know, I'm all ears!

.github/workflows/delivery/docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.10
+
+ARG VERSION
```
dfreilich, line 3: I used `VERSION` over `PACK_VERSION`, because empirically it seemed that `outputs` didn't get set for things with a `_`.

.github/workflows/delivery-docker.yml
natalieparellano: FWIW your implementation is what we are doing for the lifecycle image: https://github.com/buildpacks/lifecycle/blob/main/.github/workflows/lifecycle-image.yml

.github/workflows/delivery-docker.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Build and Push Image
+        uses: docker/build-push-action@v1
```
natalieparellano, line 19: Nice. I wonder if lifecycle should switch to using this action too.

.github/workflows/delivery/docker/Dockerfile
natalieparellano: Oh hmm... I didn't observe this before. That said, the use of `VERSION` here seems pretty clear.

.github/workflows/delivery-docker.yml
dfreilich: It seems like it is the official docker image, so I figured it was good to use

dfreilich:
> @dfreilich were you also able to verify that the resulting image can run `pack`? (Just wanting to ensure the correct permissions were set, etc.) Otherwise, this looks good to me!

image

.github/workflows/delivery-docker.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: cnbs/pack
```
jromero, line 23: Since this is a "prod" release it should be under the "buildpacksio" namespace/org. Eg. `buildpacksio/pack`

jromero:
Not sure how I feel about the tag with sha for something that has a version tag. What are some use cases for it? 

I can see it useful for "snapshots" (latest from main) but less so here.

dfreilich:
It's nice to have immutable tags; we could conceivably update an image (i.e. a hotfix for a bug), and it may be nice for consumers to have something they can always point to. I don't think it's needed now though, so I'll take it out, and if it's requested by users we can add it in?

dfreilich:
@jromero This PR would also require the docker credentials added to the secrets of the repo, which I don't have access to do.

jromero:
> @jromero This PR would also require the docker credentials added to the secrets of the repo, which I don't have access to do.

✅  secrets  `DOCKER_USERNAME` and `DOCKER_PASSWORD` created.

PR #715, opened 2020-06-29
Write stack.toml to container instead of tmp dir

### LINKED ISSUE #702: pack v0.11.2 fails on Google Cloud Build when pack v0.11.0 and v0.11.1 were ok
I use `gcr.io/k8s-skaffold/pack` to build images with `pack` on `Google Cloud Build`. v0.11.0 and v0.11.1 were just fine but v0.11.2 now failsl **when the builder is not trusted**:

To reproduce:
 + build this project: https://github.com/GoogleCloudPlatform/buildpack-samples/tree/master/sample-go
 + with this command: `gcloud builds submit --config cloudbuild.yaml`
 + with this `cloudbuild.yaml`:

```
steps:
- name: 'gcr.io/k8s-skaffold/pack:v0.11.2'
  args: ['pack', 'build', 'test', '--builder=gcr.io/buildpacks/builder:latest']
```

It will fail with this error:

```
Step #2: ===> EXPORTING
Step #2: ERROR: failed to create 'exporter' container: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/stack.toml155056902
Finished Step #2
```

jromero:
## Summary
Instead of needing to mount a temp file for `stack.toml` this PR writes the `stack.toml` directly to the container when running export. This should remove the needed to mount additional volumes in host system when using docker-in-docker. See #702.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #702


internal/build/build.go
```diff
@@ -175,3 +177,51 @@ func randString(n int) string {
 	}
 	return string(b)
 }
+
+func (l *Lifecycle) CopyApp(ctx context.Context, containerID string) error {
```
natalieparellano, line 181: It feels weird to me that `CopyApp` lives in build.go and `WriteStackToml` lives in phases.go - could they be consolidated?

internal/build/phase.go
```diff
@@ -3,31 +3,27 @@ package build
...
-	name        string
-	infoWriter  io.Writer
-	errorWriter io.Writer
-	docker      client.CommonAPIClient
-	ctrConf     *dcontainer.Config
-	hostConf    *dcontainer.HostConfig
-	ctr         dcontainer.ContainerCreateCreatedBody
-	uid, gid    int
-	appPath     string
-	appOnce     *sync.Once
```
natalieparellano, line 29: Nice job doing away with this... this is the part that was giving me trouble before :)

internal/build/phase_config_provider.go
```diff
@@ -151,3 +160,9 @@ func WithRoot() PhaseConfigProviderOperation {
 		provider.ctrConf.User = "root"
 	}
 }
+
+func WithContainerOperations(operations ...ContainerOperation) PhaseConfigProviderOperation {
```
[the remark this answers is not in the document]
jromero: :/ I'm going to voice my distaste for adding unit test for what essentially is a getter/setter. I don't see much of a value here when we have integration and acceptance tests that also verify this.

internal/build/build.go
jromero: Yeah, I didn't give the organization too much as it seemed disorganized before I got here and wanted to limit the number of changes. I'll see what I can do with more thought.

internal/build/phase.go
jromero: There is a "risk" now that IF you don't run `detect` or `create` FIRST that the app contents won't be there but I didn't see that as a reason why we need to _try_ to copy the app during every phase.

internal/build/phase_config_provider.go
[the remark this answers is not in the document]
natalieparellano: I actually tend to agree with you on this.

internal/build/phase_test.go
```diff
@@ -129,7 +129,12 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
...
-				configProvider := build.NewPhaseConfigProvider(phaseName, lifecycle, build.WithArgs("read", "/workspace/fake-app-file"))
+				configProvider := build.NewPhaseConfigProvider(
+					phaseName,
+					lifecycle,
+					build.WithArgs("read", "/workspace/fake-app-file"),
+					build.WithContainerOperations(lifecycle.CopyApp),
+				)
 				readPhase := phaseFactory.New(configProvider)
 				assertRunSucceeds(t, readPhase, &outBuf, &errBuf)
 				h.AssertContains(t, outBuf.String(), "file contents: fake-app-contents")
```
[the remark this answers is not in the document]
jromero: This has been added to `container_ops_test.go`. I opt'd to test the functionality as an independent unit (integration with docker client) than within the phases for (hopefully) better separation of concerns.

internal/build/phase_test.go
```diff
@@ -133,7 +133,15 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
 					phaseName,
 					lifecycle,
 					build.WithArgs("read", "/workspace/fake-app-file"),
-					build.WithContainerOperations(lifecycle.CopyApp),
+					build.WithContainerOperations(
+						build.CopyDir(
```
natalieparellano, line 137: I would love to see us call `build.WriteStackToml` here - is it possible? That said, with the introduction of internal/build/container_ops_test.go, perhaps it's not necessary.

internal/build/phases_test.go
```diff
@@ -889,21 +888,6 @@ func testPhases(t *testing.T, when spec.G, it spec.S) {
 				configProvider := fakePhaseFactory.NewCalledWithProvider
 				h.AssertSliceContains(t, configProvider.HostConfig().Binds, expectedBind)
 			})
-
-			it("configures the phase with bind mounts", func() {
```
natalieparellano, line 893: Instead of this test, could we add an assertion around the contents of `configProvider.ContainerOperations()`? (Btw, it looks like that getter is not being used otherwise?)

internal/build/phases.go
```diff
@@ -122,6 +100,7 @@ func (l *Lifecycle) Detect(ctx context.Context, networkMode string, volumes []st
 		),
 		WithNetwork(networkMode),
 		WithBinds(volumes...),
+		WithContainerOperations(l.CopyApp),
```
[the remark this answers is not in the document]
natalieparellano: (see my latest comment here: https://github.com/buildpacks/pack/pull/715/files#r448372663 )

internal/build/phases_test.go
jromero: You are right that the getter isn't called. I can remove it. :D. As far as testing the getter, I was under the assumption that a getter/setter test isn't valuable based on earlier [comments](https://github.com/buildpacks/pack/pull/715/files#r447937651).

As for testing the `ContainerOperations` being executed, they are tested via integration test by [`phase_test.go`](https://github.com/buildpacks/pack/pull/715/files#diff-5c8e2e3a86224438e607242a69b117aeR137).

jromero: Ok, after our slack discussion I see what you are saying now:

>  can we test that for the “detect” phase, that we actually call CopyDir  - by inspecting the contents of configProvider.ContainerOperations()

Added a few more tests for this purpose. Given they are functions the check uses reflection.

dgageot:
Awesome, thanks a lot!

PR #716, opened 2020-06-29
Register buildpack

elbandito:
## Summary

****NOTE: In the RFC, there is mention of 2 types to support: `github` and `git`.  This PR is ONLY for supporting `github`.**  

This PR adds functionality to the `pack` CLI to support registering a buildpack with github.  A new command (`register-buildpack`) was introduced (hidden behind the experimental flag)

```
Register the buildpack to a registry

Usage:
  pack register-buildpack <image> [flags]

Flags:
  -R, --buildpack-registry string   Buildpack Registry reference
  -h, --help                        Help for 'register-buildpack'

Global Flags:
```

By default, if no configuration values are defined in the `config.toml` file, buildpacks will be registered against:
`https://github.com/buildpacks/registry`.   In addition, user can specify their own registries.  For example:

```
[[registries]]
name="elbandito"
type="github"
url="https://github.com/elbandito/buildpack-registry"

[[registries]]
name="joe"
type="github"
url="https://github.com/jkutner/buildpack-registry/"
```

When there are multiple `registries`, the user can specify a specific registry via `buildpack-registry, -R` flag.

For example:
`pack register-buildpack <buildpack/image> -R elbandito`

For more details:
https://github.com/buildpacks/rfcs/blob/main/text/0022-client-side-buildpack-registry.md
https://github.com/buildpacks/rfcs/pull/75

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related
https://github.com/buildpacks/rfcs/pull/75


register_buildpack.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func parseURL(githubURL string) (*url.URL, error) {
+	if githubURL == "" {
+		return nil, errors.New("missing github URL")
+	}
+	return url.Parse(fmt.Sprintf("%s/issues/new", strings.TrimSuffix(githubURL, "/")))
+}
+
+func createGithubIssueBody(buildpack registry.Buildpack) (string, error) {
```
dfreilich, line 122: This `func` can be combined with the next one (`createGithubIssueTitle`), with the addition of a template parameter (switching between `registry.GithubIssueBodyTemplate` and `registry.GithubIssueTitleTemplate`

register_buildpack_test.go
```diff
@@ -0,0 +1,102 @@
...
+				imageFetcher: fakeImageFetcher,
+			}
+		})
+
+		it.After(func() {
+			_ = fakeAppImage.Cleanup()
+		})
+
+		when("registry type is github", func() {
+			const expectedURL = `https://github.com/jkutner/buildpack-registry/issues/new?body=%0A%23%23%23+Data%0A%0A%60%60%60toml%0Aid+%3D+%22heroku%2Fjava-function%22%0Aversion+%3D+%221.1.1%22%0Aaddr+%3D+%22buildpack-image%22%0A%60%60%60&title=ADD+heroku%2Fjava-function%401.1.1`
```
dfreilich, line 57: Any reason not to move this into the `it` block?

register_buildpack_test.go
```diff
@@ -0,0 +1,102 @@
...
+		})
+	})
+}
+
+func getFakeExecCommand(t *testing.T, expectedCmd, expectedURL string) func(command string, args ...string) *exec.Cmd {
+	return func(command string, args ...string) *exec.Cmd {
+		h.AssertEq(t, command, expectedCmd)
+		h.AssertEq(t, args[0], expectedURL)
+
+		cs := []string{"-test.run=TestHelperProcess", "--", command}
```
dfreilich, line 95: For other reviewers: 
this is a super cool way of mocking out cmd calls in tests. Used in [os.exec_test](https://golang.org/src/os/exec/exec_test.go)

internal/commands/register_buildpack_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+				it("should handle config errors", func() {
+					cfg = config.Config{
+						DefaultRegistryRef: "missing registry",
+					}
+					command = commands.RegisterBuildpack(logger, cfg, mockClient)
+					command.SetArgs([]string{buildpackImage})
+
+					err := command.Execute()
+					h.AssertNotNil(t, err)
```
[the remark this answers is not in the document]
elbandito: @dfreilich I had originally asserted on a specific error, but I made it more generic, because I really don't care what kind of error get's thrown - all I care about testing is if I propagate the thrown error whenever I receive an error from `config`.  Inside of the `config` component, I test/assert on specific errors.

internal/commands/register_buildpack.go
```diff
@@ -0,0 +1,44 @@
...
+			opts.URL = defaultRegistry.URL
+
+			if err := client.RegisterBuildpack(cmd.Context(), opts); err != nil {
+				return err
+			}
+			logger.Infof("Successfully registered %s", style.Symbol(opts.ImageName))
+			return nil
+		}),
+	}
+	cmd.Flags().StringVarP(&flags.BuildpackRegistry, "buildpack-registry", "R", "", "Buildpack Registry name")
```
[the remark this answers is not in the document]
elbandito: @dfreilich I'm in support of following conventions, so happy to make the change if that is the case. :+1:

register_buildpack_test.go
elbandito: Thanks, @dfreilich.  It was the best I can do, so I hope it's good enough.  I had to do some digging around to find this solution.  Can't take all the credit ;)

register_buildpack_test.go
elbandito: No.  :)

internal/config/config_test.go
```diff
@@ -174,4 +174,71 @@ func testConfig(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 	})
+
+	when("#GetRegistry", func() {
+		it("should return a default registry", func() {
+			cfg := config.SetRunImageMirrors(
```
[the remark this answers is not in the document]
elbandito: Good catch!

internal/commands/register_buildpack.go
```diff
@@ -0,0 +1,44 @@
...
+func RegisterBuildpack(logger logging.Logger, cfg config.Config, client PackClient) *cobra.Command {
+	var opts pack.RegisterBuildpackOptions
+	var flags RegisterBuildpackFlags
+
+	cmd := &cobra.Command{
+		Use:   "register-buildpack <image>",
+		Args:  cobra.ExactArgs(1),
+		Short: "Register the buildpack to a registry",
+		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
+			defaultRegistry, err := cfg.GetRegistry(flags.BuildpackRegistry)
```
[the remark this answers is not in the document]
elbandito: I can get behind that.

internal/commands/register_buildpack.go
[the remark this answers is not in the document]
jromero: @dfreilich  I don't think that's the ongoing rule. At least during the initial registry PR I proposed capital `R` because:

> A --buildpack-registry feels like it's of the same importance as --builder.

https://github.com/buildpacks/pack/pull/540#pullrequestreview-401105177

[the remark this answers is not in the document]
dfreilich: I understand the importance, but as a cli user, I would find it annoying to capitalize something (a whole extra button press) which is supposed to make it easier; I find capitalized arguments to be more annoying to remember/use. 

I thought the use of `-B` by `pack build` was more because we needed some way of distinguishing `--buildpack` from `--builder`, rather than to underscore the importance of the argument; and if so, we should follow that pattern in other commands (like `create-builder`, which only has lower case shortened ones)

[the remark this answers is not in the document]
jromero: @dfreilich I see your point. at the very minimum I would think we can agree that both `build` and `register-buildpack` should both use the same flag casing.

elbandito:
> Overall, it's definitely looking good. Some broader suggestions:
> 
> * This should probably have an acceptance test as well, given that it's a new command
> * There are quite a few sad paths (particularly in `register_buildpack`) that aren't tested, and should be simple to exercise
> * It would be nice if more of the errors were wrapped `errors.Wrap` to provide a bit more context to users in sad paths.
> * I assume one of the maintainers will need to create https://github.com/buildpacks/registry for this to work/test locally, which may be helpful.

I'm on the fence on the validity of an acceptance test, at least for the happy path.  The command will pop-up a browser window for the user

elbandito:
> Overall, it's definitely looking good. Some broader suggestions:
> 
> * This should probably have an acceptance test as well, given that it's a new command
> * There are quite a few sad paths (particularly in `register_buildpack`) that aren't tested, and should be simple to exercise
> * It would be nice if more of the errors were wrapped `errors.Wrap` to provide a bit more context to users in sad paths.
> * I assume one of the maintainers will need to create https://github.com/buildpacks/registry for this to work/test locally, which may be helpful.

@dfreilich Thanks again for the feedback!  

As far as acceptance tests are concerned, I don't think it should be considered a blocker.  Most of the logic for this command is centered on the ability to open a `github` issue via a browser.  There's some parsing that takes place from the `config.toml` file, but we should be able to cover most of this surface area with unit tests.   I did add more tests, but if there is anything else that I'm missing, please let me know.  Perhaps later on we can look at doing something fancy with a headless browser to test the ability to open a new github issue.

internal/commands/register_buildpack.go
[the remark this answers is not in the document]
elbandito: @jromero @dfreilich so what's the take away here?  Do we want to change `-R` to be `-r`?

elbandito:
@jromero and @dfreilich are there any more action items on this PR that you'd like resolved?

register_buildpack_test.go
```diff
@@ -0,0 +1,130 @@
...
+			h.AssertNotNil(t, subject.RegisterBuildpack(context.TODO(),
+				RegisterBuildpackOptions{
+					ImageName: "invalid/image",
+					Type:      "github",
+					URL:       "https://github.com/jkutner/buildpack-registry",
+				}))
+		})
+
+		it("should return error for missing image label", func() {
+			fakeAppImage = fakes.NewImage("misslinglabel/image", "", &fakeIdentifier{name: "buildpack-image"})
```
dfreilich, line 69:
```suggestion
			fakeAppImage = fakes.NewImage("missinglabel/image", "", &fakeIdentifier{name: "buildpack-image"})
```

internal/commands/register_buildpack.go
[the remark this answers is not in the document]
dfreilich: I'd still vote for `-r` over `-R`

register_buildpack.go
```diff
@@ -0,0 +1,149 @@
...
+	params := url.Values{}
+	params.Add("title", issue.title)
+	params.Add("body", issue.body)
+	issueURL.RawQuery = params.Encode()
+
+	c.logger.Debugf("Open URL in browser: %s", issueURL)
+	return openBrowser(issueURL.String())
+}
+
+var execCommand = exec.Command
```
dfreilich, line 85: nit: Move this `var` and `openBrowser()` to bottom of file, so the functions read in order of definition. Feel free to disregard this though

elbandito:
@jromero any final thoughts on this?

register_buildpack_test.go
```diff
@@ -0,0 +1,130 @@
...
+						ImageName: "buildpack/image",
+						Type:      "github",
+						URL:       "",
+					}), "missing github URL")
+			})
+		})
+	})
+}
+
+func getFakeExecCommand(t *testing.T, expectedCmd string, expectedArgs ...string) func(command string, args ...string) *exec.Cmd {
```
jromero, line 116: A few points here:

1. Terminology here made it confusing for me to understand what was going on. This isn't a "fake" in the way I would understand a fake to work. It's really more of a "mock" with expectations/asserting. Maybe changing the name to something like `createExpectedCommand`.
2. I wonder if we can make it more architecturally sound if this were to be moved to the `registry` package (as commented [here](https://github.com/buildpacks/pack/pull/716/files#diff-f4dc727bc4e1e13f3f5bcc35f4da913aR103)) and then instead of relying on a package variable `execCommand` we used an interface that we can properly mock. eg. `BrowserExecFactory`.

register_buildpack.go
```diff
@@ -0,0 +1,149 @@
...
+	label, err := appImage.Label(buildpackage.MetadataLabel)
+	if err != nil {
+		return err
+	}
+
+	c.logger.Debugf("Found image label %s: %s", buildpackage.MetadataLabel, label)
+	var buildpackInfo dist.BuildpackInfo
+	if err = json.Unmarshal([]byte(label), &buildpackInfo); err != nil {
+		return err
+	}
```
jromero, line 46: We have a helper function `dist.GetLabel` that could condense these operations.

register_buildpack_test.go
```diff
@@ -0,0 +1,130 @@
...
+				RegisterBuildpackOptions{
+					ImageName: "missinglabel/image",
+					Type:      "github",
+					URL:       "https://github.com/jkutner/buildpack-registry",
+				}))
+		})
+
+		when("registry type is github", func() {
+			it("should open a github issue in the browser", func() {
+				const expectedURL = `https://github.com/jkutner/buildpack-registry/issues/new?body=%0A%23%23%23+Data%0A%0A%60%60%60toml%0Aid+%3D+%22heroku%2Fjava-function%22%0Aversion+%3D+%221.1.1%22%0Aaddr+%3D+%22buildpack-image%22%0A%60%60%60&title=ADD+heroku%2Fjava-function%401.1.1`
```
jromero, line 83: All references to `jkutner/buildpacks-registry` should be replaces with the `registry.defaultRegistryURL` value.

cmd/cmd.go
```diff
@@ -82,6 +82,8 @@ func NewPackCommand(logger ConfigurableLogger) (*cobra.Command, error) {
 	rootCmd.AddCommand(commands.Version(logger, pack.Version))
 	rootCmd.AddCommand(commands.Report(logger, pack.Version))
 
+	rootCmd.AddCommand(commands.RegisterBuildpack(logger, cfg, &packClient))
```
jromero, line 85: We would want to wrap this around experimental being enabled such as [`build --registry-url` flag](https://github.com/buildpacks/pack/blob/972ed1ef930df623bd8a0bdb9351519745b8f41f/internal/commands/create_builder.go#L65-L67).

jromero:
## UA 1: TOML data format

```
$ ./out/pack register-buildpack cnbs/sample-package:hello-universe
Successfully registered cnbs/sample-package:hello-universe
```

Screen Shot 2020-07-12 at 3 30 40 PM

#### Feedback

It looks like the output doesn't match exactly what the [RFC proposes](https://github.com/buildpacks/rfcs/blob/main/text/0044-pack-publish-buildpack.md#type-github-issue). There is a missing field for the digest whereas this PR has the digest included as part of the `addr`. I can't really see the benefit of one vs the other. @jkutner do you have some insight about whether it's worth breaking out or not?

## Error cases
```
$ ./out/pack register-buildpack cnbs/sample-package:no-exist
ERROR: image cnbs/sample-package:no-exist does not exist in registry: not found

$  ./out/pack register-buildpack gcr.io/cnbs/sample-package:no-exist
ERROR: connect to repo store 'gcr.io/cnbs/sample-package:no-exist': GET https://gcr.io/v2/token?scope=repository%3Acnbs%2Fsample-package%3Apull&service=gcr.io: UNKNOWN: Project 'project:cnbs' not found or deleted.

$ ./out/pack register-buildpack -r no-exist cnbs/sample-package:hello-universe
ERROR: registry no-exist is not defined in your config file
```

elbandito:
@jromero I spoke with @jkutner and the intention is to have the `digest` as part of the `addr` field.
https://github.com/buildpacks/rfcs/blob/main/text/0032-update-json-cnb-registry.md

jromero:
> @jromero I spoke with @jkutner and the intention is to have the `digest` as part of the `addr` field.
> https://github.com/buildpacks/rfcs/blob/main/text/0032-update-json-cnb-registry.md
> 
> 

Sounds good! ✨ Approved! ✨

PR #718, opened 2020-06-30
Add Chocolatey delivery Github Action

### LINKED ISSUE #703: Publish to Chocolatey
### Description
As a Windows user, I'd like to be able to install `pack` via https://chocolatey.org/.

### Proposed solution

Use the information provided at https://chocolatey.org/docs/create-packages to create a distributable package of `pack` via https://chocolatey.org/.

### Additional context
- [x] This feature should be documented:
    - [Release notes](https://github.com/buildpacks/pack/blob/main/.github/workflows/build.yml#L186) (maintained in this repo)
    - https://buildpacks.io/docs/install-pack/

Relates to https://github.com/buildpacks/docs/issues/7

dfreilich:

## Summary
This adds a Github Action pipeline which, upon a new release, will create and publish a Chocolatey release. 

This can't be fully tested through [act](https://github.com/nektos/act), unfortunately, because `act` doesn't yet support `windows` virtual environments, and the [choco action](https://github.com/crazy-max/ghaction-chocolatey) (despite the promise otherwise) must be run on a `windows` virtual env - but I tested it on my fork [here](https://github.com/dfreilich/pack/runs/820476343?check_suite_focus=true), and it worked. 

Note: This will necessitate a maintainer adding a secret for the chocolatey API key (I referred to it as `CHOC_KEY`, but I'm happy to change that if so desired). 

## Documentation
    - [x] Yes, see https://github.com/buildpacks/docs/issues/7

## Related

Resolves #703


dfreilich:
This package is based off [this article](https://weblog.west-wind.com/posts/2017/jan/29/creating-a-portable-and-embedded-chocolatey-package), which clarified exactly what's needed if you're going to embed the binary inside. 

Note: 
- Despite what the docs say [here](https://chocolatey.org/docs/chocolatey-faqs#what-is-the-difference-between-packages-no-suffix-as-compared-toinstallportable), the package should be named `pack` and not `pack.portable`; `portable` is only relevant if a package is going to be offered in multiple ways, and we aren't planning ATM to have an installer method for chocolatey.

.github/workflows/delivery/chocolatey/pack.nuspec
```diff
@@ -0,0 +1,68 @@
...
+
+    <title>pack</title>
+    <authors>Cloud Native Buildpack Authors</authors>
+    <projectUrl>https://github.com/buildpacks/pack</projectUrl>
+    <iconUrl>https://rawcdn.githack.com/buildpacks/artwork/36f02fae98ed82bb3175918796b6cca7acb813df/light-background/logo-light.png</iconUrl>
+    <licenseUrl>https://github.com/buildpacks/pack/blob/main/LICENSE</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <projectSourceUrl>https://github.com/buildpacks/pack</projectSourceUrl>
+    <docsUrl>https://buildpacks.io/docs/</docsUrl>
+    <!--<mailingListUrl></mailingListUrl>-->
```
jromero, line 18: We do have a mailing list we can use here.

.github/workflows/delivery/chocolatey/pack.nuspec
```diff
@@ -0,0 +1,68 @@
...
+    <bugTrackerUrl>https://github.com/buildpacks/pack/issues</bugTrackerUrl>
+    <tags>pack cloud-native-buildpacks cncf</tags>
+    <summary>pack is a CLI for building apps using Cloud Native Buildpacks.</summary>
+    <description># pack - Buildpack CLI
```
jromero, line 22: This looks like the README. Could we just read it and fill it in as part of the pipeline?

.github/workflows/delivery/chocolatey/tools/LICENSE.txt
```diff
@@ -0,0 +1,206 @@
+
+From: https://github.com/buildpacks/pack/blob/main/LICENSE
```
jromero, line 2: Could we generated this file by copying it over from the repo instead of having two copies?

.github/workflows/delivery/chocolatey/pack.nuspec
dfreilich: It mostly is the readme, with the removal of the usage Gif, because that's a relative path that I wasn't sure if I should keep. I could just copy the README in, but without the usage gif, that section would just be empty except for a broken link.

.github/workflows/delivery/chocolatey/pack.nuspec
dfreilich: Added it.

.github/workflows/delivery/chocolatey/pack.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <licenseUrl>https://github.com/buildpacks/pack/blob/main/LICENSE</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <projectSourceUrl>https://github.com/buildpacks/pack</projectSourceUrl>
+    <docsUrl>https://buildpacks.io/docs/</docsUrl>
+    <mailingListUrl>cncf-buildpacks@lists.cncf.io</mailingListUrl>
+    <bugTrackerUrl>https://github.com/buildpacks/pack/issues</bugTrackerUrl>
+    <tags>pack cloud-native-buildpacks cncf</tags>
+    <summary>pack is a CLI for building apps using Cloud Native Buildpacks.</summary>
+    <description>{{README}}
+    </description>
```
dfreilich, line 23: I found that the `sed` wouldn't replace the `{{README}}` string normally if I had the description close on the same line; instead, it would insert on the next line, outside of the `<description>` tag entirely.

README.md
```diff
@@ -14,7 +14,7 @@
 
 ## Usage
 
-<img src="resources/pack-build.gif" width="600px" />
+<img src="https://github.com/buildpacks/pack/raw/main/resources/pack-build.gif" width="600px" />
```
dfreilich, line 17: I'd love to hear another suggestion if this isn't the best idea! I could also include the resource in the package, but that doesn't seem like the nicest solution, it would just needlessly clog the chocolatey package.

.github/workflows/delivery/chocolatey/pack.nuspec
```diff
@@ -0,0 +1,28 @@
...
+
+    <title>pack</title>
+    <authors>Cloud Native Buildpack Authors</authors>
+    <projectUrl>https://github.com/buildpacks/pack</projectUrl>
+    <iconUrl>https://rawcdn.githack.com/buildpacks/artwork/36f02fae98ed82bb3175918796b6cca7acb813df/light-background/logo-light.png</iconUrl>
+    <licenseUrl>https://github.com/buildpacks/pack/blob/main/LICENSE</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <projectSourceUrl>https://github.com/buildpacks/pack</projectSourceUrl>
+    <docsUrl>https://buildpacks.io/docs/</docsUrl>
+    <mailingListUrl>cncf-buildpacks@lists.cncf.io</mailingListUrl>
```
jromero, line 18: Sorry for not being specific but I think this should be the url: https://lists.cncf.io/g/cncf-buildpacks

It aligned with the initial intent proposed here: https://github.com/chocolatey/choco/issues/205#issuecomment-115306421

.github/workflows/delivery/chocolatey/pack.nuspec
jromero: `sed` is more foreign to me than standard RegEx so I've avoided using it. This here just adds to my avoidance. :)

README.md
jromero: TLDR; Might be worth NOT posting a README.md

I'm looking around at some other packages and how this file would be used and I have a few concerns:

1. I haven't found that this README.md is used anywhere by another project.
2. How would this README be presented? If it's not and it's only listed in the `files` then what's its purpose?
3. The README.md is currently tailored to be hosted on GH. Some package registries (such as npm) do a good job about resolving links but chocolatey doesn't.

Based on the above and the added complexity/concerns, I would suggest that the `<description>` is sufficient from an end-user's perspective. I'm assuming it can then refer to (link) to additional details.

.github/workflows/delivery/chocolatey/pack.nuspec
dfreilich: Ah yes, of course. I misread it as the mailing list email.

.github/workflows/delivery/chocolatey/pack.nuspec
dfreilich: I totally agree. I just couldn't get it to work otherwise, so...

README.md
dfreilich: I'm pretty sure that this is how it gets displayed, which seems pretty visible to me: https://chocolatey.org/packages/buildpacks#description

jromero: Is there a way to confirm? If I download the `nupkg`. The README.md isn't there. Instead the content is in the `<description>` field:

image
image

dfreilich: To clarify what this change is about - above, we read in the `README` and putting it into the `description` tag. We don't publish a separate `README` file in the `nupkg`.

jromero:
NOTE: I changed the target branch for this PR in order (and hopes) that this will apply to the upcoming release. The workflows (including for the publish event) only trigger when included in the branch/commit that is tagged for release.

See: https://github.community/t/actions-not-triggered-on-release/17944/6

dfreilich:
@jromero Just to confirm, was an API key added in under secrets for this line? https://github.com/buildpacks/pack/pull/718/files#diff-186c3d9e52fe0339dd81e7ce89864e4eR76

PR #720, opened 2020-06-30
Warns if user is using untrusted builder

### LINKED ISSUE #711: Make it more obvious that a builder is trusted/untrusted
### Description
When using a custom builder via `pack build`, it is unclear whether or not the builder is trusted/untrusted. I had created a custom builder with my registry CA cert, and it was failing in the analyze step because analyze occurs using a different image. I realized that this was because my builder was not trusted, but it is not immediately obvious.  

### Proposed solution
Inform the user with a log message that the builder is not trusted and will use a different image for analyze and export. Alternatively, if the build fails in analyze or export, inform the user that this may be the result of being untrusted.

I haven't checked documentation, but hopefully this is documented somewhere.

### Additional context
N/A

dfreilich:
* Adds error message attributing lifecycle failure in untrusted case to untrusted builder

## Summary

This adds a warning if building with an untrusted builder, so the user understands that there may be some issues with it. This also adds an error if lifecycle execute fails, clarifying that it may be because the builder is untrusted. 

#### Before
image

#### After
Untrusted Builder: 
image

With `--trust-builder`:
image

## Documentation
    - [x] No

## Related

Resolves #711

### DIFF
--- a/build.go
+++ b/build.go
@@ -192,7 +192,11 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		}
 	}
 
-	return c.lifecycle.Execute(ctx, lifecycleOpts)
+	if err := c.lifecycle.Execute(ctx, lifecycleOpts); err != nil {
+		return errors.Wrap(err, "executing lifecycle. This may be the result of using an untrusted builder")
+	}
+
+	return nil
 }
 
 func (c *Client) processBuilderName(builderName string) (name.Reference, error) {
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -89,21 +89,11 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 				}
 			}
 
-			var cfgTrustedBuilder bool
-			for _, trustedBuilder := range cfg.TrustedBuilders {
-				logger.Debugf("Builder %s is trusted", style.Symbol(trustedBuilder.Name))
-				if flags.Builder == trustedBuilder.Name {
-					cfgTrustedBuilder = true
-					break
-				}
-			}
-
-			var suggestedBuilder bool
-			for _, builder := range suggestedBuilders {
-				if flags.Builder == builder.Image {
-					suggestedBuilder = true
-					break
-				}
+			trustBuilder := isTrustedBuilder(cfg, flags.Builder) || flags.TrustBuilder
+			if trustBuilder {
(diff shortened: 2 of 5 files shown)


build.go
```diff
@@ -192,7 +192,11 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		}
 	}
 
-	return c.lifecycle.Execute(ctx, lifecycleOpts)
+	if err := c.lifecycle.Execute(ctx, lifecycleOpts); err != nil {
```
dfreilich, line 195: I put the error here, rather than on `lifecycle.Analyze` or `lifecycle.Export`, because that would be easiest to test (by substituting a failing lifecycle).

internal/commands/build.go
```diff
@@ -89,21 +89,11 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
...
-			for _, builder := range suggestedBuilders {
-				if flags.Builder == builder.Image {
-					suggestedBuilder = true
-					break
-				}
+			trustBuilder := isTrustedBuilder(cfg, flags.Builder) || flags.TrustBuilder
+			if trustBuilder {
+				logger.Debugf("Builder %s is trusted", style.Symbol(flags.Builder))
+			} else {
+				logger.Warnf("Builder %s is untrusted", style.Symbol(flags.Builder))
```
dfreilich, line 96: Should this be changed to a `Debugf`?

internal/commands/commands.go
```diff
@@ -81,3 +81,19 @@ func getMirrors(config config.Config) map[string][]string {
 	}
 	return mirrors
 }
+
+func isTrustedBuilder(cfg config.Config, builder string) bool {
```
dfreilich, line 85: This pattern is followed a number of times in the `commands`, and should be refactored out in all cases. If that shouldn't be part of this refactor, I'm happy to undo this change and leave it for another PR.

internal/commands/build.go
jromero: 👍 

Yeah, I think so. I don't think this should be something alarming since we do all the right things to make it none concerning. It's really just something we want them to know (in debug).

Also, might be worth adding a few more sentences about the side effects. eg. "executing sensitive phases separate containers." and/or "using a known lifecycle image to execute sensitive phases."

dfreilich: Added, let me know what you think.

PR #722, opened 2020-07-01
Add warning when building with untrusted builder and volumes

### LINKED ISSUE #529: pack when passed volume mounts should output a warning when running untrusted builders
### Description
Users may be using volume mounts to pass credentials or other sensitive data to the `detect` and `build` phases of the lifecycle. If they are also running an untrusted builder (see #528), this presents a security vulnerability.

### Proposed solution
pack when passed volume mounts should output a warning when running untrusted builders

**AC 1** `pack build my-app -B untrusted-builder --volume maybe-sensitive-data:/mount-target` should output a warning to the effect of "Warning: using untrusted builder with volume mounts. Are you sure you want to proceed?"

Do nothing. Assume users will always use trusted builders when using volume mounts in this way.

dfreilich:
* Fix 529

## Summary

This adds a warning when running `pack build test -p path -B untrusted-builder --volume a:b`

#### After
image

## Documentation

    - [x] No

## Related

Resolves #529

### DIFF
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -100,7 +100,7 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 			}
 
 			if !trustBuilder && len(flags.Volumes) > 0 {
-				logger.Warn("Using untrusted builder with volume mounts. If there is sensitive data in the volumes, this may present a security vulnerability")
+				logger.Warn("Using untrusted builder with volume mounts. If there is sensitive data in the volumes, this may present a security vulnerability.")
 			}
 
 			if err := packClient.Build(cmd.Context(), pack.BuildOptions{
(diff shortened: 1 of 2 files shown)


internal/commands/build.go
```diff
@@ -99,6 +99,10 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 				logger.Debug("For more information, see https://github.com/buildpacks/pack/issues/528")
 			}
 
+			if !trustBuilder && len(flags.Volumes) > 0 {
+				logger.Warn("Using untrusted builder with volume mounts. If there is sensitive data in the volumes, this may present a security vulnerability")
```
jromero, line 103:
```suggestion
				logger.Warn("Using untrusted builder with volume mounts. If there is sensitive data in the volumes, this may present a security vulnerability.")
```

PR #725, opened 2020-07-01
Explicitly deal with untrusting suggested builder

dfreilich:

## Summary

Currently, we have a few holes in our coverage of trusted builders, due to our implementation decision of having 2 separate slices: one written in the config, and the other our `suggestedBuilders`. This attempts to explicitly deal with one of them: 
- If a user attempts to run `pack untrust-builder <SOME_SUGGESTED_BUILDER>`, it should explicitly tell the user that it's not going to do anything, in order that users don't think it's a bug. Alternatively, if we decide on a course of action to do, I'm happy to drive that in this PR; I just think that leaving the hole open leaves it as an open bug.  

#### Before
This isn't true:
image

#### After
This is explicit about the limitation:
image

## Documentation

    - [x] Yes, when we document trust-builders in general

## Related

Relates to https://github.com/buildpacks/pack/issues/670#issuecomment-646019133

### DIFF
--- a/internal/commands/untrust_builder.go
+++ b/internal/commands/untrust_builder.go
@@ -16,19 +16,26 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 		Long:  "Stop trusting builder.\n\nWhen building with this builder, all lifecycle phases will be no longer be run in a single container using the builder image.",
 		Args:  cobra.ExactArgs(1),
 		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
-			builderName := args[0]
-			existingBuilders := cfg.TrustedBuilders
+			builder := args[0]
+
+			existingTrustedBuilders := cfg.TrustedBuilders
 			cfg.TrustedBuilders = []config.TrustedBuilder{}
-			for _, builder := range existingBuilders {
-				if builder.Name == builderName {
+			for _, trustedBuilder := range existingTrustedBuilders {
+				if trustedBuilder.Name == builder {
 					continue
 				}
 
-				cfg.TrustedBuilders = append(cfg.TrustedBuilders, builder)
+				cfg.TrustedBuilders = append(cfg.TrustedBuilders, trustedBuilder)
 			}
 
-			if len(existingBuilders) == len(cfg.TrustedBuilders) {
-				logger.Infof("Builder %s wasn't trusted", style.Symbol(builderName))
+			// Builder is not in the trusted builder list
+			if len(existingTrustedBuilders) == len(cfg.TrustedBuilders) {
+				if isSuggestedBuilder(builder) {
+					// Attempted to untrust a suggested builder
+					return errors.Errorf("Builder %s is a suggested builder, and is trusted by default. Currently pack doesn't support making these builders untrusted", style.Symbol(builder))
+				}
+
+				logger.Infof("Builder %s wasn't trusted", style.Symbol(builder))
 				return nil
 			}
 
@@ -41,7 +48,7 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 				return errors.Wrap(err, "writing config file")
 			}
 
-			logger.Infof("Builder %s is no longer trusted", style.Symbol(builderName))
+			logger.Infof("Builder %s is no longer trusted", style.Symbol(builder))
 			return nil
 		}),
 	}
--- a/internal/commands/untrust_builder_test.go
+++ b/internal/commands/untrust_builder_test.go
@@ -20,7 +20,7 @@ import (
 	h "github.com/buildpacks/pack/testhelpers"
 )
 
-func TestUntrustBuilder(t *testing.T) {
+func TestUntrustBuilderCommand(t *testing.T) {
 	color.Disable(true)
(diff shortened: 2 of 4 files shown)


internal/commands/untrust_builder.go
```diff
@@ -28,7 +34,8 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 			}
 
 			if len(existingBuilders) == len(cfg.TrustedBuilders) {
-				logger.Infof("Builder %s wasn't trusted", style.Symbol(builderName))
+				// Attempted to untrust a suggested builder
+				logger.Warnf("Builder %s currently can't be untrusted", style.Symbol(builderName))
```
simonjjones, line 38: Can we be more explicit about why? Something like `builders that are listed in suggested-builders are trusted by default. Currently pack doesn't support making these builders untrusted`

internal/commands/untrust_builder_test.go
```diff
@@ -126,6 +126,30 @@ func testUntrustBuilderCommand(t *testing.T, when spec.G, it spec.S) {
...
+				builder := "gcr.io/paketo-buildpacks/builder:base"
+				cfg := configManager.configWithTrustedBuilders()
+				command := commands.UntrustBuilder(logger, cfg)
+				command.SetArgs([]string{builder})
+
+				h.AssertNil(t, command.Execute())
+
+				b, err := ioutil.ReadFile(configPath)
+				h.AssertNil(t, err)
+				h.AssertNotContains(t, string(b), builder)
```
simonjjones, line 141: Is this assertion valuable in this context? There's no reason why the suggested builder would appear in this file, especially after all we've invoked is `UntrustBuilder`

internal/commands/untrust_builder.go
```diff
@@ -17,6 +17,12 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
 			builderName := args[0]
+
+			if !isTrustedBuilder(cfg, builderName) {
```
simonjjones, line 21: This command iterates through `cfg.TrustedBuilders` which we do again on line 28. Given discussions around the inefficiency of this data structure being a list we need to iterate through, is it possible to at least only do that once?

dfreilich: I moved `isTrustedBuilder` to within the no-trusted-builder case, and pass in an empty config to maximize efficiency.

internal/commands/untrust_builder.go
```diff
@@ -17,6 +17,7 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
 			builderName := args[0]
+
 			existingBuilders := cfg.TrustedBuilders
```
yaelharel, line 21: It doesn't relate to your PR but maybe it's better to change the variable name from `existingBuilders` to `existingTrustedBuilders`.

internal/commands/untrust_builder.go
```diff
@@ -28,6 +29,12 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 			}
 
 			if len(existingBuilders) == len(cfg.TrustedBuilders) {
+				// Use empty config to just search through suggested builders
+				if isTrustedBuilder(config.Config{}, builderName) {
+					// Attempted to untrust a suggested builder
+					return errors.Errorf("Builder %s is a suggested builder, and is trusted by default. Currently pack doesn't support making these builders untrusted", style.Symbol(builderName))
+				}
+
```
yaelharel, line 37: I would add a new function `isSuggestedBuilder` to `suggested_builders.go` and call it instead of the `isTrustedBuilder` call.
Its implementation is pretty simple:
```
func isSuggestedBuilder(builder string) bool {
	for _, sugBuilder := range suggestedBuilders {
		if builder == sugBuilder.Image {
			return true
		}
	}

	return false
}
```
and it should be called by `isTrustedBuilder` in `commands.go`.

internal/commands/untrust_builder.go
```diff
@@ -28,6 +29,12 @@ func UntrustBuilder(logger logging.Logger, cfg config.Config) *cobra.Command {
 			}
 
 			if len(existingBuilders) == len(cfg.TrustedBuilders) {
```
yaelharel, line 31: I would add a comment - "builderName isn't part of the trusted builders list" or replace this line with a simple function with a meaningful name.

internal/commands/untrust_builder.go
yaelharel: Please see my idea in [this comment](https://github.com/buildpacks/pack/pull/725#discussion_r449322973).

internal/commands/untrust_builder.go
jromero: Agreed. Having this comment "Use empty config to just search through suggested builders" points out that the interface is a little hard to comprehend.

dfreilich:
Shoot I merged main in, which I don't think I was supposed to do.

PR #726, opened 2020-07-01
Combine config.Builder with SuggestedBuilder

dfreilich:
* Standardize usage of bulider.Image, to minimize language changes
* Use command.isTrustedBuilder to reduce reused logic
* Collapsed 2 loops in commands.isTrustedBuilder to one

## Summary

As we haven't yet released a pack that writes `TrustedBuilders` into the config, I wanted to standardize the language in it, to reduce future possible tech debt. As such, I combined the `TrustedBuilder` and `SuggestedBuilder` interfaces into a common `config.Builder` interface, and used `builder.Image` in both cases, rather than having `Image` in one and `Name` in another. 

## Documentation

    - [ ] Yes, see #___
    - [x] No

### DIFF
--- a/internal/commands/build_test.go
+++ b/internal/commands/build_test.go
@@ -90,7 +90,7 @@ func testBuildCommand(t *testing.T, when spec.G, it spec.S) {
 						Build(gomock.Any(), EqBuildOptionsWithTrustedBuilder(true)).
 						Return(nil)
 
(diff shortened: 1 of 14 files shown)


dfreilich:
In this case, because the two interact with each other so often (we loop through a combined `suggestedBuilders` and `trustedBuilders` at least 3 times, and as we may have to decide further what their interaction should be (in #725), I thought it useful to standardize them into one interface, so we don't end up with more annoyances later.

PR #731, opened 2020-07-06
Use isTrustedBuilder func in inspect_builder

dfreilich:
* Add Command to all XCommand tests, to allow for pattern matching when running tests

## Summary

This is to clean up the code slightly - it removes repeated code, and it adds Command to the end of tests, so that I can use `go test -run Command` to run all XCommand tests

## Documentation

    - [x] No

## Related

Resolves #___


internal/commands/list_trusted_builders_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	h "github.com/buildpacks/pack/testhelpers"
 )
 
-func TestListTrustedBuilders(t *testing.T) {
+func TestListTrustedBuildersCommand(t *testing.T) {
```
simonjjones, line 22: Where is this change coming from across the board? We're already in the `commands` package, is it not reasonable to assume that all of our `Test*` cases are testing commands?

dfreilich: It definitely is reasonable, but we have both patterns throughout the `commands` package. In this case, I find it useful to have the tests be `TextXCommand`, because I can then test the specific package by running `go test -run Command`, which I had been doing occasionally, and which I found didn't work as well as could be, given that a number of the commands didn't have that suffix.

PR #734, opened 2020-07-07
Use markdown instead of img html for choco delivery

dfreilich:

## Summary

Fix this error: https://github.com/buildpacks/pack/runs/846975725?check_suite_focus=true

```
The element 'http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd:description' cannot contain child element 'http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd:img' because the parent element's content model is text only.
```
## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


dfreilich:
Tested on Windows VM: 

image

PR #738, opened 2020-07-09
Support default windows lifecycle uri

aemengo:
Signed-off-by: Anthony Emengo <anthonyemengojr@gmail.com>

## Summary

Allow pack to construct the correct lifecycle url when creating a _windows_ builder.

#### Before

```
Creating builder debug/builder from build-image pack-test/build
Using cached version of https://github.com/buildpacks/lifecycle/releases/download/v0.8.0/lifecycle-v0.8.0+linux.x86-64.tgz
...
```

#### After

```
Creating builder debug/builder from build-image pack-test/build
Using cached version of https://github.com/buildpacks/lifecycle/releases/download/v0.8.0/lifecycle-v0.8.0+windows.x86-64.tgz
...
```

## Documentation

    - [x] No

@ameyer-pivotal 
@micahyoung

### DIFF
--- a/create_builder.go
+++ b/create_builder.go
@@ -136,7 +136,7 @@ func (c *Client) createBaseBuilder(ctx context.Context, opts CreateBuilderOption
 		)
 	}
 
-	lifecycle, err := c.fetchLifecycle(ctx, opts.Config.Lifecycle)
+	lifecycle, err := c.fetchLifecycle(ctx, opts.Config.Lifecycle, os)
 	if err != nil {
 		return nil, errors.Wrap(err, "fetch lifecycle")
 	}
@@ -146,7 +146,7 @@ func (c *Client) createBaseBuilder(ctx context.Context, opts CreateBuilderOption
 	return bldr, nil
 }
 
-func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleConfig) (builder.Lifecycle, error) {
+func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleConfig, os string) (builder.Lifecycle, error) {
 	if config.Version != "" && config.URI != "" {
 		return nil, errors.Errorf(
 			"%s can only declare %s or %s, not both",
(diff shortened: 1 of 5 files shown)


jromero:
This is currently failing on Windows due to the windows version of lifecycle not being published.

> https://github.com/buildpacks/lifecycle/releases/download/v0.8.0/lifecycle-v0.8.0+windows.x86-64.tgz returns 404

I'm assuming this would be blocked until a Windows version is published. Is this being coordinated for 0.9.0 and should we be updating the default lifecycle version as part of this?

aemengo:
Yes, you're right that this would be blocked. I don't think anything is being coordinated as of yet, but we could make the decision.

How about we aim for this functionality not being present (or functional) until **v0.9.0**? Would you accept the PR with that caveat in order?

ameyer-pivotal:
@jromero We expect the windows artifact to be on the next release of lifecycle (0.9.0). If y'all are comfortable, it would probably make sense to bump the default lifecycle that pack downloads to 0.9.0 once that release is out. If not, we can work around it too (for instance, issue a warning that a lifecycle path must be set explicitly during builder creation).

create_builder.go
```diff
@@ -303,6 +303,10 @@ func validateBuildpack(bp dist.Buildpack, source, expectedID, expectedBPVersion
 	return nil
 }
 
-func uriFromLifecycleVersion(version semver.Version) string {
+func uriFromLifecycleVersion(version semver.Version, os string) string {
+	if os == "windows" {
```
dfreilich, line 307: Given that the only place the two urls differ is in the name of the OS, can we assign the download OS to either 'windows' or 'linux', and then `sprintf` that into the download url?

aemengo: The string being passed in there is being parsed from another object, not selected from a set of constants. I'm fine with whatever; the only benefit of the code above that I'm thinking is protection from an unsupported return value (ex: `centos`, `mac`, `Windows`, `LINUX`, etc).

create_builder.go
```diff
@@ -136,7 +136,7 @@ func (c *Client) createBaseBuilder(ctx context.Context, opts CreateBuilderOption
 		)
 	}
 
-	lifecycle, err := c.fetchLifecycle(ctx, opts.Config.Lifecycle)
+	lifecycle, err := c.fetchLifecycle(ctx, opts.Config.Lifecycle, os)
```
natalieparellano, line 139: Where does `os` come from? Is it pulled from the builder metadata?

dfreilich:
@aemengo Are we keeping this as draft until the lifecycle cuts a release?

ameyer-pivotal:
> @aemengo Are we keeping this as draft until the lifecycle cuts a release?

@dfreilich That's correct. Once lifecycle is released, we'll update this and undraft (likely early next week)

create_builder.go
aemengo: It comes from the stack build image in the lines above.

aemengo:
Lifecycle cut.
PR undrafted.
Lifecycle bumped to **v0.9.0** as part of this PR.

PR #739, opened 2020-07-10
Support Windows-based images for `build`

### LINKED ISSUE #470: `build` should be able to create a Windows image
### Description
`pack build` currently only creates Linux images. 

### Proposed solution
`pack` (compiled for Linux and Mac) and `pack.exe` (compiled for Windows) should be able to create Windows images.

### Additional context
* This work should follow / is blocked by https://github.com/buildpacks/pack/issues/469
* Leaving buildpackages out of scope for now due to their complexity
* This issue may / is intended to drive out further issues in the [lifecycle](https://github.com/buildpacks/lifecycle) for Windows compatibility
* When `pack` is compiled for Linux or Mac, it is assumed that the docker daemon being targeted is running on Windows
* When the default stack is Linux, should `--run-image` be required for Windows builds?

### Possible acceptance criteria
* Given a builder that is a Windows image
* `pack build windows-app-image --builder windows-builder --buildpack minimal-windows-buildpack --path windows-app` should succeed with and without `--publish`
* `pack.exe build windows-app-image --builder windows-builder --buildpack minimal-windows-buildpack --path windows-app` should succeed with and without `--publish`
* `docker run windows-app-image` should succeed and the app should start correctly

ameyer-pivotal:
- Supports Windows-based images for `pack build` (under experimental flag)

Resolves #470


ameyer-pivotal:
cc: @aemengo

internal/build/container_ops.go
```diff
@@ -3,57 +3,143 @@ package build
...
+	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/BurntSushi/toml"
 	"github.com/docker/docker/api/types"
+	dcontainer "github.com/docker/docker/api/types/container"
```
dfreilich, line 15: Could we do this camelCase? Is that not best practice for imports?

internal/build/build.go
```diff
@@ -26,6 +28,57 @@ type Builder interface {
 	GID() int
 	LifecycleDescriptor() builder.LifecycleDescriptor
 	Stack() builder.StackMetadata
+	Image() imgutil.Image
+}
+
+type mountPaths struct {
```
dfreilich, line 34: nit: Could we move this to the bottom of the file, below the exported funcs/interfaces?

rebase.go
```diff
@@ -20,6 +20,14 @@ type RebaseOptions struct {
 }
 
 func (c *Client) Rebase(ctx context.Context, opts RebaseOptions) error {
+	info, err := c.docker.Info(ctx)
+	if err != nil {
+		return errors.Wrap(err, "getting docker host OS")
+	}
+	if info.OSType == "windows" {
```
dfreilich, line 27: This seems to be validating the docker OS, rather than the OS of the image we are rebasing. Should we also handle the case of someone trying to rebase a Windows container on another OS? That is to say, I run `pack rebase <some-windows-container>` on my `mac`?

acceptance/acceptance_test.go
```diff
@@ -251,11 +251,6 @@ func testWithoutSpecificBuilderRequirement(
 		var tmpDir string
 
 		it.Before(func() {
-			h.SkipIf(t,
```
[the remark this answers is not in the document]
ameyer-pivotal: It was just a bit of cleanup along the way, as all pack versions now support `package-buildpack`. Arguably though, it doesn't have much to do with this PR, so happy to revert if you'd like.

acceptance/acceptance_test.go
```diff
@@ -1525,36 +1519,47 @@ include = [ "*.jar", "media/mountain.jpg", "media/person.png" ]
 				var buildRunImage func(string, string, string)
 
 				it.Before(func() {
+					h.SkipIf(t, dockerHostOS() == "windows", "rebase is not yet supported for Windows containers")
```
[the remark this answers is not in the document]
ameyer-pivotal: Good catch. We had the tests nearly working then discovered a bug for rebase in the Windows container case, so we decided to skip these and play `rebase` in a separate story. The changes below are just leftover. We can remove.

build.go
```diff
@@ -177,11 +185,14 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return c.lifecycle.Execute(ctx, lifecycleOpts)
 	}
 
-	lifecycleImageSupported := lifecycleVersion.Equal(builder.VersionMustParse(prevLifecycleVersionSupportingImage)) || !lifecycleVersion.LessThan(semver.MustParse(minLifecycleVersionSupportingImage))
+	builderOS, err := lifecycleOpts.Builder.Image().OS()
+	if err != nil {
+		return errors.Wrap(err, "getting builder OS")
+	}
+	lifecycleImageSupported := builderOS != "windows" && (lifecycleVersion.Equal(builder.VersionMustParse(prevLifecycleVersionSupportingImage)) || !lifecycleVersion.LessThan(semver.MustParse(minLifecycleVersionSupportingImage)))
```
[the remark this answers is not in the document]
ameyer-pivotal: Sure, can do

internal/archive/archive.go
```diff
@@ -81,7 +81,12 @@ func GenerateTarWithWriter(genFn func(TarWriter) error, twf TarWriterFactory) io
 		errChan <- closeErr
 	}()
 
+	closed := false
 	return ioutils.NewReadCloserWrapper(pr, func() error {
+		if closed {
```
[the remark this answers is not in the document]
ameyer-pivotal: I forget the exact details, but there was problem that occurs when this gets closed too many times, but there was no formal error surfacing, which made it hard to debug.

internal/build/build.go
ameyer-pivotal: Sure

internal/build/container_ops.go
ameyer-pivotal: I believe that the convention for package names (and hence, aliases, it seems) is all lowercase.

internal/build/container_ops.go
```diff
@@ -3,57 +3,143 @@ package build
...
+	<-doneChan
+	if err == nil {
+		err = clientErr
+	}
+
+	return err
+}
+
+// copyDirWindows provides an alternate, Windows container-specific implementation of copyDir.
+// This implementation is needed because copying directly to a mounted volume is currently buggy
```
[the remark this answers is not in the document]
ameyer-pivotal: @micahyoung Do you know if there was a formal issue filed anywhere for this?

rebase.go
ameyer-pivotal: The two are one and the same, as you can only run Windows containers on a Windows docker host (and the other direction is true too: a Windows docker host can only run Windows containers, and not Linux, etc). If you happen to run commands on your Mac that are intended for Windows containers/images, you'd need to set the `DOCKER_HOST` env var to point to a Windows docker host (in fact, this is what we do on a daily basis).

ameyer-pivotal: That said, I think this should actually not be an `ExperimentError`. That was likely a copy/paste artifact.

ameyer-pivotal:
> Overall, this looks great!
> 
> One general thought - would it be better to move some of the common variables/Windows changes (i.e. the change from `root` to `ContainerAdministrator`) to be injected build-time variables (like we use [here](https://github.com/buildpacks/pack/blob/main/acceptance/variables/variables_windows.go#L10)), or would that make it more complicated? I'm not recommending necessarily that we use it, I just wonder what people think. (cc @simonjjones )

@dfreilich It seems to me that the reason for build-time injection would be for values that are subject to change on a build-by-build basis (pack version, etc.). I'm not seeing the value for these particular variables, but open to others' thoughts.

rebase.go
ameyer-pivotal: Also, I should clarify the `info.OSType` is the docker __host__ OS, not necessarily the OS running the docker client. If you have `DOCKER_HOST` set to point to a Windows machine configured to run Windows images, the value of `OSType` will be `windows`, even if you're executing `pack` from a Mac.

internal/builder/lifecycle.go
```diff
@@ -125,7 +126,10 @@ func (l *lifecycle) validateBinaries() error {
 	for _, p := range l.binaries() {
 		_, found := headers[p]
 		if !found {
-			return fmt.Errorf("did not find '%s' in tar", p)
+			_, found = headers[p+".exe"]
```
natalieparellano, line 129: Clever :)

acceptance/acceptance_test.go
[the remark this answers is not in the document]
ameyer-pivotal: Removed.

acceptance/acceptance_test.go
```diff
@@ -901,6 +867,8 @@ func testAcceptance(
 						var buildpackTgz, tempVolume string
 
 						it.Before(func() {
+							h.SkipIf(t, os.Getenv("DOCKER_HOST") != "", "cannot mount volume when DOCKER_HOST is set")
```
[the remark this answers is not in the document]
ameyer-pivotal: Not currently, but since it's a docker usage concern, I wonder if we should link to the [docker docs](https://docs.docker.com/engine/reference/commandline/cli/) somewhere?

acceptance/acceptance_test.go
```diff
@@ -1750,6 +1757,11 @@ func createBuilder(t *testing.T, runImageMirror, configDir, packHome, packPath,
 		"read-env-buildpack",
 	}
 
+	// NOTE: Remove when Windows-based packages are supported
+	if dockerHostOS() == "windows" {
+		buildpacks = append(buildpacks, "simple-layers-buildpack")
```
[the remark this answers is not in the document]
ameyer-pivotal: Great feedback. Will do!

acceptance/testdata/mock_buildpacks/0.2/descriptor-buildpack/bin/detect.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
+
+echo ---- Detect: Descriptor Buildpack
+
+echo ---- Done
```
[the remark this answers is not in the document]
ameyer-pivotal: Done.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
ameyer-pivotal: Removed extraneous changes below.

acceptance/testdata/mock_buildpacks/0.2/not-in-builder-buildpack/bin/build.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
+
+echo ---- Build: Local Buildpack
```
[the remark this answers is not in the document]
ameyer-pivotal: Done.

dfreilich:
For tracking purposes, would you mind also creating a docs issues for documenting building Windows images?

acceptance/acceptance_test.go
```diff
@@ -2028,7 +2046,7 @@ func waitForResponse(t *testing.T, port string, timeout time.Duration) string {
 	for {
 		select {
 		case <-ticker.C:
-			resp, err := h.HTTPGetE("http://localhost:"+port, map[string]string{})
+			resp, err := h.HTTPGetE("http://"+h.RegistryHost(port), map[string]string{})
```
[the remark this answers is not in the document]
ameyer-pivotal: We created an issue in our backlog to port this to `imgutil` after this story is complete (that way it's easier to validate the implementation first).

internal/build/container_ops.go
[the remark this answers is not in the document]
micahyoung: Apologies for the late reply. Here's the issue here: https://github.com/moby/moby/issues/40771

acceptance/acceptance_test.go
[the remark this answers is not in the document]
ameyer-pivotal: Maybe this could be covered in https://github.com/buildpacks/docs/issues/180?

ameyer-pivotal:
> For tracking purposes, would you mind also creating a docs issues for documenting building Windows images?

Done. https://github.com/buildpacks/docs/issues/180

internal/archive/archive.go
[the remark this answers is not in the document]
jromero: I think we'd want to investigate this a tad further. At least have a bit better understanding as to why this was necessary. I'm afraid this could have some adverse effects. If I'm not mistaken we use this strategy in various places so making this change only here either means it's "broken" in multiple places or this particular implementation has issues.

internal/build/container_ops.go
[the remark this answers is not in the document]
jromero: Could we add this issue to the godoc?

build.go
```diff
@@ -625,18 +639,39 @@ func randString(n int) string {
 	return string(b)
 }
 
-func buildPlatformVolumes(volumes []string) ([]string, error) {
+func buildPlatformVolumes(imgOS string, volumes []string) ([]string, error) {
+	// TODO: Consider a cleaner logic for windows/linux differences
```
jromero, line 643: We should resolve this "TODO". I agree with the statement. 😄 

Suggestion: There are a lot of conditionals throughout this PR. It would be nice it it was abstracted out similar to how it was done for `mountPaths`.

testhelpers/testhelpers.go
```diff
@@ -622,6 +583,10 @@ func RunContainer(ctx context.Context, dockerCli client.CommonAPIClient, id stri
 	if err := dockerCli.ContainerStart(ctx, id, dockertypes.ContainerStartOptions{}); err != nil {
 		return errors.Wrap(err, "container start")
 	}
+
+	// wait for logs to show
+	time.Sleep(time.Second)
```
jromero, line 588: This would be dependant on the environment performance. 

Could we discuss why this was necessary to begin with? Maybe we can come up with a better solution.

internal/archive/archive.go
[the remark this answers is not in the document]
ameyer-pivotal: I'll try reverting it and see if it's still needed. It was part of the original implementation, and the PR has evolved quite a bit since then (though I don't believe it has changed much in the area we needed this).

internal/build/container_ops.go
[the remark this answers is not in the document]
ameyer-pivotal: You mean the exported method? How do we add it to the godoc?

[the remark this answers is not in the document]
dfreilich: I think he means adding the moby issue link into the documentation for the method

build.go
```diff
@@ -177,11 +181,12 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
...
-	lifecycleImageSupported := lifecycleVersion.Equal(builder.VersionMustParse(prevLifecycleVersionSupportingImage)) || !lifecycleVersion.LessThan(semver.MustParse(minLifecycleVersionSupportingImage))
-
+	builderOS, err := lifecycleOpts.Builder.Image().OS()
+	if err != nil {
+		return errors.Wrap(err, "getting builder OS")
+	}
 	if !opts.TrustBuilder {
-		switch lifecycleImageSupported {
-		case true:
+		if lifecycleImageSupported(builderOS, lifecycleVersion) {
```
dfreilich, line 189: 👍 to the simplifying func

build_test.go
```diff
@@ -1318,6 +1318,18 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 				})
 
 				when("builder is untrusted", func() {
+					when("building Windows containers", func() {
```
dfreilich, line 1321: i would've expected more unit tests, given that the degree of support this adds, but I guess the existing unit tests create Windows containers already?

build.go
```diff
@@ -625,18 +639,39 @@ func randString(n int) string {
 	return string(b)
 }
 
-func buildPlatformVolumes(volumes []string) ([]string, error) {
+func buildPlatformVolumes(imgOS string, volumes []string) ([]string, error) {
```
dfreilich, line 642: Could we add in a unit tests verifying that the volume is constructed appropriately?

internal/build/phase_config_provider.go
```diff
@@ -94,8 +105,12 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
 
 func WithDaemonAccess() PhaseConfigProviderOperation {
 	return func(provider *PhaseConfigProvider) {
-		provider.ctrConf.User = "root"
-		provider.hostConf.Binds = append(provider.hostConf.Binds, "/var/run/docker.sock:/var/run/docker.sock")
+		WithRoot()(provider)
```
dfreilich, line 108: What's happening here? This is calling `WithRoot`, and naming the return value `provider`?

.github/workflows/build.yml
```diff
@@ -67,15 +73,38 @@ jobs:
...
         run: make verify
+      - name: Test-WCOW
+        shell: powershell
+        if: matrix.config == 'windows-wcow'
+        env:
+          TEST_COVERAGE: 1
+          # https://github.com/buildpacks/lifecycle/actions/runs/159882275
+          # https://github.com/buildpacks/lifecycle/suites/883071116/artifacts/10407089a
+          # LIFECYCLE_URL temporarily retrieved from the following until the windows lifecycle binary is released
+          LIFECYCLE_URL: https://windows-bosh-temp.s3.amazonaws.com/lifecycle-windows-x86-64.zip
```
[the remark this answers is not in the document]
ameyer-pivotal: The artifacts required a login, so we opted for a simpler solution temporarily. Realistically, lifecycle is to be released today (or early next week) with Windows support, so we'll likely be using that.

Additionally/alternately, we can leave this as is for this PR and remove it as part of #738, which might make more sense.

.github/workflows/build.yml
```diff
@@ -67,15 +73,38 @@ jobs:
...
+
+          go test -v -count=1 -parallel=1 -timeout=900s -tags=acceptance ./acceptance
       - name: Test
+        if: matrix.config != 'windows-wcow'
         env:
           TEST_COVERAGE: 1
         run: make test
       - name: Upload Coverage
         uses: codecov/codecov-action@v1
+        if: matrix.config != 'windows-wcow'
```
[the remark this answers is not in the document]
ameyer-pivotal: Interesting. I see that we have TEST_COVERAGE enabled for `Test-WCOW`, so perhaps we can turn this back on.

Makefile
```diff
@@ -1,13 +1,30 @@
+ifeq ($(OS),Windows_NT)
+SHELL:=cmd.exe
+BLANK:=
+/:=\$(BLANK)
```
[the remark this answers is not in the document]
ameyer-pivotal: We're defining an make variable named `/` that you can use with `$/`. It will resolve to the correct slash direction based on OS. Unfortunately, `/:=\` doesn't work syntactically for makefiles, so this is a [workaround](https://stackoverflow.com/a/54733416/2456938)

acceptance/acceptance_test.go
```diff
@@ -641,6 +609,8 @@ func testAcceptance(
 					})
 
 					it("uses the 5 phases", func() {
+						h.SkipIf(t, dockerHostOS() == "windows", "lifecycle images are not yet supported on windows builds")
```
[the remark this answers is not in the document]
ameyer-pivotal: It's really about lack of support for untrusted builders, so we'll clarify this and move it up to the `it.Before`.

acceptance/acceptance_test.go
```diff
@@ -1222,12 +1214,20 @@ func testAcceptance(
 
 						when("the run-image has the correct stack ID", func() {
 							it.Before(func() {
+								user := func() string {
+									if dockerHostOS() == "windows" {
+										return "ContainerAdministrator"
+									}
+
+									return "root"
+								}
```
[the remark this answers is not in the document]
ameyer-pivotal: @jromero Believe me, I wonder the same thing often...

@dfreilich Unfortunately, we can't add these to those build-time files, as they are not related to the `GOOS`, but rather the OS of the docker host, which can only be resolved at runtime.

acceptance/acceptance_test.go
```diff
@@ -1720,6 +1753,12 @@ func createBuilder(
 		"read-env-buildpack",
 	}
 
+	// NOTE: Windows-based packages are not yet supported, so we'll add this buildpack in the usual way for now.
```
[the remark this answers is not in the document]
ameyer-pivotal: There is no issue on the repo, but it's in our team backlog for sure.

Makefile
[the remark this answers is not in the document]
jromero: Interesting. This makes sense. Can we add this comment to the Makefile so the knowledge is transferable.

.github/workflows/build.yml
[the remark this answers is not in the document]
ameyer-pivotal: As yet another alternative, we could prioritize getting #738 merged before this PR, since that one really has to do with `create-builder` functionality, which has been in `main` for a few weeks now.

Makefile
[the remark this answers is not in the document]
ameyer-pivotal: Will do :)

Makefile
```diff
@@ -1,13 +1,30 @@
...
+CAT=type
+RMRF=rmdir /q /s
+SRC=$(shell dir /q /s /b *.go | findstr /v $/vendor$/)
+GOIMPORTS_DIFF_OPTION="-l" # Windows can't do diff-mode because it's missing the "diff" binary
+PACK_BIN?=pack.exe
+else
+/:=/
```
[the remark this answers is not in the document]
ameyer-pivotal: We wanted to draw an ASCII Tie Fighter ;)

JK, will document, as per @jromero 's comment

build_test.go
ameyer-pivotal: Most of the new functionality is covered by acceptance tests. Because pack uses a generic `TarWriter` interface to write layers (which is swapped out based on the target container OS) there's not a lot of extra tests to write other than checking "wiring", which is covered by acceptance.

The code for erroring out when a builder is untrusted was not explicitly covered by acceptance, so we added it here.

internal/build/container_ops.go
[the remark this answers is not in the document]
ameyer-pivotal: Oh, haha, thanks @dfreilich. Will do.

internal/build/phase_config_provider.go
ameyer-pivotal: Because `WithRoot()` is a higher-order function and returns a function, we have to then call that returned function to perform its behavior.

testhelpers/testhelpers.go
ameyer-pivotal: We added this because the container logs were always showing up empty on the WCOW runner (but not WCOW locally). We don't know the root cause, but it had to do with the container perhaps quitting too soon in some scenarios. At first, we thought it was because process isolation (`docker run --isolation=process` equivalent) was not enabled on the runner (it's enabled locally on our VMs), but that was proven wrong. There's something about the runner configuration that is causing logging capture to run more slowly than than it does locally.

We were originally considering putting this under a `dockerHostOS() == "windows"` guard, but didn't want to add yet another branch for that unless necessary. Happy to do that after all if you're concerned about performance.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
natalieparellano: We have lifecycle images for Windows now! See [here](https://hub.docker.com/layers/buildpacksio/lifecycle/0.9.0/images/sha256-e33782fd03f3bd866a7ef796a095e79c461c9aa0b9b1d61919718013267218c1?context=explore) - I could see a case for implementing untrusted builder workflows for Windows as a separate unit of work. I agree with @dfreilich that the wording could be more clear.

internal/cache/volume_cache_test.go
```diff
@@ -29,6 +29,29 @@ func TestVolumeCache(t *testing.T) {
 }
 
 func testCache(t *testing.T, when spec.G, it spec.S) {
+	when("#FilterReservedNames", func() {
+		when("volume contains a reserved name", func() {
+			it("modifies the volume name", func() {
```
natalieparellano, line 34: Are we filtering even when the docker host is Linux?

testhelpers/testhelpers.go
natalieparellano: Maybe try wrapping in a context with a cancel? Like what we do [here](https://github.com/buildpacks/pack/blob/3a3bf282e703ee50c7f9e189015a3a9914b359a1/testhelpers/registry.go#L96) for the registry?

internal/build/phase_config_provider_test.go
```diff
@@ -53,6 +54,22 @@ func testPhaseConfigProvider(t *testing.T, when spec.G, it spec.S) {
...
+			it("sets process isolation", func() {
+				fakeBuilderImage := ifakes.NewImage("fake-builder", "", nil)
+				fakeBuilderImage.SetPlatform("windows", "", "")
+				fakeBuilder, err := fakes.NewFakeBuilder(fakes.WithImage(fakeBuilderImage))
+				h.AssertNil(t, err)
+				lifecycle := newTestLifecycle(t, false, fakes.WithBuilder(fakeBuilder))
+
+				phaseConfigProvider := build.NewPhaseConfigProvider("some-name", lifecycle)
+
+				h.AssertEq(t, phaseConfigProvider.HostConfig().Isolation, container.IsolationProcess)
```
natalieparellano, line 71: We might consider adding a test that process isolation is NOT set for Linux containers.

acceptance/acceptance_test.go
```diff
@@ -482,7 +482,9 @@ func testAcceptance(
 
 	it.Before(func() {
 		pack = invoke.NewPackInvoker(t, assert, subjectPackConfig, registryConfig.DockerConfigDir)
+		pack.EnableExperimental()
```
natalieparellano, line 485: At first I was going to suggest conditionalizing this on the daemon host OS being Windows, but I think this is the right call.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
natalieparellano: You might add "pack developer" to the "target audience" list in that issue.

acceptance/acceptance_test.go
```diff
@@ -1561,18 +1590,29 @@ include = [ "*.jar", "media/mountain.jpg", "media/person.png" ]
 				var buildRunImage func(string, string, string)
 
 				it.Before(func() {
+					pack.JustRunSuccessfully("trust-builder", builderName)
```
natalieparellano, line 1593: I'm guessing this was needed because we don't support untrusted builder workflows on Windows yet (though I'm confused how it relates to rebase, could you explain?). 

I'm a bit concerned about this setup because it could potentially pollute other tests (it modifies that pack config file). Is there any way to pass `--trust-builder` instead?

internal/build/phases.go
```diff
@@ -38,7 +30,7 @@ func (l *Lifecycle) Create(
 	phaseFactory PhaseFactory,
 ) error {
 	flags := []string{
-		"-cache-dir", cacheDir,
+		"-cache-dir", l.mountPaths.cacheDir(),
```
natalieparellano, line 33: I'm surprised this didn't prompt a bunch of changes in phases_test.go - did I miss something?

build_test.go
```diff
@@ -1317,6 +1317,18 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+						it("errors and mentions that builder must be trusted", func() {
+							defaultBuilderImage.SetPlatform("windows", "", "")
+							h.AssertError(t, subject.Build(context.TODO(), BuildOptions{
+								Image:        "some/app",
+								Builder:      defaultBuilderName,
+								Publish:      true,
+								TrustBuilder: false,
+							}), "does not have an associated lifecycle image. Builder must be trusted.")
```
natalieparellano, line 1328: We'll probably want to update this language to say something like "is not supported" (since there are lifecycle images for Windows now).

build.go
```diff
@@ -180,11 +184,12 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return c.lifecycle.Execute(ctx, lifecycleOpts)
 	}
 
-	lifecycleImageSupported := lifecycleVersion.Equal(builder.VersionMustParse(prevLifecycleVersionSupportingImage)) || !lifecycleVersion.LessThan(semver.MustParse(minLifecycleVersionSupportingImage))
-
+	builderOS, err := lifecycleOpts.Builder.Image().OS()
```
natalieparellano, line 187: I'm confused... what's the point of getting `imgOS` and `builderOS` - could they be different?

build.go
```diff
@@ -628,16 +642,23 @@ func randString(n int) string {
...
 			return nil, nil, errors.Wrapf(err, "platform volume %q has invalid format", v)
 		}
 
-		for _, p := range [...]string{"/cnb", "/layers"} {
+		sensitiveDirs := []string{"/cnb", "/layers"}
+		if imgOS == "windows" {
+			sensitiveDirs = []string{`c:/cnb`, `c:\cnb`, `c:/layers`, `c:\layers`}
+		}
+		for _, p := range sensitiveDirs {
 			if strings.HasPrefix(volume.Spec.Target, p) {
```
natalieparellano, line 662: Is it worth lowercasing the target, in case the user passes `C:\`? Unless `volume.Spec.Target` sanitizes this...

build_test.go
natalieparellano: Is it worth expanding the tests here: https://github.com/buildpacks/pack/blob/3a3bf282e703ee50c7f9e189015a3a9914b359a1/build_test.go#L1826 to cover more of the posix cases?

.github/workflows/build.yml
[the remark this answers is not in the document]
ameyer-pivotal: We're going to go with the latest approach and get #738 in first, then we'll rebase this

acceptance/acceptance_test.go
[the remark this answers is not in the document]
ameyer-pivotal: Is that audience covered by buildpacks.io??

acceptance/acceptance_test.go
ameyer-pivotal: I understand the confusion. Trusting the builder is needed for the reason you mentioned, but we added this line because `when("rebase")` is a sibling to the `when("build")` where this normally happens (under `when("default builder is set")`, so we duplicated it here. I don't feel strongly one way or the other regarding `--trust-builder` instead, though the pollution argument would be a problem already (due to the `when("build")`.

build.go
ameyer-pivotal: Ah, good catch. That was an oversight as the `imgOS` part was added at a later time. Will fix!

build.go
ameyer-pivotal: Good catch! Will do

build_test.go
ameyer-pivotal: > We'll probably want to update this language to say something like "is not supported" (since there are lifecycle images for Windows now).

This language comes from shared code, so we'd have to rework it to separate this scenario out. Given that this is temporary and the argument is somewhat specific (and Windows builds is experimental), would you be willing to allow us to punt on it?

ameyer-pivotal: > Is it worth expanding the tests here:
> 
> https://github.com/buildpacks/pack/blob/3a3bf282e703ee50c7f9e189015a3a9914b359a1/build_test.go#L1826
> 
> to cover more of the posix cases?

Do you mean more `windows` cases? If so, @dfreilich asked a similar question above (see collapsed convo).

internal/build/phase_config_provider_test.go
ameyer-pivotal: We do test this :) See a few lines above (L58 currently)

internal/cache/volume_cache_test.go
ameyer-pivotal: It does, but we'll take another look at doing it only for Windows, so we don't bust folks' caches

testhelpers/testhelpers.go
ameyer-pivotal: @natalieparellano Could you elaborate on that suggestion? I'm a bit confused as we want to block for 1 second, not timeout after a certain amount of time. I could be missing something though :)

testhelpers/testhelpers.go
```diff
@@ -310,36 +307,6 @@ func dockerCli(t *testing.T) client.CommonAPIClient {
 	return dockerCliVal
 }
 
-func proxyDockerHostPort(port string) error {
```
ameyer-pivotal, line 313: I'm so glad this stuff is going away!

internal/build/phases.go
ameyer-pivotal: Hmm... I'll have to look into it. I'm unsure off the top of my head why `phases_test.go` isn't failing for a Windows dockerhost. Will investigate.

testhelpers/testhelpers.go
jromero: I would hate for this to be the reason this doesn't make it to the next release. For now, can we guard it as you suggested and create an issue so that we can revisit it afterwards.

acceptance/acceptance_test.go
natalieparellano: Oh, ha! I guess I'm wondering, if you remove the setup, do the rebase tests still pass? I would assume so, but maybe I'm wrong...

build_test.go
natalieparellano: Ah, [this convo](https://github.com/buildpacks/pack/pull/739#discussion_r467103611)? I was thinking specifically about the volume mount tests - though now that I'm reading those again, I see that they don't have to do with the OS of the container, just the host. So it's probably fine... maybe consider adding Windows-type directories here: https://github.com/buildpacks/pack/blob/3a3bf282e703ee50c7f9e189015a3a9914b359a1/build_test.go#L1801-L1808

internal/cache/volume_cache_test.go
ameyer-pivotal: @natalieparellano We looked into this and we can make it Windows-container specific, however it makes the code a bit more complex as we have to key off of the image/dockerhost OS in `NewVolumeCache()`. That requires a `context.Context` object to be passed in, and changes the return values of `NewVolumeCache()` to include an `error`. Is it worth it to y'all?

If not, the only downside, is a one-time rebuild cost for the small subset of pack users who happened to have one of the reserved words in their hashed volume name

acceptance/acceptance_test.go
ameyer-pivotal: @natalieparellano If this were removed, the tests would pass in the Linux container case and fail in the Windows container case

build_test.go
ameyer-pivotal: @natalieparellano The docker host OS and the container OS always align, so you can think of them as the same thing

internal/cache/volume_cache_test.go
natalieparellano: I don't feel strongly about this. @jromero ?

acceptance/acceptance_test.go
natalieparellano: Ohhh... I think I get it now. I was struggling to see how "trusted builder" has anything to do with rebase - but it looks like the test setup has a `pack build` that would fail if the builder is not trusted. Maybe it's worth adding a comment, so that others don't also get confused. Or you could use `--trust-builder` just where it's needed.

build_test.go
natalieparellano: Hmm... in that case, I would expect the `windows` subset of tests to have the same number of cases as the `posix` ones, no? It looks like for Windows we just check that the drive is transformed, and on posix we test all sort of things.

.github/workflows/build.yml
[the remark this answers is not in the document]
ameyer-pivotal: Looks like we're all good on this now.

internal/archive/archive.go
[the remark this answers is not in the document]
jromero: Deferred to https://github.com/buildpacks/pack/issues/792

build_test.go
jromero: Deferred to https://github.com/buildpacks/pack/issues/793

acceptance/acceptance_test.go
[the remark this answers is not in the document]
jromero: It's not an I don't think it should. "pack developers" should be looking at the repo *.md or wiki.

internal/cache/volume_cache_test.go
jromero: > I don't feel strongly about this.

Same. The cost vs benefit doesn't seem worth it at this point in time.

internal/build/phases.go
ameyer-pivotal: OK, looked into this. Those tests don't actually start containers, so they pass for any type of dockerhost. There aren't any window-specific changes that need to be tested in there, as it's mostly checking that the proper `mountPaths` functions were called. The differences in `mountPaths` behavior for each dockerhost OS is covered by the acceptance tests.

You could perhaps make an argument for unit testing `mountPaths` in a whitebox test, but it wouldn't increase coverage per se.

PR #757, opened 2020-07-21
PackInvoker manages interactions with the tested pack CLI

simonjjones:

## Summary

This change creates PackInvoker object which handles the task invoking `pack` commands throughout the acceptance suite, and a PackFixtureManager object which finds and templates fixtures for use in the tests.

There is a small change to the logic of which fixture is used for one test, which I believe uses the more appropriate source. Other than this, hopefully there are no changes to the logic of what is being tested, just rearrangement for clarity.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Contributes to #606


acceptance/config/pack_assets.go
```diff
@@ -0,0 +1,25 @@
+// +build acceptance
+
+package config
+
+type PackAssetManager struct {
```
jromero, line 5: Nitpick: `Manager` in my mind entails a service class when in reality this appears to simply be a data type.

```suggestion
type PackAsset struct {
```

simonjjones: Renamed throughout

simonjjones:
@jromero Should I be worried about the codecov report for this? I don't see how it can be right, as I haven't touched the unit tests, but it's alarmingly red, and don't want to ignore it out of hand.

jromero:
@simonjjones it would be nice to know _why_ it changed although I agree with you that it shouldn't have.

jromero:
Ok, so I think I've found the issue. It appears that both Windows jobs did not complete successfully on `main` on the latest commit. This in effect raised the _average_ code coverage percentage (aggregate of all build outcomes):


This PR (and any subsequent PRs) will lower the overall code coverage average as they properly report all build outcomes. Windows results effectively dragging the code coverage average down.

image

acceptance/acceptance_test.go
```diff
@@ -799,22 +791,23 @@ func testAcceptance(
...
+
 							h.AssertEq(t, output, expectedOutput)
 						}
 					})
 
 					when("--no-color", func() {
 						it.Before(func() {
-							h.SkipIf(t,
-								packSemver.LessThan(semver.MustParse("0.12.0")) || !packSemver.Equal(semver.MustParse("0.0.0")),
+							h.SkipUnless(t,
```
dfreilich, line 801: I love this addition

acceptance/invoke/pack.go
```diff
@@ -0,0 +1,273 @@
...
+func (i *PackInvoker) RunSuccessfully(name string, args ...string) string {
+	i.testObject.Helper()
+
+	output, err := i.Run(name, args...)
+	h.AssertNil(i.testObject, err)
+
+	return output
+}
+
+func (i *PackInvoker) JustRunSuccessfully(name string, args ...string) {
```
dfreilich, line 101: I appreciate this change, rather than the `PackWithOutput` pattern before. Thanks!

PR #763, opened 2020-07-23
460 remove symlink

### LINKED ISSUE #460: Remove lifecycle symlink from lifecycle layer creation logic
The symlink from `/lifecycle -> /cnb/lifecycle` is no longer needed in builders (after releasing pack 0.8.0), as pack 0.7.0+ will invoke lifecycle from `/cnb/lifecycle`.

### Acceptance Criteria

**Given** a `pack` with this change
**When** I create a builder with `pack create-builder`
**And** I run a container from the builder image
**Then** I should not see a symlink from `/lifecycle` to `/cnb/lifecycle` in the filesystem

### Notes

See below where we are creating the symlink today:

https://github.com/buildpacks/pack/blob/070802169e288f849a9ddd7ad0ce64f5de59ee87/internal/builder/builder.go#L664-L672

dwillist:
## Summary
- Remove code to create obsolete `/lifecycle` symlink in builders

- Small fix to run image mirror tests to disable parallel test execution. 

#### Before
Directory structure of builders:
```
/
├── bin
├── ...
├── cnb
├── ...
└── lifecycle
```
#### After
Directory structure of builders:
```
/
├── bin
├── ...
├── cnb
└── ...
```

## Documentation
No documentation needed

## Related
Resolves #460

### DIFF
--- a/internal/commands/set_run_image_mirrors_test.go
+++ b/internal/commands/set_run_image_mirrors_test.go
@@ -18,7 +18,7 @@ import (
 )
 
 func TestSetRunImageMirrorsCommand(t *testing.T) {
-	spec.Run(t, "Commands", testSetRunImageMirrorsCommand, spec.Parallel(), spec.Report(report.Terminal{}))
+	spec.Run(t, "Commands", testSetRunImageMirrorsCommand, spec.Sequential(), spec.Report(report.Terminal{}))
 }
 
 func testSetRunImageMirrorsCommand(t *testing.T, when spec.G, it spec.S) {
--- a/internal/builder/builder.go
+++ b/internal/builder/builder.go
@@ -32,13 +32,12 @@ const (
 
 	cnbDir = "/cnb"
 
-	orderPath          = "/cnb/order.toml"
-	stackPath          = "/cnb/stack.toml"
-	platformDir        = "/platform"
-	lifecycleDir       = "/cnb/lifecycle"
-	compatLifecycleDir = "/lifecycle"
-	workspaceDir       = "/workspace"
-	layersDir          = "/layers"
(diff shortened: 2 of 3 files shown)


internal/commands/set_run_image_mirrors_test.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 func TestSetRunImageMirrorsCommand(t *testing.T) {
-	spec.Run(t, "Commands", testSetRunImageMirrorsCommand, spec.Parallel(), spec.Report(report.Terminal{}))
+	spec.Run(t, "Commands", testSetRunImageMirrorsCommand, spec.Sequential(), spec.Report(report.Terminal{}))
```
natalieparellano, line 21: I'm sort of confused by this as it seems we use `spec.Parallel()` more often than `spec.Sequential()`. What's the reason that this test needs to be run sequentially?

dwillist: This test is uses environment variables during setup [here](https://github.com/buildpacks/pack/blob/5b6731320073fe1451f440ac93c27edf2f792a2a/internal/commands/set_run_image_mirrors_test.go#L41) and [here](https://github.com/buildpacks/pack/blob/5b6731320073fe1451f440ac93c27edf2f792a2a/internal/commands/set_run_image_mirrors_test.go#L45), environment variables are shared by goroutines so running this test with any sort of parallelism will have race conditions. Using `make unit` runs unit tests sequentially by using a single goroutine configured [here](https://github.com/dwillist/pack/blob/5b6731320073fe1451f440ac93c27edf2f792a2a/Makefile#L5) so they end up passing.

PR #773, opened 2020-08-04
Add pull-policy flag to build/create-builder/package-buildpack/rebase commands

### LINKED ISSUE #686: Support `pack build --pull-if-needed`
By default, `pack` will always try to pull the builder and run images if newer versions exist.
With the `--no-pull` flag, it'll never pull and fail if no image is found locally.
We could have a new `--pull-if-needed` flag in pack that doesn't pull an image if there's one with the same tag locally, no matter if it's outdated.

dfreilich:
## Summary

This deprecates the `--no-pull` flag, in favor of `--pull-policy`, which as three options at the moment:
* `always` &rarr; Always pulls the images necessary (the default behavior of pack)
* `never` &rarr; Never pulls the images necessary (equivalent to `--no-pull`)
* `if-not-present` &rarr; Pulls the image only if it is not present

It also adds the possibility for future policies to be added, by centralizing the `PullPolicy`s. 

#### Before
Screen Shot 2020-08-05 at 9 06 42 AM

#### After
Screen Shot 2020-08-05 at 9 12 23 AM

#### Before
Screen Shot 2020-08-05 at 9 02 12 AM

#### After
Screen Shot 2020-08-05 at 9 04 40 AM

#### Before
image

#### After
image

#### Before
Screen Shot 2020-08-05 at 9 20 24 AM

#### After
image

## Documentation

    - [x] Yes, and it will be automatically through the generated pack documentation

## Related

Resolves #686 
Implements https://github.com/buildpacks/rfcs/pull/80

### DIFF
--- a/build.go
+++ b/build.go
@@ -92,7 +92,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "invalid builder '%s'", opts.Builder)
 	}
 
-	rawBuilderImage, err := c.imageFetcher.NewFetch(ctx, builderRef.Name(), true, opts.PullPolicy)
+	rawBuilderImage, err := c.imageFetcher.Fetch(ctx, builderRef.Name(), true, opts.PullPolicy)
 	if err != nil {
 		return errors.Wrapf(err, "failed to fetch builder image '%s'", builderRef.Name())
 	}
@@ -183,7 +183,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 	if !opts.TrustBuilder {
 		switch lifecycleImageSupported {
 		case true:
-			lifecycleImage, err := c.imageFetcher.NewFetch(
+			lifecycleImage, err := c.imageFetcher.Fetch(
 				ctx,
 				fmt.Sprintf("%s:%s", lifecycleImageRepo, lifecycleVersion.String()),
 				true,
@@ -254,7 +254,7 @@ func (c *Client) validateRunImage(context context.Context, name string, pullPoli
 	if name == "" {
 		return nil, errors.New("run image must be specified")
 	}
-	img, err := c.imageFetcher.NewFetch(context, name, !publish, pullPolicy)
+	img, err := c.imageFetcher.Fetch(context, name, !publish, pullPolicy)
 	if err != nil {
 		return nil, err
 	}
--- a/build.go
+++ b/build.go
@@ -53,7 +53,6 @@ type BuildOptions struct {
 	AdditionalMirrors  map[string][]string // only considered if RunImage is not provided
 	Env                map[string]string
 	Publish            bool
-	NoPull             bool
 	ClearCache         bool
 	TrustBuilder       bool
 	Buildpacks         []string
@@ -93,7 +92,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "invalid builder '%s'", opts.Builder)
 	}
 
-	rawBuilderImage, err := c.imageFetcher.Fetch(ctx, builderRef.Name(), true, !opts.NoPull)
+	rawBuilderImage, err := c.imageFetcher.NewFetch(ctx, builderRef.Name(), true, opts.PullPolicy)
 	if err != nil {
 		return errors.Wrapf(err, "failed to fetch builder image '%s'", builderRef.Name())
 	}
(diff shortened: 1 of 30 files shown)


config/pull_policy.go
```diff
@@ -0,0 +1,41 @@
...
+func ParsePolicyFromPull(pull bool) PullPolicy {
+	if pull {
+		return PullAlways
+	}
+
+	return PullNever
+}
+
+func (p PullPolicy) String() string {
+	return [...]string{"always", "never", "if-not-present"}[p]
```
jromero, line 40: What are these hieroglyphics? 🤔

dfreilich: Haha. I wanted to use these as `Enums`, but also be able to access the string form of them, so that I could assign that value to a string that would then be parsed into a `PullPolicy`. It's defining an array of strings that is accessed by the value of the `Policy` – `PullAlways` is `0`, `PullNever` is `1`, and `PullIfNotPresent` is `2` (still not implemented, btw, which is why it's still a draft)

build.go
```diff
@@ -92,7 +94,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "invalid builder '%s'", opts.Builder)
 	}
 
-	rawBuilderImage, err := c.imageFetcher.Fetch(ctx, builderRef.Name(), true, !opts.NoPull)
+	rawBuilderImage, err := c.imageFetcher.Fetch(ctx, builderRef.Name(), true, opts.PullPolicy)
```
dfreilich, line 97: From a code readability standpoint, I appreciate the removal of double negatives (from here, and many other place throughout the code).

config/pull_policy.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/pkg/errors"
+)
+
+// PullPolicy defines a policy for how to manage images
+type PullPolicy int
+
+const (
+	// PullAlways images, even if they are present
+	PullAlways PullPolicy = iota
```
dfreilich, line 12: `iota` is useful for defining enums. For more on iotas, see https://github.com/golang/go/wiki/Iota or  https://splice.com/blog/iota-elegant-constants-golang/

On the other hand, I also debated just defining these as strings, which would've obviated the need for a secondary `String()` method. If people would rather change it to a string type to make it simpler, I'm happy to do that.

config/pull_policy.go
dfreilich: But I totally understand if it's not clear, and you'd want me to remove it/clarify it.

create_builder.go
```diff
@@ -64,7 +66,7 @@ func (c *Client) validateRunImageConfig(ctx context.Context, opts CreateBuilderO
 	var runImages []imgutil.Image
 	for _, i := range append([]string{opts.Config.Stack.RunImage}, opts.Config.Stack.RunImageMirrors...) {
 		if !opts.Publish {
-			img, err := c.imageFetcher.Fetch(ctx, i, true, false)
+			img, err := c.imageFetcher.Fetch(ctx, i, true, opts.PullPolicy)
```
dfreilich, line 69: There were times where we had ignored the `--no-pull` option, which I think we shouldn't do, once we explicitly ask people what they want to do.

internal/commands/build.go
```diff
@@ -141,20 +148,24 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
...
 	cmd.Flags().StringArrayVar(&buildFlags.Volumes, "volume", nil, "Mount host volume into the build container, in the form '<host path>:<target path>'. Target path will be prefixed with '/platform/'"+multiValueHelp("volume"))
 	cmd.Flags().StringVarP(&buildFlags.DefaultProcessType, "default-process", "D", "", "Set the default process type")
+	cmd.Flags().StringVar(&buildFlags.Policy, "pull-policy", "", "Pull policy to use. Accepted values are always, never, and if-not-present. The default is always")
+	// TODO: Remove --no-pull flag after v0.13.0 released. See https://github.com/buildpacks/pack/issues/775
+	cmd.Flags().BoolVar(&buildFlags.NoPull, "no-pull", false, "Skip pulling builder and run images before use")
+	cmd.Flags().MarkHidden("no-pull")
 }
 
-func validateBuildFlags(flags BuildFlags, logger logging.Logger, cfg config.Config, packClient PackClient) error {
+func validateBuildFlags(flags *BuildFlags, cfg config.Config, packClient PackClient, logger logging.Logger) error {
```
dfreilich, line 168: From a readability perspective, I found it confusing for the logger to be so early in the parameters, so I took the liberty of moving it to the end of the parameter list

config/pull_policy.go
jromero: This seems slightly error prone since it's tied to the order in which the enumeration is declared. Granted that the tests are in place to prevent any misalignment it still doesn't seem like the most ideal setup.

dfreilich: Makes sense. That's why I made the tests, but it definitely doesn't seem the most secure.

PR #774, opened 2020-08-04
Add AssertionManager and OutputAssertionManager

simonjjones:
## Summary

This change creates an AssertionManager, which provides a slightly different abstraction for common types of assertion, an OutputAssertionManager which provides a way to abstract the actual content of messages away from tests checking whether the messages are printed in various scenarios

Hopefully there are no changes to the logic of what is being tested, just rearrangement for clarity.

## Documentation
    - [ ] Yes, see #___
    - [x] No

## Related

Contributes to #606


acceptance/assertions/output.go
```diff
@@ -0,0 +1,276 @@
...
+func (o OutputAssertionManager) ReportsConnectedToInternet() {
+	o.testObject.Helper()
+
+	o.assert.Contains(o.output, "RESULT: Connected to the internet")
+}
+
+func (o OutputAssertionManager) ReportsDisconnectedFromInternet() {
+	o.testObject.Helper()
+
+	o.assert.Contains(o.output, "RESULT: Disconnected from the internet")
```
jromero, line 179: It feels like this "manager" is a little overloaded. I can see the value in common pack specific output given that is what the project repo is for. 

On the other hand, abstracting other output that is more specific to a particular test makes it just an abstraction leading to a burden with little to no value if we were to expect the pattern to be followed.

acceptance/assertions/output.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func (o OutputAssertionManager) ReportsExporterReusingUnchangedLayer(layer string) {
+	o.testObject.Helper()
+	o.testObject.Log("exporter reuses unchanged layers")
+
+	o.assert.Matches(o.output, regexp.MustCompile(fmt.Sprintf(`(?i)Reusing layer '%s'`, layer)))
+}
+
+func (o OutputAssertionManager) ReportsCacheReuse(layer string) {
```
jromero, line 78: Related to the `OutputAssertionManager` being "overloaded".

Another potential improvement here may be to create a separation for the parties producing the output. For instance this output is produced by the lifecycle and not pack so something like `LifecycleOutputAssertionManager` would further encapsulate and clarify responsibilities.

acceptance/acceptance_test.go
```diff
@@ -55,16 +55,18 @@ func TestAcceptance(t *testing.T) {
 	h.RequireDocker(t)
 	rand.Seed(time.Now().UTC().UnixNano())
 
+	assert := h.NewAssertionManager(t)
+
 	dockerCli, err = client.NewClientWithOpts(client.FromEnv, client.WithVersion("1.38"))
-	h.AssertNil(t, err)
+	assert.Nil(err)
```
jromero, line 61: Visually this new `assert.X` style looks like a 🏆 .

simonjjones: It feels nice to work with, I've put it in `testhelpers` too, so immediately usable in all our test suites.

acceptance/assertions/output.go
simonjjones: Given the `LifecycleOutputAssertionManager` idea floated in a related comment, do you think something like a `TestAppOutputAssertionManager` be a worthwhile abstraction, that would ensure the separation of concerns?

simonjjones: Or more accurately `TestBuildpackOutputAssertionManager`... becoming somewhat of a mouthful

jromero: Hahaha. Yeah, `NewTestBuildpackOutputAssertionManager` is so Java-ey. I think the use of packages may help.

Naming concerns aside, it's still debatable. Without working in it first hand with these set of change I can't say that I have too strong of an opinion. I'm just wondering whether it's reasonable to expect the creation of a new method solely to call another singleline when you could have just called `assert.Contains(...)`. Specially when that method is only used once.

Either way, looking at the latest changes. I'm okay with merging this in and iterating on it if need be.

simonjjones: Perfect

PR #777, opened 2020-08-05
Update "suggest-builders" to use dockerhub paketo builders.

dumez-k:
[#173928543]

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -162,7 +162,7 @@ func testWithoutSpecificBuilderRequirement(
 
 	when("set-default-builder", func() {
 		it("sets the default-stack-id in ~/.pack/config.toml", func() {
-			builderName := "gcr.io/paketo-buildpacks/builder:base"
+			builderName := "paketobuildpacks/builder:base"
 			output := pack.RunSuccessfully("set-default-builder", builderName)
 
 			assertions.NewOutputAssertionManager(t, output).ReportsSettingDefaultBuilder(builderName)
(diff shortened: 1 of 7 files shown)


dfreilich:
@dumez-k Thanks for the PR! There are a few tests throughout Pack (both unit and acceptance) that test for the specific output of inspect-builder; please update those as well, to ensure the tests pass. 

Additionally, if you are updating the name of the builder, we had recently updated the tektoncd tasks for [buildpacks](https://github.com/tektoncd/catalog/tree/master/task/buildpacks/0.1) and [buildpacks-phases](https://github.com/tektoncd/catalog/tree/master/task/buildpacks-phases/0.1) to point to the `paketo-buildpacks` builder. We'll have to update that to point to the dockerhub builders.

dumez-k:
Hey @dfreilich thanks for the quick feedback! 

Sorry definitely should have been more thorough in looking through those tests, will update accordingly!

dumez-k:
@dfreilich updated every instance I could find of `gcr.io/paketo-buildpacks` to use the dockerhub builders. 

However in the tektoncd tasks, the only references I could find to the gcr builders were in the README.md. I'm fairly unfamiliar with tekton, is this the only place you updated the builder?

jromero:
Hi @dumez-k, seems like the commits weren't signed. Unfortunately, we would need all commits to be [signed](https://github.com/buildpacks/.github/blob/main/CONTRIBUTING.md#sign-off-process) as part of our 
 contribution guidelines before we can merge this in. This takes place instead of the typical CLA signing.

You can learn more about this process here: https://github.com/buildpacks/.github/blob/main/CONTRIBUTING.md#sign-off-process

PR #782, opened 2020-08-09
Git type for register-buildpack

supra08:
## Summary
This PR adds the functionality to add a git type option in the `register-buildpack` command. Previously there was a type-Github that would open an issue in the browser with the to-be-registered buildpack. This git-type addition would enable us to create a buildpack file in the local cache and create a commit. 
The registry URL and the user should be provided in the `config.toml` :
```
[[registries]]
name="supra08"
type="git"
url="https://github.com/supra08/buildpack-registry/"
```
If nothing is provided then it fall backs to github type with the `https://github.com/buildpacks/registry` as the registry URL.
As already been implemented, the local cache registries will be maintained in `<packHome>/registry-<hash>`

Currently, in the commits, the github username is used as the author, and the email field is omitted. The commit message is as per the RFC. Also automated pushing to the registry is not implemented because this may lead to complications dealing with from merges and rebases. The author needs to manually go and push from the local cache currently.

Reference:
https://github.com/buildpacks/rfcs/blob/main/text/0044-pack-publish-buildpack.md

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related
buildpacks/rfcs#75

Resolves #___


dfreilich:
Hey @supra08 ! Glad to see the PR. Would you mind fixing the DCO sign off? Instructions for that are [here](https://github.com/buildpacks/pack/pull/782/checks?check_run_id=963791044).

supra08:
Hi @dfreilich sorry to miss that earlier. Done the DCO. Needs to fix the coverage though.

internal/registry/git.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"text/template"
+
+	"github.com/pkg/errors"
+)
+
+func CreateGitCommit(b Buildpack, username string, registryCache Cache) error {
+	_, err := os.Stat(registryCache.Root)
+	if err != nil {
+		if os.IsNotExist(err) {
```
dfreilich, line 14: You can combine these lines, to make it a bit simpler:
```suggestion
	if err != nil && if os.IsNotExist(err) {
```
You'd have to take out a closing bracket in addition.

internal/registry/git.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"bytes"
+	"os"
+	"text/template"
+
+	"github.com/pkg/errors"
+)
+
+func CreateGitCommit(b Buildpack, username string, registryCache Cache) error {
```
dfreilich, line 11: nit: Add GoDocs

internal/registry/git.go
```diff
@@ -0,0 +1,37 @@
...
+			err = registryCache.CreateCache()
+			if err != nil {
+				return errors.Wrap(err, "creating registry cache")
+			}
+		}
+	}
+
+	commitTemplate, err := template.New("buildpack").Parse(GitCommitTemplate)
+	if err != nil {
+		return err
```
dfreilich, line 24: nit: I prefer to wrap my errors whenever possible, to provide more helpful context on the error. Something like
```suggestion
		return errors.Wrap(err, "creating template")
```
(here, as well as in the next few errors) would be great

internal/registry/git_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/sclevine/spec/report"
+
+	ilogging "github.com/buildpacks/pack/internal/logging"
+	"github.com/buildpacks/pack/internal/registry"
+	"github.com/buildpacks/pack/logging"
+	h "github.com/buildpacks/pack/testhelpers"
+)
+
+func TestGit(t *testing.T) {
+	color.Disable(true)
```
dfreilich, line 21: nit: If you're not testing any logging, disabling and reenabling color isn't needed here

internal/registry/git_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	it.After(func() {
+		err := os.RemoveAll(tmpDir)
+		h.AssertNil(t, err)
+	})
+
+	when("#createGitCommit", func() {
+		it("should work with a proper buildpack and registry cache", func() {
+			err := registry.CreateGitCommit(bp, username, registryCache)
+			h.AssertNil(t, err)
```
dfreilich, line 64: Can you verify that the commit was created?

internal/registry/git_test.go
```diff
@@ -0,0 +1,78 @@
...
+		registryCache, err = registry.NewRegistryCache(logger, tmpDir, registryFixture)
+		h.AssertNil(t, err)
+	})
+
+	it.After(func() {
+		err := os.RemoveAll(tmpDir)
+		h.AssertNil(t, err)
+	})
+
+	when("#createGitCommit", func() {
```
dfreilich, line 61: Could you add a test where the RegistryCache wasn't created, to verify that it will call `CreateCache`? Or is that tested here already?

internal/registry/registry_cache.go
```diff
@@ -204,6 +208,125 @@ func (r *Cache) validateCache() error {
...
+	if err != nil {
+		return errors.Wrapf(err, "reading (%s)", r.Root)
+	}
+
+	index, err := r.writeEntry(b)
+	if err != nil {
+		return errors.Wrapf(err, "writing (%s)", index)
+	}
+
+	if index == "" {
```
dfreilich, line 233: Would you mind explaining this case? When would writeEntry return ""?

internal/registry/registry_cache.go
```diff
@@ -204,6 +208,125 @@ func (r *Cache) validateCache() error {
...
+
+			availableBuildpacks := entry.Buildpacks
+
+			if len(availableBuildpacks) != 0 {
+				if availableBuildpacks[len(availableBuildpacks)-1].Version == b.Version {
+					return "", errors.Wrapf(err, "same version exists, upgrade the version to add")
+				}
+			}
+		}
+	}
```
dfreilich, line 297: A lot of this seems to be duplicated from the `readEntry()` func. If we could pull out some or all of this into funcs that are used in both `readEntry` and `writeEntry`, or if we could call `readEntry()` within `writeEntry` to simplify a lot of this, that would be great

internal/registry/registry_cache.go
```diff
@@ -204,6 +208,125 @@ func (r *Cache) validateCache() error {
...
+
+	f, err := os.OpenFile(index, os.O_APPEND|os.O_CREATE|os.O_RDWR, 0644)
+	if err != nil {
+		return "", errors.Wrapf(err, "creating buildpack file: %s/%s", ns, name)
+	}
+
+	var newline string
+
```
dfreilich, line 305: nit: delete line

internal/registry/registry_cache_test.go
```diff
@@ -220,4 +220,77 @@ func testRegistryCache(t *testing.T, when spec.G, it spec.S) {
...
+			registryCache, err = NewRegistryCache(logger, tmpDir, registryFixture)
+			h.AssertNil(t, err)
+
+			err = registryCache.CreateCache()
+			h.AssertNil(t, err)
+		})
+
+		when("correct buildpack and commit message is passed", func() {
+			it("creates a file and a commit", func() {
+				h.AssertNil(t, registryCache.CreateCommit(bp, username, msg))
```
dfreilich, line 249: Can you verify that the file/commit was created?

internal/registry/registry_cache.go
supra08: Yeah actually we can separate out some parts of readEntry() and reuse it here. Would definitely a lot cleaner.

jromero:
Sorry @supra08 for the lack of an update. The release of pack 0.13.1 and KubeCon have taken a toll on our review cycle. Looking at this now.

internal/registry/git_test.go
jromero: Side tantrum: I have issues with the architecture decision. I'm not seeing the value of lazy loading the cache in this case. I think a better pattern may be to differentiate between a RegistryCache "config" and an "initialized" RegistryCache. Any time a `RegistryCache` is provided it should already be initialized. A `RegistryCacheConfig` can be used in cases where we just want to provide configuration without necessary interaction.

All this to say, I vote we omit this check if the hope is to change it soon.

jromero: Issue (#818) created.

PR #785, opened 2020-08-10
Enable multi-api support from lifecycle

jromero:
## Summary

This is the initial PR to enable multi-api support. Full support of Platform API 0.4 to follow.

This PR encompasses the following:

- Read new format
- Set new metadata to builder (incl old fields - [backwards compatibility][backward-compat])
- Read new metadata from builder (falling back to old fields - [backwards compatibility][backward-compat])
- Remove support for platform 0.2 (**breaking-change**)
- Update inspect builder output to include `deprecated` and `supported`

## Documentation

- Should this change be documented?
    - [x] Yes, see https://github.com/buildpacks/docs/issues/183
    - [ ] No

## Related

Resolves #780 

[backward-compat]: https://github.com/buildpacks/rfcs/blob/main/text/0049-multi-api-lifecycle-descriptor.md#backwards-compatibility

PR #786, opened 2020-08-10
Update golangci-lint

jromero:
## Summary

Update golangci-lint


jromero:
> This should make it much much much easier to review #785. 🥳

Yup, that's why I'm doing this. Probably going to unvendor as well to prevent this from being a massive change when updating deps. Lifecycle already made this move. It's our turn. :D

PR #796, opened 2020-08-11
Use CNB_PLATFORM_API to adjust what default process to show

### LINKED ISSUE #789: Omit CNB_PROCESS_TYPE when listing processes
### Description

As of platform API 0.4, the `CNB_PROCESS_TYPE` on the app image no longer determines what the default process should be. We should stop detecting it an listing the process when using `pack inspect-image`. See [RFC 45](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md#selecting-a-process-type-at-runtime)

### Proposed solution

```gherkin
Given an app image was created using a builder using platform API 0.4
And there is a process type "web" with command "web-command"
And there is a process type "other-process" with command "other-command"
And an environment variable `CNB_PROCESS_TYPE` with value "other-process" is set on the the image
When I run `pack inspect-image`
Then I should see the default process be "web-command"
And I should see "other-process" listed
```

```gherkin
Given an app image was created using a builder using platform API 0.3
And there is a process type "web" with command "web-command"
And there is a process type "other-process" with command "other-command"
And an environment variable `CNB_PROCESS_TYPE` with value "other-process" is set on the the image
When I run `pack inspect-image`
Then I should see the default process be "other-process"
And I should see "web" listed
```

dfreilich:

## Summary

This is an attempt to change the default process type Pack displays for images created with a lifecycle implementing Platform API 0.4. 

## Documentation

    - [x] No

## Related

Resolves #789

### DIFF
--- a/inspect_image.go
+++ b/inspect_image.go
@@ -34,7 +34,12 @@ type layersMetadata struct {
 	Stack    lifecycle.StackMetadata    `json:"stack" toml:"stack"`
 }
 
-const PlatformAPIEnv = "CNB_PLATFORM_API"
+const (
+	PlatformAPIEnv      = "CNB_PLATFORM_API"
+	CNBProcessEnv       = "CNB_PROCESS_TYPE"
+	defaultProcess      = "web"
+	fallbackPlatformAPI = "0.3"
+)
 
 func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
 	img, err := c.imageFetcher.Fetch(context.Background(), name, daemon, config.PullNever)
@@ -73,7 +78,7 @@ func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
 	}
 
 	if platformAPI == "" {
-		platformAPI = "0.3"
+		platformAPI = fallbackPlatformAPI
 	}
 
 	platformAPIVersion, err := semver.NewVersion(platformAPI)
@@ -83,12 +88,12 @@ func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
 
 	var defaultProcessType string
 	if platformAPIVersion.LessThan(semver.MustParse("0.4")) {
-		defaultProcessType, err = img.Env("CNB_PROCESS_TYPE")
+		defaultProcessType, err = img.Env(CNBProcessEnv)
 		if err != nil || defaultProcessType == "" {
-			defaultProcessType = "web"
+			defaultProcessType = defaultProcess
 		}
 	} else {
-		defaultProcessType = "web"
+		defaultProcessType = defaultProcess
 	}
 
 	var processDetails ProcessDetails
--- a/inspect_image.go
+++ b/inspect_image.go
@@ -36,12 +36,15 @@ type layersMetadata struct {
 }
 
 const (
-	PlatformAPIEnv      = "CNB_PLATFORM_API"
-	CNBProcessEnv       = "CNB_PROCESS_TYPE"
-	defaultProcess      = "web"
-	fallbackPlatformAPI = "0.3"
-	LauncherEntrypoint  = "/cnb/lifecycle/launcher"
-	BpEntrypointPrefix  = "/cnb/process/"
+	platformAPIEnv            = "CNB_PLATFORM_API"
+	cnbProcessEnv             = "CNB_PROCESS_TYPE"
+	launcherEntrypoint        = "/cnb/lifecycle/launcher"
+	windowsLauncherEntrypoint = `c:\cnb\lifecycle\launcher.exe`
+	entrypointPrefix          = "/cnb/process/"
+	windowsEntrypointPrefix   = `c:\cnb\process\`
+	defaultProcess            = "web"
+	fallbackPlatformAPI       = "0.3"
+	windowsPrefix             = "c:"
 )
 
 func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
@@ -75,7 +78,7 @@ func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
 		return nil, err
 	}
 
-	platformAPI, err := img.Env(PlatformAPIEnv)
+	platformAPI, err := img.Env(platformAPIEnv)
 	if err != nil {
 		return nil, errors.Wrap(err, "reading platform api")
 	}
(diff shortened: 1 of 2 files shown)


dfreilich:
@jromero I'm not sure whether or not this solves the problem. I based this off this line in the [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md#platforms):
> We can use CNB_PLATFORM_API (API "modes") to toggle between the current and new launcher process resolution flow. CNB_PLATFORM_API will be set in the exported image to the platform API provided at build time, but can be explicitly overridden at runtime.

The lifecycle release (https://github.com/buildpacks/lifecycle/releases/tag/v0.9.0) and code seems to show that they set it.

jromero:
@dfreilich This makes a lot of sense. I think this is the right solution.

inspect_image_test.go
```diff
@@ -199,42 +199,181 @@ func testInspectImage(t *testing.T, when spec.G, it spec.S) {
...
+											Direct:  true,
+										},
+									},
+								},
+							)
+						})
 					})
 
-					it("returns a nil default process", func() {
+					it("returns a nil default process when there is no web process", func() {
```
natalieparellano, line 361: This doesn't seem correct to me :/

inspect_image_test.go
```diff
@@ -199,42 +199,181 @@ func testInspectImage(t *testing.T, when spec.G, it spec.S) {
...
+										Direct:  true,
+									},
+								},
+							},
+						)
+					})
+				})
+
+				when("Platform API >= 0.4", func() {
+					when("CNB_PROCESS_TYPE is set", func() {
```
natalieparellano, line 299: I don't think we expect this variable to be set in the image if platform api >= 0.4, but I suppose there is value in checking the edge cases.

inspect_image_test.go
natalieparellano: Looking here: https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md 

Specifically here: 

> The exporter will prepend /cnb/process to the image PATH. The exporter will prepend /cnb/lifecycle to the image PATH. The exporter will set the app image entrypoint to the symlink matching the default process.

But, perhaps I'm misunderstanding something. I'm not very familiar with how it worked before.

dfreilich: I'm also not. I just lifted that test from the pre Platform API 0.4, but I don't know if there was a change re: to the above.

jromero: Could you elaborate on what doesn't seem right?

I'm looking at [this](https://github.com/buildpacks/spec/blob/platform/v0.4/platform.md#outputs-4):

```
- If <process-type> matches a buildpack-provided process:
    - MUST have ENTRYPOINT=/cnb/process/<process-type>
- If <process-type> does not match a buildpack-provided process:
    - If there is exactly one buildpack-provided process:
        - MUST have ENTRYPOINT=/cnb/process/<type> where <type> matches the type of the process
           Else:
         -MUST have ENTRYPOINT set to /cnb/lifecycle/launcher
```

which leads me to believe that there is a potential of having no default process type.

inspect_image.go
```diff
@@ -65,9 +72,28 @@ func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
...
+	}
+
+	var defaultProcessType string
+	if platformAPIVersion.LessThan(semver.MustParse("0.4")) {
+		defaultProcessType, err = img.Env(CNBProcessEnv)
+		if err != nil || defaultProcessType == "" {
+			defaultProcessType = defaultProcess
+		}
+	} else {
+		defaultProcessType = defaultProcess
```
dfreilich, line 96: to be fair, I'm not sure on this either. I just don't see in the spec how to know what process type was set, outside of the `web` default.

inspect_image_test.go
natalieparellano: I agree. I guess my point is that we shouldn't be looking at `CNB_PROCESS_TYPE` in the app image to determine the default process type. I don't think the exporter sets that for platform API >= 0.4: https://github.com/buildpacks/lifecycle/blob/f65216e2ad4692027fb64649fb1187a1b502e8be/exporter_test.go#L1021

We should be looking at the entrypoint. I don't see that in this PR...

inspect_image_test.go
dfreilich: I only check it once, now, just to verify that we ignore it and follow entrypoint.

inspect_image_test.go
dfreilich: I now use the entrypoint! Let me know if this looks more correct

inspect_image.go
```diff
@@ -39,6 +40,8 @@ const (
 	CNBProcessEnv       = "CNB_PROCESS_TYPE"
 	defaultProcess      = "web"
 	fallbackPlatformAPI = "0.3"
+	LauncherEntrypoint  = "/cnb/lifecycle/launcher"
+	BpEntrypointPrefix  = "/cnb/process/"
```
natalieparellano, line 44: We'll need different variables for Windows images.

inspect_image.go
```diff
@@ -93,7 +96,16 @@ func (c *Client) InspectImage(name string, daemon bool) (*ImageInfo, error) {
...
 	} else {
-		defaultProcessType = defaultProcess
+		inspect, _, err := c.docker.ImageInspectWithRaw(context.TODO(), name)
+		if err != nil {
+			return nil, errors.Wrap(err, "reading image")
+		}
+
+		entrypoint := inspect.Config.Entrypoint
+		if len(entrypoint) > 0 && entrypoint[0] != LauncherEntrypoint {
+			process := strings.Replace(entrypoint[0], BpEntrypointPrefix, "", 1)
```
natalieparellano, line 106: On Windows, we might also need to strip `.exe`

inspect_image.go
dfreilich: Meaning, in Windows images, it would be `\cnb\lifecycle\launcher` and `\cnb\process`?

inspect_image.go
dfreilich: 👍

dfreilich:
Added support for inspecting Windows images, even on other machines

inspect_image_test.go
```diff
@@ -248,49 +294,258 @@ func testInspectImage(t *testing.T, when spec.G, it spec.S) {
...
 								},
-							},
-						},
-					)
+							)
+						})
+					})
+
+					when("Inspecting Windows images", func() {
+						when(`ENTRYPOINT set to \cnb\lifecycle\launcher.exe`, func() {
```
jromero, line 460: "\cnb\lifecycle\launcher.exe" seems to be misaligned with the test. The test uses `WindowsLauncherEntrypoint` which has the `c:` drive. Which one is the image actually expected to have?

We should probably not share production values for testing to prevent accidentally changing the tests.

Issue #800, opened 2020-08-12
WCOW: build produces an unrunnable app image


jromero:
### Summary

When building a Windows container using the latest version of lifecycle 0.9.0 (which uses platform API 0.4).

---

##### Steps

1. On a WCOW setup run the task `make build-windows-apps` from https://github.com/buildpacks/samples/pull/85/files
2. `docker run -it --rm sample-batch-script-app:nanoserver-1809`

##### Current behavior

```
ERROR: failed to launch: cmd execute: exec: "cmd": executable file not found in %PATH%
```

##### Expected behavior

Application runs as expected.

---

##### pack info

```
Pack:
  Version:  0.0.0+git-5e9a21f
  OS/Arch:  windows/amd64

Default Lifecycle Version:  0.9.0

Config:
  experimental=true
```

##### docker info

```
Client:
 Debug Mode: false

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 38
 Server Version: 19.03.12
 Storage Driver: windowsfilter
  Windows:
 Logging Driver: json-file
 Plugins:
  Volume: local
  Network: ics internal l2bridge l2tunnel nat null overlay private transparent
  Log: awslogs etwlogs fluentd gcplogs gelf json-file local logentries splunk syslog
 Swarm: inactive
 Default Isolation: hyperv
 Kernel Version: 10.0 18363 (18362.1.amd64fre.19h1_release.190318-1202)
 Operating System: Windows 10 Pro Version 1909 (OS Build 18363.1016)
 OSType: windows
 Architecture: x86_64
 CPUs: 2
 Total Memory: 7.999GiB
 Name: DESKTOP-D0G2N78
 ID: BKPH:BDOG:5IE3:DKQA:5BYF:D56J:QUHQ:AIUM:NRJR:YYL6:ZWKT:IYNW
 Docker Root Dir: C:\ProgramData\Docker
 Debug Mode: true
  File Descriptors: -1
  Goroutines: 93
  System Time: 2020-08-12T15:03:46.9912819-07:00
  EventsListeners: 1
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Product License: Community Engine
```

jromero:
Related: https://github.com/buildpacks/rfcs/pull/114

TLDR; won't fix

micahyoung:
[...]

PR #804, opened 2020-08-13
Fix unrunnable Windows app images due to incomplete PATH env var

micahyoung:
- Run images need to explicitly set `Config.Env.PATH=C:\\windows...` to workaround differences in conventions for well-known Windows (mcr.microsoft.com/windows/nanoserver:1809) and Linux (alpine/bionic) base images
- Windows base image configs set `Config.Env=null` and use a default setting when `null`
- Linux base image configs set `Config.Env=["PATH=/usr/local/bin...",]`

- Works-around #800 though a proper fix in lifecycle will be needed to allow unmodified stacks

Signed-off-by: Micah Young <ymicah@vmware.com>

## Summary
Adds an explicit `PATH` env var image config entry for run image mock stacks. This allows Windows stack images to work like Linux stack images which always inherit a `PATH` env var from their base images. On Windows, a `null` path gets converted to a default setting of `c:\windows\system32;c:\windows` so the explicit value here is the same. This only affected platform v4 which now adds new entries to `PATH` for Windows, which v3 did not (resulting in `null`).

## Output
N/A

#### Before
Acceptance tests failed
#### After
Acceptance tests pass

## Documentation
There is a reference implementation pending-merge in samples https://github.com/buildpacks/samples/pull/85/commits/dbfc2e96a1cadc20f87bed83db0207c45f754cef

Besides that, we should probably document this unless we want to support `PATH=null` image configs in lifecycle.

    - [x] Yes, see #800 
    - [] No

## Related
Related to #800 but we should discuss how to proceed to accommodate stacks that do not include this workaround

### DIFF
--- a/acceptance/testdata/mock_stack/windows/run/Dockerfile
+++ b/acceptance/testdata/mock_stack/windows/run/Dockerfile
@@ -21,3 +21,6 @@ LABEL io.buildpacks.stack.id=pack.test.stack
 LABEL io.buildpacks.stack.mixins="[\"mixinA\", \"netcat\", \"mixin3\"]"
 
 USER pack
+
+# launcher requires a non-empty PATH to workaround https://github.com/buildpacks/pack/issues/800
+ENV PATH c:\\Windows\\system32;C:\\Windows


dfreilich:
Rebase on master, there's a fix to a failing acceptance test merged in that should make some of these pass.

PR #809, opened 2020-08-19
Add missing volume tests, refactor to test to test both wcow and lcow…

### LINKED ISSUE #793: Expand volume tests to occur on Windows
Hmm... in that case, I would expect the `windows` subset of tests to have the same number of cases as the `posix` ones, no? It looks like for Windows we just check that the drive is transformed, and on posix we test all sort of things.

_Originally posted by @natalieparellano in https://github.com/buildpacks/pack/pull/739#discussion_r468687677_

dwillist:
… volume mounting

## Summary
Adding in some missing tests.

## Documentation
N/A

    - [ ] Yes, see #___
    - [x] No

## Related

~~Wait for buildpack/imgutil#57 to be merged~~
Resolves #793


project/project_test.go
```diff
@@ -123,8 +123,8 @@ name = "gallant"
 
 			expected := 0
 			if len(projectDescriptor.Build.Env) != 0 {
-				t.Fatalf("Expected\n-----\n%#v\n-----\nbut got\n-----\n%#v\n",
```
dwillist, line 126: go `1.15` doesn't like the old syntax here

build_test.go
```diff
@@ -51,19 +53,25 @@ func TestBuild(t *testing.T) {
...
-		fakeLifecycleImage    *fakes.Image
+		subject                      *Client
+		fakeImageFetcher             *ifakes.FakeImageFetcher
+		fakeLifecycle                *ifakes.FakeLifecycle
+		defaultBuilderStackID        string
+		defaultWindowsBuilderStackID string
+		defaultBuilderImage          *fakes.Image
+		defaultWindowsBuilderImage   *fakes.Image
+		defaultBuilderName           string
+		defaultWindowsBuilderName    string
```
simonjjones, line 64: I don't think these default names ever change do they? If so they should be constants. This will help shrink the large `it.Before` method too.

build_test.go
```diff
@@ -1978,24 +2008,148 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+								Volumes: []string{volume},
+							},
+							TrustBuilder: true,
+						})
+						expected := []string{
+							fmt.Sprintf("%s:/x:ro", strings.ToLower(dir)),
+						}
+						h.AssertNil(t, err)
+						t.Log(fakeLifecycle.Opts.Volumes)
+						t.Log(expected)
```
simonjjones, line 2028: These logs should be removed.

build_test.go
```diff
@@ -1978,24 +2008,148 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+								Volumes: []string{volume},
+							},
+							TrustBuilder: true,
+						})
+						expected := []string{
+							fmt.Sprintf("%s:c:\\x:ro", strings.ToLower(dir)),
+						}
+						h.AssertNil(t, err)
+						t.Log(fakeLifecycle.Opts.Volumes)
+						t.Log(expected)
```
simonjjones, line 2099: As should these

internal/fakes/fake_images.go
```diff
@@ -15,8 +16,10 @@ import (
 	h "github.com/buildpacks/pack/testhelpers"
 )
 
-func NewFakeBuilderImage(t *testing.T, tmpDir, name string, stackID, uid, gid string, metadata builder.Metadata, bpLayers dist.BuildpackLayers, order dist.Order) *fakes.Image {
-	fakeBuilderImage := fakes.NewImage(name, "", nil)
+type FakeImageCreator func(string, string, imgutil.Identifier) *fakes.Image
```
simonjjones, line 19: We should name the input variables in the function type, to help when creating new ones.

build_test.go
simonjjones: I notice you've assigned fixed variables in your latest PR, but those strings should still probably be `const`'s

jromero (approved):

PR #816, opened 2020-08-25
WARN when a deprecated buildpack API version is being used

### LINKED ISSUE #788: WARN when a deprecated buildpack API version is being used
### Description

Since the new [lifecycle descriptor](https://github.com/buildpacks/pack/issues/780) has the potential of deprecating buildpack API (and platform API) versions we should warn when a buildpack is using a deprecated buildpack API so that it's less surprising in future versions.

### Proposed solution

```gherkin
Given there is buildpack 'id@v1' with buildpack API 0.2
And lifecycle has a descriptor with:
  [apis]
  [apis.buildpack]
  deprecated = ["0.2"]
When I run 'pack create-builder ...' including buildpack 'id@v1'
Then I should see a warning "buildpack 'id@v1' is using deprecated Buildpacks API version 0.2"
```

danielcooperxyz:
## Summary
WARN when a deprecated buildpack API version is being used

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #788


dfreilich (changes requested):
Thanks for the PR, and sorry this hasn't gotten seen until now! Looks simple, and definitely worth pulling in. If you'd be able to add a test verifying this, that would be fantastic.

PR #820, opened 2020-08-25
update inspect-builder tests

dwillist:
- factor out large multiline strings
 - use an `AssertionManager` object for assertions

## Summary
Refactor `inspect-builder` tests:
- use new `AssertionManager`
- factor multi line strings out of test cases and into constants


inspect_builder_test.go
```diff
@@ -76,17 +77,17 @@ func testInspectBuilder(t *testing.T, when spec.G, it spec.S) {
 
 				when("only deprecated lifecycle apis are present", func() {
 					it.Before(func() {
-						h.AssertNil(t, builderImage.SetLabel(
+						assert.Nil(builderImage.SetLabel(
```
simonjjones, line 80: Should we maybe alias `Nil` with `Succeeds` for this use case? I'm not sure it reads especially well why we want this to be `nil`.

testhelpers/assertions.go
```diff
@@ -175,3 +176,23 @@ func (a AssertionManager) Error(actual error) {
...
+}
+
+func cutSuffixWhitespace(input string) string {
+	index := len(input) - 1
+	for ; 0 <= index; index-- {
+		if !unicode.IsSpace(rune(input[index])) {
+			break
+		}
+	}
+	return input[:index+1]
```
simonjjones, line 197: I feel like we can probably solve this trimming problem with a regex. We have [regexp.ReplaceAllString](https://golang.org/pkg/regexp/#Regexp.ReplaceAllString) available to us. I'd be happy to help out having trying to make this all a little more elegant.

inspect_builder_test.go
```diff
@@ -49,13 +50,13 @@ func testInspectBuilder(t *testing.T, when spec.G, it spec.S) {
...
-		h.AssertNil(t, builderImage.SetLabel(
+		assert.Nil(builderImage.SetLabel("io.buildpacks.stack.id", "test.stack.id"))
+		assert.Nil(builderImage.SetLabel(
 			"io.buildpacks.stack.mixins",
 			`["mixinOne", "build:mixinTwo", "mixinThree", "build:mixinFour"]`,
 		))
-		h.AssertNil(t, builderImage.SetEnv("CNB_USER_ID", "1234"))
-		h.AssertNil(t, builderImage.SetEnv("CNB_GROUP_ID", "4321"))
+		assert.Nil(builderImage.SetEnv("CNB_USER_ID", "1234"))
+		assert.Nil(builderImage.SetEnv("CNB_GROUP_ID", "4321"))
```
dfreilich, line 59: any reason not to use `assert.Succeeds`?

internal/commands/inspect_builder_test.go
```diff
@@ -30,6 +31,141 @@ func TestInspectBuilderCommand(t *testing.T) {
 	spec.Run(t, "Commands", testInspectBuilderCommand, spec.Parallel(), spec.Report(report.Terminal{}))
 }
 
+const inspectBuilderRemoteOutputSection = `
```
dfreilich, line 34: 👍 on pulling this out

internal/commands/inspect_builder_test.go
```diff
@@ -406,46 +446,46 @@ Detection Order:
 			})
 
 			it("missing creator info is skipped", func() {
-				h.AssertNil(t, command.Execute())
-				h.AssertNotContains(t, outBuf.String(), "Created By:")
+				assert.Succeeds(command.Execute())
+				assert.NotContain(outBuf.String(), "Created By:")
```
dfreilich, line 450: I know that you didn't add `NotContain`, but given that the pattern has been plural for asserts (`assert.Succeeds`, `assert.Contains`, `assert.ErrorContains`), I find it odd for this not to have an `s` at the end.

dfreilich (approved):
Good work using the cleaner pattern! The problem is, now we probably should update all of the tests to use it, if we want to keep that pattern going...

But good stuff, like always 🔥

inspect_builder_test.go
dwillist: Nope totally missed this.

testhelpers/assertions.go
```diff
@@ -175,3 +187,18 @@ func (a AssertionManager) Error(actual error) {
...
+
+	if actual == nil {
+		a.testObject.Fatalf("Expected %q an error but got nil", expected)
+	}
+
+	a.Contains(actual.Error(), expected)
+}
+
+func cutSuffixWhitespace(input string) string {
+	re := regexp.MustCompile(`  +$`)
```
simonjjones, line 202: I think it would be clearer to use the whitespace character `\s` here. I'm also not sure if this would match a situation where there was only one space character at the end of the line, as it appears there are two spaces in this regex. Perhaps I'm missing a case where that's useful?

jromero: @simonjjones brings up some good points about this being a little confusing. 

From a performance perspective, I would think that compiling a regex everytime this is called isn't optimal either.

Is there a reason something like [`strings.TrimRight(input, "\t \n")`](https://golang.org/pkg/strings/#TrimRight) wouldn't work here? From what I can tell, inputs to this method are lines (not multiline).


If we wanted to keep it exactly inline with the previous implementation the following should be equivalent:

```
strings.TrimRightFunc(input, func(c rune) bool { return unicode.IsSpace(c); })
```

Additional references: 

* https://www.admfactory.com/how-to-trim-whitespace-from-string-in-golang/

jromero (changes requested):

jromero (approved):

PR #822, opened 2020-08-27
Output image name as provided by user

### LINKED ISSUE #692: don't expand my docker image name in the logs
### Description
I say
`pack build app`

and then the logs say:

```
...
[analyzer] Previous image with name "index.docker.io/library/app1:latest" not found
...
[exporter] *** Images (1f00a0e5507f):
[exporter]       index.docker.io/library/app1:latest
...
```

It's unnerving to see index.docker.io on there, since this is my local, private code. Could it accidentally get pushed to docker.io? I don't even know.

### Proposed solution

I understand `app1:latest` but I wish pack could not show me the `index.docker.io/library/` prefix (wherever that comes from).

dwillist:
- New tag implementation is used to determine imageName passed to lifecycle phases.

## Summary
This changes the image name that we pass the the lifecycle when building. Previously when building `namespace/image` or `image` a default registry and namespace would be prepended to the image. This would result in the following fully qualified names, `index.docker.io/namespace/image` or `index.docker.io/library/image`. This change strips off these prefixes unless explicit specified by the user.

#### Before

```
...
[analyzer] Previous image with name "index.docker.io/library/app1:latest" not found
...
[exporter] *** Images (1f00a0e5507f):
[exporter]       index.docker.io/library/app1:latest
...
```

#### After
```
...
[analyzer] Previous image with name "app1:latest" not found
...
[exporter] *** Images (1f00a0e5507f):
[exporter]       app1:latest
...

```

## Documentation

    - [x] No

## Related

Resolves #692


dfreilich:
I'm not sure if this quality change is necessary - I think that we can also do it by just changing https://github.com/buildpacks/pack/blob/c9bb6afa3e89ab6b27cfe16a63fd4f2c09226372/internal/build/build.go#L150 from 
`opts.Image.Name()` to `opts.Image.String()`, which would be a significantly smaller change. (I had proven that locally, I just hadn't validated it with tests because it was going to take a bit of rearchitecting and it was right around the release time).

dwillist:
@dfreilich I think if we use the `String()` function we are no longer given `:latest` as a default tag name when none is provided. We need to either write a function to strip off the `index.docker.io/library` prefix (what is done here) or one to conditionally add the `:latest`.

dfreilich:
I tested running 
```
	repo := opts.Image
	l.logger.Infof("string: %s", repoName)
	l.logger.Infof("Context: %s", repo.Context())
	l.logger.Infof("Identifier: %s", repo.Identifier())
	l.logger.Infof("Name: %s", repo.Name())
	l.logger.Infof("Context.Name: %s", repo.Context().Name())
	l.logger.Infof("Context.RegistryStr: %s", repo.Context().RegistryStr())
	l.logger.Infof("Context.RepositoryStr: %s", repo.Context().RepositoryStr())
	l.logger.Infof("Context.Scheme: %s", repo.Context().Scheme())
	l.logger.Infof("Context.String: %s", repo.Context().String())
```

and results I got for `pack build test1...`
```
string: test1
Context: index.docker.io/library/test1
Identifier: latest
Name: index.docker.io/library/test1:latest
Context.Name: index.docker.io/library/test1
Context.RegistryStr: index.docker.io
Context.RepositoryStr: library/test1
Context.Scheme: https
Context.String: index.docker.io/library/test1
```

vs `pack build test1:not-latest
```
String: test1:not-latest
Context: index.docker.io/library/test1
Identifier: not-latest
Name: index.docker.io/library/test1:not-latest
Context.Name: index.docker.io/library/test1
Context.RegistryStr: index.docker.io
Context.RepositoryStr: library/test1
Context.Scheme: https
Context.String: index.docker.io/library/test1
```

I'm not denying that this could solve the problem, I just feel like it's over-architecting the solution, when we could just append `opts.Image.Identifier()` to `opts.Image.String()` if it doesn't already end with it.

jromero:
Sorry to weigh in on this a little late. Here are my thoughts.

1. @dfreilich  I'm typically opposed to using `.String()` as anything other than to be printed on logs given that there is typically little to know contract. For GGCR though, it does seem like there is an explicit documented contract (see ["String returns the original input string"](https://github.com/google/go-containerregistry/blob/v0.1.2/pkg/name/digest.go#L58)) and it would satisfy our requirements.
2. @dwillist, I think the original user was _ok_ with the `:latest` tag but I don't believe it should be a requirement.

dwillist:
@jromero @dfreilich. Thanks for the feedback. I guess there are some assumptions that lead me to this implementation which might help explain why this logging change is so large. Would like to understand if any of these are excessive/incorrect. 

1) Logs should surface values used by a program. 
   - As per dfrelich's suggestion we could simplify the implementation by changing the output the user gets to see. However, this leads to a disconnect between the value we use in our program and what show users. 

2) Single source of truth. If we want all of the uses of `opts.Image.Name()` to have the `docker.io/library` prefix omitted we should probably do this is a single place, hence the new `pack.Tag` struct.

Totally agree its a lot of code for a small change. But couldn't find a way to satisfy the above to points without over-riding the `opts.Image.Name()` function.

jromero:
> @jromero @dfreilich. Thanks for the feedback. I guess there are some assumptions that lead me to this implementation which might help explain why this logging change is so large. Would like to understand if any of these are excessive/incorrect.
> 
> 1. Logs should surface values used by a program.
>    
>    * As per dfrelich's suggestion we could simplify the implementation by changing the output the user gets to see. However, this leads to a disconnect between the value we use in our program and what show users.

I think this is where it starts to break down. Ultimately, there is a distinction between logs and user output. The way GGCRs reference implementation works I think is the _right_ approach in that from a programmatic perspective the object is filled with "default" presumptions that may or may not be known by the user but we might need. From the user's perspective, seeing these defaults created the confusion mentioned in the original issue (and have heard elsewhere). So going back to your original statement. If we separate logs (for troubleshooting which in our implementation would be "debug" logs) vs output (for user consumption), then we should be able to determine when to use which representation.

> 2. Single source of truth. If we want all of the uses of `opts.Image.Name()` to have the `docker.io/library` prefix omitted we should probably do this is a single place, hence the new `pack.Tag` struct.
> 

I feel like my previous answer may cover this as well. The "source of truth" in my mind is an object, not a single property of that object.

dfreilich:
@jromero I was reading this over again, and I'm a bit confused which way you are suggesting we go, as a result. Are you arguing that there is always an inherent disconnect between the internal state of an object and what we display to users (in logs), and therefore it doesn't bother you to have an object floating around with the tag being `index.docker.../image:tag` while we display the name `image:tag`? Or are you arguing that the above statement is problematic, something we should change, and therefore advocating to continue along with this change?

dfreilich:
At the same time, @dwillist , I just want to reiterate that this change is totally fine, and really great work 🥇, especially with its 👌  test coverage. I just want to caution against overall bloating if something isn't totally necessary, but totally not a hill I'll pick a stand on.

jromero:
@dfreilich thanks for the feedback. I had a side chat with @dwillist making it more explicit that my desire would be to rely on `String()`. 

We also discovered that some of the confusion was around there not being a spec'd contract around how image name defaults would behave. We've spoken to the implementation team and  created this issue in aim to address it: https://github.com/buildpacks/spec/issues/146

dwillist:
Re-wrote this with @jromero and @dfreilich's feedback. The code-cov patch is failing, despite there being < 10 lines of code changed. Let me know if you think this is worth addressing by adding some new patterns in the `build` package.d

dwillist:
Re-wrote this with @jromero and @dfreilich's feedback. The code-cov patch is failing, despite there being < 10 lines of code changed. Let me know if you think this is worth addressing by adding some new patterns in the `build` package.

PR #823, opened 2020-08-27
Enable usage of lifecycle image for windows

ameyer-pivotal:
Signed-off-by: Victoria Henry <vhenry@pivotal.io>
Signed-off-by: Andrew Meyer <meyeran@vmware.com>
Signed-off-by: Malini Valliath <mvalliath@vmware.com>

## Summary
Enables use of lifecycle images for Windows builds

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -616,8 +616,6 @@ func testAcceptance(
 					var untrustedBuilderName string
 
 					it.Before(func() {
-						h.SkipIf(t, dockerHostOS() == "windows", "untrusted builders are not yet supported for windows builds")
-
 						var err error
 						untrustedBuilderName, err = createBuilder(
 							t,
@@ -630,9 +628,6 @@ func testAcceptance(
 					})
 
 					it.After(func() {
-						if dockerHostOS() == "windows" {
-							return
-						}
 						h.DockerRmi(dockerCli, untrustedBuilderName)
 					})
 
@@ -2360,6 +2355,7 @@ func assertMockAppRunsWithOutput(t *testing.T, assert h.AssertionManager, repoNa
 	ctrID := runDockerImageExposePort(t, assert, containerName, repoName)
 	defer dockerCli.ContainerKill(context.TODO(), containerName, "SIGKILL")
 	defer dockerCli.ContainerRemove(context.TODO(), containerName, dockertypes.ContainerRemoveOptions{Force: true})
+
 	logs, err := dockerCli.ContainerLogs(context.TODO(), ctrID, dockertypes.ContainerLogsOptions{
 		ShowStdout: true,
 		ShowStderr: true,
--- a/build_test.go
+++ b/build_test.go
@@ -1447,18 +1447,6 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 				})
 
 				when("builder is untrusted", func() {
-					when("building Windows containers", func() {
-						it("errors and mentions that builder must be trusted", func() {
-							defaultBuilderImage.SetPlatform("windows", "", "")
-							h.AssertError(t, subject.Build(context.TODO(), BuildOptions{
-								Image:        "some/app",
-								Builder:      defaultBuilderName,
-								Publish:      true,
-								TrustBuilder: false,
-							}), "does not have an associated lifecycle image. Builder must be trusted.")
-						})
-					})
-
 					when("lifecycle image is available", func() {
 						it("uses the 5 phases with the lifecycle image", func() {
 							h.AssertNil(t, subject.Build(context.TODO(), BuildOptions{
--- a/internal/build/container_ops.go
+++ b/internal/build/container_ops.go
@@ -8,13 +8,14 @@ import (
 	"io/ioutil"
 	"os"
 	"runtime"
-	"strings"
 
 	"github.com/BurntSushi/toml"
 	"github.com/docker/docker/api/types"
 	dcontainer "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
 	"github.com/pkg/errors"
+
+	"github.com/buildpacks/pack/internal/paths"
 
 	"github.com/buildpacks/pack/internal/archive"
 	"github.com/buildpacks/pack/internal/builder"
@@ -24,25 +25,22 @@ import (
 type ContainerOperation func(ctrClient client.CommonAPIClient, ctx context.Context, containerID string, stdout, stderr io.Writer) error
 
 // CopyDir copies a local directory (src) to the destination on the container while filtering files and changing it's UID/GID.
-func CopyDir(src, dst string, uid, gid int, fileFilter func(string) bool) ContainerOperation {
+func CopyDir(src, dst string, uid, gid int, os string, fileFilter func(string) bool) ContainerOperation {
 	return func(ctrClient client.CommonAPIClient, ctx context.Context, containerID string, stdout, stderr io.Writer) error {
-		info, err := ctrClient.Info(ctx)
-		if err != nil {
-			return err
+		tarPath := dst
+		if os == "windows" {
+			tarPath = paths.WindowsToPosixPath(dst)
 		}
-		if info.OSType == "windows" {
-			reader, err := createReader(src, winPathToTarPath(dst), uid, gid, fileFilter)
-			if err != nil {
-				return errors.Wrapf(err, "create tar archive from '%s'", src)
-			}
-			defer reader.Close()
-			return copyWindows(ctx, ctrClient, containerID, reader, dst, stdout, stderr)
-		}
-		reader, err := createReader(src, dst, uid, gid, fileFilter)
+
+		reader, err := createReader(src, tarPath, uid, gid, fileFilter)
 		if err != nil {
 			return errors.Wrapf(err, "create tar archive from '%s'", src)
 		}
 		defer reader.Close()
+
+		if os == "windows" {
+			return copyDirWindows(ctx, ctrClient, containerID, reader, dst, stdout, stderr)
+		}
 		return copyDir(ctx, ctrClient, containerID, reader)
 	}
 }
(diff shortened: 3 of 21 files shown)


ameyer-pivotal:
@jromero Also note that after this is released, we should update the Windows build guide to remove `--trust-builder` from the example

acceptance/acceptance_test.go
```diff
@@ -615,8 +615,6 @@ func testAcceptance(
 					var untrustedBuilderName string
 
 					it.Before(func() {
-						h.SkipIf(t, dockerHostOS() == "windows", "untrusted builders are not yet supported for windows builds")
```
dfreilich, line 618: 🥳

build_test.go
```diff
@@ -1345,18 +1345,6 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 				})
 
 				when("builder is untrusted", func() {
-					when("building Windows containers", func() {
-						it("errors and mentions that builder must be trusted", func() {
```
dfreilich, line 1349: is there another test in the section that tries to build Windows containers? If not, can we add one?

internal/build/container_ops.go
```diff
@@ -71,24 +69,26 @@ func copyDir(ctx context.Context, ctrClient client.CommonAPIClient, containerID
...
 }
 
-// copyDirWindows provides an alternate, Windows container-specific implementation of copyDir.
+// copyWindows provides an alternate, Windows container-specific implementation of copyDir.
 // This implementation is needed because copying directly to a mounted volume is currently buggy
 // for Windows containers and does not work. Instead, we perform the copy from inside a container
 // using xcopy.
 // See: https://github.com/moby/moby/issues/40771
-func copyDirWindows(ctx context.Context, ctrClient client.CommonAPIClient, containerID string, appReader io.Reader, dst string, stdout, stderr io.Writer) error {
+func copyWindows(ctx context.Context, ctrClient client.CommonAPIClient, containerID string, reader io.Reader, dst string, stdout, stderr io.Writer) error {
```
dfreilich, line 77: What motivated the rename? Was it just length/clarity?

internal/build/container_ops.go
```diff
@@ -97,10 +97,9 @@ func copyDirWindows(ctx context.Context, ctrClient client.CommonAPIClient, conta
 		&dcontainer.Config{
 			Image: info.Image,
 			Cmd: []string{
-				"xcopy",
-				fmt.Sprintf(`c:\windows\%s`, baseName),
-				dst,
-				"/e", "/h", "/y", "/c", "/b",
+				"cmd",
+				"/c",
+				fmt.Sprintf(`echo %s|xcopy /e /h /y /c /b c:\windows\%s %s`, fileOrDir, dst[3:], dst),
```
dfreilich, line 102: Would you mind clarifying what the flags are for me?

micahyoung: Is there a way to make `xcopy /i` work here instead of piping in the answer?

> /i  -- If Source is a directory or contains wildcards and Destination does not exist, xcopy assumes Destination specifies a directory name and creates a new directory. Then, xcopy copies all specified files into the new directory. By default, xcopy prompts you to specify whether Destination is a file or a directory.

https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/xcopy#:~:text=If%20Source%20is%20a%20directory

ameyer-pivotal: We tried `/i` and it didn't work for us for some reason. Piping seemed to be the suggestion based on some research. Can try again though. Is there a reason why one is better, other than a simpler command?

build_test.go
ameyer-pivotal: These tests wouldn't exercise any new code paths due to mocking. Now that we've removed guard that returned an error for windows containers, there are no more windows-specific code paths.

internal/build/container_ops.go
ameyer-pivotal: Now this is used to copy both files and dirs

internal/build/container_ops.go
```diff
@@ -149,14 +148,29 @@ func WriteStackToml(dstPath string, stack builder.StackMetadata) ContainerOperat
...
 		tarBuilder.AddFile(dstPath, 0755, archive.NormalizedDateTime, buf.Bytes())
 		reader := tarBuilder.Reader(archive.DefaultTarWriterFactory())
 		defer reader.Close()
-
 		return ctrClient.CopyToContainer(ctx, containerID, "/", reader, types.CopyToContainerOptions{})
 	}
 }
 
+func winPathToTarPath(path string) string {
+	return strings.ReplaceAll(path, `\`, "/")[2:] // strip volume, convert slashes
```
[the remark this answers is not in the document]
micahyoung: Oh nvm. That's used in lifecycle b/c it's `runtime.GOOS == "windows"` so `filepath` works.

internal/container/run.go
```diff
@@ -18,6 +19,17 @@ func Run(ctx context.Context, docker client.CommonAPIClient, ctrID string, out,
...
+
+	info, err := docker.Info(ctx)
+	if err != nil {
+		return errors.Wrap(err, "getting docker info")
+	}
+
+	if info.OSType == "windows" {
+		// wait for logs to show
+		time.Sleep(time.Second)
+	}
```
jromero, line 31: Based on our slack conversation can we create an issue to keep track of this work and add a reference to the code so we can track this workaround.

```suggestion
	// FIXME: <issue number>
	if info.OSType == "windows" {
		// wait for logs to show
		time.Sleep(time.Second)
	}
```

jromero (changes requested):

internal/build/container_ops.go
micahyoung: Yeah, if I change to `/i` on your branch, it fails acceptance tests for me when I try it, though not the `container_opts_test` nor other units. I feel like that's an issue, we probably need some better test coverage, if only to serve as better documentation for our `xcopy` stuff. I'm worried that since this is such a hotspot for arbitrary input, we should probably bolster the coverage a bit.

Additionally, it might make it easier to iterate on this implementation in the future as well.

micahyoung: Oh, nvm, there's a `SkipIf` in `container_ops_test`. Let's take that out as part of this work.

micahyoung: I started a branch tweaking the tests as bit. It's stuck at what may be a bug in this implementation, but either way, adding some coverage to `container_ops_test` this PR shouldn't be too hard: 

https://github.com/micahyoung/buildpacks-pack/commit/a4182c87553c4eb671c5e2c3d9102254a320b1c1

internal/build/container_ops.go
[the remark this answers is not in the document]
ameyer-pivotal: @micahyoung This won't work because most `filepath` functions, including these are OS-dependent.

internal/build/container_ops.go
micahyoung: After trying a few different alternatives, it appears the current implementation is the best one. I tried plain `copy` for individual files but it did not preserve original ownership info, I also tried `tar -c | tar -x` but it doesn't handle symlinks properly.

micahyoung: I added comments for the flags to a follow-up commit.

internal/build/container_ops.go
```diff
@@ -97,10 +89,16 @@ func copyDirWindows(ctx context.Context, ctrClient client.CommonAPIClient, conta
...
+				"cmd",
+				"/c",
+
+				//xcopy args
+				// e - recursively create subdirectories
+				// h - copy hidden and system files
+				// b - copy symlinks, do not dereference
+				// x - copy attributes
+				// y - suppress prompting
+				fmt.Sprintf(`xcopy c:\windows\%s %s /e /h /b /x /y`, baseName, dst),
```
micahyoung, line 101: Added better documentation for xcopy command

internal/paths/paths.go
```diff
@@ -15,31 +15,31 @@ func IsURI(ref string) bool {
 	return schemeRegexp.MatchString(ref)
 }
 
-func IsDir(path string) (bool, error) {
-	fileInfo, err := os.Stat(path)
```
micahyoung, line 19: avoid clobbering golang `path` package

testhelpers/testhelpers.go
```diff
@@ -629,31 +629,22 @@ func SkipUnless(t *testing.T, expression bool, reason string) {
...
+	if err := dockerCli.ContainerStart(ctx, id, dockertypes.ContainerStartOptions{}); err != nil {
+		return errors.Wrap(err, "container start")
 	}
 
 	copyErr := make(chan error)
 	go func() {
-		_, err := stdcopy.StdCopy(stdout, stderr, logs)
+		_, err := stdcopy.StdCopy(stdout, stderr, logs.Reader)
 		copyErr <- err
 	}()
```
micahyoung, line 649: Use ContainerAttach + ContainerStart instead of ContainerStart + ContainerLogs. The current implementation has race conditions for Windows and is more heavy-weight for all OSes as the stdout/stderr have to flow through the logs subsystem. This implementation follows the docker/cli run [implementation](https://github.com/docker/cli/blob/f8696535e9e516a5a404661697fef6f1228ada64/cli/command/container/run.go#L157-L168) which effectively does the same (though with additional signal handling and tty options).

This fixes Windows, potentially avoids race conditions for both platforms by attaching before start, and I feel like having the Attach handle gives us more option for signalling that ContainerLogs handle in the future.

micahyoung (commented):
Everything is ready for review again. I added some notes inline.

internal/container/run.go
micahyoung: No time.Sleep is needed with the subsequent changes to `container.Run` below.

internal/build/container_ops.go
micahyoung: In my latest changes, this is now reverted back to `copyDirWindows` as requiring `dst` to be a directory greatly simplifies `xcopy` behavior.

jromero:
This overall looks great! Thank you for the added effort. There seems to be a small conflict. Once that is resolved we can happily bring this in.

micahyoung:
@jromero should be RTM/review now, I think.

jromero (approved):

PR #825, opened 2020-08-27
BuildpackManager prepares all types of buildpacks for use in tests

simonjjones:

## Summary

This change adds a BuildpackManager, and defines a number of _type_ of buildpacks which are used in tests.

Hopefully there are no changes to the logic of what is being tested, just rearrangement for clarity.

## Documentation
    - [ ] Yes, see #___
    - [x] No

## Related

Contributes to #606


acceptance/acceptance_test.go
```diff
@@ -917,10 +936,14 @@ func testAcceptance(
 								volumeRoot = `c:\`
 								slash = `\`
 							}
-							readBuildpackTgz = h.CreateTGZ(t, filepath.Join(bpDir, "read-volume-buildpack"), "./", 0755)
-							readWriteBuildpackTgz = h.CreateTGZ(t, filepath.Join(bpDir, "read-write-volume-buildpack"), "./", 0755)
 
 							var err error
+
```
dfreilich, line 941: nit: rm line?

acceptance/buildpacks/archive_buildpack.go
```diff
@@ -0,0 +1,99 @@
...
+	ReadVolume          = &archiveBuildpack{name: "read-volume-buildpack"}
+	ReadWriteVolume     = &archiveBuildpack{name: "read-write-volume-buildpack"}
+	ArchiveNotInBuilder = &archiveBuildpack{name: "not-in-builder-buildpack"}
+	Noop                = &archiveBuildpack{name: "noop-buildpack"}
+	Noop2               = &archiveBuildpack{name: "noop-buildpack-2"}
+	OtherStack          = &archiveBuildpack{name: "other-stack-buildpack"}
+	ReadEnv             = &archiveBuildpack{name: "read-env-buildpack"}
+	NestedLevelOne      = &archiveBuildpack{name: "nested-level-1-buildpack"}
+	NestedLevelTwo      = &archiveBuildpack{name: "nested-level-2-buildpack"}
+)
```
dfreilich, line 99: nit: What do you think about moving this underneath the consts? They are the variables that we interact most with in this package.

testhelpers/testhelpers.go
```diff
@@ -567,44 +567,87 @@ func PullImageWithAuth(dockerCli client.CommonAPIClient, ref, registryAuth strin
...
 
 func CopyFile(t *testing.T, src, dst string) {
-	fi, err := os.Stat(src)
+	t.Helper()
+
+	err := CopyFileE(src, dst)
 	AssertNil(t, err)
+}
+
+func CopyFileE(src, dst string) error {
```
dfreilich, line 576: Why are `CopyFileE` and `RecursiveCopyE` useful? Are there places where we actually want to return errors?

dfreilich (approved):
Overall, really really good. I have a bit of a quibble with `buildpacks.FullPathIn`, but I'm happy to get push-back on that.

testhelpers/testhelpers.go
simonjjones: In the new code added as part of this PR, I didn't want the `FolderBuildpack` to need a `*testing.T` as then I wouldn't be able to instantiate them at build time. An alternative is to copy this behavior in that context.

acceptance/buildpacks/archive_buildpack.go
simonjjones: I wondered about this, and generally agree, with the exception that this would then define variables based on types that are declared later in the same file. We could use a separate `buildpacks` file that defines all of the buildpacks that can be instantiated at build time, but that would be another level of indirection. Which do you prefer?

acceptance/acceptance_test.go
```diff
@@ -839,28 +854,32 @@ func testAcceptance(
...
 						})
 
 						when("the network mode is not provided", func() {
-							it("reports that build and detect are online", func() {
+							it("reports buildpack access to internet", func() {
 								output := pack.RunSuccessfully(
 									"build", repoName,
 									"-p", filepath.Join("testdata", "mock_app"),
-									"--buildpack", buildpackTgz,
+									"--buildpack", buildpacks.InternetCapable.FullPathIn(tmpDir),
```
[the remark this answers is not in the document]
simonjjones: What do you have in mind? Something like `buildpackManager.PathFor(buildpacks.InternetCapable)` looks OK, but in reality we want the `buildpackManager` to be able to operate in a consistent way for all type of buildpacks.

They don't all have a path, but maybe there's always **something** that makes sense?

acceptance/acceptance_test.go
simonjjones: Happy to, the thinking was that the `err` defined here is used four times in the file. That's not really different to reassigning an `err` first assigned with `:=` though. Seems like a little for one commit, so let's figure out what other changes we want to make.

simonjjones: I found a cleanup issue, so removed this with that commit

acceptance/acceptance_test.go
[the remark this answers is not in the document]
dfreilich: How would introducing `manager.PackFor` make it operate inconsistently for different buildpacks?

acceptance/buildpacks/archive_buildpack.go
dfreilich: I'd definitely rather keep it in this file. I feel like it's alright to have them defined prior to the type declared later, since it's immediately following it, but I'm fine if you decide to keep it where it is.

acceptance/acceptance_test.go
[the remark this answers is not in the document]
simonjjones: I think about the implementation differences between buildpack sources based on the local file system and those based on images.

The manager's implementation of `PathFor(TestBuildpack)` would end up being a wrapper around this existing implementation, potentially presenting an extra layer of abstraction.

acceptance/buildpacks/manager.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return m
+}
+
+type TestBuildpack interface {
+	Prepare(source, destination string) error
+}
+
+func (b BuildpackManager) PrepareBuildpacks(destination string, buildpacks ...TestBuildpack) {
```
jromero, line 46: I'm trying to understand what "prepare" means in this context. Is the term we should be looking to use here "Install" as in it installs the provided buildpacks at the provided destination?

simonjjones: Prepare is the best term I could think of for whatever an actual end user might do to get buildpacks ready for running their pack command. This could be creating a new buildpack from scratch, tarring a folder into an archive, or using pack to create a package. If I'm mapping the user action here, I don't think `install` is very descriptive for these scenarios, but happy to think about alternative terms.

jromero: Maybe it's simply the fact that I don't associate prepare with a "finalizing" state. Prepare to me indicates that there would be additional steps after the fact to get it into its final desired state. That doesn't seem to be the case here. In either way, I wouldn't say it's a blocker.

jromero (approved):

PR #830, opened 2020-09-02
Fix for GHA to work with default Windows runners

micahyoung:
## Summary
 - Change `build.yml` to ignore self-hosted worker to use GHA default runner
 - Change `build.yml` to setup docker hostname helpers `host.docker.internal` so Docker for Windows Server matches those made by Docker Desktop [docs](https://docs.docker.com/docker-for-windows/networking/#:~:text=host.docker.internal)
 - Fix windows mock stack Dockerfile that failed to build on Docker for Windows Server

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related
* Original context for creating self-hosted runner [slack](https://buildpacks.slack.com/archives/CSRHS92NP/p1595881563207600)
* Self-hosted runner [issue](https://github.com/buildpacks/ci/issues/20)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,11 +58,20 @@ jobs:
           git config --global core.autocrlf false
           git config --global core.eol lf
           git config --global core.symlinks true
-      - name: Set docker host dns helpers
+      - name: Set docker to allow local registry using helpers dns names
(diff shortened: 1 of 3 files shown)


jromero:
why

Did something not work out?

PR #831, opened 2020-09-03
Print only image name (and digest) when running `pack build` in quiet mode

### LINKED ISSUE #285: Only output full image tag or image digest from build
I find myself running pack builds in bash scripts and having to grep for the resultant image digest quite a lot. It would be nice to be able to pass a flag to pack build so that it only prints the resulting full image tag or the image digest. For example:
```
pack build something -q (--quiet)
```
outputs only (on success)
```
sha256:abcdefg123456
```
or
```
example-registry.io/my-image@sha256:abcdefg123456
```

Using `-q` is just an example and I would be happy with anything that made sense.

dfreilich:
* Fetcher used to log directly to the Writer, and bypass quiet mode. This fixes that hole.
* This implements a request (#285) for having build -q to print the app name @ sha. It only adds it into the quiet mode; in a normal build, there's no need to print it, since the information is printed by the lifecycle. It is only in cases where we suppress that output (quiet mode) where we need to print it ourselves. 

#### Before
image

#### After
image

## Documentation

    - [x] No

## Related

Resolves #285

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -813,6 +813,7 @@ func testAcceptance(
 						it("doesn't have color", func() {
 							appPath := filepath.Join("testdata", "mock_app")
 
+							//--no-color is set as a default option in our tests, and doesn't need to be explicitly provided
 							output := pack.RunSuccessfully("build", repoName, "-p", appPath)
 							imgId, err := imgIDForRepoName(repoName)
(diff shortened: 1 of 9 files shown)


jromero (approved):

jromero:
🎉 LGTM

Feel free to merge once re-testing is successful

Issue #745, opened 2020-07-16
Buildpack Registry config commands


jromero:
### Description

As a `pack` user it would be nice to have commands to manage the registries in the `config` file.

### Proposed solution

- [x] `pack list-buildpack-registry` - a new command that will list all the `[[registries]]` config in `.pack/config.toml`.
- [x] `pack add-buildpack-registry` - a new command that will add to the `[[registries]]` config in `.pack/config.toml`.
- [ ] `pack remove-buildpack-registry` - a new command that will remove a registry from the `[[registries]]` config in `.pack/config.toml`.
- [ ] `pack set-buildpack-registry` - a new command that will mutate the `default-registry` config in `.pack/config.toml`.

For more details on what the data structure of the `.pack/config.toml` see [RFC #44](https://github.com/buildpacks/rfcs/blob/main/text/0044-pack-publish-buildpack.md#pack-register-buildpack-url).

### Additional Refactor/Clean-up
`add-buildpack-registry `-> `add-registry`
`list-buildpack-registry` -> `list-registries`

### Additional context
- [x] This feature should be documented somewhere

- These commands should be hidden via `experimental` config option.

elbandito:
@jromero and @dfreilich I'm curious what your opinion is on removing `buildpack` from the command?  For example, `add-buildpack-registry` would become `add-registry`?  I personally like any opportunity to type less characters, but I also don't know if there could be other types of `registries`, making it too vague.  Thoughts?

dfreilich:
Hmmm. I would personally find it confusing, assuming that it's talking about OCI registries, rather than anything buildpack related, and then wonder whether it relates to #821. I'd personally vote for keeping buildpacks in (and then just hope people use `pack completion`, so it's less to type)

dfreilich:
@jromero In light of https://github.com/buildpacks/pack/issues/597, what should these commands look like?

dfreilich:
And should that impact what the commands in #839 is as well?

jromero:
@elbandito I'm in favor of making the buildpack registries the primary registry types we talk about. Hence they shouldn't need the explicit denotation, as you suggested. We can then use "image registries" when talking about oci/docker registries.

CC: @jkutner / @ekcasey

elbandito:
@jromero sounds good!  I'll create an issue to refactor the following:

`add-buildpack-registry` -> `add-registry`
`list-buildpack-registyr` -> `list-registries`

Or do we want to get these ^ in as a patch?

[...]

PR #839, opened 2020-09-17
List buildpack registries

elbandito:
## Summary

Allow the user to see what current Buildpack Registries they already have configured in their `config.toml` file.

## Output

```
➜  pack list-buildpackregistries
* official
  joe
  dpack
```

```
➜  pack list-buildpack-registries --verbose
  official   https://github.com/elbandito/buildpack-registry
* joe        https://github.com/jkutner/buildpack-registry/
  dpack      https://github.com/dfreilich/pack
  travis     github
Tip: Run pack add-buildpack-registry to add additional registries

```

## Documentation

    - [ ] Yes, see #___
    - [X] No (not yet officially - still experimental)

## Related
https://github.com/buildpacks/pack/issues/745


internal/commands/list_registries.go
```diff
@@ -0,0 +1,32 @@
...
+			for _, registry := range cfg.Registries {
+				if registry.Name == cfg.DefaultRegistryName {
+					logger.Infof("* %s", registry.Name)
+				} else {
+					logger.Infof("  %s", registry.Name)
+				}
+				if logger.IsVerbose() {
+					logger.Infof("    Type: %s", registry.Type)
+					logger.Infof("    URL:  %s", registry.URL)
+				}
```
dfreilich, line 25: Would adding this as a debug comment work here? I think it should also only print if verbose, and simplify this a bit.

internal/commands/list_registries_test.go
```diff
@@ -0,0 +1,96 @@
...
+			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/personal-registry")
+		})
+
+		it("should list nothing when NO registries exist", func() {
+			cfg = config.Config{}
+			command = commands.ListRegistries(logger, cfg)
+
+			h.AssertNil(t, command.Execute())
+			h.AssertEq(t, outBuf.String(), "")
+		})
```
dfreilich, line 94: What do you think about logging something like "No registries have been added yet. You can do that by `pack add-buildpacks-registry`"? I'd be confused if I ran this and nothing happened.

elbandito: @dfreilich I could go either way.  Part of me is inclined to think that if the user is confused they could look at the command's help menu to better understand its behavior.  I'll let @jromero be the tie breaker :smiley:

internal/commands/list_registries.go
elbandito: @dfreilich I have a question.  Do you know if all the log functions e.g. `Info`, `Debug`, etc.  print to `stderr`?

jromero: Only `Error` prints to stderr:

https://github.com/buildpacks/pack/blob/main/internal/logging/logging.go#L90-L92

elbandito: Ok, good to know.  Generally speaking, I try and only reserve parse-able output to `stdout`.  I don't know if this output would necessarily classify as parse-able, but at the same time, I don't want `Error:` showing-up in the log output.

elbandito: @dfreilich I realise this has been outdated, but do you still have any thoughts on the new changes?

elbandito: @jromero I do wonder if I should be outputting to `stderr` instead of `stdout`

jromero: As a CLI, I don't understand why we would want to. What would we be trying to achieve by printing to stderr?

elbandito: @jromero best practices suggest reserving `stdout` for things that are easily parse-able.  This is especially helpful when our users write scripts that use `pack`.

As I think more about this particular case it's probably to write to `stdout`.

internal/commands/list_registries_test.go
elbandito: @jromero do you have any strong thoughts on this?

internal/commands/list_registries.go
dfreilich: I definitely agree that this should be printed to `stdout`, rather than `stderr`.

elbandito:
@jromero and @dfreilich, thank you for your review.   Are there any outstanding items that you'd like for me to address?

internal/commands/list_buildpack_registries_test.go
```diff
@@ -0,0 +1,94 @@
...
+			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/public-registry")
+
+			h.AssertContains(t, outBuf.String(), "private registry")
+			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/private-registry")
+
+			h.AssertContains(t, outBuf.String(), "personal registry")
+			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/personal-registry")
+		})
+
+		it("should list nothing when NO registries exist", func() {
```
jromero, line 86: Is there not the idea of a built-in default registry? Aka "official"? In my mind, I had envisioned it as being set automatically (out-of-the-box). It was up in the air as to whether it could be removed or not. If it can't be removed then would we ever have this case?

Also, I agree with @dfreilich that we should "Hint" as to how to add a registry when listing the registries and there are no custom registries set.

elbandito: @jromero I can add hint.  

With regards to always having a "`official`" registry, I would have to double check or maybe @jkutner knows off the top of his head.  However, we don't have complete control over this file, so if the user wanted to they could intentionally (or accidentally) delete it.  If this were the case what's the expected behavior?  Should we automatically re-create the `official` registry and always display it at a minimum?

jromero: > With regards to always having a "official" registry, I would have to double check or maybe @jkutner knows off the top of his head. However, we don't have complete control over this file, so if the user wanted to they could intentionally (or accidentally) delete it. If this were the case what's the expected behavior? Should we automatically re-create the official registry and always display it at a minimum?

It may be worth discussing in a separate issue as to not overload this one. I believe a strategy that _may_ work is to hardcode it in memory instead of appending it to the config file. A similar example of this would be trusted-builders in which case we list trusted-builders as "user-defined trusted builder" (from config.toml) + "suggested builders" (from hard coded values).

jromero: Created https://github.com/buildpacks/pack/issues/842 to continue conversation there.

jkutner: I commented on the issue, but this test should check that the _default_ registry is listed when there are no entries in the config file. That default registry (the official registry) should be hard-coded into pack. We can allow it to be overridden in the config though.

jromero (approved):

PR #840, opened 2020-09-17
Enable package-buildpack for windows

TisVictress:
## Summary
This change enables creation of Windows buildpackages. Windows buildpackages differ from Linux buildpackages in these ways:

#### Linux buildpackage image (no change from main)
* The image's `os` [property](https://github.com/opencontainers/image-spec/blob/master/config.md#:~:text=os%20string) is `linux`
* Image buildpack layers are tar files with entry paths at `/cnb/buildpacks/...`

#### Linux buildpackage file (no change from main)
* The *Linux buildpackage image* packaged as as a standard `.cnb` 

#### Windows buildpackage image (new)
* The image's `os` property is `windows` (depends on: https://github.com/buildpacks/imgutil/pull/66)
* Image buildpack layers are tar files with entry paths at `Files/cnb/buildpacks/...`
* Image base layer is a minimal set of layer files required to support pulling and storing buildpackages on a WCOW daemon using `pack` or `docker` (depends on: https://github.com/buildpacks/imgutil/pull/64) 

#### Windows buildpackage file (new)
* The *Windows buildpackage image* packaged as a standard `.cnb`

## Before
The `package-buildpack` command always generated buildpackage files/images as Linux OCI images.

## After

The `package-buildpack` command preserves existing behavior but supports creating files/images as Windows OCI images based on `--os=windows` flags or daemon-based OS detection.

Preserves existing behavior (with or without pack config `experimental=true`):
* `package-buildpack --format=file ...` => Linux buildpackage file (No daemon required)
* `package-buildpack --format=image --publish ...` => Linux buildpackage  image on registry (No daemon required)
* `package-buildpack --format=image ...` => Linux buildpackage image on daemon (when Linux daemon present)

With pack config `experimental=true`:
* `package-buildpack --format=file --os=windows ...` => Windows buildpackage file (No daemon required.)
* `package-buildpack --format=image --publish --os=windows...` => Windows buildpackage  image on registry (No daemon required)
* `package-buildpack --format=image --os=windows ...` => Windows buildpackage image on daemon (when WCOW daemon present, otherwise fail)
* `package-buildpack --format=image ...` => Linux/Windows buildpackage image on daemon (when Linux/WCOW daemon present, resp)

With pack config `experimental` not enabled:
* `package-buildpack --os=linux` => Fails with experimental support missing for "os" flag
* `package-buildpack --format=image ...` => Fails with experimental support missing for Windows packages (when Windows daemon present)

## Example artifacts:
cnbs/sample-package:hello-world-windows (built from [draft PR](https://github.com/buildpacks/samples/pull/93)):
  * `hello-universe-windows.cnb` ([Google Drive download](https://drive.google.com/file/d/1VBRBIqS_J7RvVj85fcQnuRyUXWi0FtIH/view?usp=sharing))
  * `dive micahyoung/sample-package:hello-universe-windows` (requires WCOW daemon)
```
┃ ● Layers ┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ │ Current Layer Contents ├──────────────────────
Cmp   Size  Command                              ├── Files
     12 kB  FROM 8a4dd6e94861fde                 │   ├── Windows
    1.5 kB                                       │   │   └── System32
    1.5 kB                                       │   │       └── config
     782 B                                       │   │           ├── DEFAULT
                                                 │   │           ├── SAM
│ Layer Details ├─────────────────────────────── │   │           ├── SECURITY
                                                 │   │           ├── SOFTWARE
Tags:   (unavailable)                            │   │           └── SYSTEM
Id:     bc0d21bd19f46c9aec44955ebc8db42808a9303a │   └── cnb  
9d52e279080d281c4ba20b0d                         │       └── buildpacks
Digest: sha256:47b97ff94b56931f42be9d1b5a454751c │           ├── samples_hello-moon-windows
8a6ce95c7fdf29db54ebcefa6858787                  │           │   └── 0.0.1
Command:                                         │           │       ├── README.md
                                                 │           │       ├── bin
                                                 │           │       │   ├── build.bat
│ Image Details ├─────────────────────────────── │           │       │   └── detect.bat
                                                 │           │       └── buildpack.toml
                                                 │           ├── samples_hello-universe-windows
Total Image size: 16 kB                          │           │   └── 0.0.1
Potential wasted space: 0 B                      │           │       ├── README.md
Image efficiency score: 100 %                    │           │       └── buildpack.toml
                                                 │           └── samples_hello-world-windows
Count   Total Space  Path                        │               └── 0.0.1             
    3           0 B  /Hives                      │                   ├── README.md     
                                                 │                   ├── bin
                                                 │                   │   ├── build.bat
                                                 │                   │   └── detect.bat
                                                 │                   └── buildpack.toml
                                                 ├── Hives                
                                                 └── UtilityVM           
                                                     └── Files              
                                                         └── EFI
                                                             └── Microsoft
                                                                 └── Boot
                                                                     └── BCD
``` 
Signed-off-by: Victoria Henry <vhenry@vmware.com>
Signed-off-by: Andrew Meyer <meyeran@vmware.com>
Signed-off-by: Micah Young <meyeran@vmware.com>


micahyoung:
Awesome to see this working! Just to start the discussion, is there a detailed behavior description we can give for the behavior here? What's images are created for local, publish, file? Which layers are created in them and in which format for each case?

acceptance/acceptance_test.go
```diff
@@ -346,6 +347,7 @@ func testWithoutSpecificBuilderRequirement(
 						"package-buildpack", packageName,
 						"-c", aggregatePackageToml,
 						"--publish",
+						"--os", dockerHostOS(),
```
ameyer-pivotal, line 350: I wonder if there's a way to infer this without a flag... I understand why we might not want to look at the dockerhost OS, as you might be cross-building. Though wouldn't that only work in a `--publish` scenario?

Back to the original question, is there some way we could derive this from the buildpack? Seems unlikely but thought I'd ask.

create_builder_test.go
```diff
@@ -563,9 +568,9 @@ func testCreateBuilder(t *testing.T, when spec.G, it spec.S) {
 				bldr := successfullyCreateBuilder()
 
 				h.AssertEq(t, bldr.LifecycleDescriptor().Info.Version.String(), "0.0.0")
-				//nolint:staticcheck
+				// nolint:staticcheck
```
ameyer-pivotal, line 571: Guessing this was from some auto-formatting. Wanted to call it out in case the linter will expect a certain format.

ameyer-pivotal (changes requested):
This looks really good. I did notice that Windows buildpackage support seems to be missing the experimental flag requirement. Other than that, this is looking great!

acceptance/acceptance_test.go
micahyoung: Yeah, that's a good thing to discuss. Right now, for backward compatibility reasons, this implementation does still always default to `linux` for `--publish` or `--format=file` w/o the `--os` flag (it emits a warning telling you this) but for local images (`package-buildpack` w/o options) it will autodetect the image type (without a warning) and save it on the dameon. Also for backward compatibility, it can't _require_ a daemon for `--publish` or `--format=file`. 

With that in mind, the autodetection options for `--publish` or `--format=file` might be to detect based on a _optional_ daemon, but depending on if your daemon is running/stopped, you have a reachable/unreachable `DOCKER_HOST` set, or some other environment difference, you'd generate different images with the same command.

Moreover, autodetection might get surprising for new users - especially on a Windows workstation, which depending on whether you picked LCOW or WCOW, would make different remote images and `.cnb` files for the same command.

I feel like autodetecting the image type only really feels right in the case of a local image on the daemon, where it has to match the running daemon anyways and it unlikely to be pushed up to a registry or copied around.

I'd love to hear any holes in that logic though

micahyoung: We *could* make the `os` an optional field in the package config though, instead of the `--os` flag. Then `package-buildpack` would consistently generate the same image time regardless of daemon. The only potential catch there is that to-date, the package config tends to describe CNB-related things (buildpack paths and images) inside a package and not things about the way it's packaged (ex: `--format` is specified on the command line, not in the `package.toml`).

create_builder_test.go
micahyoung: Fixed

go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/buildpacks/pack
 
+//TODO remove before merge
```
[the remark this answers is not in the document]
micahyoung: Bumped!

micahyoung:
> This looks really good. I did notice that Windows buildpackage support seems to be missing the experimental flag requirement. Other than that, this is looking great!

Great catch. Now fixed! I added the experimental flag for both `--os` and when Windows is autodetected for local images without the `--os` flag

jromero (approved):
This is great! _Clean_ implementation.

Issue #515, opened 2020-02-29
Spike: Implementations of Cloud Native Buildpacks on CI/CD platforms


jromero:
### Description
Cloud Native Buildpacks primary function is to turn source code into runnable image and because of that it's natural for it to be used within various CI/CD platform pipelines.

The purpose of this issue is produce various proof of concept implementations for the top 5 to 10 platforms as ranked by [stackshare](https://stackshare.io/continuous-integration).

- [ ] Jenkins
- [ ] Travis CI
- [ ] Circle CI
- [ ] Gitlab CI
- [ ] Codeship
- [ ] TeamCity
- [ ] Bamboo
- [ ] Drone.io
- [ ] wercker
- [ ] Concourse

2pai:
hello @jromero , I would like to take this as my GSOC project. Can you help me with it? :D

zmackie:
Hi @2pai 

Here are a few resources I was able to put together:

- [lifecycle phases](https://youtu.be/SK6e_ZatOaw?t=358) (video) - A high-level explanation of the lifecyle and it's phases.
- [pack build process](https://github.com/buildpacks/pack/blob/master/internal/build/build.go) (code) -  How these phases are executed by pack.
- [buildpack ecosystem](https://www.youtube.com/watch?v=j9Ak5YLrihU&feature=youtu.be&t=151) (video) - How lifecycle plays a part in the bigger system.

These probably provide a higher-level view of the lifecycle then what you might be asking for. I'm looping in @ekcasey who is a maintainer of the lifecyle and can give you a bit more direct insight into contributing. 

Also, @jromero  who originally proposed this story can give you more guidance into the outcomes.

jromero:
We can break each CI platform into its own issue given the scope of this issue and the potential for parallel work.

@2pai do you have one in mind that you'd like to look into?

2pai:
Thank you @zmackie 

I think I would take Jenkins @jromero

jromero:
@2pai https://github.com/buildpacks/pack/issues/531 has been created. 😄

safakozdek:
I am interested in working for GitLab CI integration as a GSoC participant. Where can I start ?

chialin-liu:
Dear @jromero, I had the experience about using docker associated with Travis CI to deploy multiple container applications to AWS. May I take this as a GSoC project? Thank you so much

jromero:
Given this is an epic. It probably should have a milestone associated and instead the individual CI efforts should.

[...]

Issue #844, opened 2020-09-23
Publish with multiple tags


dfreilich:
### Description

I'd like to publish an image with pack, with multiple tags. For instance, `dfreilich/pack:latest` and `dfreilich/pack:v0.0.15`. 

### Proposed solution

Add a flag `--tag` to allow for inputting multiple tags. This should work in cases where we call the creator, and can pass in the `-tag` flag. https://github.com/buildpacks/spec/blob/main/platform.md#creator

### Additional context
- [x] This feature should be documented somewhere

ekcasey:
> This should work in cases where we call the creator, and can pass in the -tag flag

This should also work with untrusted builders, as `exporter` can accept multiple tags as positional arguments.

PR #846, opened 2020-09-23
Replace docker with pack when building pack image

dfreilich:
## Summary

As part of our delivery process, we push an image of pack to `buildapcksio/pack`. While we had been using docker for that, this PR uses [dfreilich/pack-action](https://github.com/dfreilich/pack-action) to release the image using `pack build....`

The release worked when I tried it out on my fork: [GH Action](https://github.com/dfreilich/pack/actions/runs/268939564), and the images are visible on DockerHub [here](https://hub.docker.com/r/dfreilich/pack/tags). 

#### Before
Docker

#### After
dogfood

## Documentation

    - [x] No

## Related

Relates to #515

### DIFF
--- a/.github/workflows/delivery-docker.yml
+++ b/.github/workflows/delivery-docker.yml
@@ -1,19 +1,14 @@
 name: delivery / docker
 
 on:
-  workflow_dispatch:
-    inputs:
-      tag_name:
-        description: 'tag'
-        required: true
-#  release:
-#    types:
-#      - published
+  release:
+    types:
+      - published
 
 env:
   BUILDER: "paketobuildpacks/builder:tiny"
   IMG_NAME: 'pack'
-  USERNAME: 'dfreilich'
+  USERNAME: 'buildpacksio'
 
 jobs:
   deliver-docker:
@@ -24,7 +19,7 @@ jobs:
         id: version
         shell: bash
         run: |
-          echo "::set-output name=VERSION::$(echo ${{ github.event.inputs.tag_name }} | cut -d "v" -f2)"
+          echo "::set-output name=VERSION::$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)"
       - name: Setup Working Dir
         shell: bash
         run: |
--- a/.github/workflows/delivery-docker.yml
+++ b/.github/workflows/delivery-docker.yml
@@ -1,26 +1,57 @@
 name: delivery / docker
 
 on:
-  release:
-    types:
-      - published
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'tag'
+        required: true
+#  release:
+#    types:
+#      - published
+
+env:
+  BUILDER: "paketobuildpacks/builder:tiny"
+  IMG_NAME: 'pack'
+  USERNAME: 'dfreilich'
 
 jobs:
   deliver-docker:
     runs-on: ubuntu-latest
+    container:
+      image: docker:stable
+      volumes:
+        - /home/runner:/var/www
     steps:
       - uses: actions/checkout@v2
       - name: Determine version
         id: version
         shell: bash
         run: |
-          echo "::set-output name=VERSION::$(echo ${{ github.event.release.tag_name }} | cut -d "v" -f2)"
-      - name: Build and Push Image
-        uses: docker/build-push-action@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/delivery-docker.yml
```diff
@@ -15,12 +20,31 @@ jobs:
...
+      - name: Login to Dockerhub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Tag Image as Latest
+        run: |
+          docker pull ${{ env.IMG_NAME }}:${{ steps.version.outputs.VERSION }}
+          docker tag ${{ env.IMG_NAME }}:${{ steps.version.outputs.VERSION }} ${{ env.IMG_NAME }}:latest
+          docker push ${{ env.IMG_NAME }}:latest
```
dfreilich, line 50: This is relevant to #844. We currently publish the image to both tags: `latest` and `<VERSION>`, and we can only do one when we push an image. Therefore, we tag it afterwards. 

In order to access/push the image, though, we need to re-verify, because the credentials from pack aren't transferred to the rest of the workflow (it is restricted to the container that pack runs in)

.github/workflows/delivery/docker/buildpack.yml
```diff
@@ -0,0 +1,16 @@
...
+  targets:
+    - ./cmd/pack
+
+  build:
+    # The go.build.flags property allows you to override the default build
+    # flags when compiling your program.
+    flags:
+      - -buildmode=default
+      - -tags=paketo
+      - -ldflags="-s -w -X 'github.com/buildpacks/pack.Version=#{PACK_VERSION}#' -extldflags ${LDFLAGS}"
```
dfreilich, line 11: There is currently an open issue (https://github.com/paketo-buildpacks/go-build/issues/60) for interpolating as part of pack build, so we don't have to do it in the workflow.

.github/workflows/delivery/docker/buildpack.yml
```diff
@@ -0,0 +1,16 @@
...
+
+    # The go.build.import-path property allows you to specify an import path
+    # for your application. This is necessary if you are building a $GOPATH
+    # application that imports its own sub-packages.
+    import-path: github.com/buildpacks/pack
```
ryanmoran, line 16: You should not need this as this is a `go.mod` app. The buildpack can find the import path from the `go.mod` file.

.github/workflows/delivery/docker/buildpack.yml
```diff
@@ -0,0 +1,16 @@
+go:
+  targets:
+    - ./cmd/pack
+
+  build:
+    # The go.build.flags property allows you to override the default build
+    # flags when compiling your program.
+    flags:
+      - -buildmode=default
+      - -tags=paketo
```
ryanmoran, line 10: These flags are not necessary if you don't need them for a specific reason. We show them in docs as examples of what you might want to override, but you don't NEED them.

dfreilich: After experimenting, it seems like I needed the `buildmode=default` (otherwise I got [this](https://github.com/dfreilich/pack/runs/1156366130?check_suite_focus=true) error), but not tags

dfreilich:
I was able to verify that the current version works [here](https://github.com/dfreilich/pack/actions/runs/269189131)

.github/workflows/delivery/docker/buildpack.yml
ryanmoran: Seems like that is caused the the `${LDFLAGS}` env var that is not interpolated. It looks like the build flags you include are trying to pass some flags to the external linker, `gcc` in this case, and you haven't specified what those are. Do you really need to set `-extldflags`? And if so, can you not hard-code those values?

dfreilich: @jromero looks like you added that here: https://github.com/buildpacks/pack/commit/968445647b9b99d995a718208706fdfab5723f9e#diff-b67911656ef5d18c4ae36cb6741b7965. Any chance you remember why?

jromero: This was a feature request by @ttpcodes, the contributor/collaborator, of the Arch Linux delivery flow. It shouldn't be necessary for this use case since it's a controlled build environment.

dfreilich: I'll remove it then.

PR #847, opened 2020-09-23
feat: add buildpack registry

elbandito:
## Summary (WIP)
This is a new command that allows users to add new buildpack registries to their `config.toml` file.   The user also has the option via `--default` flag to set the newly added registry as the new default registry.

## Output
```
➜  pack add-buildpack-registry --help                                          
Adds a new buildpack registry to your pack config file

Usage:
  pack add-buildpack-registry <name> <url> <type> (github | git) [flags]

Examples:
pack add-buildpack-registry myregistry github https://github.com/buildpacks/mybuildpack

Flags:
      --default   Set this buildpack registry as the default
  -h, --help      Help for 'add-registry'

Global Flags:
      --no-color     Disable color output
  -q, --quiet        Show less output
      --timestamps   Enable timestamps in output
  -v, --verbose      Show more output
```

```
➜ pack add-buildpack-registry buildpack https://github.com/buildpacks/registry github
Successfully added buildpack to buildpack registries
```

```
➜ pack add-buildpack-registry buildpack https://github.com/buildpacks/registry github
ERROR: add buildpack registry: Buildpack registry buildpack already exists.  First run remove-buildpack-registry buildpack and try again.
```

## Documentation

- Should this change be documented?
    - [X] Yes, see #___
    - [ ] No

## Related

Resolves #___

PR #849, opened 2020-09-24
Prepend experimental to experimental commands

dfreilich:
* Adds Experimental prior to experimental command help flags
* Add long description for add_buildpacks_registry
* Fix spacing in config.go

(essentially, some small fixes for #847)

#### Before
image

#### After
image

## Documentation

    - [x] No

### DIFF
--- a/internal/commands/commands.go
+++ b/internal/commands/commands.go
@@ -77,6 +77,10 @@ func multiValueHelp(name string) string {
 	return fmt.Sprintf("\nRepeat for each %s in order,\n  or supply once by comma-separated list", name)
 }
 
+func prependExperimental(short string) string {
+	return fmt.Sprintf("(%s) %s", style.Warn("experimental"), short)
+}
+
 func getMirrors(config config.Config) map[string][]string {
 	mirrors := map[string][]string{}
 	for _, ri := range config.RunImages {
--- a/internal/commands/add_registry.go
+++ b/internal/commands/add_registry.go
@@ -22,9 +22,12 @@ func AddBuildpackRegistry(logger logging.Logger, cfg config.Config, cfgPath stri
 	)
 
 	cmd := &cobra.Command{
-		Use:   "add-registry <name> <url>",
-		Args:  cobra.ExactArgs(2),
-		Short: "Add buildpack registry to config file",
+		Use:     "add-registry <name> <url>",
+		Args:    cobra.ExactArgs(2),
+		Short:   prependExperimental("Add buildpack registry to your pack config file"),
+		Example: "pack add-registry my-registry https://github.com/buildpacks/my-registry",
+		Long: "A Buildpack Registry is a (still experimental) place to publish, store, and discover buildpacks. " +
+			"Users can add buildpacks registries using add-registry, and publish/yank buildpacks from it, as well as use those buildpacks when building applications.",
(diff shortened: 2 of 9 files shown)


internal/commands/add_buildpack_registry_test.go
```diff
@@ -27,49 +28,62 @@ func TestAddBuildpackRegistry(t *testing.T) {
...
 		)
 
 		it.Before(func() {
 			var err error
-			tmpDir, err = ioutil.TempDir("", "pack-home-*")
+			tmpDir, err = ioutil.TempDir("", "pack-home-")
 			assert.Nil(err)
 
-			configFile = filepath.Join(tmpDir, "config.toml")
+			assert.Nil(os.Setenv("PACK_HOME", tmpDir))
```
jromero, line 43: Parallel test runs don't run in their own environment. I had it implemented this way but flakes started to occur due to the `PACK_HOME` being overridden in concurrent executions when `spec.Parallel()` is set. We could move it to `spec.Sequential` and live with it but ultimately it steered me into the thought that we _shouldn't_ be calling `DefaultConfigPath()` during our tests at all. Instead we should be relying on dependency injection instead of global state.

internal/commands/commands.go
```diff
@@ -77,6 +77,10 @@ func multiValueHelp(name string) string {
 	return fmt.Sprintf("\nRepeat for each %s in order,\n  or supply once by comma-separated list", name)
 }
 
+func prependExperimental(short string) string {
+	return fmt.Sprintf("%s: %s", style.Symbol("Experimental"), short)
```
jromero, line 81: I love this feature as a whole. I would suggest a few tweaks to the visual output.

1. I would use yellow or no color.
2. Instead of `Experimental:` it may look better as `(experimental)` given it's surroundings.

jromero (changes requested):

internal/commands/add_buildpack_registry_test.go
dfreilich: Ah interesting. I thought this was an overthought, so I switched it back to our normal pattern. If we switch it to dependency injection, we should probably be doing it throughout though...

internal/commands/commands.go
dfreilich: Yellow, just to make the contrast higher?

internal/commands/add_buildpack_registry_test.go
jromero: We should. I didn't want to grow scope on someone else's PR. :D

internal/commands/commands.go
jromero: `Symbol` in my mind is used for "values" where this to me feels more like an annotation that should call attention similar to a warning message.

internal/commands/add_buildpack_registry_test.go
dfreilich: Makes sense. I'll revert this change, then; and I opened https://github.com/buildpacks/pack/issues/852 for fixing this entirely.

internal/commands/commands.go
dfreilich: image

Thoughts?

dfreilich:
Current output: 
image

jromero (approved):

PR #855, opened 2020-09-26
Add option to install pack CLI using scoop on Windows

goostleek:
## Summary
An additional option to install pack CLI (using scoop) on Windows has been documented on the release page.
Scoop's packages database is refreshed automatically once a day so it will always install the latest version available.

#### Before

> #### Windows
>
>On Windows, you can use [Chocolatey](https://chocolatey.org/):
>
>```bash
>choco install pack
>```

## Related
buildpacks/docs#217

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -204,10 +204,16 @@ jobs:
 
             On Windows, you can use [Chocolatey](https://chocolatey.org/):
 
-            ```bash
+            ```cmd
             choco install pack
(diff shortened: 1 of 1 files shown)


jromero (approved):
👍 Good changes! Appreciate the work put in to add this to scoop!

PS: We _might_ be replacing these installation instructions with simply a link to the docs (which you updated). I'll still merge this in in case the change doesn't go through.

PR #857, opened 2020-09-29
Ensure there is always an official default registry

### LINKED ISSUE #842: Add default official registry
> With regards to always having a "official" registry, I would have to double check or maybe @jkutner knows off the top of his head. However, we don't have complete control over this file, so if the user wanted to they could intentionally (or accidentally) delete it. If this were the case what's the expected behavior? Should we automatically re-create the official registry and always display it at a minimum?

I believe a strategy that _may_ work is to hardcode it in memory instead of appending it to the config file. A similar example of this would be trusted-builders in which case we list trusted-builders as "user-defined trusted builder" (from config.toml) + "suggested builders" (from hard coded values).

_Originally posted by @jromero in https://github.com/buildpacks/pack/pull/839#discussion_r492211992_

elbandito:
## Summary
Given our limited control of the `.pack/config.toml` file, we should always ensure that a default registry is always being communicated to the user.   If, for some reason, no registries defined, the default registry will be displayed to the customer. 

#### Before

**When no registries exist**
```
❯ pack list-buildpack-registries -v
Tip: Run pack add-buildpack-registry to add additional registries
```

**When registries already exist**
```
❯ pack list-buildpack-registries -v
* travis     https://github.com/elbandito/buildpack-registry
  joe        https://github.com/jkutner/buildpack-registry
```

#### After
**When no registries exist**
```
❯ pack list-buildpack-registries -v
* official   https://github.com/buildpacks/registry-index
Tip: Run pack add-buildpack-registry to add additional registries
```

**When registries already exist**
```
❯ pack list-buildpack-registries -v
* travis     https://github.com/elbandito/buildpack-registry
  joe        https://github.com/jkutner/buildpack-registry
  official   https://github.com/buildpacks/registry-index
```

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #842

### DIFF
--- a/internal/commands/list_buildpack_registries_test.go
+++ b/internal/commands/list_buildpack_registries_test.go
@@ -76,11 +76,39 @@ func testListBuildpackRegistriesCommand(t *testing.T, when spec.G, it spec.S) {
 			h.AssertContains(t, outBuf.String(), "public registry")
 			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/public-registry")
 
-			h.AssertContains(t, outBuf.String(), "private registry")
+			h.AssertContains(t, outBuf.String(), "* private registry")
 			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/private-registry")
 
 			h.AssertContains(t, outBuf.String(), "personal registry")
 			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/personal-registry")
+
+			h.AssertContains(t, outBuf.String(), "official")
+			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/registry-index")
+		})
+
+		it("should indicate official as the default registry by default", func() {
+			logger := ilogging.NewLogWithWriters(&outBuf, &outBuf)
+			cfg.DefaultRegistryName = ""
+
+			command = commands.ListBuildpackRegistries(logger, cfg)
+
+			h.AssertNil(t, command.Execute())
+
+			h.AssertContains(t, outBuf.String(), "* official")
+			h.AssertContains(t, outBuf.String(), "public registry")
+			h.AssertContains(t, outBuf.String(), "private registry")
+			h.AssertContains(t, outBuf.String(), "personal registry")
+		})
+
+		it("should use official when no registries are defined", func() {
+			logger := ilogging.NewLogWithWriters(&outBuf, &outBuf)
+			cfg.DefaultRegistryName = ""
+
+			command = commands.ListBuildpackRegistries(logger, config.Config{})
+
+			h.AssertNil(t, command.Execute())
+
+			h.AssertContains(t, outBuf.String(), "* official")
 		})
 	})
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -36,6 +36,16 @@ type TrustedBuilder struct {
 	Name string `toml:"name"`
 }
 
+const OfficialRegistryName = "official"
+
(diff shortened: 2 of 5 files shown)


internal/commands/list_buildpack_registries_test.go
```diff
@@ -76,11 +76,21 @@ func testListBuildpackRegistriesCommand(t *testing.T, when spec.G, it spec.S) {
...
 
-			h.AssertContains(t, outBuf.String(), "private registry")
+			h.AssertContains(t, outBuf.String(), "* private registry")
 			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/private-registry")
 
 			h.AssertContains(t, outBuf.String(), "personal registry")
 			h.AssertContains(t, outBuf.String(), "https://github.com/buildpacks/personal-registry")
 		})
+
+		it("should ALWAYS list an official default registry", func() {
```
jromero, line 86: This test says "ALWAYS" but it's only testing when there are not other registries.

Can we have two tests? One when no registries are added and another for when 1 or more registries are added.

FWIW: I agree that official should always be displayed. It currently doesn't _always_ display. It only display when there are no others:

```
$ ./out/pack list-buildpack-registries -v
  my-registry https://github.com/buildpacks/my-buildpack
  my-registry2 https://github.com/buildpacks/my-buildpack
* my-registry3 https://github.com/buildpacks/my-buildpack
```

jromero, line 86: We should also add a test that "official" cannot be redeclared via `add-buildpack-registry`. We can discuss more intricate edge-cases later but for now we should allow for "official" to be listed twice.

internal/config/config.go
```diff
@@ -107,9 +115,5 @@ func GetRegistry(cfg Config, registryName string) (Registry, error) {
 		return Registry{}, errors.Errorf("registry %s is not defined in your config file", style.Symbol(registryName))
```
jromero, line 115: Wouldn't this fail if I call `GetRegistry(cfg, "official")`? Where does the amendment of the in-memory "official" happen?

jromero (changes requested):

internal/config/config.go
elbandito: This code path that you're referring to is for the `register-buildpack` command.  If there isn't a defined registry in `config.toml` file, `GetRegistry()` will just return a single default registry to use for opening up the Github issue against.    As of now, if I called `GetRegistry(cfg, "official")` it would look for "official", but it shouldn't find it, and will return the hardcoded default registry which would be the `official` registry.

There isn't any amending that happening.  This is only for when no registries are defined, then it will default to "official".

internal/commands/list_buildpack_registries_test.go
elbandito: My approach was to treat the `official` registry as a fallback in case there weren't any registries.  Similiar to what we're already doing with https://github.com/buildpacks/pack/blob/4c29bf83e989fdf0843183a3b6c2d21d1ce24f6d/internal/config/config.go#L105.   But maybe we should always include it with the others.  I need to think about this some more.

internal/config/config.go
elbandito: We aren't doing any amending, never had before either.  This code path is responsible for returning a single registry that should be used for creating the Github issue against.  When a user runs `register-buildpack` if no registries are defined, a default registry will be used instead, which is the `"official"` registry.

jromero (approved):

PR #870, opened 2020-10-05
PPA publish

### LINKED ISSUE #704: Publish to a PPA (for Ubuntu)
### Description
As an Ubuntu user, I'd like to use my standard package manager to install `pack`.

### Proposed solution
Provide a [PPA](https://help.ubuntu.com/stable/ubuntu-help/addremove-ppa.html.en) from which `pack` can be installed.

### Additional context
- [x] This feature should be documented:
    - [Release notes](https://github.com/buildpacks/pack/blob/main/.github/workflows/build.yml#L186) (maintained in this repo)
    - https://buildpacks.io/docs/install-pack/

#### References
- [Guide: All about PPA](https://itsfoss.com/ppa-guide/)
- [Stackoverflow: Resources to create a PPA](https://askubuntu.com/a/71516)

dwillist:
## Summary
Add ubuntu ppa publishing as part of the release process

## Output
This will publish ubuntu PPA release of the `pack-cli` at the following ppa location https://launchpad.net/~cncf-buildpacks/+archive/ubuntu/pack-cli

#### After

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #704


.github/workflows/scripts/create_ppa.sh
```diff
@@ -0,0 +1,70 @@
...
+
+    # build a source based debian package, Ubuntu ONLY accepts source packages.
+    debuild -S
+
+}
+
+# import gpg keys from env
+function import_gpg() {
+    gpg --import <(echo "$GPG_PUBLIC_KEY")
+    gpg --allow-secret-key-import --import <(echo "$GPG_PRIVATE_KEY")
```
jromero, line 67: Can we add checks to make sure these env variables are set and fail early so that we get an concise error.

Makefile
```diff
@@ -47,18 +47,24 @@ endif
...
+build: out
+	@echo "> Building..."
+	$(GOCMD) build -ldflags "-s -w -X 'github.com/buildpacks/pack.Version=${PACK_VERSION}' -extldflags ${LDFLAGS}" -trimpath -o ./out/$(PACK_BIN) -a ./cmd/pack
+
 all: clean verify test build
 
+install:
+	mkdir -p ${DESTDIR}${BINDIR}
+	cp ./out/$(PACK_BIN) ${DESTDIR}${BINDIR}/
+
```
jromero, line 61: Is this necessary? If so, can we add a comment as to why and where we can "learn more".

Makefile
```diff
@@ -47,18 +47,24 @@ endif
 export GOFLAGS:=$(GOFLAGS)
 export CGO_ENABLED=0
 
+BINDIR:=/usr/bin/
+
+build: out
+	@echo "> Building..."
+	$(GOCMD) build -ldflags "-s -w -X 'github.com/buildpacks/pack.Version=${PACK_VERSION}' -extldflags ${LDFLAGS}" -trimpath -o ./out/$(PACK_BIN) -a ./cmd/pack
```
jromero, line 54: I'm assuming this has to do with the build tools executing `make` by default so it `build` needs to be the first target. Is there no way of tweaking it? If it's important that this be first, we should comment on it.

.github/workflows/deliver-ubuntu.yml
```diff
@@ -0,0 +1,110 @@
...
+            if: matrix.target == 'focal'
+            uses: docker://ubuntu:20.04
+            env:
+              PACKAGE_NAME: "pack-cli"
+            with:
+              entrypoint: .github/workflows/delivery/ubuntu/test-install-package.sh
+
+          - name: Upload
+            env:
+              PPA_DEST: "ppa:thorntondwt/testppa"
```
jromero, line 106: We'll want to change this before merging it in.

.github/workflows/scripts/create_ppa.sh
jromero: Additionally, we'll want to add these values to LastPass.

jromero (changes requested):

Makefile
dwillist: We can tweak this, but it means we have to add a lot of complexity to the `/.github/workflows/delivery/ubuntu/rules` file. To manually run other build tools. Figured this was a work while trade off.

Makefile
dwillist: Not `needed` but I figure that users want `pack` on the path after running and `apt-get install`.

.github/workflows/scripts/create_ppa.sh
dwillist: Should already be present in lastpass

Makefile
jromero: Ok, so it is _somewhat_ necessary for `apt-get` to work the way we want it to. 

How does it know to run this target? Is there a convention or can we name it something else? Are there some docs or a place we use it in our workflow? Those are the things I think should be answer/noted in the target.

Makefile
jromero: That makes sense. Can we add some notes about it on the target?

For example, "# NOTE: this target must be listed first in order for it to be the default target so that tools like `apt-get` can build it without additional parameters".

Makefile
```diff
@@ -47,18 +47,27 @@ endif
...
+BINDIR:=/usr/bin/
+
+#
+build: out
+	@echo "> Building..."
+	$(GOCMD) build -ldflags "-s -w -X 'github.com/buildpacks/pack.Version=${PACK_VERSION}' -extldflags ${LDFLAGS}" -trimpath -o ./out/$(PACK_BIN) -a ./cmd/pack
+
 all: clean verify test build
 
+# used by ap-get install when installing ubuntu ppa.
```
jromero, line 59:
```suggestion
# used by apt-get install when installing ubuntu ppa.
```

jromero (approved):

jromero:
Left a few nitpicky things. Other than that this is great 👍 ✅. Feel free to merge in.

.github/workflows/deliver-ubuntu.yml
```diff
@@ -0,0 +1,110 @@
+name: delivery / ubuntu
+
+on:
+  release:
+    types:
+      - released
```
dfreilich, line 6: Could we add mechanism for triggering it with a workflow-dispatch, similar to what we have here: https://github.com/buildpacks/pack/blob/1bd3718c7479bc85b59d584108c056055024be16/.github/workflows/delivery-docker.yml#L7-L42?

.github/workflows/deliver-ubuntu.yml
```diff
@@ -0,0 +1,110 @@
...
+            - target: focal
+              runner: ubuntu-20.04
+              image: docker://ubuntu:20.04
+      name: create-ppa
+      runs-on:  ${{ matrix.runner }}
+
+      steps:
+          - name: Checkout code
+            uses: actions/checkout@v2
+          - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/* || true
```
dfreilich, line 30: Why is this necessary? Isn't this fetched by actions/checkout? https://github.com/actions/checkout#usage

.github/workflows/deliver-ubuntu.yml
```diff
@@ -0,0 +1,110 @@
...
+              ARCH: "any"
+              PACKAGE_NAME: "pack-cli"
+              UBUNTU_VERSION:  ${{ matrix.target }}
+              HOMEPAGE: "https://buildpacks.io"
+              PACK_VERSION: ${{ steps.metadata.outputs.version }}
+              REPO: "buildpacks/pack"
+              DESCRIPTION: "CLI for building apps using Cloud Native Buildpacks"
+              DATE_TIME: ${{ steps.metadata.outputs.date }}
+              MAINTAINER_NAME: "cncf-buildpacks"
+              MAINTAINER_EMAIL: "cncf-buildpacks-maintainers@lists.cncf.io"
```
dfreilich, line 68: nit: Can we put the ones that are static in the files, to reduce the complexity?

.github/workflows/deliver-ubuntu.yml
```diff
@@ -0,0 +1,110 @@
...
+
+          - name: Create PPA
+            env:
+              PACKAGE_NAME: "pack-cli"
+              PACKAGE_VERSION: ${{ steps.metadata.outputs.version }}
+              MAINTAINER: "cncf-buildpacks"
+              MAINTANER_EMAIL: "cncf-buildpacks-maintainers@lists.cncf.io"
+              GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+              GPG_PUBLIC_KEY: ${{ secrets.GPG_PUBLIC_KEY }}
+            run: ./.github/workflows/scripts/create_ppa.sh
```
dfreilich, line 78: nit: I feel like the other commands had the command defined before the envs. Any reason you didn't want to do that here? You thought it would improve readability?

.github/workflows/deliver-ubuntu.yml
```diff
@@ -0,0 +1,110 @@
...
+            if: matrix.target == 'bionic'
+            uses: docker://ubuntu:18.04
+            env:
+              PACKAGE_NAME: "pack-cli"
+            with:
+              entrypoint: .github/workflows/delivery/ubuntu/test-install-package.sh
+
+          - name: Test Focal
+            if: matrix.target == 'focal'
+            uses: docker://ubuntu:20.04
```
dfreilich, line 98: can you use `matrix.runner` here, to simplify it if we want to change them? or is that not possible in `uses:`?

dfreilich (approved):
super

A few nits and comments, but nothing that would this back for me, though maybe room for a future clean-up PR.

.github/workflows/deliver-ubuntu.yml
dwillist: Ya not possible to dynamically set the `uses:` fields :(

dfreilich: cry

.github/workflows/deliver-ubuntu.yml
dwillist: Ya its not, looks like we get the release tag via other methods in our other jobs, copying that format here.

.github/workflows/deliver-ubuntu.yml
dwillist: Could do, but these will be useful to help debug when this job fails.

PR #887, opened 2020-10-06
Add output format option to inspect-builder command

### LINKED ISSUE #387: pack inspect-builder --json/yaml/toml
I'd love to see `pack inspect-builder --format <json/yaml/toml>` so we could machine parse the output (of commands generally)

simonjjones:

## Summary
Add output format option to inspect-builder command

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #387


dwillist:
One quirk about this approach is that the `json` `yaml` and `toml` representations are not all equivalent due to differences in the underlying `json`, `yaml` and `toml` libraries. This could be corrected but will require structured outputs to go though multiple rounds of reading/writing `output` -> re-read -> `output in new format`

builder/detection_order.go
```diff
@@ -0,0 +1,18 @@
+package builder
+
+import (
+	"github.com/buildpacks/pack/internal/dist"
+)
+
+type DetectionOrderEntry struct {
+	dist.BuildpackRef
```
jromero, line 8:
```suggestion
	dist.BuildpackRef `yaml:",inline"`
```

^ this fixes the output so that there is no empty `buildpackref: {}`

builder/detection_order.go
```diff
@@ -0,0 +1,18 @@
...
+	Cyclical            bool           `json:"cyclic,omitempty" yaml:"cyclic,omitempty" toml:"cyclic,omitempty"`
+	GroupDetectionOrder DetectionOrder `json:"nested_buildpacks,omitempty" yaml:"nested_buildpacks,omitempty" toml:"nested_buildpacks,omitempty"`
```
jromero, line 10:
```suggestion
	GroupDetectionOrder DetectionOrder `json:"buildpacks,omitempty" yaml:"buildpacks,omitempty" toml:"buildpacks,omitempty"`
```

The term "nested" doesn't make sense when it's a top-level group. This proposed general term works well in both cases IMO.

After:

```json
{
  "detection_order": [
      {
        "buildpacks": [
          {
            "id": "samples/java-maven",
            "version": "0.0.1"
          }
        ]
      },
      {
        "buildpacks": [
          {
            "id": "samples/hello-universe",
            "version": "0.0.1",
            "buildpacks": [
              {
                "id": "samples/hello-world",
                "version": "0.0.1"
              },
              {
                "id": "samples/hello-moon",
                "version": "0.0.1"
              }
            ]
          }
        ]
      },
  ]
}
```

internal/dist/buildpack.go
```diff
@@ -40,9 +40,9 @@ type Buildpack interface {
 }
 
 type BuildpackInfo struct {
-	ID       string `toml:"id" json:"id,omitempty"`
-	Version  string `toml:"version" json:"version,omitempty"`
-	Homepage string `toml:"homepage,omitempty" json:"homepage,omitempty"`
+	ID       string `toml:"id" json:"id,omitempty" yaml:"id,omitempty"`
+	Version  string `toml:"version" json:"version,omitempty" yaml:"version,omitempty"`
+	Homepage string `toml:"homepage,omitempty" json:"homepage,omitempty" yaml:"homepage,omitempty"`
```
jromero, line 45:
```suggestion
	ID       string `toml:"id,omitempty" json:"id,omitempty" yaml:"id,omitempty"`
	Version  string `toml:"version,omitempty" json:"version,omitempty" yaml:"version,omitempty"`
	Homepage string `toml:"homepage,omitempty" json:"homepage,omitempty" yaml:"homepage,omitempty"`
```

testhelpers/assertions.go
```diff
@@ -122,6 +137,155 @@ func (a AssertionManager) Contains(actual, expected string) {
...
+
+	var expected interface{}
+	err = toml.Unmarshal([]byte(expectedTOML), &expected)
+	if err != nil {
+		a.testObject.Fatalf(
+			"Unable to unmarshal 'expectedTOML': %q", err,
+		)
+	}
+
+	// TODO: convert to a toml based diff. Unfortunately marshalling toml
```
jromero, line 255: Complete TODO.

testhelpers/comparehelpers/deep_compare.go
```diff
@@ -0,0 +1,106 @@
...
+	v2 := reflect.ValueOf(containee)
+
+	return deepContains(v1, v2, 0)
+}
+
+func deepContains(v1, v2 reflect.Value, depth int) bool {
+	if depth > 200 {
+		panic("deep Contains depth exceeded, likely a circular reference")
+	}
+	if !v1.IsValid() || !v2.IsValid() { // This is an odd case TODO: when does this happen??
```
jromero, line 37: Complete TODO.

testhelpers/comparehelpers/deep_compare_test.go
```diff
@@ -0,0 +1,155 @@
...
+			containerJSON := `[
+	{"Name": "Platypus", "Order": "Monotremata"},
+	{"Name": "Quoll",    "Order": "Dasyuromorphia"}
+]`
+			//containerJSON := `["a"]`
+			var container interface{}
+			assert.Succeeds(json.Unmarshal([]byte(containerJSON), &container))
+
+			containedJSON := `[{"Name": "Notapus", "Order": "Monotremata"}]`
+			//containedJSON := `["a"]`
```
jromero, line 146:
```suggestion
```

testhelpers/comparehelpers/deep_compare_test.go
```diff
@@ -0,0 +1,155 @@
...
+			})
+		})
+	})
+	when("json is not contained", func() {
+		it("return false", func() {
+			containerJSON := `[
+	{"Name": "Platypus", "Order": "Monotremata"},
+	{"Name": "Quoll",    "Order": "Dasyuromorphia"}
+]`
+			//containerJSON := `["a"]`
```
jromero, line 141:
```suggestion
```

jromero (changes requested):
Gave it a high-level pass and some manual testing. Overall, I like how this turned out. A bit of clean up requested.

acceptance/acceptance_test.go
```diff
@@ -1678,9 +1682,9 @@ exclude = [ "*.sh", "secrets/", "media/metadata" ]
 									"--buildpack", buildpackTgz,
 									"--descriptor", excludeDescriptorPath,
 								)
-								assert.NotContains(output, "api_keys.json")
-								assert.NotContains(output, "user_token")
-								assert.NotContains(output, "test.sh")
+								assert.NotContain(output, "api_keys.json")
```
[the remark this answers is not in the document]
simonjjones: FWIW, I'm in a similar situation of not feeling too strongly. I believe the method was added in divergent branches and I had to pick one when bringing everything back together, so chose the one I had added. Happy to converge on either.

internal/commands/inspect_builder.go
```diff
@@ -1,430 +1,83 @@
...
-	buildpacksTabWidth = 8
-	defaultTabWidth    = 4
-	writerPadChar      = ' '
-	writerFlags        = 0
-	none               = "(none)"
-)
+type BuilderWriter interface {
+	Print(
+		logger logging.Logger,
+		localRunImages []config.RunImage,
```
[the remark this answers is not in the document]
simonjjones: I believe these are run images defined in the system's config, rather than either the local or remote builder. It feels distinct from just `runImages` which would be the superset of run images from all possible sources. In this context there is no correlated concept of a `remote` run image. I think a different term could be helpful to avoid overloading the term `local` in this context, but it would be useful to continue to reference these as a subset of run images, and imply their source as being local config.

testhelpers/assertions.go
dwillist: Think this one might be a bit out of scope, for now. But should be addressed when there is a `toml` library that supports uninstalling into an arbitrary `interface`. Should I just remove the `TODO` part of this?

internal/builder/inspect.go
```diff
@@ -0,0 +1,147 @@
...
+
+type InspectableFetcher interface {
+	Fetch(ctx context.Context, name string, daemon bool, pullPolicy config.PullPolicy) (Inspectable, error)
+}
+
+type LabelManagerFactory interface {
+	BuilderLabelManager(inspectable Inspectable) LabelInspector
+}
+
+type LabelInspector interface {
```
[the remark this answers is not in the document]
dwillist: Could you give some additional clarification on this point? Each of the methods `Metadata` `StackID`, etc indicates the  type of info being retrieved.

testhelpers/assertions.go
jromero: The TODO comment didn't make too much sense to me to begin with. Based on your comment, I would agree that it's probably not something worth having a TODO. I'm all for deleting it.

internal/builder/inspect.go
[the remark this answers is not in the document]
jromero: My initial instinct was to propose something like `BuilderInspector` since it's the information of a builder that you are trying to inspect as opposed to where the data is stored (Label). Upon further inspection, no pun intended, I do see that it's part of the builder package so it's less ideal to have `builder.BuilderInspector` vs `builder.LabelInspector` and `builder.Inspector` is already taken. There is probably still a bit of clean up that can be done here but given that this is an internal package and it's simply semantics I'm okay with pushing this forward as is.

jromero (approved):

jromero:
@dwillist this is approved ✅ . If you can resolve the conflicts and get it passing in CI then please feel free to merge.

inspect_builder_test.go
```diff
@@ -336,33 +342,62 @@ func testInspectBuilder(t *testing.T, when spec.G, it spec.S) {
...
+												BuildpackInfo: dist.BuildpackInfo{ID: "test.bp.two"},
+												Optional:      true,
+											},
+										},
+									},
+								},
+							}
+
+							if diff := cmp.Diff(want, builderInfo.Order); diff != "" {
+								t.Errorf("InspectBuilder() Info.Oerswe miamrxh (-want +got):\b%s", diff)
```
dfreilich, line 392: What does this error message mean?

dwillist: woah, ya that should be in english.

PR #892, opened 2020-10-10
Remove support for `--no-pull` flag

### LINKED ISSUE #775: Remove support for `--no-pull` flag
### Description

https://github.com/buildpacks/rfcs/pull/80 proposed adding in a `pull-policy` flag for `pack build`, `pack create-builder`, `pack package-buildpack`, and `pack rebase`, with `pull-policy never` being equivalent to `--no-pull`. That is being implemented in #773. 

Having duplicated flags makes a confusing UX experience. 

### Proposed solution

Once we have released pack `v0.13.0`, we should remove the `--no-pull` flag.

anshlykov:
Resolves #775


internal/commands/build_test.go
```diff
@@ -136,32 +134,9 @@ func testBuildCommand(t *testing.T, when spec.G, it spec.S) {
...
-		when("--no-pull", func() {
-			it("sets pull-policy=never and logs warning", func() {
-				mockClient.EXPECT().
-					Build(gomock.Any(), EqBuildOptionsWithPullPolicy(pubcfg.PullNever)).
-					Return(nil)
-
-				command.SetArgs([]string{"image", "--builder", "my-builder", "--no-pull"})
-				h.AssertNil(t, command.Execute())
-				h.AssertContains(t, outBuf.String(), "Warning: Flag --no-pull has been deprecated")
-			})
```
dfreilich, line 148: It looks like we aren't testing any successful paths for setting `pull-policy never` at this point. Would you be able to port this test to the `when("--pull-policy...` block, replacing the `--no-pull` here with `--pull-policy never` and removing the Warning assertion (and keeping the `EqBuildOptionsWithPullPolicy()` func)?

package_buildpack_test.go
```diff
@@ -197,7 +197,7 @@ func testPackageBuildpack(t *testing.T, when spec.G, it spec.S) {
 				return img
 			}
 
-			when("publish=false and no-pull=false", func() {
+			when("publish=false and pull-policy=always", func() {
```
dfreilich, line 200: Nice catch!

internal/commands/build.go
```diff
@@ -160,9 +159,6 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
 	cmd.Flags().StringArrayVar(&buildFlags.Volumes, "volume", nil, "Mount host volume into the build container, in the form '<host path>:<target path>[:<mode>]'."+multiValueHelp("volume"))
 	cmd.Flags().StringVarP(&buildFlags.DefaultProcessType, "default-process", "D", "", `Set the default process type. (default "web")`)
 	cmd.Flags().StringVar(&buildFlags.Policy, "pull-policy", "", `Pull policy to use. Accepted values are always, never, and if-not-present. (default "always")`)
-	// TODO: Remove --no-pull flag after v0.13.0 released. See https://github.com/buildpacks/pack/issues/775
-	cmd.Flags().BoolVar(&buildFlags.NoPull, "no-pull", false, "Skip pulling builder and run images before use")
-	cmd.Flags().MarkHidden("no-pull")
 }
 
 func validateBuildFlags(flags *BuildFlags, cfg config.Config, packClient PackClient, logger logging.Logger) error {
```
dfreilich, line 164: Can we remove the `logger` from this func now, given that it looks like it's not used anymore?

dfreilich (changes requested):
Thanks for the help with this! Overall, it's great, just a few small fixes

internal/commands/build.go
anshlykov: The `logger` is still using in this function as the parameter for the `suggestSettingBuilder` func.
https://github.com/buildpacks/pack/blob/7268c89ac5a6bbc6f843f1fdddc0ae598ccf54bd/internal/commands/build.go#L168-L172

internal/commands/build_test.go
anshlykov: done

jromero (approved):

jromero:
Thank you @anshlykov for the contribution!

PR #901, opened 2020-10-20
Update lifecycle to 0.9.1

jromero:
## Summary

This PR updates lifecycle library and it's dependencies. 

## Related

Resolves #888


jromero:
More docker dependency woes similar to [this one](https://github.com/buildpacks/pack/pull/895#issuecomment-713069370) but in this case related to Windows. 

```
github.com/docker/docker/pkg/system (C:\\Users\\Admin\\go\\pkg\\mod\\github.com\\docker\\docker@v17.12.0-ce-rc1.0.20200
916142827-bd33bbf0497b+incompatible\\pkg\\system\\filesys_windows.go:111:26: cannot use 
uintptr(unsafe.Pointer(&sd[0])) (value of type uintptr) as *windows.SECURITY_DESCRIPTOR value in assignment))))"
```


Using `go get -u github.com/docker/docker@master` may work but we would need to explore the potential risks.

CC: @ekcasey @micahyoung

PR #902, opened 2020-10-20
Handle more types of naked locators

jkutner:
## Summary

Naked locators are locator strings that do not have a prefix that
matches any known prefixes (like `docker://` or `https://`). When presented
with one of these, pack must try to determine whether the user intended
for the locator to reference a file, id, or docker image.

This topic was discussed as part of https://github.com/buildpacks/rfcs/pull/56#discussion_r388716172 but we didn't resolve the order as part of the RFC.

The ability to handle more types of naked locators is important for [RFC-0022: Client Side Buildpack Registry](https://github.com/buildpacks/rfcs/blob/main/text/0022-client-side-buildpack-registry.md)

## Output

N/A

## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

N/A


internal/buildpack/locator_type.go
```diff
@@ -74,8 +81,14 @@ func GetLocatorType(locator string, buildpacksFromBuilder []dist.BuildpackInfo)
 		return IDLocator, nil
 	}
 
-	if _, err := name.ParseReference(locator); err == nil {
-		return PackageLocator, nil
+	if strings.ContainsAny(locator, "@sha") || strings.Count(locator, "/") > 1 {
```
jromero, line 84: We should add some tests for these changes.

jromero (changes requested):

internal/buildpack/locator_type.go
jkutner: whoops, forgot to add them to the commit. The existing tests mostly cover it

jromero:
It appears as if there are various failures in the acceptance tests for both Linux and Windows platforms. It would be interesting to know exactly what the issue/impact from this change may be...

jkutner:
@jromero yea, if you have any advice on how to get some output from CI that would be great. I haven't been able to figure this out locally, so I'll need to fire up a Windows laptop.

jromero:
The logs give you a decent amount of information. If you view the results, click on the "test" job you should be able to see the full logs. Additionally, I typically search for "FAIL:" to see what tests failed. For acceptance, due to how `go test` works, the actual error output may be higher in the logs since the results are collected and printed after all of the exectution.

For example: https://github.com/buildpacks/pack/pull/902/checks?check_run_id=1289728328

```
=== CONT  TestAcceptance/acceptance_suite/p_current_cb_current_lc_default/stack_is_created/builder_is_created/rebase/--publish/--run-image/uses_provided_run_image
    acceptance_test.go:2235: creating builder image...
    acceptance_test.go:2269: creating package image from: testdata/mock_buildpacks/0.2
    acceptance_test.go:2280: adding lifecycle version '0.9.1' to builder config
    acceptance_test.go:2303: expected nil: exit status 1: Pulling image 'localhost:32770/pack-test/run'
        latest: Pulling from pack-test/run
        Digest: sha256:f1981c8abc7c3618299b2eeda6de95d57cb3756f9af4cfb215f2a7fe62959cce
        Status: Image is up to date for localhost:32770/pack-test/run:latest
        Creating builder 'localhost:32770/test/builder-zlyzkferqm' from build-image 'pack-test/build'
        Downloading from 'https://github.com/buildpacks/lifecycle/releases/download/v0.9.1/lifecycle-v0.9.1+linux.x86-64.tgz'
        0 B/5.2 MB
16.4 KB/5.2 MB
5.2 MB/5.2 MB 
        Looking up buildpack 'read/env@read-env-version'
        Downloading buildpack from URI: 'file:///tmp/create-test-builder163934492/read-env-buildpack.tgz'
        Looking up buildpack ''
        Downloading buildpack from URI: 'file:///tmp/create-test-builder163934492/noop-buildpack.tgz'
        Looking up buildpack ''
        Downloading buildpack from URI: 'file:///tmp/create-test-builder163934492/noop-buildpack-2.tgz'
        Looking up buildpack ''
        Warning: The 'image' key is deprecated. Use 'uri="docker://..."' instead.
        Downloading buildpack from registry: ''
        Refreshing registry cache for github.com//buildpacks/registry-index
        Creating registry cache for github.com//buildpacks/registry-index
        Validating registry cache for github.com//buildpacks/registry-index
        Creating registry cache for github.com//buildpacks/registry-index
        ERROR: failed to add buildpacks to builder: locating in registry '': parsing buildpacks registry id: invalid registry ID: 
```

internal/buildpack/locator_type.go
```diff
@@ -74,8 +81,14 @@ func GetLocatorType(locator string, buildpacksFromBuilder []dist.BuildpackInfo)
...
-		return PackageLocator, nil
+	if strings.Contains(locator, "@sha") || strings.Count(locator, "/") > 1 {
+		if _, err := name.ParseReference(locator); err == nil {
+			return PackageLocator, nil
+		}
+	}
+
+	if strings.Count(locator, "/") == 1 {
+		return RegistryLocator, nil
 	}
```
dfreilich, line 92: If this is all trying to parse a naked locator, can we pull it out to a separate func, to make that a bit more clear?

jkutner: addressed in d127f425

create_builder.go
```diff
@@ -239,6 +239,9 @@ func (c *Client) addBuildpacksToBuilder(ctx context.Context, opts CreateBuilderO
 				return errors.Wrapf(err, "extracting from registry %s", style.Symbol(b.URI))
 			}
 		case buildpack.URILocator:
+			if b.URI == "" {
```
jromero, line 242: nit: I see that we get `locator` from `b.URI` in line 203, then we mutate it based on the deprecated key. To make this more comprehensible couldn't we do one of the following: set `b.URI` to locator in line 206 OR use `locator` only after line 203.

jromero (approved):
Left a minor comment but otherwise looks great! Feel free to merge.

create_builder.go
jkutner: I wouldn't want to do the first recommendation, because we may only want to replace `b.URI` _if_ we know the locator is a `URILocator`. I opted not to use `locator` only after line 243 because it would break the encapsulation provided by the `b` object (which I believe is of `BuildpackConfig` type).

So there's tradeoffs all around, but i think this is best approach.

PR #926, opened 2020-10-27
Move package `os` to be configured via `package.toml`

aemengo:
## Summary
The following PR is intended as an addendum to https://github.com/buildpacks/pack/pull/840, which enables `pack package-buildpack` for windows. It removes the recently added `--os` parameter in favor a **platform.os** declaration in one's `package.toml` . This new declaration is **optional**: defaulting to linux when omitted, and thus required for specifying windows. The intention is a more consistent (and backward compatible) user experience.

## Output

```bash
$ cat package.toml
[buildpack]
uri = "simple-layers-buildpack.tgz"

[platform]
os = "windows"

$ pack package-buildpack /tmp/package.tgz -f file -c package.toml
Successfully created package /tmp/package.tgz
```

## Documentation

Should this change be documented?

[x] Yes, see https://github.com/buildpacks/docs/pull/242

## Related
- https://github.com/buildpacks/pack/pull/840

cc: @micahyoung @TisVictress @ameyer-pivotal

### DIFF
--- a/buildpackage/config_reader.go
+++ b/buildpackage/config_reader.go
@@ -16,6 +16,7 @@ import (
 type Config struct {
 	Buildpack    dist.BuildpackURI `toml:"buildpack"`
 	Dependencies []dist.ImageOrURI `toml:"dependencies"`
+	Platform     dist.Platform     `toml:"platform"`
 }
 
 // NewConfigReader returns an instance of ConfigReader. It does not take any parameters.
@@ -49,6 +50,10 @@ func (r *ConfigReader) Read(path string) (Config, error) {
 
 	if packageConfig.Buildpack.URI == "" {
 		return packageConfig, errors.Errorf("missing %s configuration", style.Symbol("buildpack.uri"))
+	}
+
+	if packageConfig.Platform.OS == "" {
+		packageConfig.Platform.OS = "linux"
 	}
 
 	configDir, err := filepath.Abs(filepath.Dir(path))
--- a/buildpackage/config_reader.go
+++ b/buildpackage/config_reader.go
@@ -56,6 +56,11 @@ func (r *ConfigReader) Read(path string) (Config, error) {
 		packageConfig.Platform.OS = "linux"
 	}
 
+	if packageConfig.Platform.OS != "linux" && packageConfig.Platform.OS != "windows" {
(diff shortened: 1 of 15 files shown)


buildpackage/config_reader.go
```diff
@@ -51,6 +52,10 @@ func (r *ConfigReader) Read(path string) (Config, error) {
 		return packageConfig, errors.Errorf("missing %s configuration", style.Symbol("buildpack.uri"))
 	}
 
+	if packageConfig.Platform.OS == "" {
```
ameyer-pivotal, line 55: One pattern I've seen for setting defaults is to set it on the object *before* decoding. Then it will be retained after decoding if nothing overwrote it.

package_buildpack.go
```diff
@@ -144,9 +132,11 @@ func (c *Client) PackageBuildpack(ctx context.Context, opts PackageBuildpackOpti
 
 	switch opts.Format {
 	case FormatFile:
-		return packageBuilder.SaveAsFile(opts.Name, opts.OS)
+		return packageBuilder.SaveAsFile(opts.Name, opts.Config.Platform.OS)
 	case FormatImage:
-		_, err = packageBuilder.SaveAsImage(opts.Name, opts.Publish, opts.OS)
+		//TODO: should check if saving linux image on windows daemon ???
```
ameyer-pivotal, line 137: It would fail already, but if you wanted a friendlier message then checking here would be a good route

buildpackage/config_reader.go
```diff
@@ -16,6 +16,7 @@ import (
 type Config struct {
 	Buildpack    dist.BuildpackURI `toml:"buildpack"`
 	Dependencies []dist.ImageOrURI `toml:"dependencies"`
+	Platform     dist.Platform     `toml:"platform"`
```
ameyer-pivotal, line 19: Do you think we should have some extra validation that the contents of the `platform.os` is only either `windows` or `linux`, and anything else produces an error?

micahyoung (approved):
LGTM. Checks all the boxes for me, based on the Slack conversation: https://buildpacks.slack.com/archives/CD61YAG69/p1603485561046900

jromero (approved):

jromero:
💯 @aemengo. Thanks for driving this through!

PR #927, opened 2020-10-28
Update github actions

dfreilich:
## Summary

This PR updates a number of things with out delivery github actions:
- Bumps versions of actions/download-artifact and actions/upload-artifact (closing #879 and closing #881) (tested [here](https://github.com/dfreilich/pack/runs/1320723996?check_suite_focus=true))
- Uses the new form of setting env vars, due to the deprecation of `set-env` discussed [here](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)
- Uses dfreilich/pack-action@v1 for delivery/docker, after the merging of https://github.com/dfreilich/pack-action/pull/9 (tested [here](https://github.com/dfreilich/pack/actions/runs/333721398))
- Adds `workflow dispatch` trigger for chocolatey delivery (tested [here](https://github.com/dfreilich/pack/actions/runs/333454542))

## Documentation

    - [x] No

## Related

Resolves #879 
Resolves #881


.github/workflows/delivery-docker.yml
```diff
@@ -35,27 +35,24 @@ jobs:
             return tag.replace(/^v/, '');
       - name: Checkout workflow
         uses: actions/checkout@v2
-      - name: Checkout source at tag
-        uses: actions/checkout@v2
         with:
-          path: ./src
```
jromero, line 41: I wonder if we'd still like to keep this. Not the local action (`./.github/workflows/actions/pack`), but the ability to update the workflow when re-running a release via `workflow_dispatch`. For example, we can update `.github/workflows/delivery/docker/buildpack.yml` if we need to without needing to bump our release version.

.github/workflows/delivery-chocolatey.yml
```diff
@@ -4,25 +4,49 @@ on:
...
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
-  prepare-chocolatey:
-    runs-on: ubuntu-latest
+  deliver-chocolatey:
+    runs-on: windows-latest
```
jromero, line 22: Was this tested on Windows? IIRC, the default shell would be powershell in which case I'm not sure the `run` commands would work as expected.

PS: There is a way to change the default shell: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun

.github/workflows/delivery-chocolatey.yml
```diff
@@ -4,25 +4,49 @@ on:
...
+      tag_name:
+        description: The release tag to distribute
+        required: true
 
 env:
   CHOCO_PATH: chocolatey
 
+defaults:
+  run:
+    shell: bash
```
dfreilich, line 18: > Was this tested on Windows? IIRC, the default shell would be powershell in which case I'm not sure the run commands would work as expected. (from https://github.com/buildpacks/pack/pull/927/files#r514303529)

@jromero I'm using the default shell set to `bash` over here, which is why it works.

.github/workflows/delivery-docker.yml
dfreilich: We would still need to bump our release version in order to get a change to `buildpack.yml`, though – because it copies from the `src` dir, which is grabbed by the git ref? Or am I misunderstanding this?

.github/workflows/delivery-docker.yml
```diff
@@ -33,29 +33,30 @@ jobs:
...
 
             return tag.replace(/^v/, '');
-      - name: Checkout workflow
-        uses: actions/checkout@v2
-      - name: Checkout source at tag
+      - name: Checkout delivery configuration
         uses: actions/checkout@v2
         with:
           path: ./src
+      - name: Checkout workflow
```
jromero, line 40: Was this not clearer as "Checkout source at tag"?  workflow and delivery config sound to be the same.

dfreilich: I reverted that. Is it better to check out the src to `/src` and the configuration to `/`, or to have the source be the home, while the configuration is `/somewhere`?

.github/workflows/delivery-docker.yml
```diff
@@ -33,29 +33,30 @@ jobs:
             }
 
             return tag.replace(/^v/, '');
-      - name: Checkout workflow
+      - name: Checkout delivery configuration
         uses: actions/checkout@v2
+        with:
+          path: ./main
```
jromero, line 39:
```suggestion
          path: ./delivery
```

Sorry, I'm being a little pedantic here but if `main` is suppose to represent what branch is being checked out then I think that's incorrect. In the GHA UI you are able to specify what branch you would like to run as your `HEAD`. 

I suggested "delivery" but may "head"...? idk.

dfreilich: I think head instead, because that's what it's doing - it's the head of the branch you are using

jromero (approved):
improvement!

PR #928, opened 2020-10-28
Add codeql analysis

dfreilich:
## Summary

This PR enables CodeQL code scanning for the repository, and any PRs to `main` or `release` branches. It could be helpful in ensuring that common mistakes are nipped at the bud. 

## Documentation

    - [x] No


jromero:
Very interesting idea! I have two points I'd like to simple log:

1. This is an experimental integration. The outcome is yet to be determined. If we run into any issues we should be able to easily disable it.
1. If we decide to keep this integration, we may want to provide a simple way of running the analysis locally. According to the docs there is a [CLI](https://help.semmle.com/codeql/codeql-cli.html).

jromero (approved):

PR #932, opened 2020-10-30
Use a default package config for a buildpackage 

jkutner:
## Summary

This change causes pack to create a default configuration when no `package.toml` file is provided to the `package-buildpack` command. The default is the equivalent of:

```toml
[buildpack]
uri = "."
```

This is important for buildpack authors who want to publish to the Buildpack Registry. In order to publish, you must create a buildpackage, but learning about the `package.toml` file in order to share your buildpack is an unnecessary hurdle for a simple buildpack.

#### Before

```
$ pack package-buildpack my-cnb
ERROR: Please provide a package config path, using --config
```

#### After

If the buildpack is a normal buildpack, it just works:
```
$ pack package-buildpack my-cnb
Successfully created package my-cnb
```

If the buildpack is a meta-buildpack it will create an error:

```
$ pack package-buildpack my-meta-cnb
ERROR: saving image: buildpack jkutner/minecraft@0.1.0 references buildpack heroku/jvm@0.1 which is not present
```

## Documentation

    - [x] Yes, see #TBD
    - [ ] No

### DIFF
--- a/buildpackage/config_reader.go
+++ b/buildpackage/config_reader.go
@@ -12,11 +12,24 @@ import (
 	"github.com/buildpacks/pack/internal/style"
 )
 
+const defaultOS = "linux"
+
 // Config encapsulates the possible configuration options for buildpackage creation.
 type Config struct {
 	Buildpack    dist.BuildpackURI `toml:"buildpack"`
 	Dependencies []dist.ImageOrURI `toml:"dependencies"`
 	Platform     dist.Platform     `toml:"platform"`
+}
+
+func DefaultConfig() Config {
+	return Config{
+		Buildpack: dist.BuildpackURI{
+			URI: ".",
+		},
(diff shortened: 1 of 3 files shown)


dfreilich (approved):
I really like this. I think it helps the UX a lot, and it should make it more similar to `pack build`, which builds the application dir you are in at the moment. 

plus one

dfreilich (changes requested):
Unfortunately, UA for it fails, due to the changes from #926. 
Attempting to use it results in:
```
$  ~/workspace/pack/out/pack package-buildpack hello-moon
ERROR: invalid platform.os specified: DOCKER_OS is linux
```

jkutner:
@dfreilich fixed and rebased. thanks

jromero (approved):

PR #935, opened 2020-11-05
Add pull-buildpack command

jkutner:
## Summary

The `pull-buildpack` command accepts a buildpack URI as an argument, in the form of a URL, CNB URN, docker tag, or buildpack ID and version in a registry. It then pulls the buildpackage, extracts it into a buildpack image if necessary, and stores it in the local daemon.

This command is useful for users who may need to run the `build` or `create-builder` command with `--pull-policy never`. They can run the `pull-buildpack` at an earlier time to load the buildpackage image into the local daemon.

## Output

```
$ pack pull-buildpack heroku/spring-boot@0.2.2
sha256:d5db24e9b2d1b2662e0712089c2934556e15d8280473d9d1bac46c3ef4c9d877: Pulling from heroku-examples/buildpacks/heroku_spring-boot
Digest: sha256:d5db24e9b2d1b2662e0712089c2934556e15d8280473d9d1bac46c3ef4c9d877
Status: Image is up to date for ghcr.io/heroku-examples/buildpacks/heroku_spring-boot@sha256:d5db24e9b2d1b2662e0712089c2934556e15d8280473d9d1bac46c3ef4c9d877
Successfully pulled heroku/spring-boot@0.2.2
```

## Documentation

    - [x] Yes, see #TBD
    - [ ] No


pull_buildpack.go
```diff
@@ -0,0 +1,80 @@
...
+			_, _, err = buildpackage.BuildpacksFromOCILayoutBlob(blob)
+			if err != nil {
+				return errors.Wrapf(err, "extracting buildpacks from %s", style.Symbol(opts.URI))
+			}
+		} else {
+			return errors.Errorf("buildpack is in an unsupported layout")
+		}
+	case buildpack.PackageLocator:
+		imageName := buildpack.ParsePackageLocator(opts.URI)
+		_, _, err := extractPackagedBuildpacks(ctx, imageName, c.imageFetcher, false, config.PullAlways)
```
jromero, line 56: Given that we don't need the buildpacks extracted is there a need to call `extractPackagedBuildpacks`? Seems like a call to the `Fetcher` should suffice:

```
        pkgImage, err := fetcher.Fetch(ctx, pkgImageRef, !publish, pullPolicy)
	if err != nil {
		return nil, nil, errors.Wrapf(err, "fetching image")
	}
```

pull_buildpack.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "invalid registry '%s'", opts.RegistryName)
+		}
+
+		registryBp, err := registryCache.LocateBuildpack(opts.URI)
+		if err != nil {
+			return errors.Wrapf(err, "locating in registry %s", style.Symbol(opts.URI))
+		}
+
+		_, _, err = extractPackagedBuildpacks(ctx, registryBp.Address, c.imageFetcher, false, config.PullAlways)
```
jromero, line 71: Ditto: https://github.com/buildpacks/pack/pull/935/files#r517773673

pull_buildpack.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (c *Client) PullBuildpack(ctx context.Context, opts PullBuildpackOptions) error {
+	locatorType, err := buildpack.GetLocatorType(opts.URI, []dist.BuildpackInfo{})
+	if err != nil {
+		return err
+	}
+
+	switch locatorType {
+	case buildpack.URILocator:
```
jromero, line 30: If I'm not mistaken this code path would download the files, for example `.cnb`, but wouldn't actually load them into the daemon.

Also, some tests that prove that the docker daemon has the images would be useful.

jromero (changes requested):
This is a great idea!

It looks like there are a few issues in the implementation but should be easy enough to address. Looking forward to the update.

pull_buildpack.go
```diff
@@ -0,0 +1,80 @@
...
+
+		if isOCILayout {
+			_, _, err = buildpackage.BuildpacksFromOCILayoutBlob(blob)
+			if err != nil {
+				return errors.Wrapf(err, "extracting buildpacks from %s", style.Symbol(opts.URI))
+			}
+		} else {
+			return errors.Errorf("buildpack is in an unsupported layout")
+		}
+	case buildpack.PackageLocator:
```
dfreilich, line 54: Can you add a test for this? I see a test for a `URI` and a registry BP, but none for a PackageLocator (unless I missed it)

pull_buildpack.go
jkutner: Actually, I realize now that we don't need to support `URILocator` in this command (pullPolicy doesn't apply to it, so there's not much of a use case for needing it). Give that saving the blob to the daemon would be a fairly new chunk of code, I'm going to omit this.

internal/commands/pull_buildpack.go
```diff
@@ -0,0 +1,45 @@
...
+
+func PullBuildpack(logger logging.Logger, cfg config.Config, client PackClient) *cobra.Command {
+	var opts pack.PullBuildpackOptions
+	var flags PullBuildpackFlags
+
+	cmd := &cobra.Command{
+		Use:     "pull-buildpack <uri>",
+		Args:    cobra.ExactArgs(1),
+		Short:   prependExperimental("Pull the buildpack and store it locally"),
+		Example: "pack pull-buildpack my-buildpack",
```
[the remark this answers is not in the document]
jkutner: Fixed in ce60779a

pull_buildpack.go
jkutner: Fixed in ce60779a

pull_buildpack.go
jkutner: Fixed in ce60779a

pull_buildpack.go
jkutner: The "pulling from a docker registry" test is for `PackageLocator`. I've removed the `URILocator` code path.

internal/commands/pull_buildpack.go
```diff
@@ -0,0 +1,45 @@
...
+
+func PullBuildpack(logger logging.Logger, cfg config.Config, client PackClient) *cobra.Command {
+	var opts pack.PullBuildpackOptions
+	var flags PullBuildpackFlags
+
+	cmd := &cobra.Command{
+		Use:     "pull-buildpack <uri>",
+		Args:    cobra.ExactArgs(1),
+		Short:   prependExperimental("Pull the buildpack and store it locally"),
+		Example: "pack pull-buildpack example/my-buildpack@1.0.0",
```
dfreilich, line 24: nit: Could you add a `Long:` description clarifying why people would use it, and what valid inputs look like?

dfreilich (approved):
Overall, looks great. Just a small comment on adding some of the information from the PR to the command, to make it a bit clearer for users.

jromero (approved):

dfreilich:
I'm going to rework this command as part of #962 to make it `pack buildpack pull`.

PR #936, opened 2020-11-05
Add Workflow Dispatch trigger to archlinux/homebrew pipelines

### LINKED ISSUE #875: Add manual execution capability to Arch Linux pack-cli delivery pipeline
### Description
As part of https://github.com/buildpacks/pack/pull/874, we added manual execution capabilities for **`pack-cli-bin`**. We should also enable it for the **`pack-cli`** job.

### Proposed solution

For `pack-cli`, we should seperate the workflow source from the pack source in the pipeline similar to delivery-docker. This needs to be done so that we can update the release workflow and be able to checkout the source code for pack at any given tag. As an example, we do this for [`delivery/docker`](https://github.com/buildpacks/pack/blob/main/.github/workflows/delivery-docker.yml#L39-L42)

dfreilich:
## Summary

This PR adds a [workflow_dispatch](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) trigger to archlinux and homebrew pipelines, to allow for manual triggering. 

I also took out any occurrence of `set-env` or `add-path` GHA commands, both of which were deprecated by Github (see [here](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) for details)
## Documentation

    - [x] Yes

## Related

Resolves #875

### DIFF
--- a/.github/workflows/delivery-archlinux-git.yml
+++ b/.github/workflows/delivery-archlinux-git.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   pack-cli-git:
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -49,8 +49,8 @@ jobs:
           go-version: '1.14'
       - name: Set up go env
         run: |
-          echo "::set-env name=GOPATH::$(go env GOPATH)"
-          echo "::add-path::$(go env GOPATH)/bin"
+          echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
         shell: bash
       - name: Acceptance
         env:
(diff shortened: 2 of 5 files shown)


.github/workflows/delivery-archlinux-git.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
```
dfreilich, line 7: This pipeline just picks up the code from main, and shouldn't need to be versioned at all. This just serves as a trigger to start the pipeline if something goes wrong.

dfreilich (commented):
For `archlinux (pack-cli)`, there shouldn't be a need to do anything else, given that it already just downloads the code from a given version:
```
url=https://github.com/buildpacks/pack/archive/v${{ env.PACK_VERSION }}.tar.gz
```
and passes that along to the archlinux build script.

jromero (approved):

Issue #58, opened 2018-10-17
Enumerating the stacks that a buildpack runs on seems dumb


dsyer:
Maybe this is all part of some plan, but isn't it kind of dumb to have to enumerate all the stacks that a buildpack can run on in `buildpack.toml`? What if there's a new stack which is epsilon different from bionic? Shouldn't my buildpack be able to run on that?

josegonzalez:
A buildpack should probably be vetted against the stack you are attempting to run it on. For instance, its possible that I create a buildpack that can be run against an Alpine Linux stack, but would not be able to run on an Ubuntu Bionic stack. In those cases, the Bionic stack would probably be considered an invalid configuration, and rather than increase support headaches, being able to unequivocally state compatibility is useful. 

Just my two cents :)

jchesterpivotal:
>  What if there's a new stack which is epsilon different from bionic? Shouldn't my buildpack be able to run on that?

Maybe, maybe not. There's no way for us to tell, short of being able to accurately forecast whether buildpack A will work on stack B as well as stack B+epsilon. That smells like the halting problem to me. More loosely it means providing heuristics which would misfire periodically and cause surprise and distress.

I think a better way to look at this is not that it's "a list of things the buildpack can run on", because as you point out, that's not strictly true. Instead it's "a list of things the buildpack author asserts they have checked", which might not be strictly true either, but perhaps less confusing as to scope.

sclevine:
Nothing prevents you from modifying the bionic run and build images and labeling them with `io.buildpacks.stacks.bionic`. This is how I imagine individual platform operators would make extra packages available to developers.

We should probably introduce another label for modified stacks though. That way we can validate that the correct run image is always used for rebase operations.

nebhale:
I view this as a way to indicate that a buildpack has been tested to work on a specified stack.  Given how reliant buildpacks are on the underlying stack below them, I lean towards being very conservative and certifying that a buildpack works on a given stack.  If we find that that can be loosened in the future, and that field can be used to generate warnings rather than outright enforcement errors, that might be a path, but I don't think we're that flexible today.

sclevine:
https://github.com/buildpack/spec/pull/18

sclevine:
Closing as mixins is merged into the spec.

PR #937, opened 2020-11-05
Add stack subcommand in place of suggest-stacks

### LINKED ISSUE #919: pack stack subcommand
### Description

Based on [RFC #58](https://github.com/buildpacks/rfcs/blob/main/text/0058-pack-subcommands.md), pack should have a subcommand `stack`.

### Proposed solution

The following existing commands should be mapped accordingly with all options preserved.

* `pack suggest-stacks ...` -> `pack stack suggest ...`

_**NOTE: Previous commands should be hidden and if used deprecation warning should be printed.**_

### Additional context

Relates to: https://github.com/buildpacks/pack/issues/597
Relates to: https://github.com/buildpacks/rfcs/pull/93 ([RFC](https://github.com/buildpacks/rfcs/blob/main/text/0058-pack-subcommands.md))

anshlykov:
## Summary

Based on RFC #58 and #919, I hid the old `suggest-stacks` command and added the `stack` subcommand.

## Documentation

    - [X] Yes, see #___
    - [ ] No

## Related

Resolves #919


internal/commands/stack/suggest.go
```diff
@@ -0,0 +1,82 @@
...
+	{
+		ID:          "io.paketo.stacks.tiny",
+		Description: "A tiny Paketo stack based on Ubuntu 18.04, similar to distroless",
+		Maintainer:  "Paketo Project",
+		BuildImage:  "paketobuildpacks/build:tiny-cnb",
+		RunImage:    "paketobuildpacks/run:tiny-cnb",
+	},
+}
+
+func newSuggestCmd(logger logging.Logger) *cobra.Command {
```
dfreilich, line 52:
```suggestion
func suggest(logger logging.Logger) *cobra.Command {
```
Our command pattern is typically to do the command name itself (e.g. `command.Build`), so I think this would fit the pattern a bit more.

internal/commands/stack/suggest_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+	"github.com/spf13/cobra"
+
+	"github.com/buildpacks/pack/internal/logging"
+	h "github.com/buildpacks/pack/testhelpers"
+)
+
+func TestSuggestStacksCommand(t *testing.T) {
```
dfreilich, line 15:
```suggestion
func TestStacksSuggestCommand(t *testing.T) {
```

internal/commands/stack/suggest_test.go
```diff
@@ -0,0 +1,61 @@
...
+	spec.Run(t, "Commands", testSuggestStacksCommand, spec.Parallel(), spec.Report(report.Terminal{}))
```
dfreilich, line 16:
```suggestion
	spec.Run(t, "StacksSuggestCommand", testSuggestStacksCommand, spec.Parallel(), spec.Report(report.Terminal{}))
```

internal/commands/stack/suggest_test.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func testSuggestStacksCommand(t *testing.T, when spec.G, it spec.S) {
```
dfreilich, line 19:
```suggestion
func testStacksSuggestCommand(t *testing.T, when spec.G, it spec.S) {
```

internal/commands/suggest_stacks_test.go
```diff
@@ -31,7 +31,8 @@ func testSuggestStacksCommand(t *testing.T, when spec.G, it spec.S) {
 		it("displays stack information", func() {
 			command.SetArgs([]string{})
 			h.AssertNil(t, command.Execute())
-			h.AssertEq(t, outBuf.String(), `Stacks maintained by the community:
+			h.AssertEq(t, outBuf.String(), "\x1b\x5b\x33\x33\x3b\x31\x6dWarning: \x1b\x5b\x30\x6dCommand 'pack suggest-stacks' has been deprecated, please use 'pack stack suggest' instead\n"+
```
dfreilich, line 34: nit: Instead of having the codes, add 
```
	color.Disable(true)
	defer color.Disable(false)
```
to `TestSuggestStacksCommand`, like we do here: https://github.com/buildpacks/pack/blob/1efc582d28151f115f874f5a5372aad1e833ffd2/internal/commands/set_default_builder_test.go#L25-L29

dfreilich (changes requested):
Thanks for the help! Looks great, but I left a few nits to take care of.

anshlykov:
@dfreilich thanks for the feedback! fixed

dfreilich (approved):
This looks great to me, thanks for the help @anshlykov ! 

super

dfreilich:
# UA
## `pack help`
image

## `pack help stack`
image

## `pack stack suggest`
image

## `pack suggest-stacks`
image

@jromero Do you think it should be `pack stack suggest`, or `pack suggests stack`? Also, what do you think about the `pack stack` help test: `Displays stack information`?

jromero:
@dfreilich:

> Do you think it should be `pack stack suggest`, or `pack suggests stack`? 

Based on the initial proposal (RFC),  it would be the former, as implemented. It may not read well, in English terms, but its intent is to "group" based on resources. Open to more conversation via slack or other means.

> Also, what do you think about the `pack stack help` te[x]t: "Displays stack information?"

I think that's a bit misleading. Don't know the exact wording we should use but maybe something like: "Commands available to stacks"? As it stands, I think it's mergeable though.

jromero (approved):

PR #942, opened 2020-11-09
Add support for fish shell completion

jromero:
## Summary

Add support for fish shell completion

#### Before

image

#### After

image

image

## Documentation

    - [ _ ] Yes, see #___
    - [ x ] No

## Related

Resolves #___

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -93,7 +93,12 @@ func NewPackCommand(logger ConfigurableLogger) (*cobra.Command, error) {
 		rootCmd.AddCommand(commands.YankBuildpack(logger, cfg, &packClient))
 	}
 
-	rootCmd.AddCommand(commands.CompletionCommand(logger))
+	packHome, err := config.PackHome()
+	if err != nil {
+		return nil, err
+	}
+
+	rootCmd.AddCommand(commands.CompletionCommand(logger, packHome))
 
 	rootCmd.Version = pack.Version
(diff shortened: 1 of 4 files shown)


dfreilich (changes requested):
Looks like there is also [powershell completion](https://github.com/spf13/cobra/blob/master/powershell_completions.md) in cobra (with instructions [here](https://github.com/spf13/cobra/blob/master/shell_completions.md#generating-shell-completions)). Given our Windows work, can we add that to this PR as well? 

Also, do you think there is merit to ensuring that the completion is actually present in the file? Currently, the tests just verify that the file was outputted by the command, but don't verify anything about what the command actually did.

jromero:
> Looks like there is also [powershell completion](https://github.com/spf13/cobra/blob/master/powershell_completions.md) in cobra (with instructions [here](https://github.com/spf13/cobra/blob/master/shell_completions.md#generating-shell-completions)). Given our Windows work, can we add that to this PR as well?

I wouldn't want to add powershell completion to this PR. As it stands, this PR is nicely single purposed. In addition, I'd hate to add functionality without going through and properly testing it. Something I'm not currently set up to do.

> Also, do you think there is merit to ensuring that the completion is actually present in the file? Currently, the tests just verify that the file was outputted by the command, but don't verify anything about what the command actually did.

I went ahead and added some tests for the files created. I went as far as testing that the file was created and that it wasn't empty. I didn't go much further since it doesn't feel like the value wouldn't be worth the effort.

dfreilich (approved):

PR #943, opened 2020-11-09
Add archive tests

### LINKED ISSUE #814: Extract internal archive package into a public package
### Description
The archive package contains a zip to tar conversion utility that is useful outside of the `pack` CLI itself. We have a use case where we would like to convert user-submitted `zip` files containing app source code into `.tar` files so that we can construct single-layer OCI images.

The `pack` CLI is battle-tested and includes logic for handling a wide-variety of zip files (e.g. those created on Windows machines, `FAT` filesystems, files with strange permissions, etc.) so we would like to build upon this rather than reinvent the wheel. 🎡 🌝 

### Proposed solution
The code within [`internal/archive`](https://github.com/buildpacks/pack/tree/10f5f0a357ccc0b8cc4cabf7ee9a70814386f0cd/internal/archive) is part of a public package that others can consume.

### Additional context
If you're curious, currently we're vendoring in these files wholesale into our repo [here](https://github.com/cloudfoundry/capi-k8s-release/tree/07eb5c8a2205253257681c8060b67e122cb5c77c/src/package-image-uploader/archive). This works right now (especially after https://github.com/buildpacks/pack/issues/813 adds copyright info to the license), but it would be much more convenient if we could just import these packages. That way we could easily keep up as you improve the code, too! 😁 

Thanks!

Tim && @piyalibanerjee

dfreilich:
## Summary

This is a draft PR, exhibiting the code diff, to simplify a PR to https://github.com/buildpacks/imgutil

## Related

Resolves #814

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -184,7 +184,8 @@ func WriteDirToTar(tw TarWriter, srcDir, basePath string, uid, gid int, mode int
 				return err
 			}
 
-			header, err = tar.FileInfoHeader(fi, target)
+			// Ensure that symlinks have Linux link names, independent of source OS
+			header, err = tar.FileInfoHeader(fi, filepath.ToSlash(target))
 			if err != nil {
 				return err
 			}
--- a/internal/archive/archive_test.go
+++ b/internal/archive/archive_test.go
@@ -389,9 +389,7 @@ func testArchive(t *testing.T, when spec.G, it spec.S) {
 				verify := h.NewTarVerifier(t, tr, 1234, 2345)
 				verify.NextFile("/nested/dir/dir-in-archive/some-file.txt", "some-content", 0777)
 				verify.NextDirectory("/nested/dir/dir-in-archive/sub-dir", 0777)
-				if runtime.GOOS != "windows" {
-					verify.NextSymLink("/nested/dir/dir-in-archive/sub-dir/link-file", "../some-file.txt")
-				}
+				verify.NextSymLink("/nested/dir/dir-in-archive/sub-dir/link-file", "../some-file.txt")
 			})
 		})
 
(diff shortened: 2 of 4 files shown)


internal/archive/archive.go
```diff
@@ -184,7 +184,8 @@ func WriteDirToTar(tw TarWriter, srcDir, basePath string, uid, gid int, mode int
 				return err
 			}
 
-			header, err = tar.FileInfoHeader(fi, target)
+			// Ensure that symlinks have Linux link names, independent of source OS
+			header, err = tar.FileInfoHeader(fi, filepath.ToSlash(target))
```
jromero, line 188: This doesn't seem quite right because we already transform the path at line 206. Could you explain why this exception is necessary just for symlinks here and not on line 193?

dfreilich: It's necessary, because in getting the head path (the so-called "target"), we call `os.Readlink()`. That has differing implementations for `unix` and `windows` systems, and returns the OS dependent paths. 

With re: to line 206, when we change the path names in 206, we're doing it for `header.Name`, but not `header.Linkname` (only valid for symlinks), which is set using the `tar.FileInfoHeader` method here. This is a fix for a bug that came out when I checked the symlinks read from zip files, and saw that the Linknames was wrong.

jromero: > It's necessary, because in getting the head path (the so-called "target"), we call os.Readlink(). That has differing implementations for unix and windows systems, and returns the OS dependent paths.

I don't know how relevant that is here since `fi` and `file` would also be OS dependent anywhere within this block.

> With re: to line 206, when we change the path names in 206, we're doing it for header.Name, but not header.Linkname (only valid for symlinks), which is set using the tar.FileInfoHeader method here. This is a fix for a bug that came out when I checked the symlinks read from zip files, and saw that the Linknames was wrong.

Ah, I see. It's all about `header.Linkname`. That's not very obvious. I wonder if it would be more obvious to add the following below line 206 instead:

```golang
if header.Linkname != "" {
  header.Linkname = filepath.ToSlash(header.Linkname)
}
```

dfreilich: We could, but Linkname is only used in a symlink case, which is why I had just put it in the symlink Header initialization. Happy to move it there, though, if that would be a clearer place.

jromero: I do think it would help but not a blocker now that I understand it's purpose.

jromero (approved):

PR #986, opened 2020-12-14
Provide better help text for --volume flag

### LINKED ISSUE #851: Provide better help text for `mode` in `--volume`
### Description
When a user uses `--volume` on `pack build`. Pack defaults to the mode to `ro`. The user doesn't know what the default is and what values are possible. Additionally, given that's it controls more than just the "mode", we should change it to "options".

### Proposed solution

Make the following adjustments to the help text for the `--volume` flag.

1. Rename the template from `[mode]` to `[options]`
1. Add `(default options: "ro")`
1. Add `options` information. Similar to [our docs](https://buildpacks.io/docs/app-developer-guide/mouting-volumes/#mode).

### Acceptance Criteria

```
When a user runs `pack build --help`
Then they should see the following text:

      --volume stringArray          Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.
                                    - 'host path': Name of the volume or absolute directory path to mount.
                                    - 'target path': The path where the file or directory is available in the container.
                                    - 'options' (default "ro"): An optional comma separated list of mount options. 
                                        - "ro", volume contents are read-only.
                                        - "rw", volume contents are readable and writeable.
                                        - "volume-opt=<key>=<value>", can be specified more than once, takes a key-value pair consisting of the option name and its value.
                                    Repeat for each volume in order,
                                      or supply once by comma-separated list


```

Clivern:
## Summary

Provide better help text for --volume flag

#### Before

```
$ go run cmd/pack/main.go build --help

....etc


      --volume stringArray       Mount host volume into the build container, in the form '<host path>:<target path>[:<mode>]'.
                                 Repeat for each volume in order,
                                   or supply once by comma-separated list
```
#### After

```
$ go run cmd/pack/main.go build --help

....etc


--volume stringArray       Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.
                                 - 'host path': Name of the volume or absolute directory path to mount.
                                 - 'target path': The path where the file or directory is available in the container.
                                 - 'options' (default "ro"): An optional comma separated list of mount options.
                                     - "ro", volume contents are read-only.
                                     - "rw", volume contents are readable and writeable.
                                     - "volume-opt=<key>=<value>", can be specified more than once, takes a key-value pair consisting of the option name and its value.
                                 Repeat for each volume in order,
                                   or supply once by comma-separated list
```

## Documentation

    - [ ] Yes, See #___
    - [X] No, See https://buildpacks.io/docs/app-developer-guide/mouting-volumes/#mode

## Related

Resolves #851


dfreilich (approved):
super

Issue #966, opened 2020-11-26
Support podman v.2


abitrolly:
**UPDATE**:

List of issues with `podman` 3.0.0-dev to make it work with `pack`.

- https://github.com/containers/podman/issues/8697
- https://github.com/buildpacks/pack/pull/988

---

Looks like there should be more specific issue than #564 to track interaction between `pack` and `podman`.

As of `podman 2.1.1` it is still impossible to use it with `pack`. I am not sure which bug is this. I don't see any errors in `podman` logs.
```bash
✗ DOCKER_HOST=unix:///run/user/1000/podman/podman.sock pack -v build json2openapi -B heroku/buildpacks:18
Builder heroku/buildpacks:18 is trusted
Pulling image index.docker.io/heroku/buildpacks:18
d32588c49e27fa2d8cfc3ff16eb411487b12f6116054445b5c026112925a9911: pulling image () from docker.io/heroku/buildpacks:18
ERROR: failed to build: invalid builder heroku/buildpacks:18: provided image OS '' must be either 'linux' or 'windows'
✗ podman --version
podman version 2.1.1
```

##### pack info
```
Pack:
  Version:  0.15.0+git-49ad805.build-1613
  OS/Arch:  linux/amd64

Default Lifecycle Version:  0.9.3

Supported Platform APIs:  0.3, 0.4

Config:
(no config file found at /home/anatoli/.pack/config.toml)
```

##### test session for `tmux`

If you're not being paid to work on this like me, then the following script will save you time on checking `podman` debug logs. 
```bash
#!/bin/bash

LEFT="./bin/podman system service --log-level debug -t 0"
RIGHT="DOCKER_HOST=unix://$XDG_RUNTIME_DIR/podman/podman.sock pack -v build json2openapi -B heroku/buildpacks:18"

tmux new-session\; send-keys "$LEFT" Enter\; split-window -h\; send-keys "$RIGHT" Enter
```

matejvasek:
@abitrolly could you please try this with podman build from master?

matejvasek:
I guess `v2.2.0` would be OK too, it's just 4 days old.

matejvasek:
Nope, the fix of mine will be merged after `v2.2` so you need master.

matejvasek:
What is the project you tried to build?

matejvasek:
Nevermind I was using older version of `pack` at first, now I can reproduce it on `podman` master.

abitrolly:
[...]

PR #988, opened 2020-12-14
Implement new build flag --docker-host

matejvasek:
## Summary
Introducing new build flag: --docker-host.
The purpose of the flag is to expose custom docker deamon to build
container. This is useful when running docker daemon on non standard
location e.g. "tcp://localhost:1234", or if resulting image should be
stored in different daemon than you are running locally.

### DIFF
--- a/internal/build/lifecycle_execution_test.go
+++ b/internal/build/lifecycle_execution_test.go
@@ -230,19 +230,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 			fakePhase := &fakes.FakePhase{}
 			fakePhaseFactory := fakes.NewFakePhaseFactory(fakes.WhichReturnsForNew(fakePhase))
 
-			err := lifecycle.Create(
-				context.Background(),
-				false,
-				false,
-				"test",
-				"test",
-				"test",
-				fakeBuildCache,
-				fakeLaunchCache,
-				[]string{},
-				[]string{},
-				fakePhaseFactory,
-			)
+			err := lifecycle.Create(context.Background(), false, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 			h.AssertNil(t, err)
 
 			h.AssertEq(t, fakePhase.CleanupCallCount, 1)
@@ -255,19 +243,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 			expectedRepoName := "some-repo-name"
 			expectedRunImage := "some-run-image"
 
-			err := verboseLifecycle.Create(
-				context.Background(),
-				false,
-				false,
-				expectedRunImage,
-				expectedRepoName,
-				"test",
-				fakeBuildCache,
-				fakeLaunchCache,
-				[]string{},
-				[]string{},
-				fakePhaseFactory,
-			)
+			err := verboseLifecycle.Create(context.Background(), false, "", false, expectedRunImage, expectedRepoName, "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 			h.AssertNil(t, err)
 
 			lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -288,19 +264,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 			fakePhaseFactory := fakes.NewFakePhaseFactory()
 			expectedNetworkMode := "some-network-mode"
 
-			err := lifecycle.Create(
-				context.Background(),
-				false,
-				false,
-				"test",
-				"test",
-				expectedNetworkMode,
-				fakeBuildCache,
-				fakeLaunchCache,
-				[]string{},
-				[]string{},
-				fakePhaseFactory,
-			)
+			err := lifecycle.Create(context.Background(), false, "", false, "test", "test", expectedNetworkMode, fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 			h.AssertNil(t, err)
 
 			lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -315,19 +279,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				verboseLifecycle := newTestLifecycleExec(t, true)
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-				err := verboseLifecycle.Create(
-					context.Background(),
-					false,
-					true,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := verboseLifecycle.Create(context.Background(), false, "", true, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -347,19 +299,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				verboseLifecycle := newTestLifecycleExec(t, true)
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-				err := verboseLifecycle.Create(
-					context.Background(),
-					false,
-					false,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := verboseLifecycle.Create(context.Background(), false, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -383,19 +323,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				verboseLifecycle := newTestLifecycleExec(t, true)
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-				err := verboseLifecycle.Create(
-					context.Background(),
-					false,
-					true,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := verboseLifecycle.Create(context.Background(), false, "", true, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -419,19 +347,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 				additionalTags := []string{"additional-tag-1", "additional-tag-2"}
 
-				err := lifecycle.Create(
-					context.Background(),
-					false,
-					false,
-					"test",
-					"test",
-					"test",
-					fakes.NewFakeCache(),
-					fakes.NewFakeCache(),
-					additionalTags,
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := lifecycle.Create(context.Background(), false, "", false, "test", "test", "test", fakes.NewFakeCache(), fakes.NewFakeCache(), additionalTags, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -466,19 +382,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				volumeMount := "custom-mount-source:/custom-mount-target"
 				expectedBinds := []string{volumeMount, "some-cache:/cache"}
 
-				err := lifecycle.Create(
-					context.Background(),
-					true,
-					false,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{volumeMount},
-					fakePhaseFactory,
-				)
+				err := lifecycle.Create(context.Background(), true, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{volumeMount}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -492,19 +396,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				lifecycle := newTestLifecycleExec(t, false)
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-				err := lifecycle.Create(
-					context.Background(),
-					true,
-					false,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := lifecycle.Create(context.Background(), true, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -519,19 +411,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 				expectedRepos := "some-repo-name"
 
-				err := lifecycle.Create(
-					context.Background(),
-					true,
-					false,
-					"test",
-					expectedRepos,
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := lifecycle.Create(context.Background(), true, "", false, "test", expectedRepos, "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -551,19 +431,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 					verboseLifecycle := newTestLifecycleExec(t, true)
 					fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-					err := verboseLifecycle.Create(
-						context.Background(),
-						true,
-						true,
-						"test",
-						"test",
-						"test",
-						fakeBuildCache,
-						fakeLaunchCache,
-						[]string{},
-						[]string{},
-						fakePhaseFactory,
-					)
+					err := verboseLifecycle.Create(context.Background(), true, "", true, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 					h.AssertNil(t, err)
 
 					lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -602,7 +470,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 					lifecycle := newTestLifecycleExec(t, true, fakes.WithBuilder(fakeBuilder))
 					fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-					err = lifecycle.Export(context.Background(), "test", "test", true, "test", fakeBuildCache, fakeLaunchCache, []string{}, fakePhaseFactory)
+					err = lifecycle.Export(context.Background(), "test", "test", true, "", "test", fakeBuildCache, fakeLaunchCache, []string{}, fakePhaseFactory)
 					h.AssertNil(t, err)
 
 					lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -620,7 +488,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 					lifecycle := newTestLifecycleExec(t, true, fakes.WithBuilder(fakeBuilder))
 					fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-					err = lifecycle.Export(context.Background(), "test", "test", true, "test", fakeBuildCache, fakeLaunchCache, []string{}, fakePhaseFactory)
+					err = lifecycle.Export(context.Background(), "test", "test", true, "", "test", fakeBuildCache, fakeLaunchCache, []string{}, fakePhaseFactory)
 					h.AssertNil(t, err)
 
 					lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -650,19 +518,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				verboseLifecycle := newTestLifecycleExec(t, true)
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-				err := verboseLifecycle.Create(
-					context.Background(),
-					false,
-					false,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := verboseLifecycle.Create(context.Background(), false, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
@@ -681,19 +537,7 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				lifecycle := newTestLifecycleExec(t, false)
 				fakePhaseFactory := fakes.NewFakePhaseFactory()
 
-				err := lifecycle.Create(
-					context.Background(),
-					false,
-					false,
-					"test",
-					"test",
-					"test",
-					fakeBuildCache,
-					fakeLaunchCache,
-					[]string{},
-					[]string{},
-					fakePhaseFactory,
-				)
+				err := lifecycle.Create(context.Background(), false, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
 				h.AssertNil(t, err)
 
 				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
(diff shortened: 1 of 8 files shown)


matejvasek:
Using TCP via `--net=host` and envvar might be advantageous when using on system with SELinux since SELinux doesn't like mounting things into container without proper policies set.

matejvasek:
This is also of interest for https://github.com/buildpacks/pack/issues/966.

matejvasek:
Docker doc also mentions `fd://` protocol but I never used it.

internal/build/phase_config_provider.go
```diff
@@ -108,11 +110,25 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+
 		bind := "/var/run/docker.sock:/var/run/docker.sock"
+
+		if dockerHost, ok := os.LookupEnv("DOCKER_HOST"); ok {
+			if strings.HasPrefix(dockerHost, "unix://") {
+				bind = fmt.Sprintf("%s:/var/run/docker.sock", strings.TrimPrefix(dockerHost, "unix://"))
+			}
+			if strings.HasPrefix(dockerHost, "tcp://") {
+				bind = ""
+				provider.hostConf.NetworkMode = "host"
```
matejvasek, line 122: I am not sure how well this would interact with `WithNetwork()`.

matejvasek: Maybe I should check if the NetworkMode is non-empty string. If it's empty then it's ok, if not I should print warning?

matejvasek: Also this probably will only work with Linux.

jromero:
Hi @matejvasek, thanks for this PR. Just wanted to give you an update and heads up. Given the holiday's it'll probably take a little bit longer to get through the process.  I'm setting up an environment to test this. I have a few comments I'll add to the PR but it's great to see this feature finally coming in.

Adding some context for others, `pack` does support `DOCKER_HOST` for managing the containers it uses to execute the `lifecycle` (aka build process). The problem this PR is attempting to address is the fact that we currently blindly try to mount the docker socket when `--publish` is false. This ultimately fails when the `lifecycle` tries to use the daemon since *it* doesn't have the `DOCKER_HOST` information. In other words, this PR provides the `DOCKER_HOST` information to the `lifecycle`.

jromero:
**Update:** So I ran into a case where this conflicts with the use of Docker Desktop. Currently the mount `/var/run/docker.sock:/var/run/docker.sock` works when the remote docker daemon is "Docker Desktop" because it mounts the internal socket from the VM of the daemon host (not the client). 

The proposed solution in this PR currently adds complexity that isn't quite a drop-in replacement. For example, if you use a `tcp://docker:1234` remote daemon and `docker` hostname is local to the client via `/etc/hosts` the lifecycle container wouldn't be able to resolve the hostname.

I do wonder if this solution could be optionally applied via detecting the host daemon process (docker desktop vs podman) or (less preferable) by using an optional flag/configuration element. Overall, I think it's going in the right direction but we'll want to make sure it's not disruptive out of the gate.

internal/build/phase_config_provider.go
```diff
@@ -108,11 +111,33 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
 	return func(provider *PhaseConfigProvider) {
 		WithRoot()(provider)
+
 		bind := "/var/run/docker.sock:/var/run/docker.sock"
+
+		if dockerHost, ok := os.LookupEnv("DOCKER_HOST"); ok {
+			if strings.HasPrefix(dockerHost, "unix://") {
+				bind = fmt.Sprintf("%s:/var/run/docker.sock", strings.TrimPrefix(dockerHost, "unix://"))
+			}
+			if strings.HasPrefix(dockerHost, "tcp://") {
```
dfreilich, line 121:
```suggestion
			} else if strings.HasPrefix(dockerHost, "tcp://") {
```

internal/build/phase_config_provider.go
```diff
@@ -108,11 +111,33 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+				bind = ""
+				if runtime.GOOS == "linux" {
+					provider.hostConf.NetworkMode = "host"
```
dfreilich, line 124: So, to clarify, this is specifically for SELinux? (https://github.com/buildpacks/pack/pull/988#issuecomment-744619639)

internal/build/phase_config_provider.go
```diff
@@ -108,11 +111,33 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+					provider.ctrConf.Env = append(provider.ctrConf.Env, fmt.Sprintf("DOCKER_HOST=%s", dockerHost))
+				} else {
+					w := provider.errorWriter
+					if w == nil {
+						w = os.Stderr
+					}
+					fmt.Fprintln(w, "Don't know how to expose Docker API via TCP to the container.")
```
dfreilich, line 131: Writing directly to the errorWriter would bypass our typical styles. We should probably look for another way of sending that error message that would be a more seamless fit with our normal practices.

internal/build/phase_config_provider.go
```diff
@@ -159,7 +184,16 @@ func WithLifecycleProxy(lifecycleExec *LifecycleExecution) PhaseConfigProviderOp
...
 func WithNetwork(networkMode string) PhaseConfigProviderOperation {
 	return func(provider *PhaseConfigProvider) {
-		provider.hostConf.NetworkMode = container.NetworkMode(networkMode)
+		if networkMode != "" {
+			if provider.hostConf.NetworkMode != "" {
+				w := provider.errorWriter
+				if w == nil {
+					w = os.Stderr
+				}
+				fmt.Fprintf(w, "Overriding network from %q to %q.", provider.hostConf.NetworkMode, networkMode)
```
dfreilich, line 193: This seems to be more of a debug message than an error message. This, together with the comment above, may point us in the direction of threading the logger directly into the provider.

dfreilich (changes requested):
Thank you for the PR! I have a few reservations about the PR as is: 
1. Tests added for these cases would help the robustness
2. We'd probably want to thread the logger through the provider, so we don't have to write directly to `errorWriter`

I'm happy to take care of both of those in a polish PR, but I do want to clarify the specific logical questions I had – notably, which OS's we think we need to set `host` (so we can have an appropriate comment clarifying what we're doing, and setting `provider.ctrConf.Env` even for non `linux` environments. 

Definitely hoping to get this in in the next few days!

internal/build/phase_config_provider.go
dfreilich, line 121: If we're adding in support for DOCKER_HOST with `tcp`, should we also add support in for `DOCKER_CERT_PATH`? 

https://github.com/moby/moby/blob/3ea762b9f6ba256cf51bd2c35988f0c48bccf0b0/client/client.go#L40

internal/build/phase_config_provider.go
matejvasek: My primary motivation to support `TCP` is `SELinux`. I found it easier to expose `docker API` to container that way. Of course user can set proper `SELinux` policy to make it work with `Unix` socket (some docker installation pkgs do that), but I think that's little bit less convenient. Beside `TCP` may be useful if user want's to run build on remote machine.

internal/build/phase_config_provider.go
```diff
@@ -108,11 +111,33 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+
+		if dockerHost, ok := os.LookupEnv("DOCKER_HOST"); ok {
+			if strings.HasPrefix(dockerHost, "unix://") {
+				bind = fmt.Sprintf("%s:/var/run/docker.sock", strings.TrimPrefix(dockerHost, "unix://"))
+			}
+			if strings.HasPrefix(dockerHost, "tcp://") {
+				bind = ""
+				if runtime.GOOS == "linux" {
+					provider.hostConf.NetworkMode = "host"
+					provider.ctrConf.Env = append(provider.ctrConf.Env, fmt.Sprintf("DOCKER_HOST=%s", dockerHost))
```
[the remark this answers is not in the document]
matejvasek: Yes, we probably should set the `provider.ctrConf.Env` in any case here. The thing is that there can be issue if the `TCP` destination is not reachable from internet, for instance if the `docker API` is running on localhost of the host. On Linux we can use `host` network, I don't know how to solve that on other supported `OS`es.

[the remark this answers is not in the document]
matejvasek: > The proposed solution in this PR currently adds complexity that isn't quite a drop-in replacement. For example, if you use a `tcp://docker:1234` remote daemon and `docker` hostname is local to the client via `/etc/hosts` the lifecycle container wouldn't be able to resolve the hostname.

@jromero I know that's problematic, on Linux this can be solved by using `--network=host` I don't know how to fix that on other `OS`es.

matejvasek:
> The proposed solution in this PR currently adds complexity that isn't quite a drop-in replacement. For example, if you use a `tcp://docker:1234` remote daemon and `docker` hostname is local to the client via `/etc/hosts` the lifecycle container wouldn't be able to resolve the hostname.

@jromero I know that's problematic, on Linux this can be solved by using `--network=host` I don't know how to fix that on other `OS`es.

internal/build/phase_config_provider.go
matejvasek: It could be useful sometimes, but I think that most of the time user will run the `docker API`  on localhost with just `HTTP` not `HTTPS`. If user runs `docker API` accessible from the outside internet then the user should ensure it uses valid certificate  signed by global authority like `Let's Encrypt`.

internal/build/phase_config_provider.go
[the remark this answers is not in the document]
matejvasek: Maybe we could just use `os.Stderr` and not `errorWriter`.

internal/build/phase_config_provider.go
micahyoung: The `runtime.GOOS` would need to be `provider.os` instead, though I think the check could be omitted entirely (and the implementation would otherwise work on Windows) without the `provider.hostConf.NetworkMode=`.

Additionally, I feel like we may want to omit the `provider.hostConf.NetworkMode=`, since `pack build` already has a `--network` flag, and any user setting `DOCKER_HOST=tcp://` will need to consider remote host/container concerns like volumes, bind mounts anyways, in addition to networking.

Also, I wonder if we lose a use-case by always setting the mode as `host`: a remote docker host whose containers should have no network access.

internal/build/phase_config_provider.go
micahyoung: Good catch! This change makes this no longer work: https://medium.com/buildpacks/pack-with-a-remote-docker-daemon-41aab804b839

I regularly use `DOCKER_HOST` + `DOCKER_CERT_PATH` to target remote Windows (and `arm`) daemons with `pack` and that does get lost with this implementation. I'm really not sure there's a secure way to expose the client's mTLS certs and keys to a container, even it's it's through a containers trusted store.

This is leading me to preferring an explicit `pack build` option to enable an alternative `DOCKER_HOST` for a run container instead of inferring from the `pack`'s `DOCKER_HOST`.

micahyoung (changes requested):
To summarize my thinking, I love the idea of being able to specify an alternative to the default bind-mounted socket. I'm a bit concerned with this implementation right now because it's very unencrypted-TCP-on-Linux oriented and breaks the currently working, TCP-over-mTLS which is platform agnostic and much safer for general use.

I wonder, what if this were done with `pack build` option like `--docker-host=`
which, if specified, would *always disable the default binding* and sets a "universal" docker host for `pack` and `lifecycle`, supporting values like:

#### Use-case 1: localhost-over-unencrypted-TCP
Just like pack's current behavior (respect `DOCKER_HOST`) but also copy it to down to lifecycle

```
export DOCKER_HOST=tcp://localhost:2375
pack build my-app --docker-host=inherit --network=host

# equivalent
pack build my-app --docker-host=tcp://localhost:2375 --network=host
```

#### Use-case 2: Non-standard socket location
This would bind-mount anything with a prefix of `unix:` (or `npipe:` for Windows) and use it as `pack` and `lifecycle`'s `DOCKER_HOST` (though maybe using the standard socket path in the container). This *should* enable [rootless docker](https://docs.docker.com/engine/security/rootless/) (which as of last year, did not work due to the hard-coded host socket path)
```
pack build my-app --docker-host=unix:///run/user/1001/docker.sock
```
#### Use-case 3: mTLS-secured daemon
Parse any TCP URI for optional cert options (syntax could be tweaked but a cert paths would also set `DOCKER_TLS_VERIFY` and `DOCKER_CERT_PATH` for lifecycle). [More details on use-case](https://medium.com/buildpacks/pack-with-a-remote-docker-daemon-41aab804b839).
```
pack build my-app --docker-host=tcp://my-secure-host:2376?tlscertpath=/docker-certs --volume $HOME/.docker:/docker-certs
```

We could start with just *Use-case 1* which could expand into the other cases with future contributions.

matejvasek:
Thanks for the feedback. I hope I'll get back to this next week.

dfreilich:
Hey @matejvasek ! Just wanted to check on in this – this is something we're definitely interested in enabling, and it would be great to get this done. Do you think you'll have some time for this over the next few weeks? If not, I know @micahyoung had mentioned that he had some ideas for doing it as well.

matejvasek:
@dfreilich sorry for not getting to this earlier, I had to fix some `podman` issues to make this feature useful for me. I've just pushed new commit adding new `docker-host` flag as @micahyoung suggested. Tests are TODO.

matejvasek:
@micahyoung @dfreilich PTAL

internal/build/lifecycle_execution_test.go
```diff
@@ -1121,12 +986,45 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+				var hasDockerHostMount bool
+				for _, bind := range configProvider.HostConfig().Binds {
+					if strings.HasSuffix(bind, ":/var/run/docker.sock") {
+						hasDockerHostMount = true
+					}
+					if strings.HasSuffix(bind, `:\\.\pipe\docker_engine`) {
+						hasDockerHostMount = true
+					}
+				}
+				h.AssertTrue(t, !hasDockerHostMount)
```
dfreilich, line 1012: I assume we don't need both of these. Overall, the first pattern is definitely shorter – can't we just repeat it to ensure that it doesn't have the windows docker_engine bind?

internal/build/lifecycle_execution_test.go
```diff
@@ -1121,12 +986,45 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+				var hasDockerHostEnv bool
+				for _, e := range configProvider.ContainerConfig().Env {
+					if strings.HasPrefix(e, "DOCKER_HOST=") {
+						hasDockerHostEnv = true
+					}
+				}
+				h.AssertTrue(t, hasDockerHostEnv)
```
dfreilich, line 1019: can we directly test with h.AssertSliceContains with the full expected string?

internal/build/phase_config_provider.go
```diff
@@ -122,14 +123,38 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
 
-func WithDaemonAccess() PhaseConfigProviderOperation {
+func WithDaemonAccess(dockerHost string) PhaseConfigProviderOperation {
 	return func(provider *PhaseConfigProvider) {
 		WithRoot()(provider)
-		bind := "/var/run/docker.sock:/var/run/docker.sock"
-		if provider.os == "windows" {
-			bind = `\\.\pipe\docker_engine:\\.\pipe\docker_engine`
+		if dockerHost == "inherit" {
+			dh, ok := os.LookupEnv("DOCKER_HOST")
```
dfreilich, line 130: Is there any case we're getting by using `LookupEnv` here (and the ensuing if/else), as opposed to just doing something like:
```
dockerHost = os.GetEnv("DOCKER_HOST")
```
?

It seems like whether someone set it to `""`, or it wasn't set, we're going to assign dockerHost to `""`, and then it will be given the default values in the next for loops, so we may as well simplify it.

internal/build/phase_config_provider.go
```diff
@@ -122,14 +123,38 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+				bind = `\\.\pipe\docker_engine:\\.\pipe\docker_engine`
+			}
+		} else {
+			switch {
+			case strings.HasPrefix(dockerHost, "tcp://"):
+				provider.ctrConf.Env = append(provider.ctrConf.Env, fmt.Sprintf(`DOCKER_HOST=%s`, dockerHost))
+			case strings.HasPrefix(dockerHost, "unix://"):
+				bind = fmt.Sprintf("%s:/var/run/docker.sock", strings.TrimPrefix(dockerHost, "unix://"))
+			case strings.HasPrefix(dockerHost, "npipe://"):
+				bind = fmt.Sprintf(`%s:\\.\pipe\docker_engine`, strings.TrimPrefix(dockerHost, "npipe://"))
```
dfreilich, line 150: As you mentioned, it would be good to see tests for these cases as well, though not necessarily in all 3 phases that call it (`Analyze`, `Export` and `Creater`), I'd be happy to just see it in `Creator`

internal/build/phase_test.go
```diff
@@ -446,3 +475,62 @@ func CreateFakeLifecycleExecution(logger logging.Logger, docker client.CommonAPI
...
+		}
+		defer unixCon.Close()
+		go func() {
+			<-ctx.Done()
+			unixCon.Close()
+		}()
+
+		buffA := make([]byte, 4096)
+		buffB := make([]byte, 4096)
+		go io.CopyBuffer(tcpCon, unixCon, buffA)
```
matejvasek, line 504: Here I am not checking errors because it's not easy to check if it really error. For instance `connection closed` is not really error but expected state at some point.

internal/build/phase_test.go
```diff
@@ -446,3 +475,62 @@ func CreateFakeLifecycleExecution(logger logging.Logger, docker client.CommonAPI
...
+		io.CopyBuffer(unixCon, tcpCon, buffB)
+	}
+
+	listener, err := net.Listen("tcp4", ":12345")
```
matejvasek, line 508: @dfreilich shall I randomize port or can we live with having it hardcoded?

matejvasek:
@dfreilich PTAL I did fixes you requested + updated doc. for new flag + added few tests.

internal/commands/build.go
```diff
@@ -143,6 +145,13 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
 	cmd.Flags().StringArrayVar(&buildFlags.EnvFiles, "env-file", []string{}, "Build-time environment variables file\nOne variable per line, of the form 'VAR=VALUE' or 'VAR'\nWhen using latter value-less form, value will be taken from current\n  environment at the time this command is executed\nNOTE: These are NOT available at image runtime.\"")
 	cmd.Flags().StringVar(&buildFlags.Network, "network", "", "Connect detect and build containers to network")
 	cmd.Flags().BoolVar(&buildFlags.Publish, "publish", false, "Publish to registry")
+	cmd.Flags().StringVar(&buildFlags.DockerHost, "docker-host", "inherit",
```
micahyoung, line 148: Apologies, but I realize I missed one aspect of the workflow in my [earlier comment](https://github.com/buildpacks/pack/pull/988#pullrequestreview-564491641), that the previous behavior needs to be the new default (namely, when `--docker-host` is not provided, don't pass through `DOCKER_HOST` env vars into the container and always using a bind mount). Unfortunately, `inherit` isn't sufficient for a backward compatible default. 

I was able to fix this in your PR by setting the default value for the flag to `""` instead of `"inherit"`. I'd be happy with this change for now and if needs any further cleanup, I can do another pass when I add in the TLS behavior.

micahyoung (changes requested):

matejvasek:
@micahyoung this is odd `build / test (windows-lcow)` fails after minor change.

matejvasek:
@micahyoung could you please try restarting the action?

matejvasek:
I'll have to squash the commits anyway.

internal/commands/build.go
```diff
@@ -165,6 +174,16 @@ func validateBuildFlags(flags *BuildFlags, cfg config.Config, packClient PackCli
 		return errors.New("cache-image flag requires the publish flag")
 	}
 
+	dockerHostValid := strings.HasPrefix(flags.DockerHost, `tcp://`) ||
+		strings.HasPrefix(flags.DockerHost, `unix://`) ||
+		strings.HasPrefix(flags.DockerHost, `pipe://`) ||
```
matejvasek, line 179: should be npipe!

micahyoung:
Thank you for the quick implementation and changes. Looking forward to using the feature. It appears the last run went green as well.

matejvasek:
squashed

internal/build/phase_config_provider.go
```diff
@@ -122,14 +123,34 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+		}
+		var bind string
+		if dockerHost == "" {
+			bind = "/var/run/docker.sock:/var/run/docker.sock"
+			if provider.os == "windows" {
+				bind = `\\.\pipe\docker_engine:\\.\pipe\docker_engine`
+			}
+		} else {
+			switch {
+			case strings.HasPrefix(dockerHost, "tcp://"):
```
dfreilich, line 140: This seems to say that `fd` can be handled the same way as `tcp`: https://stackoverflow.com/a/43408869

internal/build/phase_config_provider.go
```diff
@@ -122,14 +123,34 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+			dockerHost = os.Getenv("DOCKER_HOST")
+		}
+		var bind string
+		if dockerHost == "" {
+			bind = "/var/run/docker.sock:/var/run/docker.sock"
+			if provider.os == "windows" {
+				bind = `\\.\pipe\docker_engine:\\.\pipe\docker_engine`
+			}
+		} else {
+			switch {
```
dfreilich, line 139: It also looks like `ssh` is supported (https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option), although I couldn't see how it's handled in the code: https://github.com/moby/moby/blob/41be7293f54f15dc04f024bf2b0f09e1a697208b/daemon/listeners/listeners_linux.go#L22-L56

I think that `ssh` is beyond the scope of this PR, though, if it's unclear how to handle it.

internal/build/phase_config_provider.go
```diff
@@ -122,14 +123,34 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+			case strings.HasPrefix(dockerHost, "tcp://"):
+				provider.ctrConf.Env = append(provider.ctrConf.Env, fmt.Sprintf(`DOCKER_HOST=%s`, dockerHost))
+			case strings.HasPrefix(dockerHost, "unix://"):
+				bind = fmt.Sprintf("%s:/var/run/docker.sock", strings.TrimPrefix(dockerHost, "unix://"))
+			case strings.HasPrefix(dockerHost, "npipe://") || strings.HasPrefix(dockerHost, `npipe:\\`):
+				sub := ([]rune(dockerHost))[8:]
+				bind = fmt.Sprintf(`%s:\\.\pipe\docker_engine`, string(sub))
+			default:
+				panic(fmt.Sprintf("don't know how to handle docker-host value: %s", dockerHost))
```
dfreilich, line 148: I don't think we should panic here, it's not the most graceful exit. How about we just have it be a noop, and write to the errorWriter that we are ignoring the request?

internal/build/phase_test.go
```diff
@@ -261,13 +264,63 @@ func testPhase(t *testing.T, when spec.G, it spec.S) {
...
+						build.WithArgs("daemon"),
+						build.WithDaemonAccess("tcp://127.0.0.1:12345"),
+						build.WithNetwork("host"))
+					phase := phaseFactory.New(configProvider)
+					assertRunSucceeds(t, phase, &outBuf, &errBuf)
+					h.AssertContains(t, outBuf.String(), "daemon test")
+					cancelForward()
+					<-forwardExited
+				})
+			})
```
dfreilich, line 322: nit: Move these inside the #WithDaemonAccess `when` block above it, so that the tests are nested

internal/build/phase_test.go
dfreilich: Hmmm. Hardcoded should work for now, though I worry that it could somehow cause a collision with another test, though we don't really listen on any other ports. Can you pull the port out and make it an explicit variable in the test file though, to make it clearer?

internal/build/lifecycle_execution_test.go
```diff
@@ -704,25 +548,94 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+			it("configures the phase with daemon access with inherited docker-host", func() {
+				oldDH, oldDHExists := os.LookupEnv("DOCKER_HOST")
+				os.Setenv("DOCKER_HOST", "tcp://example.com:1234")
+				defer func() {
+					if oldDHExists {
+						os.Setenv("DOCKER_HOST", oldDH)
+					} else {
+						os.Unsetenv("DOCKER_HOST")
+					}
+				}()
```
dfreilich, line 604: nit: Can you wrap this test with a `when`, and do this in an `it.Before` and `it.After` block?

internal/build/lifecycle_execution_test.go
```diff
@@ -704,25 +548,94 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertSliceContains(t, configProvider.ContainerConfig().Env, "DOCKER_HOST=tcp://example.com:1234")
+			})
+
+			it("configures the phase with daemon access with invalid docker-host", func() {
+				lifecycle := newTestLifecycleExec(t, false)
+				fakePhaseFactory := fakes.NewFakePhaseFactory()
+
+				defer func() {
+					if r := recover(); r == nil {
+						t.Errorf("bad host dind't caused panic")
```
dfreilich, line 625:
```suggestion
						t.Errorf("bad host didn't caused panic")
```

dfreilich (commented):
Thank you so much for doing this, this looks really good, and really close! I'm sorry for sending this back, just a few small things I'd like to see added to this.

internal/build/phase_config_provider.go
```diff
@@ -122,14 +123,34 @@ func WithBinds(binds ...string) PhaseConfigProviderOperation {
...
+				bind = `\\.\pipe\docker_engine:\\.\pipe\docker_engine`
+			}
+		} else {
+			switch {
+			case strings.HasPrefix(dockerHost, "tcp://"):
+				provider.ctrConf.Env = append(provider.ctrConf.Env, fmt.Sprintf(`DOCKER_HOST=%s`, dockerHost))
+			case strings.HasPrefix(dockerHost, "unix://"):
+				bind = fmt.Sprintf("%s:/var/run/docker.sock", strings.TrimPrefix(dockerHost, "unix://"))
+			case strings.HasPrefix(dockerHost, "npipe://") || strings.HasPrefix(dockerHost, `npipe:\\`):
+				sub := ([]rune(dockerHost))[8:]
```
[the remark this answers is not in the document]
matejvasek: Yes, it removes first eight characters. I might as well use `[]byte` instead of `[]rune` since all the character at beginning are one byte wide. Or maybe I could write
```go
([]rune(dockerHost))[len(`npipe://`):]
```

internal/build/phase_config_provider.go
matejvasek: Create an issues for this please.

internal/build/phase_config_provider.go
matejvasek: It is validate on CLI flag level that we won't get here under normal circumstances. Is that OK?

internal/build/phase_config_provider.go
matejvasek: @dfreilich I read http://0pointer.de/blog/projects/socket-activation.html just very briefly, it seems that `fd://` is special value for docker demon (when running as systemd service), not for docker client. For some reason client accepts it and treats `fd://` as `tcp://`.

matejvasek: I changed behaviour. For any protocol that is not `unix://` nor `npipe://` the dockerHost flag is simply passed to build container via envvar. This should allow `fd://`.

internal/build/phase_config_provider.go
matejvasek: I changed behaviour. For any protocol that is not `unix://` nor `npipe://` the dockerHost flag is simply passed to build container via envvar. So the method never panics.

matejvasek:
@dfreilich PTAL

dfreilich (approved):
LGTM

Thank you so much for this, this ended up being considerably more complex than we initially thought, and I appreciate all the work that went into this! 

@micahyoung I'll give you time to check it out as well (if you have the time).

internal/commands/build.go
```diff
@@ -143,6 +145,13 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
 	cmd.Flags().StringArrayVar(&buildFlags.EnvFiles, "env-file", []string{}, "Build-time environment variables file\nOne variable per line, of the form 'VAR=VALUE' or 'VAR'\nWhen using latter value-less form, value will be taken from current\n  environment at the time this command is executed\nNOTE: These are NOT available at image runtime.\"")
 	cmd.Flags().StringVar(&buildFlags.Network, "network", "", "Connect detect and build containers to network")
 	cmd.Flags().BoolVar(&buildFlags.Publish, "publish", false, "Publish to registry")
+	cmd.Flags().StringVar(&buildFlags.DockerHost, "docker-host", "",
+		`Address to docker demon that will be exposed to the build container.
```
matejvasek, line 149: I probably should change spelling to daemon.

internal/build/phase_test.go
```diff
@@ -422,3 +478,64 @@ func CreateFakeLifecycleExecution(logger logging.Logger, docker client.CommonAPI
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer unixCon.Close()
+		go func() {
+			<-ctx.Done()
+			unixCon.Close()
+		}()
+
+		buffA := make([]byte, 4096)
```
matejvasek, line 506: Using buffers here is probably ineffectual, `io.Copy()` should work same job.

internal/commands/build.go
```diff
@@ -143,6 +145,13 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
...
+If not set (or set to empty string) the standard socket location will be used.
+The value has to start with one of following: 'unix://', 'tcp://', 'npipe://', 'npipe:\\'.
```
matejvasek, line 151: I should remove this line.

matejvasek:
@micahyoung what is your use-case for this? Are you pushing images into remote daemon?

micahyoung (approved):
LGTM. Great contribution, many thanks!

micahyoung:
> @micahyoung what is your use-case for this? Are you pushing images into remote daemon?

Yeah, I am using Docker Desktop Windows daemons (and recently Docker Desktop MacOS/ARM64 daemons) on local/remote VMs mostly, with the daemon API either unencrypted or [secured with mTLS](https://docs.docker.com/engine/security/protect-access/), but running `pack` commands from my non-Windows Workstation. I develop buildpacks and stacks mostly for Windows OCI images (and contribute support for this on pack/lifecycle). Having various daemons (Linux/Windows, ARM/AMD64, podman/docker) and just swapping the `DOCKER_HOST` makes it easy to try out various combinations of things against lots of different daemons:

My current workflow, before this feature, was based on swapping around the `DOCKER_HOST`/`DOCKER_CERT_PATH`:

* Targeting Windows daemon insecurely on local VM
    ```
    source ~/.docker/docker-windows-local/env.sh

    docker build ...

    # contents of env.sh
    export DOCKER_HOST=tcp://<local VM IP>:2375
    unset DOCKER_TLS_VERIFY
    unset DOCKER_CERT_PATH
    ```

* Targeting Windows daemon securely over an untrusted network:
    ```
    source ~/.docker/docker-windows-gcp/env.sh

    docker build ...

    # contents of env.sh
    export DOCKER_HOST=tcp://<windows daemon IP>:2376
    export DOCKER_TLS_VERIFY=1
    export DOCKER_CERT_PATH=~/.docker/docker-windows-gcp/  # self-signed client/server certs
    ```

With the functionality of this feature, I should be able to now do this from the command line:

* Targeting Windows daemon on local VM
    ```
    docker build --docker-host=tcp://<local VM IP>:2375 ...
    ```

* Targeting Windows daemon on an untrusted network:
    ... until, the `tls*` options for `--docker-host=tcp://...`, are supported, I'll still need to connect to secured TLS daemons with:
    ```
    source ~/.docker/docker-windows-gcp/env.sh
    
    docker build --docker-host=""  ...  # or unset
    ```

    ... but eventually, with TLS cert options (syntax TBD)
    ```
    docker build --docker-host=tcp://<windows daemon IP>:2376 --docker-host-options=tlscertpath=~/.docker/docker-windows-gcp,tlsverify=1
    ```
    There's a few details here that may be tricky: how to copy in certs in a way that only `lifecycle` can access them but not buildpacks (maybe root/Administrator-only read permissions?). Or perhaps an option to only use `--docker-host="tcp://..."`  
and certs for pack, but continue to bind-mount socket in the container for lifecycle.

But overall, this feature is a great step to making the implicit `DOCKER_HOST` functionality that I've been relying on become more explicit, while enabling more rootless daemon options which is awesome.

PR #1001, opened 2021-01-04
Acceptance: use --publish with untrusted builder

natalieparellano:
## Summary
The pack acceptance tests should run a build with an untrusted builder and the `--publish` flag. The code paths are very different for publish vs. not publish, so this seems reasonable to appropriately test the 5 phases.

This is also very helpful for testing the lifecycle.

## Output
One more `pack build` during acceptance suite.

## Documentation

    - [ ] Yes, see #___
    - [X] No

## Related

Resolves #___


dfreilich (approved):
test all the things

PR #1008, opened 2021-01-06
Add global pull policy configuration command

### LINKED ISSUE #991: Allow global configuration of default pull-policy
### Description
I use `pack` many times per day and I almost always want `--pull-policy if-not-present`. Sometimes I forget to pass this flag and the default behavior clobbers the dev version of a buildpack or builder image I was attempting to use. Appending this flag to every command also make usage of `pack` more cumbersome.

It would be very convenient if I could set this default once in `config.toml`. Then, I would only need to pass the `--pull-policy` flag on the occasions when I _do_ want to pull down the newest version of every image (once a week or so).

### Proposed solution
Add the following option in `config.toml`
```
pull-policy = "<pull policy>"
```

I considered requesting that pack's default behavior change to `if-not-present` but this might not be best because it could introduce vulnerabilities in cases where pack is used to create production images. Also, this default may not be the most convenient UX for app developers rather than folks like me who use `pack` mainly to develop buildpacks and/or the lifecycle.

### Additional context

nkuik:
## Summary

Adds the following commands to set, list, or unset a global container pull policy:

* `pack config pull-policy` (lists the configured policy--default is "always")
* `pack config pull-policy <if-not-present | always | never>` (set the default pull-policy)
* `pack config pull-policy --unset` (unset a configured global policy and reset to default--"always)

The changes still allow `--pull-policy` to be provided as an option in the command line. If `--pull-policy` is provided, it takes precedence to any globally configured pull-policy

### Before

These commands did not exist, so I can't think of any output to show.

#### `pack config -h`

```bash
$  out/pack config -h
Interact with Pack's configuration

Usage:
  pack config [command]

Available Commands:
  default-builder   List, set and unset the default builder used by other commands
  experimental      Print and set the current 'experimental' value from the config
  trusted-builders  Interact with trusted builders
  run-image-mirrors Interact with run image mirrors
  pull-policy       List, set and unset the global pull policy used by other commands
  registries        (experimental) Interact with registries

Flags:
  -h, --help   Help for 'config'

Global Flags:

Use "pack config [command] --help" for more information about a command.
```

#### `pack config pull-policy -h`

```bash
$ out/pack config pull-policy -h
A Buildpack Registry is a (still experimental) place to publish, store, and discover buildpacks.

You can use this command to list, set, and unset the default pull policy that will be used when working with containers:
* To list your pull policy, run `pack config pull-policy`.
* To set your pull policy, run `pack config pull-policy <pull-policy>`.
* To unset your pull policy, run `pack config pull-policy --unset`.
Unsetting the pull policy will reset the policy to the default, which is always

Usage:
  pack config pull-policy <if-not-present | always | never> [flags]

Aliases:
  pull-policy, pull-policy

Flags:
  -h, --help    Help for 'pull-policy'
  -u, --unset   Unset pull policy, and set it back to the default pull-policy, which is always

Global Flags:
```

#### `pack config pull-policy`
When no other policy has been configured.

```bash
$ out/pack config pull-policy
The current pull policy is always
```

#### `pack config pull-policy never`
When no other policy has been configured.

```bash
$ out/pack config pull-policy never
Successfully set never as the pull policy
```

#### `pack config pull-policy if-not-present`
When no other policy has been configured.

```bash
$ out/pack config pull-policy if-not-present
Successfully set if-not-present as the pull policy
```

#### `pack config pull-policy always`
When no other policy has been configured.

```bash
$ out/pack config pull-policy always
Successfully set always as the pull policy
```

#### `pack config pull-policy never`
When never was already the configured policy.

```bash
$ out/pack config pull-policy always
Pull policy is already set to never
```

#### `pack config pull-policy invalid-policy`
When never was already the configured policy.

```bash
$ out/pack config pull-policy invalid-policy
ERROR: invalid pull policy invalid-policy
```

#### `pack config pull-policy --unset`

```bash
$ out/pack config pull-policy --unset
Successfully unset pull policy never
Pull policy has been set to always
```

#### `pack config pull-policy never --unset` (invalid command)

```bash
$ out/pack config pull-policy never --unset
ERROR: pull policy and --unset cannot be specified simultaneously
```

## Documentation

    - [ ] I would imagine, but not totally sure? see #991 (If it should be documented, I could do that as well)

## Related

Resolves #991

### DIFF
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -87,7 +87,11 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 				logger.Warn("Using untrusted builder with volume mounts. If there is sensitive data in the volumes, this may present a security vulnerability.")
 			}
 
-			pullPolicy, err := pubcfg.ParsePullPolicy(flags.Policy)
+			stringPolicy := flags.Policy
+			if stringPolicy == "" {
+				stringPolicy = cfg.PullPolicy
+			}
+			pullPolicy, err := pubcfg.ParsePullPolicy(stringPolicy)
 			if err != nil {
 				return errors.Wrapf(err, "parsing pull policy %s", flags.Policy)
 			}
--- a/internal/commands/builder_create_test.go
+++ b/internal/commands/builder_create_test.go
@@ -92,6 +92,20 @@ func testCreateCommand(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 
+		when("--pull-policy is not specified", func() {
+			when("configured pull policy is invalid", func() {
+				it("errors when config set with unknown policy", func() {
+					cfg = config.Config{PullPolicy: "unknown-policy"}
+					command = commands.BuilderCreate(logger, cfg, mockClient)
+					command.SetArgs([]string{
+						"some/builder",
+						"--config", builderConfigPath,
+					})
+					h.AssertError(t, command.Execute(), "parsing pull policy")
+				})
+			})
+		})
+
 		when("--buildpack-registry flag is specified but experimental isn't set in the config", func() {
 			it("errors with a descriptive message", func() {
 				command.SetArgs([]string{
--- a/internal/commands/buildpack_package.go
+++ b/internal/commands/buildpack_package.go
@@ -10,6 +10,7 @@ import (
 	"github.com/buildpacks/pack"
 	pubbldpkg "github.com/buildpacks/pack/buildpackage"
 	pubcfg "github.com/buildpacks/pack/config"
+	"github.com/buildpacks/pack/internal/config"
 	"github.com/buildpacks/pack/internal/style"
 	"github.com/buildpacks/pack/logging"
 )
@@ -33,7 +34,7 @@ type PackageConfigReader interface {
 }
 
 // BuildpackPackage packages (a) buildpack(s) into OCI format, based on a package config
-func BuildpackPackage(logger logging.Logger, client BuildpackPackager, packageConfigReader PackageConfigReader) *cobra.Command {
+func BuildpackPackage(logger logging.Logger, client BuildpackPackager, cfg config.Config, packageConfigReader PackageConfigReader) *cobra.Command {
 	var flags BuildpackPackageFlags
 	cmd := &cobra.Command{
 		Use:     "package <name> --config <config-path>",
@@ -51,15 +52,20 @@ func BuildpackPackage(logger logging.Logger, client BuildpackPackager, packageCo
 			}
 
 			var err error
-			pullPolicy, err := pubcfg.ParsePullPolicy(flags.Policy)
+			var pullPolicy pubcfg.PullPolicy
+			stringPolicy := flags.Policy
+			if stringPolicy == "" {
+				stringPolicy = cfg.PullPolicy
+			}
+			pullPolicy, err = pubcfg.ParsePullPolicy(stringPolicy)
 			if err != nil {
 				return errors.Wrap(err, "parsing pull policy")
 			}
 
-			cfg := pubbldpkg.DefaultConfig()
+			pubbldpkgCfg := pubbldpkg.DefaultConfig()
 			relativeBaseDir := ""
 			if flags.PackageTomlPath != "" {
-				cfg, err = packageConfigReader.Read(flags.PackageTomlPath)
+				pubbldpkgCfg, err = packageConfigReader.Read(flags.PackageTomlPath)
 				if err != nil {
 					return errors.Wrap(err, "reading config")
 				}
@@ -75,7 +81,7 @@ func BuildpackPackage(logger logging.Logger, client BuildpackPackager, packageCo
 				RelativeBaseDir: relativeBaseDir,
 				Name:            name,
 				Format:          flags.Format,
-				Config:          cfg,
+				Config:          pubbldpkgCfg,
 				Publish:         flags.Publish,
 				PullPolicy:      pullPolicy,
 			}); err != nil {
--- a/internal/commands/buildpack_package.go
+++ b/internal/commands/buildpack_package.go
@@ -34,7 +34,7 @@ type PackageConfigReader interface {
 }
 
 // BuildpackPackage packages (a) buildpack(s) into OCI format, based on a package config
-func BuildpackPackage(logger logging.Logger, client BuildpackPackager, cfg config.Config, packageConfigReader PackageConfigReader) *cobra.Command {
+func BuildpackPackage(logger logging.Logger, cfg config.Config, client BuildpackPackager, packageConfigReader PackageConfigReader) *cobra.Command {
 	var flags BuildpackPackageFlags
 	cmd := &cobra.Command{
 		Use:     "package <name> --config <config-path>",
@@ -51,21 +51,19 @@ func BuildpackPackage(logger logging.Logger, client BuildpackPackager, cfg confi
 				return err
 			}
 
-			var err error
-			var pullPolicy pubcfg.PullPolicy
 			stringPolicy := flags.Policy
 			if stringPolicy == "" {
 				stringPolicy = cfg.PullPolicy
 			}
-			pullPolicy, err = pubcfg.ParsePullPolicy(stringPolicy)
+			pullPolicy, err := pubcfg.ParsePullPolicy(stringPolicy)
 			if err != nil {
 				return errors.Wrap(err, "parsing pull policy")
 			}
 
-			pubbldpkgCfg := pubbldpkg.DefaultConfig()
+			bpPackageCfg := pubbldpkg.DefaultConfig()
 			relativeBaseDir := ""
 			if flags.PackageTomlPath != "" {
-				pubbldpkgCfg, err = packageConfigReader.Read(flags.PackageTomlPath)
+				bpPackageCfg, err = packageConfigReader.Read(flags.PackageTomlPath)
 				if err != nil {
 					return errors.Wrap(err, "reading config")
 				}
@@ -81,7 +79,7 @@ func BuildpackPackage(logger logging.Logger, client BuildpackPackager, cfg confi
 				RelativeBaseDir: relativeBaseDir,
 				Name:            name,
 				Format:          flags.Format,
-				Config:          pubbldpkgCfg,
+				Config:          bpPackageCfg,
 				Publish:         flags.Publish,
 				PullPolicy:      pullPolicy,
 			}); err != nil {
--- a/internal/commands/config.go
+++ b/internal/commands/config.go
@@ -20,10 +20,10 @@ func NewConfigCommand(logger logging.Logger, cfg config.Config, cfgPath string, 
 	cmd.AddCommand(ConfigExperimental(logger, cfg, cfgPath))
 	cmd.AddCommand(ConfigTrustedBuilder(logger, cfg, cfgPath))
 	cmd.AddCommand(ConfigRunImagesMirrors(logger, cfg, cfgPath))
+	cmd.AddCommand(ConfigPullPolicy(logger, cfg, cfgPath))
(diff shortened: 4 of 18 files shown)


internal/commands/build.go
```diff
@@ -111,7 +111,11 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 				logger.Warn("Using untrusted builder with volume mounts. If there is sensitive data in the volumes, this may present a security vulnerability.")
 			}
 
-			pullPolicy, err := pubcfg.ParsePullPolicy(flags.Policy)
+			stringPolicy := flags.Policy
```
nkuik, line 114: I have tried adding this new pattern (check for `--pull-policy` and if it's not set as a parameter, use the configured PullPolicy), but I may have missed some places where it should be changed.

internal/commands/builder_create_test.go
```diff
@@ -92,6 +92,20 @@ func testCreateCommand(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 
+		when("--pull-policy is not specified", func() {
```
nkuik, line 95: Would prefer to add some more thorough testing, but it seems the tests in this file are only for checking errors?

internal/commands/buildpack_package.go
```diff
@@ -50,16 +51,21 @@ func BuildpackPackage(logger logging.Logger, client BuildpackPackager, packageCo
...
+			if stringPolicy == "" {
+				stringPolicy = cfg.PullPolicy
+			}
+			pullPolicy, err = pubcfg.ParsePullPolicy(stringPolicy)
 			if err != nil {
 				return errors.Wrap(err, "parsing pull policy")
 			}
 
-			var cfg pubbldpkg.Config
+			var pubbldpkgCfg pubbldpkg.Config
```
nkuik, line 64: New name on this variable is likely not the best.

internal/commands/config.go
```diff
@@ -20,6 +20,7 @@ func NewConfigCommand(logger logging.Logger, cfg config.Config, cfgPath string,
 	cmd.AddCommand(ConfigExperimental(logger, cfg, cfgPath))
 	cmd.AddCommand(ConfigTrustedBuilder(logger, cfg, cfgPath))
 	cmd.AddCommand(ConfigRunImagesMirrors(logger, cfg, cfgPath))
+	cmd.AddCommand(ConfigPullPolicy(logger, cfg, cfgPath))
```
nkuik, line 23: I wasn't sure if this new command should be under "experimental" or not. If it is still considered "experimental," I can move it there.

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/spf13/cobra"
+
+	pubcfg "github.com/buildpacks/pack/config"
+
+	"github.com/buildpacks/pack/internal/config"
+	"github.com/buildpacks/pack/internal/style"
+	"github.com/buildpacks/pack/logging"
+)
+
+func ConfigPullPolicy(logger logging.Logger, cfg config.Config, cfgPath string) *cobra.Command {
```
nkuik, line 14: I went with `ConfigPullPolicy`, even though the purpose of the function is a combination of configuring and setting a default. I could change it to `SetDefaultPullPolicy`, or possibly split it into two. Let me know if there might be preferences.

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+					return errors.Errorf("pull policy and --unset cannot be specified simultaneously")
+				}
+				oldPullPolicy := cfg.PullPolicy
+				cfg.PullPolicy = ""
+				if err := config.Write(cfg, cfgPath); err != nil {
+					return errors.Wrapf(err, "writing config to %s", cfgPath)
+				}
+
+				pullPolicy, err := pubcfg.ParsePullPolicy(cfg.PullPolicy)
+				if err != nil {
```
nkuik, line 40: Don't necessarily need to check for error here, as the pull policy has been set to one (line 34 this file) that we know is valid. But there could easily be more succinct ways to do this.

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
+package commands
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	pubcfg "github.com/buildpacks/pack/config"
```
nkuik, line 7: I played around with moving the `PullPolicy` type to places other than this public config, but none of those places ended up necessarily making more sense than keeping it where it is.

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+
+func ConfigPullPolicy(logger logging.Logger, cfg config.Config, cfgPath string) *cobra.Command {
+	var unset bool
+
+	cmd := &cobra.Command{
+		Use:     "pull-policy <if-not-present | always | never>",
+		Args:    cobra.MaximumNArgs(1),
+		Short:   "List, set and unset the global pull policy used by other commands",
+		Aliases: []string{"pull-policy"},
+		Long: bpRegistryExplanation + "\n\nYou can use this command to list, set, and unset the default pull policy that will be used when working with containers:\n" +
```
dfreilich, line 22:
```suggestion
		Long: "You can use this command to list, set, and unset the default pull policy that will be used when working with containers:\n" +
```

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+			"* To list your pull policy, run `pack config pull-policy`.\n" +
+			"* To set your pull policy, run `pack config pull-policy <pull-policy>`.\n" +
+			"* To unset your pull policy, run `pack config pull-policy --unset`.\n" +
+			"Unsetting the pull policy will reset the policy to the default, which is always",
```
dfreilich, line 26:
```suggestion
			"Unsetting the pull policy will reset the policy to the default, which is "+style.Symbol("always"),
```

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+func ConfigPullPolicy(logger logging.Logger, cfg config.Config, cfgPath string) *cobra.Command {
+	var unset bool
+
+	cmd := &cobra.Command{
+		Use:     "pull-policy <if-not-present | always | never>",
+		Args:    cobra.MaximumNArgs(1),
+		Short:   "List, set and unset the global pull policy used by other commands",
+		Aliases: []string{"pull-policy"},
```
dfreilich, line 21:
```suggestion
		Aliases: []string{"policy"},
```
We only need to set aliases for commands that aren't the typical command use; in this case, cobra selects the first word of `Use` as the command, so there is no need to set it again in `Aliases`. I would set `policy` there, though, to try and make it simpler for users

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+				logger.Infof("The current pull policy is %s", style.Symbol(pullPolicy.String()))
+			default: // set
+				newPullPolicy := args[0]
+
+				if newPullPolicy == cfg.PullPolicy {
+					logger.Infof("Pull policy is already set to %s", newPullPolicy)
+					return nil
+				}
+
+				pullPolicy, err := pubcfg.ParsePullPolicy(newPullPolicy)
```
dfreilich, line 61: nit: 
```suggestion
				cfg.PullPolicy, err := pubcfg.ParsePullPolicy(newPullPolicy)
```

So you can remove the assignment below

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,81 @@
...
+				}
+
+				logger.Infof("Successfully set %s as the pull policy", style.Symbol(pullPolicy.String()))
+			}
+
+			return nil
+		}),
+	}
+
+	cmd.Flags().BoolVarP(&unset, "unset", "u", false, "Unset pull policy, and set it back to the default pull-policy, which is always")
```
dfreilich, line 78:
```suggestion
	cmd.Flags().BoolVarP(&unset, "unset", "u", false, "Unset pull policy, and set it back to the default pull-policy, which is " + style.Symbol("always"))
```

internal/commands/config_pull_policy_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		command = commands.ConfigPullPolicy(logger, cfg, configFile)
+		command.SetOut(logging.GetWriterForLevel(logger, logging.InfoLevel))
+	})
+
+	it.After(func() {
+		h.AssertNil(t, os.RemoveAll(tempPackHome))
+	})
+
+	when("#ConfigPullPolicy", func() {
```
dfreilich, line 55: nit: It could make the tests a bit easier to read if you group the different functions (`list`, `set`, and `unset`) in different `when` blocks

internal/commands/config_pull_policy_test.go
```diff
@@ -0,0 +1,198 @@
...
+				h.AssertNil(t, command.Execute())
+
+				output := outBuf.String()
+				h.AssertEq(t, strings.TrimSpace(output), `Pull policy is already set to if-not-present`)
+			})
+			it("it does not change the configured policy", func() {
+				command = commands.ConfigPullPolicy(logger, cfg, configFile)
+				command.SetArgs([]string{"never"})
+				assert.Succeeds(command.Execute())
+
```
dfreilich, line 117: I would perhaps read and verify the config file here as well, to make sure that it doesn't need to be run twice in order to set it.

internal/commands/config_pull_policy_test.go
```diff
@@ -0,0 +1,198 @@
...
+				command.SetArgs([]string{"invalid-policy"})
+				command.Execute()
+
+				readCfg, err := config.Read(configFile)
+				assert.Nil(err)
+				assert.Equal(readCfg.PullPolicy, "never")
+			})
```
dfreilich, line 147: nit: Combine these tests

dfreilich (changes requested):
Overall, this is phenomenal, and really helpful. I had a bunch of nit suggestions, mostly stylistic, but this is super helpful. 

One addition location to make this change is in `internal/commands/package_buildpack.go`
Also, one additional location that I'd add a test in is in `internal/commands/confg_test.go`, specifically [here](https://github.com/buildpacks/pack/blob/3d165d5e203e3980701d39e6466ee3874a9c6c40/internal/commands/config_test.go#L66), just to provide assurance that the command is displayed as we would expect. 

Thanks

internal/commands/config_pull_policy.go
nkuik: I would like to do this, but here `cfg.PullPolicy` is a string and `ParsePullPolicy` returns a `(PullPolicy, error)`, so I couldn't think of way to do away with the weird `string -> PullPolicy -> string` kind of conversions.

internal/commands/config_pull_policy.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/buildpacks/pack/internal/config"
+	"github.com/buildpacks/pack/internal/style"
+	"github.com/buildpacks/pack/logging"
+)
+
+func ConfigPullPolicy(logger logging.Logger, cfg config.Config, cfgPath string) *cobra.Command {
+	var unset bool
+
+	cmd := &cobra.Command{
+		Use:   "pull-policy <if-not-present | always | never>",
```
dfreilich, line 20:
```suggestion
		Use:   "pull-policy <always | if-not-present | never>",
```

dfreilich (changes requested):
This is super close. Just a few more small changes I'd like to see, if you have the time.

dfreilich (approved):
fantastic

This should go out in the  release (`0.17.0`), not the immediately upcoming release (`0.16.0`), to give us a bit more time to test it and make sure no 🐛  are slipping by us. Thanks again for the help!

PR #1022, opened 2021-01-14
Added new keys to BuildpackInfo to support new keys in buildpack.toml

jkutner:
## Summary

Implementation of https://github.com/buildpacks/rfcs/pull/127

**Important:** 
* this makes `dist.BuildpackInfo` no longer acceptable as a key because it contains types that are not compareable, so I've changed it to a pointer where the actual value is required, and a string where the `FullName()` can be used.
* I had to move the `License` struct into `internal/dist` to avoid a circular reference in the `project` tests (which import `testhelper`, which imports `internal/dist`) 

#### Before

N/A

#### After

N/A

## Documentation

    - [X] Yes, see #___
    - [ ] No

## Related

Resolves #___

### DIFF
--- a/inspect_buildpack.go
+++ b/inspect_buildpack.go
@@ -137,7 +137,7 @@ func extractOrder(buildpackMd buildpackage.Metadata) dist.Order {
 
 func extractBuildpacks(layersMd dist.BuildpackLayers) []dist.BuildpackInfo {
 	result := []dist.BuildpackInfo{}
-	buildpackSet := map[dist.BuildpackInfo]bool{}
+	buildpackSet := map[*dist.BuildpackInfo]bool{}
 
 	for buildpackID, buildpackMap := range layersMd {
 		for version, layerInfo := range buildpackMap {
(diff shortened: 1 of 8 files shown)


dfreilich (approved):
I guess that if these are for the buildpack registry, there's no need to add them into the `inspect-buildpack`/`inspect-builder` output...

I've verified that the information is present on a created buildpack:
```
$ out/pack buildpack package samples/hello-world -c ~/workspace/samples/packages/hello-world/package.toml
...
$ docker inspect --format "{{ json .Config.Labels}}" samples/hello-world | jq ."\"io.buildpacks.buildpackage.metadata\"" -r | sed 's#\\##g' | jq .
{
  "id": "samples/hello-world",
  "version": "0.0.1",
  "homepage": "https://github.com/buildpacks/samples/tree/main/buildpacks/hello-world",
  "keywords": [
    "something",
    "somethingb"
  ],
  "licenses": [
    {
      "Type": "MIT",
      "URI": "https://example.com/license"
    },
    {
      "Type": "MIT2",
      "URI": "https://example.com/license"
    }
  ],
  "stacks": [
    {
      "id": "io.buildpacks.samples.stacks.bionic"
    },
    {
      "id": "io.buildpacks.samples.stacks.alpine"
    },
    {
      "id": "io.buildpacks.stacks.bionic"
    }
  ]
}

```

@jkutner Can you also make a PR to add in https://github.com/buildpacks/samples/issues/102?

PR #1027, opened 2021-01-19
Remove usage of deprecated module.

kramarz:
## Summary

v1util was deprecated and removed in latest version of go-containerregistry.
Removed in: https://github.com/google/go-containerregistry/commit/3584fa0df9f1bd51a7cf559c97126a20212fa7cd#

Function itself was already copied into this module, so it is easy to use it instead.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

This is blocking update of this dependency: https://github.com/buildpacks/pack/pull/1018

### DIFF
--- a/testhelpers/tar_assertions.go
+++ b/testhelpers/tar_assertions.go
@@ -10,7 +10,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/google/go-containerregistry/pkg/v1/v1util"
 	"github.com/pkg/errors"
 
 	"github.com/buildpacks/pack/internal/archive"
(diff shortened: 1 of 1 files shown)


dfreilich:
Thank you for the help, @kramarz ! I just made the same change in #1018 , so I think this was already taken care of. We appreciate it though!

kramarz:
Ah, thanks. I wrote it yesterday, but had trouble running tests locally and didn't check updated PR.

dfreilich:
Ah ok! What issues did you have with the tests, if I may? I'd like to make sure that the development process is as easy for new contributors as possible, and we'll take whatever feedback we can get 😄

PR #1029, opened 2021-01-21
Auto add labels/milestones to new PRs

dfreilich:
* For labelling, we use Githubs built in labeler action, together with a config file (.github/labeler.yml), describing a number of paths.
* For milestones, this uses the github-script to get the list of milestones for the repo, select the first (the oldest open), and assigns that to the PR

## Documentation

    - [x] No

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -8,5 +8,5 @@ type/chore:
   - ./go.sum
 
 type/enhancement:
-  - any: ['./*.go', './**/*.go', '!./acceptance/**/*', '!./test*/**/*']
+  - any: ['./*.go', './**/*.go', '!./acceptance/**/*', '!./test*/**/*', '!./**/*_test.go']
     all: ['!.github/**/*']
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,12 +1,12 @@
 # Rules defined here: https://github.com/actions/labeler
 type/chore:
-  - ./*.md
-  - ./**/*.yml
-  - ./acceptance/**/*
-  - ./.github/**/*
-  - ./go.mod
-  - ./go.sum
+  - '*.md'
+  - '**/*.yml'
+  - 'acceptance/**/*'
+  - '.github/**/*'
+  - 'go.mod'
+  - 'go.sum'
 
 type/enhancement:
-  - any: ['./*.go', './**/*.go', '!./acceptance/**/*', '!./test*/**/*', '!./**/*_test.go']
-    all: ['!.github/**/*']
+  - '*.go'
+  - '**/*.go'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,12 @@
+# Rules defined here: https://github.com/actions/labeler
+type/chore:
+  - ./*.md
+  - ./**/*.yml
+  - ./acceptance/**/*
+  - ./.github/**/*
+  - ./go.mod
+  - ./go.sum
+
+type/enhancement:
+  - any: ['./*.go', './**/*.go', '!./acceptance/**/*', '!./test*/**/*']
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
...
+type/chore:
+  - ./*.md
+  - ./**/*.yml
+  - ./acceptance/**/*
+  - ./.github/**/*
+  - ./go.mod
+  - ./go.sum
+
+type/enhancement:
+  - any: ['./*.go', './**/*.go', '!./acceptance/**/*', '!./test*/**/*']
```
dfreilich, line 11: This (I believe) should match any adjustment to a `.go` file, as long as it isn't just a change to the `acceptance` dir, or any top-level dir starting with `test` (like `testdata`, `testhelpers`, or `testmocks`). 

However, if any of the files in the PR are a change in the `.github` directory, it won't be labeled `type/enhancement`

.github/workflows/privileged-pr-process.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          script: |
+            var milestone = await github.issues.listMilestones({
+                      owner: github.context.repo.owner,
+                      repo: github.context.repo.repo,
+                      state: 'open',
+                      sort: 'due_on',
+                      direction: 'asc'
+                   })
+                  .then(result => result.data[0])
```
dfreilich, line 29: This assumes that milestones will be closed appropriately when we deliver them, and therefore new PRs will be tagged with the next milestone, which is the most relevant.

.github/workflows/privileged-pr-process.yml
```diff
@@ -0,0 +1,42 @@
...
+                  .catch(err => {throw "ERROR: " + err.message});
+
+            if (milestone) {
+                await client.issues.update({
```
dfreilich, line 33: PRs are considered `issues` for the github API.

.github/workflows/privileged-pr-process.yml
```diff
@@ -0,0 +1,42 @@
+name: "Process Pull Request"
+# NOTE: This workflow is privileged, and shouldn't be used to checkout the repo or run any of the PR code
+# See more in: https://securitylab.github.com/research/github-actions-preventing-pwn-requests
+on:
+  - pull_request_target
```
jromero, line 5: 💯

.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+# Rules defined here: https://github.com/actions/labeler
+type/chore:
+  - ./*.md
+  - ./**/*.yml
+  - ./acceptance/**/*
```
jromero, line 5: Are the labels checked based on the order they are defined in this file? If so, wouldn't a **feature** with an added acceptance test or new dependency by marked as `type/chore` as oppose to `type/feature`?

.github/workflows/privileged-pr-process.yml
```diff
@@ -0,0 +1,42 @@
...
+            if (milestone) {
+                await github.issues.update({
+                  owner: github.context.repo.owner,
+                  repo: github.context.repo.repo,
+                  issue_number: github.event.number,
+                  milestone: milestone.number,
+                  });
+                core.setOutput('milestone', addMilestone);
+              } else {
+              console.log(`Milestone not found, Please create it first.`);
```
jromero, line 41:
```suggestion
              console.log(`No milestone found. Please create one first.`);
```

.github/labeler.yml
dfreilich: So what I think it does is it would add both labels. For example, that's what it did here: https://github.com/dfreilich/pack/pull/13. I opted to just make the rules simpler (so we at least have labels), and then we can adjust them as needed.

jromero (approved):

PR #1039, opened 2021-01-26
Add labels to cache volumes

### LINKED ISSUE #317: Add labels to pack volumes (to help association/purging)
One feature Dokku has is the ability to run `repo:purge-cache`, which effectively deletes the cache for given app. Pack (libbuildpack?) uses a docker volume to store the cache but only exposes a mechanism to delete that cache under the `build` command.

It would be great if either of the following were implemented:

- labels on the volumes so we can associate them with a given app image
- a command that can be run which will purge the cache for a given app image

dwillist:

## Summary
Adds human readable info to pack volumes.

## Output
previously pack volumes were named with a `sha256` and had the form `<sha256>.<prefix>`
now they will be of the form 
`pack-cache-<repo>-<name>-<tag>-sha256.<prefix>`

## Documentation

    - [ ] Yes, see #___
    - [ x ] No

Resolves #317


dfreilich (approved):
## Before: 
```
 $  docker volume ls
DRIVER    VOLUME NAME
local     pack-cache-b07b36ffb1de.build
local     pack-cache-b07b36ffb1de.launch
```

## After: 
```
local     pack-cache-dfreilich_test-noop_latest-4e450adbfda1.build # ghcr publish
local     pack-cache-dfreilich_test-noop_latest-a17094201004.build # docker publish
local     pack-cache-library_test-noop_latest-b07b36ffb1de.build # local docker build
local     pack-cache-library_test-noop_latest-b07b36ffb1de.launch. #local docker build
```

Overall, this definitely seems to do what it set out to do. I have some quibbles about the presentation (`library` isn't super clear, and perhaps we should keep the entire registry string), but overall I think it's definitely a great start, and we can iterate on it as we need. 

name

PR #1041, opened 2021-01-26
Enable GID override in Build

### LINKED ISSUE #961: Add -gid Flag to Pack to Support OpenShift
### Description

In OpenShift, [the default security rules dictate that images are able to run as a random user](https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#use-uid). Many organizations have this rule strictly enforced.

RedHat's recommended approach to granting user read/write/execute permissions is to grant group `0` those permissions instead because the random user is guaranteed to be in that group.

```bash
RUN chgrp -R 0 /some/directory && \
    chmod -R g+rwX /some/directory
```

In order to support this, Buildpack stack run images must set the `CNB_GROUP_ID` to `0`.

### Proposed solution

Since having to customize every stack that needs to run in OpenShift is a lot of work, it would probably be better to add a `-gid` flag to `pack` in order to allow the group to be overridden specifically at build time.

jsncmgs1:
## Summary

Allows group ID to be set via a `-gid` flag.
#961

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #961

### DIFF
--- a/build.go
+++ b/build.go
@@ -149,6 +149,9 @@ type BuildOptions struct {
 
 	// ProjectDescriptor describes the project and any configuration specific to the project
 	ProjectDescriptor project.Descriptor
+
+	// TBD
+	GroupID config.GroupID
 }
 
 // ProxyConfig specifies proxy setting to be set as environment variables in a container.
@@ -241,7 +244,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		buildEnvs[k] = v
 	}
 
-	ephemeralBuilder, err := c.createEphemeralBuilder(rawBuilderImage, buildEnvs, order, fetchedBPs)
+	ephemeralBuilder, err := c.createEphemeralBuilder(rawBuilderImage, buildEnvs, order, fetchedBPs, opts)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


jsncmgs1:
I think I'm close here (minus test changes and additions), but I could use a nudge in the right direction on this.

So far I've added the flag and a type representing the flag, and I pass the flag value down into `builder.userAndGroupIDs`, overriding the value set on `CNB_GROUP_ID`.

When I run this locally, it picks up the group ID change for `/layers`, but not `/workspace`:

```
$ out/pack build test_img --path ../go_stuff --builder heroku/buildpacks:18 --gid=3 --pull-policy if-not-present
```

Screen Shot 2021-01-26 at 4 30 26 PM

I'm having some trouble locating where I need to make the change for the `/workspace` group ID and prevent the change for `/layers`. I think they're created here? https://github.com/buildpacks/pack/blob/860396d2086977a48921667d996b6141cda3953d/internal/builder/builder.go#L547

But it seems like at some point `/app` is copied or symlinked into `/workspace`.

dfreilich:
@jsncmgs1 Thanks for the PR and work on it! Your path right now seems to be changing the user environment of the `ephemeral builder`, which is definitely interesting! What I'd recommend is rather than changing the builder, pass the `gid` argument (if provided) to the relevant lifecycle phases (`analyze`, `restore`, `export`, and `create`).  You can see examples of how we pass flags to the lifecycle phases here (in `internal/build/lifecycle_execution.go`):

https://github.com/buildpacks/pack/blob/c911002176ac402cbd50ea18112e3bedd29917e9/internal/build/lifecycle_execution.go#L191-L212

dfreilich:
Hey @jsncmgs1  - let me know if I can be of help with this (assuming you're still interested). The fix for this would be similar to what's being done in https://github.com/buildpacks/pack/pull/988, I'd be happy to walk through it with you.

PR #1056, opened 2021-02-03
Correct homebrew formula

dwillist:

## Summary
small syntax fix in our homebrew formula that prevented the `OS.mac?` line from evaluating


jromero (approved):

dfreilich (approved):
Thanks for the catch! 

facepalm

PR #1066, opened 2021-02-09
Use dependency injection for config in commands

### LINKED ISSUE #852: Use Dependency Injection for Config Paths
> Parallel test runs don't run in their own environment. I had it implemented this way but flakes started to occur due to the `PACK_HOME` being overridden in concurrent executions when `spec.Parallel()` is set. We could move it to `spec.Sequential` and live with it but ultimately it steered me into the thought that we _shouldn't_ be calling `DefaultConfigPath()` during our tests at all. Instead we should be relying on dependency injection instead of global state.

https://github.com/buildpacks/pack/pull/849#discussion_r494422835

Use dependency injection when creating commands for config path if needed, instead of using `DefaultConfigPath()`. Follow the pattern used in #847.

dfreilich:
## Summary

This carries on with the work of #847, and ensures all commands (though not all code paths, to simplify this PR) which depend upon the config path are provided it using dependency injection. 

## Related

Resolves #852


jromero (approved):

jromero:
image

robot

PR #1076, opened 2021-02-17
Move image assertions to new ImageAssertionManager

dfreilich:
## Summary

This PR, inspired and based on #673, introduces `ImageAssertionManager` and `ImageManager` to our acceptance tests. These concepts are helpful for pulling out some of the hardcoded dependencies we had on Docker methods, and simplifies a number of the acceptance tests. It also pulls out some inline functions we had in the file, and dedicates the acceptance_tests folder more to just the tests, without surrounding functions. 

Additionally, this PR adds an acceptance build annotation to suite_manager. 

## Documentation

    - [x] No

## Related

Pulls in changes from #673


acceptance/managers/image_manager.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+type TestContainer struct {
+	testObject *testing.T
+	dockerCli  client.CommonAPIClient
+	assert     h.AssertionManager
+	name       string
+	id         string
+}
```
dfreilich, line 126: Should this be pulled to a separate file? Is it a useful abstraction at all?

acceptance/managers/image_manager.go
```diff
@@ -0,0 +1,176 @@
...
+		dockerCli:  dockerCli,
+	}
+}
+
+func (im ImageManager) CleanupImages(imageNames ...string) {
+	im.testObject.Helper()
+	err := h.DockerRmi(im.dockerCli, imageNames...)
+	if err != nil {
+		im.testObject.Logf("%s: Failed to remove image from %s", err, imageNames)
+	}
```
dfreilich, line 41: There are a number of times where we try to remove images that don't exist – for example, we do a `Cleanup` in the `it.After` for virtually all `pack build` acceptance tests, but our tests have some failure cases where a build is supposed to fail, and the resulting image wasn't created. As such, it didn't make sense to fail if we can't RM the image.

acceptance/assertions/image.go
```diff
@@ -0,0 +1,102 @@
...
+	defer container.Cleanup()
+
+	output := container.RunWithOutput()
+	a.assert.ContainsAll(output, expectedOutputs...)
+}
+
+func (a ImageAssertionManager) CanBePulledFromRegistry(name string) {
+	a.testObject.Helper()
+	a.imageManager.PullImage(name, a.registry.RegistryAuth())
+	a.ExistsLocally(name)
```
dwillist, line 83: We should probably delete the image after we check that it can be pulled

acceptance/managers/image_manager.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (im ImageManager) Inspect(image string) (dockertypes.ImageInspect, error) {
```
dwillist, line 44: nit: Could this be `InspectLocal` (or something similar) as this will only work for images on the docker daemon

dwillist (approved):

acceptance/managers/image_manager.go
dfreilich: addressed in ff6aa1d356a5e5c1cb5c03b69d86741519eb4a95

Issue #1055, opened 2021-02-03
[RFC 0073] Add Support for Asset Cache


dwillist:
### Description
[RFC 0073](https://github.com/buildpacks/rfcs/blob/main/text/0073-offline-buildpackages.md)
[buildpacks/rfcs#121](https://github.com/dwillist/rfcs/blob/offline-buildpackages/text/0000-offline-buildpackages.md)
buildpacks/spec#184

Implement the following features in order to satisfy the requirements in RFC 0073. This issue will link out to others that implement the features outlined below.
- Pack can create an asset cache `asset-cache` artifact
  - From implementation buildpacks
  - from metabuildpacks
- Pack stores asset-cache metadata on builders
- `pack build` provides assets to the build image.

Issue #597, opened 2020-04-28
Use resource-based subcommands


jromero:
### Description

As the list of commands grows, the need for grouping related `pack` commands starts to show. There could be an argument made that this pattern is common and expected.

This is what our current list of commands looks like:

```
pack
  build                 
  rebase                
  inspect-image         
  create-builder        
  package-buildpack     
  set-run-image-mirrors 
  inspect-builder       
  set-default-builder   
  suggest-builders      
  suggest-stacks        
  version               
  report                
  completion            
  help                  
```

### Proposed solution

```
pack 
  image
    build
    inspect
    rebase
  builder
    create
    inspect
    default
    suggest
    trust       [upcoming]
    untrust     [upcoming]
  buildpack
    package
    publish     [upcoming]
  stack
    suggest
    run-image-mirror
      set
  version
  report
  completion
  help
```

#### Open Questions
- Is `run-image-mirror` good enough? Not sure how to fit `set-run-image-mirror` into this new structure.
- Is there an alternative to `image`? Maybe `project`?

#### Alias `image` subcommands as base commands

Some commands such as `build` and `rebase` might benefit from remaining top-level for the foreseeable future. A simple way to make sense of this is to consider any top-level command (aside from helper commands `version`, `report`, `completion`, `help`) a subcommand of `image`.

### Additional context
- [x] This feature should be documented [somewhere](https://github.com/buildpacks/docs/issues/119)

#### Migration

In efforts to make this the least disruptive possible we would likely want to keep the existing commands untouched for a few version.

#### Prior Art
- [docker](https://docs.docker.com/engine/reference/commandline/cli/)
- [gcloud](https://cloud.google.com/sdk/docs/scripting-gcloud)
- [kubectl](https://kubernetes.io/docs/reference/kubectl/overview/)

jromero:
RFC: https://github.com/buildpacks/rfcs/pull/93

jromero:
Issues for implementing this change have been created.

NOTE: The [accepted RFC](https://github.com/buildpacks/rfcs/pull/93) has some differences from the proposed solution posted on this issue. See RFC for specific details.

PR #1085, opened 2021-02-19
Add asset package creation

dwillist:

## Summary
Implement `asset-cache` creation as specified by #1055

This adds the experimental `create-asset-cache` command that works as follows:
`pack create-asset-cache <image-name> --buildpack <buildpack locator> --publish(optional) --pull-policy(optional)`

Additionally it adds support for using asset caches when
- Running `pack build ` with the `--asset-cache` flag
- Adding asset caches to builders.

~~(side UI bug, local `buildpack` require a `file://` prefix...~~

## Documentation
    - [x] Yes,  (whole feature to be documented upon completion) 
    - [ ] No

## Related

Implements first part of #1055


dfreilich:
This is more of a meta-point, but what do you think about calling the command `pack asset-cache create`, instead of `create-asset-cache`? Do we expect any further `asset-cache` commands? 

As we're close to finished with #597, I'm hesitant to add more commands which don't fit in with the new paradigm we're trying to follow.

internal/blob/downloader.go
```diff
@@ -25,18 +27,25 @@ const (
...
+	return downloader{
 		logger:       logger,
 		baseCacheDir: baseCacheDir,
 	}
 }
 
-func (d *downloader) Download(ctx context.Context, pathOrURI string) (Blob, error) {
+type DownloadOption func(*downloader)
+
+func (d downloader) Download(ctx context.Context, pathOrURI string, options ...DownloadOption) (Blob, error) {
```
aemengo, line 45: 🤔 I'm really curious why we're removing the pointer receiver, through this whole file.

build.go
```diff
@@ -745,8 +745,8 @@ func appendBuildpackToOrder(order dist.Order, bpInfo dist.BuildpackInfo) (newOrd
 	return newOrder
 }
 
-// decomposeBuildpack decomposes a buildpack blob into the main builder (order buildpack) and it's dependencies buildpacks.
-func decomposeBuildpack(blob blob.Blob, imageOS string) (mainBP dist.Buildpack, depBPs []dist.Buildpack, err error) {
+// DecomposeBuildpack decomposes a buildpack blob into the main builder (order buildpack) and it's dependencies buildpacks.
+func DecomposeBuildpack(blob blob.Blob, imageOS string) (mainBP dist.Buildpack, depBPs []dist.Buildpack, err error) {
```
aemengo, line 749: Should this be exported? It's not referenced outside the package.

internal/blob/downloader.go
dwillist: Ya this is so that we can set `downloader.blobOptions` & `downloader.validationSha256` in the download method & prevent it mutating the underlying `downloader` object

acceptance/acceptance_test.go
```diff
@@ -2253,6 +2253,208 @@ include = [ "*.jar", "media/mountain.jpg", "/media/person.png", ]
 				})
 			})
 
+			when("creating asset cache", func() {
```
dwillist, line 2256: There is definitely some simplification here, but I just want to push off doing a refactor of these tests because there will be many additions when testing asset-cache usage

create_asset_cache.go
```diff
@@ -0,0 +1,133 @@
...
+	pair   dist.BlobAssetPair
+	sha256 string
+	err    error
+}
+
+type downloadJob struct {
+	asset dist.Asset
+	index int
+}
+
```
dwillist, line 56: Note the cache poisoning & inability to exit using `ctrl+c` is not new behavior

create_asset_cache.go
```diff
@@ -39,20 +42,84 @@ func (c *Client) CreateAssetCache(ctx context.Context, opts CreateAssetCacheOpti
...
+	resultIndiciesMap := map[string]int{}
+	errors := []error{}
+	for i := 0; i < len(assets); i++ {
+		r := <-results
+		prevIdx, ok := resultIndiciesMap[r.sha256]
+		switch {
+		case r.err != nil:
+			errors = append(errors, r.err)
+		case !ok || prevIdx < r.index:
```
aemengo, line 84: I'm struggling to see why we need `r.index` here. Our final return type is a mapping so preserving order doesn't matter.
Are we expecting duplicates to be returned from the channel with a given `r.sha256`?

dwillist: > Are we expecting duplicates to be returned from the channel with a given r.sha256?

Exactly. Two buildpacks share an asset, it has the same `r.sha256` but different metadata. If we don't always pick the same metadata we are missing a little bit of reproducibility (though this should be rare...).

If we remove this constraint and fiddle with the timings [here](https://github.com/buildpacks/pack/pull/1085/files#diff-27a7971b2d46c736fecf53ce767289b1b4121623503b68948654b51f8333a98aR96-R101), we can get failures.

ekcasey:
@dwillist I wonder if this feature should only be supported for buildpack API `0.7`+ (which doesn't exist yet). Older buildpack APIs don't support assets in `buildpack.toml`.

https://github.com/buildpacks/spec/issues/190

internal/blob/downloader.go
jromero: This sounds like a smell. What is the purpose of associating `blobOptions` to the downloader when it's really only an option for a single operation? Same can be said for `validationSha256`. It looks like it's adding unnecessary state to a service (`Downloader`) which now isn't reusable because of its state....

I would propose looking for alternatives to providing optional arguments.

For example, `DownloadOptions` can be a `func (*downloadOptions)` and `downloadOptions` be an unexported struct with the two fields you added to `Downloader`. 

Then instead of

```
	for _, option := range options {
		option(&d)
	}
```

do something like:

```
        do := &downloadOptions{}
	for _, option := range options {
		option(&do)
	}
```

jromero:
Blocked by pending spec changes. If we want to get this feature in, then we'd want to gate it behind experimental.

dwillist:
### PR overview
As things stand I think this is ready for PR review, I'll be updating this comment with additional acceptance testing materials for folks if the want to use them.

#### Whats implemented
Note that all of the below are hidden behind `experimental` (even the new `pack build ...` flags)
- Add a `pack asset-package create` command
  - this command takes list of buildpacks and creates an asset cache image as an `image` or `tarball`.
  - TODO: gate the creation of packages based on a `buildpack API` version. We cannot really test this until we have a lifecycle that supports the new API.
- Add the ability to add `assets` to a builder by specifying them in the `builder.toml` configuration file. (Add an example file here)
  - TODO gate the creation of builders with assets on the `platform API` version. Again we need to have the lifecycle release to get the actual platform API we should be gating on. 
- Add the ability to add `assets` to a pack build 
  - again gate this ability based on the platform API that will be used by pack & the underlying builder.

### Whats left to do.
Once there is a lifecycle release that solidifies the `platform API` and `buildpack API` needed to use assets we can gate the usage of asset caches around the following support table:

| pack used to build supports platform api >= 0.8 | pack used to create builder supports platform api >= 0.8 | lifecycle supports platform api >= 0.8 | conclusion                                                                                                                                                                                                                                      |
|-------------------------------------------------|----------------------------------------------------------|----------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| T                                               | T                                                        | T                                      | CNB_ASSETS can be set in build env! and buildpacks can use it                                                                                                                                                                                   |
| T                                               | T                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API Do not allow assets to be added to builders like this.                                                                                                                              |
| T                                               | F                                                        | T                                      | CNB_ASSETS will not be set on the builder :( and assets were not allowed on the  builder when it was created. But we should support adding assets using: "pack build ... --asset-package" at build time by adding CNB_ASSETS to the environment |
| T                                               | F                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API...                                                                                                                                                                                  |
| F                                               | T                                                        | T                                      | Cannot use assets as pack used to build does not support platform api 0.8                                                                                                                                                                       |
| F                                               | T                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API...                                                                                                                                                                                  |
| F                                               | F                                                        | T                                      | Cannot use assets as pack used to build does not support platform api 0.8                                                                                                                                                                       |
| F                                               | F                                                        | F                                      | CNB_ASSETS cannot be used as lifecycle does not support API...                                                                                                                                                                                  |

dwillist:
@dfreilich any chance you could take a look at this? I know its pretty large but would be nice to try and get this into the next release if possible.

acceptance/invoke/pack_fixtures.go
```diff
@@ -79,6 +79,25 @@ func (m PackFixtureManager) TemplateFixtureToFile(name string, destination *os.F
...
+	m.testObject.Helper()
+
+	outputTemplate, err := ioutil.ReadAll(file)
+	m.assert.Nil(err)
+
+	_, err = file.Seek(0, 0)
+	m.assert.Nil(err)
+
+	err = file.Truncate(0)
+	m.assert.Nil(err)
```
dfreilich, line 92: What do these do, together?

builder/config_reader.go
```diff
@@ -30,6 +31,23 @@ type BuildpackConfig struct {
 	dist.ImageOrURI
 }
 
+type Assets struct {
+	Packages []Package `toml:"package"`
+}
+
+type Package struct {
+	AssetImageOrURI
```
dfreilich, line 39: Do we need a dedicated `AssetImageOrURI`, or would `dist.ImageOrUri` work?

client.go
```diff
@@ -163,6 +182,13 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 		client.imageFactory = image.NewFactory(client.docker, authn.DefaultKeychain)
 	}
 
+	if client.assetFetcher == nil {
+		localAssetFetcher := asset.NewPackageFileFetcher()
+		imageAssetFetcher := asset.NewPackageImageFetcher(client.imageFetcher)
```
dfreilich, line 187: I find this line pretty odd, given that `client.assetFetcher` is `nil` in this block.

config/os.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import "fmt"
+
+const (
+	WindowsOS = "windows"
+	LinuxOS   = "linux"
+)
+
+func ValidateOS(os string) error {
```
dfreilich, line 10: This is Docker OS? ImageOS? or host machine OS? It's a bit unclear from the file alone

create_asset_package.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Minimum interface needed to successfully write an Asset Package in some format.
+type AssetPackage interface {
+	Save(additionalNames ...string) error
+	AddAssetBlobs(pairs ...asset.Blob)
+}
+
+// CreateAssetPackage writes a new Asset Package image using options specified in opts.
+// This image can be used to add assets to builds.
+func (c *Client) CreateAssetPackage(ctx context.Context, opts CreateAssetPackageOptions) error {
```
dfreilich, line 49: Overall, this func reads really well. It's decomposed super nicely. Great job!

create_asset_package.go
```diff
@@ -0,0 +1,139 @@
...
+		default:
+			result = append(result, asset)
+			prevAssetSha = asset.Sha256
+		}
+	}
+
+	return result
+}
+
+// this method mutates the given assetImg
```
dfreilich, line 125: This comment doesn't really add much

internal/asset/fetcher.go
```diff
@@ -0,0 +1,150 @@
...
+
+	pubcfg "github.com/buildpacks/pack/config"
+	"github.com/buildpacks/pack/internal/oci"
+)
+
+//go:generate mockgen -package testmocks -destination testmocks/mock_image_fetcher.go github.com/buildpacks/pack/internal/asset ImageFetcher
+// ImageFetcher is implemented by Fetcher which allows work with remote and local images,
+// as well as control when images are used locally vs pulled remotely.
+type ImageFetcher interface {
+
```
dfreilich, line 18:
```suggestion
```
nit

internal/asset/locator.go
```diff
@@ -0,0 +1,58 @@
...
+	return []string{
+		"InvalidLocator",
+		"URILocator",
+		"FilepathLocator",
+		"ImageLocator",
+	}[l]
+}
+
+// GetLocatorType parses a locator and returns the LocatorType
+func GetLocatorType(locator string, relativeBaseDir string) LocatorType {
```
dfreilich, line 34: Is there some way we can combine this with the existing `locator` searching mechanism, or is it totally unfit/unintuitive to do so?

builder/config_reader.go
dwillist: Ya I made a new object here because `dist.ImageOrURI` has a `buildpackURI` object inside of it. Which reads awkwardly when using assets.

client.go
dwillist: Could you elaborate? I'm having trouble understanding whats off here.

internal/asset/locator.go
dwillist: Definitly could but it would likely take quite a bit of work here to pull out all the `buildpack` locator specific logic. Was thinking this would be part of the larger refactoring story.

acceptance/invoke/pack_fixtures.go
dwillist: They delete all the files contents so that the order of operations in the `TemplateFile` function is 
1) read the file's contents into memory
2) delete them
3) re-write out the newly templated contents to the file. 

Note this is only done on files that have been moved into a `tmp` directory for testing

client.go
dfreilich: Ah never mind, I thought you were passing `client.assetFetcher` into the function, not `client.imageFetcher`

dfreilich (approved):
I've checked it over (not quite as in depth as I would like) and it definitely seems to be good. 

To be honest, I don't have the time allocated to the project right now to give it the in depth review it really needs (on the scale of 4+ hours, I would assume), and its current iteration as one massive PR isn't the most conducive for it being a comprehensively reviewed PR, unfortunately. At the same time, since it's gated with an `experimental` flag, I'm ok merging it in, assuming that there will be further refactors and simplifications/combinations (for instance, seeing whether we can combine some of the primitives which seem to be redefined, such as the separate Fetchers, locators, etc). 

Given that (and how long it's been waiting), I'll give it a approval now, but would definitely appreciate smaller PRs in the future 😅

dfreilich:
@dwillist Looks like there's some issues with acceptance tests which needs to be fixed before merging

PR #1103, opened 2021-03-08
Allow pack buildpack package to create buildpackages without package.…

### LINKED ISSUE #1082: pack buildpack package shouldn't require package.toml in all circumstances
### Description
We shouldn't require a `package.toml` file when we are:
1. packaging a single component buildpack
2. packaging a metabuildpack when all dependent buildpacks are available in the registry

This would be convenient for buildpack authors because there is one less file to maintain and update. A `package.toml` should not be necessary because the `buildpack.toml` file already contains all the information required to package the buildpack.

### Component Buildpack
Given a buildpack in a local directory `<dir>` _without_ an `order` key. I should be able to
```
pack buildpack package --buildpack <dir> my/buildpack
```
If no `--buildpack` flag is provided it should look for a `buildpack.toml` in the current directory.

### Multi Buildpack
Given a `buildpack.toml` in directory `<dir>` with an `order` that exclusively contains buildpacks that exist in an available registry. I should be able to:
```
pack buildpack package --buildpack <dir> my/buildpack
```

In this case pack should attempt to find all the dependent in the registry and fail with a helpful error message if it cannot locate one or more of them, suggesting that users provide a `package.toml` file.

### Partial `package.toml`
If a buildpack order contains multiple buildpacks some of which are available in a registry and some of which are not. Only buildpacks that cannot be located by id/version from the registry should be require an entry in `package.toml`.

NOTE: I have used directories in the above examples, but this should probably work for uri pointing to an unpackaged buildpack that we currently support in `package.toml`.

Keep creating `package.toml` files as we do today.

### Additional context

samj1912:
…toml for component buildpacks

## Summary

This change allows users to create buildpackages without a package.toml

It follows the behavior specified in https://github.com/buildpacks/pack/issues/1082

i.e.

If no package.toml is specified - 

it uses the current dir as the buildpack dir

if `--path` is specified it uses that dir

if both are specified it returns an error saying that their usage is incompatible together

#### Before

```
pack buildpack package --help                                                                                                                                                                                             
buildpack package allows users to package (a) buildpack(s) into OCI format, which can then to be hosted in image repositories. You can also package a number of buildpacks together, to enable easier distribution of a set of buildpacks. Packaged buildpacks can be used as inputs to `pack build` (using the `--buildpack` flag), and they can be included in the configs used in `pack builder create` and `pack buildpack package`. For more on how to package a buildpack, see: https://buildpacks.io/docs/buildpack-author-guide/package-a-buildpack/.

Usage:
  pack buildpack package <name> --config <config-path> [flags]

Examples:
pack buildpack package my-buildpack --config ./package.toml

Flags:
  -r, --buildpack-registry string   Buildpack Registry name
  -c, --config string               Path to package TOML config (required)
  -f, --format string               Format to save package as ("image" or "file")
  -h, --help                        Help for 'package'
      --publish                     Publish to registry (applies to "--format=image" only)
      --pull-policy string          Pull policy to use. Accepted values are always, never, and if-not-present. The default is always

Global Flags:
```

#### After

```
pack buildpack package --help    
buildpack package allows users to package (a) buildpack(s) into OCI format, which can then to be hosted in image repositories. You can also package a number of buildpacks together, to enable easier distribution of a set of buildpacks. Packaged buildpacks can be used as inputs to `pack build` (using the `--buildpack` flag), and they can be included in the configs used in `pack builder create` and `pack buildpack package`. For more on how to package a buildpack, see: https://buildpacks.io/docs/buildpack-author-guide/package-a-buildpack/.

Usage:
  pack buildpack package <name> --config <config-path> [flags]

Examples:
pack buildpack package my-buildpack --config ./package.toml

Flags:
  -r, --buildpack-registry string   Buildpack Registry name
  -c, --config string               Path to package TOML config
  -f, --format string               Format to save package as ("image" or "file")
  -h, --help                        Help for 'package'
  -p, --path string                 Path to the Buildpack that needs to be packaged
      --publish                     Publish to registry (applies to "--format=image" only)
      --pull-policy string          Pull policy to use. Accepted values are always, never, and if-not-present. The default is always

Global Flags:
```

## Documentation

    - [x] Yes, see https://github.com/buildpacks/docs/issues/321
    - [ ] No

## Related

Resolves part of #1082

### DIFF
--- a/internal/commands/buildpack_package.go
+++ b/internal/commands/buildpack_package.go
@@ -22,6 +22,7 @@ type BuildpackPackageFlags struct {
 	Publish           bool
 	Policy            string
 	BuildpackRegistry string
+	Path              string
 }
 
 // BuildpackPackager packages buildpacks
@@ -60,8 +61,14 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, client Buildpack
 			if err != nil {
 				return errors.Wrap(err, "parsing pull policy")
 			}
-
 			bpPackageCfg := pubbldpkg.DefaultConfig()
+			var bpPath string
+			if flags.Path != "" {
+				if bpPath, err = filepath.Abs(flags.Path); err != nil {
+					return errors.Wrap(err, "resolving buildpack path")
+				}
+				bpPackageCfg.Buildpack.URI = bpPath
+			}
 			relativeBaseDir := ""
 			if flags.PackageTomlPath != "" {
 				bpPackageCfg, err = packageConfigReader.Read(flags.PackageTomlPath)
@@ -74,7 +81,6 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, client Buildpack
 					return errors.Wrap(err, "getting absolute path for config")
 				}
 			}
-
 			name := args[0]
 			if err := client.PackageBuildpack(cmd.Context(), pack.PackageBuildpackOptions{
 				RelativeBaseDir: relativeBaseDir,
@@ -98,10 +104,11 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, client Buildpack
 		}),
 	}
 
-	cmd.Flags().StringVarP(&flags.PackageTomlPath, "config", "c", "", "Path to package TOML config (required)")
+	cmd.Flags().StringVarP(&flags.PackageTomlPath, "config", "c", "", "Path to package TOML config")
 	cmd.Flags().StringVarP(&flags.Format, "format", "f", "", `Format to save package as ("image" or "file")`)
 	cmd.Flags().BoolVar(&flags.Publish, "publish", false, `Publish to registry (applies to "--format=image" only)`)
 	cmd.Flags().StringVar(&flags.Policy, "pull-policy", "", "Pull policy to use. Accepted values are always, never, and if-not-present. The default is always")
+	cmd.Flags().StringVarP(&flags.Path, "path", "p", "", "Path to the Buildpack that needs to be packaged")
 	cmd.Flags().StringVarP(&flags.BuildpackRegistry, "buildpack-registry", "r", "", "Buildpack Registry name")
 
 	AddHelpFlag(cmd, "package")
@@ -111,6 +118,9 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, client Buildpack
 func validateBuildpackPackageFlags(p *BuildpackPackageFlags) error {
 	if p.Publish && p.Policy == pubcfg.PullNever.String() {
 		return errors.Errorf("--publish and --pull-policy never cannot be used together. The --publish flag requires the use of remote images.")
+	}
+	if p.PackageTomlPath != "" && p.Path != "" {
+		return errors.Errorf("--config and --path cannot be used together. Please specify the relative path to the Buildpack directory in the package config file.")
 	}
 
 	return nil
--- a/go.sum
+++ b/go.sum
@@ -95,7 +95,6 @@ github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+Ce
 github.com/bgentry/speakeasy v0.1.0/go.mod h1:+zsyZBPWlz7T6j88CTgSN5bM796AkVf0kBD4zp0CCIs=
 github.com/bketelsen/crypt v0.0.3-0.20200106085610-5cbc8cc4026c/go.mod h1:MKsuJmJgSg28kpZDP6UIiPt0e0Oz0kqKNGyRaWEPv84=
 github.com/blang/semver v3.5.0+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
-github.com/buildpacks/imgutil v0.0.0-20201211223552-8581300fe2b2 h1:tQWsNfyiBTEB6BaEs+sRxsbbo/XkNxCEqSWVwTq7I00=
 github.com/buildpacks/imgutil v0.0.0-20201211223552-8581300fe2b2/go.mod h1:NC93OGDehA2ksqgTzugeQcPqmTpilMPYRO+XaFsDyts=
 github.com/buildpacks/imgutil v0.0.0-20210209163614-30601e371ce3 h1:NuBOSLnHIE+gTjeRC/PAopad83eNT1YLRclBSIRYhKU=
 github.com/buildpacks/imgutil v0.0.0-20210209163614-30601e371ce3/go.mod h1:JiWXkmkaiXrr5nq77JUtu7ys/5OAznS4VI7TwKthP7s=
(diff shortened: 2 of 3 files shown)


dwillist:
## Acceptance Notes

### Setup
used the following buildpacks
- `node-engine` tarball [buildpack.tar.gz](https://github.com/buildpacks/pack/files/6103089/buildpack.tar.gz)
    unzip this file then run `pack buildpack package test-node-engine --buildpack /path/to/unarchived/buildpack` to create the `test-node-engine` buildpack
- `paketobuildpacks/yarn:0.1.2`
- `paketobuildpacks/yarn-install:0.2.3`
- `paketobuildpacks/yarn-start:0.0.4`

Build [this](https://github.com/paketo-buildpacks/samples/tree/main/nodejs/yarn) sample app using the above buildpacks.

### Output
a successfully built image 👍 


### Conclusion
This looks great!

internal/commands/buildpack_package.go
```diff
@@ -98,10 +106,11 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, client Buildpack
 		}),
 	}
 
-	cmd.Flags().StringVarP(&flags.PackageTomlPath, "config", "c", "", "Path to package TOML config (required)")
+	cmd.Flags().StringVarP(&flags.PackageTomlPath, "config", "c", "", "Path to package TOML config")
 	cmd.Flags().StringVarP(&flags.Format, "format", "f", "", `Format to save package as ("image" or "file")`)
 	cmd.Flags().BoolVar(&flags.Publish, "publish", false, `Publish to registry (applies to "--format=image" only)`)
 	cmd.Flags().StringVar(&flags.Policy, "pull-policy", "", "Pull policy to use. Accepted values are always, never, and if-not-present. The default is always")
+	cmd.Flags().StringVarP(&flags.Buildpack, "buildpack", "b", "", "Path to the Buildpack that needs to be packaged. If config is specified, the Buildpack URI is overridden to this value.")
```
jromero, line 113: Some questions about why this flag is necessary. 

Could we not just rename `config` to `path` and be able to provide a directory, buildpack.toml, or package.toml file?

I'm also concerned by it's behaviour "if config is specified, the buildpack URI is overriden". What's the use case in mind?

samj1912: Don't think I had a particular use case in mind. I tried to replicate the behavior that was specified in the issue. Should buildpack and config be mutually exclusive?

Also re:path, are there use cases where we would want to specify a path to the buildpack.toml instead of the buildpack dir? And for distinguishing the use case, should we just provide a path to a folder in general? if that contains a package.toml that would be picked, otherwise it would treat it like a folder with buildpack.toml, would this behavior work?

jromero: > Don't think I had a particular use case in mind. I tried to replicate the behavior that was specified in the issue. Should buildpack and config be mutually exclusive?

Ah, I see "If `--buildpack` and `--config` are provided, `--buildpack` should take precendence over `buildpack.uri` in `package.toml`.".  @ekcasey what's the thought behind that?

Would supporting multiple reference types using one flag support the desired use cases (while keeping a sensible UX)?

I'm thinking of `pack build ... --path` where path can be a path to a `jar` file that then gets treated as source code. 

In this particular case `--path` may be:

- a buildpack directory (in that directory we can search for a `package.toml` first, then a `buildpack.toml`)
- a `package.toml` file
- a `buildpack.toml` file

If we do end up deciding to go with `--path`, then I would think we would deprecate `--config`/`-c` which feels sort of wrong as well. Thoughts?

Related: https://github.com/buildpacks/pack/pull/932

samj1912: I would prefer to keep it to a directory instead of paths to specific toml files. That way the path config variable is consistent. It should then be upto pack to determine how to package the given directory. If it contains package.toml use that. If it contains buildpack.toml that is valid and can be standalone, it should be able to figure out the default package.toml (which is the behavior in this PR). And by default this path variable will be the current dir. This will make it consistent with something like pack build. Thoughts?

jromero: The reason why I proposed supporting pointing to files is because I don't think we should restrict what those files should be named. Could there be use cases for multiple flavors of a `package.toml`?

- `package.stack-a.toml` - packaging information specific for stack-a.
- `package.stack-b.toml` - packaging information specific for stack-b.

My example is pretty edgy but I think it's a principle of not forcing a filename that is important to me. Whether that's a shared sentiment or we feel is worth the trade off is up for debate.

Stepping back a little bit the alternative isn't bad either.

Let me try to summarize the options:

### Option 1: `--path` for all

- Support `--path` to point to any of the following:
    - a buildpack directory
    - a `package.toml`-formatted file
    - a `buildpack.toml`-formatted file
- When a directory is provided, search for `package.toml`, if not found fallback to `buildpack.toml` (#932 behaviour)
- Deprecate `--config` flag

### Option 2: `--config` for `project.toml`, `--path` for buildpack dir

- Support `--config` to point to a specific `package.toml`-formatted file
- Support `--path` to point to a specific buildpack **directory**
- Make `--config` and `--path` mutually exclusive
- When `--path` is provided, search for `package.toml`, if not found fallback to `buildpack.toml` (#932 behaviour)

samj1912: That makes sense, I never realized that package.toml was not a special file per the spec. I have a couple of questions WRT Option 1 - how will we detect if the given path is a package.toml file or not if it can be named something other than package.toml? If that is a usecase we wish to support then it might be easy to provide support for option #2. Thoughts?

ryanmoran: The `buildpack.uri` field in `package.toml` also supports a tarball buildpack file. Can we support that as well?

jromero: Based on our conversation during Sub-Team Sync, we are circling around Option 2 as the prefered method to eliminate the added complexity needed to detect `project.toml` vs `buildpack.toml`.

@ryanmoran, that sounds like a great idea. We'd probably want to punt that into a separate following issue though.

samj1912: Should be all done now - the flag has been renamed to --path and it has been made mutually exclusive to --config

jromero:
## ✅ Accepted

```
❯ pack-dev buildpack package --help
buildpack package allows users to package (a) buildpack(s) into OCI format, which can then to be hosted in image repositories. You can also package a number of buildpacks together, to enable easier distribution of a set of buildpacks. Packaged buildpacks can be used as inputs to `pack build` (using the `--buildpack` flag), and they can be included in the configs used in `pack builder create` and `pack buildpack package`. For more on how to package a buildpack, see: https://buildpacks.io/docs/buildpack-author-guide/package-a-buildpack/.

Usage:
  pack buildpack package <name> --config <config-path> [flags]

Examples:
pack buildpack package my-buildpack --config ./package.toml

Flags:
  -r, --buildpack-registry string   Buildpack Registry name
  -c, --config string               Path to package TOML config
  -f, --format string               Format to save package as ("image" or "file")
  -h, --help                        Help for 'package'
  -p, --path string                 Path to the Buildpack that needs to be packaged
      --publish                     Publish to registry (applies to "--format=image" only)
      --pull-policy string          Pull policy to use. Accepted values are always, never, and if-not-present. The default is always

Global Flags:

samples on  main [$!] on ☁️ rjavier@vmware.com
❯ pack-dev buildpack package my-package -c packages/hello-universe/
ERROR: reading config: decoding toml: read packages/hello-universe/: is a directory

samples on  main [$!] on ☁️ rjavier@vmware.com
❯ pack-dev buildpack package my-package -c packages/hello-universe/package.toml
Successfully created package my-package

samples on  main [$!] on ☁️ rjavier@vmware.com
❯ pack-dev buildpack package my-buildpack -p buildpacks/hello-world/
Successfully created package my-buildpack

samples on  main [$!] on ☁️ rjavier@vmware.com
❯ cat buildpacks/hello-universe/buildpack.toml
# Buildpack API version
api = "0.2"

# Buildpack ID and metadata
[buildpack]
id = "samples/hello-universe"
version = "0.0.1"
name = "Hello Universe Buildpack"
homepage = "https://github.com/buildpacks/samples/tree/main/buildpacks/hello-universe"

# Order used for detection
[[order]]
[[order.group]]
id = "samples/hello-world"
version = "0.0.1"

[[order.group]]
id = "samples/hello-moon"
version = "0.0.1"⏎

samples on  main [$!] on ☁️ rjavier@vmware.com
❯ pack-dev buildpack package my-buildpack -p buildpacks/hello-universe/
ERROR: saving image: buildpack samples/hello-universe@0.0.1 references buildpack samples/hello-world@0.0.1 which is not present

samples on  main [$!] on ☁️ rjavier@vmware.com
❯ pack-dev buildpack package my-buildpack -p buildpacks/hello-universe/ -c packages/hello-universe/package.toml
ERROR: --config and --path cannot be used together. Please specify the relative path to the Buildpack directory in the package config file.

```

ah yes

jromero (approved):

PR #1112, opened 2021-03-12
Build & release pipeline fixes

dwillist:

## Summary
This updates our build pipeline to work as there is no longer a `go 1.16.0-rc` version available. 

Additionally we should now correctly build `windows` and `macos_arm` binaries with versioning info

    - [x] No

## Related

Resolves #___

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,19 +54,39 @@ jobs:
           git config --global core.eol lf
           git config --global core.symlinks true
       - uses: actions/checkout@v2
-      - name: Derive pack version from branch name
+      - name: Derive pack version from branch name Unix
+        if: runner.os != 'Windows'
         run: |
           [[ $GITHUB_REF =~ ^refs\/heads\/release/(.*)$ ]] && version=${BASH_REMATCH[1]} || version=0.0.0
           echo "PACK_VERSION=${version}" >> $GITHUB_ENV
         shell: bash
+      - name: Derive pack version from branch name Windows
+        if: runner.os == 'Windows'
+        run: |
+          if ($Env:GITHUB_REF -match '^refs\/heads\/release/(.*)$') {
+            $refmatch=$Matches[1]
+            echo "PACK_VERSION=${refmatch}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+          }
+          else {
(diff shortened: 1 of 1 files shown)


dwillist:
@dfreilich @jromero hey if either of you get a chance could you review this, I need it merged in order to get a release candidate out.

jromero:
Could you explain why it was necessary to switch over to powershell for Windows? bash is available and keeping it the same for both platforms would be a good maintenance benefit. 

Additionally, why are the go variables needed to be set on Windows? Does the GitHub action for go not do that?

dwillist:
We have always been using powershell (as a fallback when pwsh) to run `make build` on windows, we were just only setting env vars for `bash`. 

:+1: on the gopath/gobin setup, Looks like a holdover from an old version of the action.

.github/workflows/build.yml
```diff
@@ -54,19 +54,27 @@ jobs:
           git config --global core.eol lf
           git config --global core.symlinks true
       - uses: actions/checkout@v2
-      - name: Derive pack version from branch name
+      - name: Derive pack version from branch name Unix
```
natalieparellano, line 57: On lifecycle we wrote a helper binary to do this: https://github.com/buildpacks/lifecycle/blob/11ca8ef99c1626fa34ea087e74993d8746c7cf61/tools/version/main.go#L13-L24 to avoid the cross-os headache. IDK if it's worth doing here though since you already have something working.

.github/workflows/build.yml
```diff
@@ -89,10 +97,22 @@ jobs:
...
+          echo "*Building*"
+          echo "${PACK_VERSION}"
+          make build
         env:
           PACK_BUILD: ${{ github.run_number }}
+        shell: bash
+      - name: Build Windows
+        if: runner.os == 'Windows'
+        run: |
+          make build
```
natalieparellano, line 112: Why not print the version here too?

natalieparellano (approved):

.github/workflows/build.yml
dwillist: Ya just going to remove this,  running the action will show these in the log already.

.github/workflows/build.yml
dwillist: Ohhh this is cool 👍

sclevine (approved):

PR #1141, opened 2021-04-20
Added workspace mount point as option to build

### LINKED ISSUE #1121: Need support in buildpacks to change the location where all the source code gets copied.
**On Page:** [Getting Started](https://buildpacks.io/docs/)

As of now by default , it seems there is no way to specify where in docker container soource code should be copied, default is  '/workspace' Can we please get a feature where we can override this as not everybody wants to maintain there code in /workspace.
https://github.com/buildpacks/pack/blob/aa49ae6d845f40bd576cc488a3105845e9e77506/internal/build/mount_paths.go#L36

jkutner:
## Summary

Allow the `workspace` directory to be configurable. 

#### Before

```
$ pack build ...
```

#### After

```
$ pack build --workspace app ...
```

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1121

### DIFF
--- a/build.go
+++ b/build.go
@@ -156,7 +156,7 @@ type BuildOptions struct {
 	// when using an untrusted builder.
 	LifecycleImage string
 
-	// The location at with to mount the AppDir in the build image.
+	// The location at which to mount the AppDir in the build image.
 	Workspace string
 }
 
--- a/build.go
+++ b/build.go
@@ -155,6 +155,9 @@ type BuildOptions struct {
 	// The lifecycle image that will be used for the analysis, restore and export phases
 	// when using an untrusted builder.
 	LifecycleImage string
+
+	// The location at with to mount the AppDir in the build image.
+	Workspace string
 }
 
 // ProxyConfig specifies proxy setting to be set as environment variables in a container.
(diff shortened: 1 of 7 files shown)


dwillist:
@jkutner Hey thanks for the PR, any chance we could get some associated testing here at whatever level you're comfortable with?

jkutner:
@dwillist done

build.go
```diff
@@ -155,6 +155,9 @@ type BuildOptions struct {
 	// The lifecycle image that will be used for the analysis, restore and export phases
 	// when using an untrusted builder.
 	LifecycleImage string
+
+	// The location at with to mount the AppDir in the build image.
```
dfreilich, line 159:
```suggestion
	// The location at which to mount the AppDir in the build image.
```

internal/build/lifecycle_execution_test.go
```diff
@@ -183,6 +183,37 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+				})
+				h.AssertNil(t, err)
+
+				h.AssertEq(t, len(fakePhaseFactory.NewCalledWithProvider), 1)
+
+				for _, entry := range fakePhaseFactory.NewCalledWithProvider {
+					if entry.Name() == "creator" {
+						h.AssertSliceContains(t, entry.ContainerConfig().Cmd, "/some/image")
+					}
+				}
```
dfreilich, line 213: Is there any way to verify in the phase creator that the mount path was returned correctly? This assertion doesn't seem super relevant to the change.

dfreilich (approved):
This is great! Happy to see this driven out, given the recent feedback on it. If you could just change the assertions in the test a bit to more accurately test the change, that would be 🤯

PR #1146, opened 2021-04-28
Pass mountPaths.appDirName to lifecycle as -app arg

jkutner:
## Summary

Fast follow fix for #1141

#### Before

app dir was mounted, but lifecycle was not `cd`-ing into it.

#### After

The provide workspace becomes the `CNB_APP_DIR`

## Documentation

    - [ ] Yes, see #___
    - [x] No

### DIFF
--- a/internal/build/lifecycle_execution.go
+++ b/internal/build/lifecycle_execution.go
@@ -193,7 +193,7 @@ func (l *LifecycleExecution) Create(ctx context.Context, publish bool, dockerHos
 
 	opts := []PhaseConfigProviderOperation{
 		WithFlags(l.withLogLevel(flags...)...),
-		WithArgs(repoName),
+		WithArgs("-app", l.mountPaths.appDirName(), repoName),
 		WithNetwork(networkMode),
 		cacheOpts,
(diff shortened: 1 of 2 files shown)


internal/build/lifecycle_execution.go
```diff
@@ -193,7 +193,7 @@ func (l *LifecycleExecution) Create(ctx context.Context, publish bool, dockerHos
 
 	opts := []PhaseConfigProviderOperation{
 		WithFlags(l.withLogLevel(flags...)...),
-		WithArgs(repoName),
+		WithArgs("-app", l.mountPaths.appDirName(), repoName),
```
dfreilich, line 196: Does lifecycle expect it in the format `\appdir`, or in `c:\appdir`? If the later, we can use `mountPaths.appDir()` instead

dfreilich (commented):
I should've thought of this when I merged in the other PR. Good catch!

internal/build/lifecycle_execution.go
jkutner: oh, i might need to revert 51887a3

jkutner: fixed in 9cc6613

dfreilich:
Closing this in favor of #1149

PR #1152, opened 2021-05-03
Removing redundant error checks before return

sreeharimohan:
Signed-off-by: Sreehari Mohan <sreeharimohan11@gmail.com>

## Summary
Simple bug fix to ensure there is no redundant code while returning `error`. Since the last function was only returning an `error` the error check is superfluous.

## Output
No change to the output

## Documentation
NA

    - [ ] Yes, see #___
    - [x] No

## Related
NA

Resolves #___


jromero (approved):
bug squashed

PR #1155, opened 2021-05-04
(bugsbash) fixes potential file inclusion via variable

rahulgrover99:
## Summary
gosec was showing the following warning "Potential file inclusion via variable". This is because we are trying to open files using dynamic variables. Hence, I've cleaned the bad file paths using filepath.Clean()

#### Before
ioutil.ReadFile(file)
#### After
ioutil.ReadFile(filepath.Clean(etagFile))
## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___
G304 Potential file inclusion via variable


dwillist (changes requested):
Hey @rahulgrover99 appreciate the PR! Looks like there are a few CI failures do to permission bits changing.

rahulgrover99:
I think it might be a good idea to revert the permission changes.

PR #1165, opened 2021-05-07
Removed weak random number generator

pankaj892:
Signed-off-by: Pankaj Biradar <pankaj28988@gmail.com>

### DIFF
--- a/testhelpers/testhelpers.go
+++ b/testhelpers/testhelpers.go
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
+	"crypto/rand"
 	"net/http"
 	"os"
 	"os/exec"


dwillist:
@pankaj892 think usage in the tests could stay the same. We do have a place where `math/rand` is used in the codebase [here](https://github.com/buildpacks/pack/blob/bfd0c734d95d44684ee9a520a04ffe1b23b41e32/internal/build/lifecycle_execution.go#L77) if you'd be up for swapping this out for `crypto/rand` that'd be 💯 .

pankaj892:
@dwillist should i create a new PR for the changes you suggested or should i add them in this PR itself

dwillist:
@pankaj892 just adding them here would be 🔥

dfreilich:
I was going to create a polish PR for this, but after looking at it some more, I don't think we should change (at this point) to using `crypto/rand` (at least in the example presented here). The reason being, we use `rand` here to generate strings, but we don't need them to be entirely secure for cryptographic reasons, and if we change to using it, we need to deal with the error and pass it up the call stack any time we use this method, and we would also be swallowing a slight performance cost in so doing (this is a bit of a red herring, but I assume there is some cost involved). 

I'll close it at this point, as a result, but we deeply appreciate your looking at and helping make our codebase secure, @pankaj892!

PR #1166, opened 2021-05-07
Added support for inline buildpacks

jkutner:
## Summary

This is an implementation of Inline Buildpacks described in [RFC-0048](https://github.com/buildpacks/rfcs/blob/main/text/0048-inline-buildpack.md).

## Output

Give an app with a `project.toml` containing:

```
[project]
id = "my-app"

[[build.buildpacks]]
id = "my/inline"

  [build.buildpacks.script]
  api = "0.4"
  inline = "echo 'hello!'"
```

#### Before

```
$ pack build inline-test
...
ERROR: failed to build: invalid buildpack string my/inline@
```

#### After

```
$ pack build inline-test
...
===> DETECTING
my/inline 0.0.0
===> ANALYZING
===> RESTORING
===> BUILDING
hello!
...
Successfully built image inline-test
```

## Documentation

    - [x] Yes, see https://github.com/buildpacks/docs/pull/365
    - [ ] No

### DIFF
--- a/build.go
+++ b/build.go
@@ -670,10 +670,10 @@ func (c *Client) processBuildpacks(ctx context.Context, builderImage imgutil.Ima
 				}
 				println(pathToInlineBuildpack)
 				declaredBPs = append(declaredBPs, pathToInlineBuildpack)
-			case bp.URI == "":
-				declaredBPs = append(declaredBPs, fmt.Sprintf("%s@%s", bp.ID, bp.Version))
-			case bp.Version != "":
+			case bp.URI != "":
 				declaredBPs = append(declaredBPs, bp.URI)
+			case bp.ID != "" && bp.Version != "":
+				declaredBPs = append(declaredBPs, fmt.Sprintf("%s@%s", bp.ID, bp.Version))
 			default:
 				return nil, nil, errors.New("Invalid buildpack defined in project descriptor")
 			}
--- a/build.go
+++ b/build.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
@@ -235,7 +236,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return err
 	}
 
-	fetchedBPs, order, err := c.processBuildpacks(ctx, bldr.Image(), bldr.Buildpacks(), bldr.Order(), opts)
+	fetchedBPs, order, err := c.processBuildpacks(ctx, bldr.Image(), bldr.Buildpacks(), bldr.Order(), bldr.StackID, opts)
 	if err != nil {
 		return err
 	}
@@ -645,7 +646,7 @@ func (c *Client) processProxyConfig(config *ProxyConfig) ProxyConfig {
 // 	----------
 // 	- group:
 //		- A
-func (c *Client) processBuildpacks(ctx context.Context, builderImage imgutil.Image, builderBPs []dist.BuildpackInfo, builderOrder dist.Order, opts BuildOptions) (fetchedBPs []dist.Buildpack, order dist.Order, err error) {
+func (c *Client) processBuildpacks(ctx context.Context, builderImage imgutil.Image, builderBPs []dist.BuildpackInfo, builderOrder dist.Order, stackID string, opts BuildOptions) (fetchedBPs []dist.Buildpack, order dist.Order, err error) {
 	pullPolicy := opts.PullPolicy
 	publish := opts.Publish
 	registry := opts.Registry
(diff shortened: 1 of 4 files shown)


build.go
```diff
@@ -641,10 +642,24 @@ func (c *Client) processBuildpacks(ctx context.Context, builderImage imgutil.Ima
...
+			case bp.ID != "" && bp.Script.Inline != "" && bp.Version == "" && bp.URI == "":
+				if bp.Script.API == "" {
+					return nil, nil, errors.New("Missing API version for inline buildpack")
+				}
+
+				pathToInlineBuildpack, err := createInlineBuildpack(bp, stackID)
+				if err != nil {
+					return nil, nil, errors.Wrap(err, "Could not create temporary inline buildpack")
+				}
+				println(pathToInlineBuildpack)
```
jromero, line 655: What is this? 🤔

jkutner: code gnomes put that there

build.go
```diff
@@ -889,3 +903,39 @@ func parseDigestFromImageID(id imgutil.Identifier) string {
...
+	if err = createBuildpackTOML(pathToInlineBuilpack, bp.ID, "0.0.0", bp.Script.API, []dist.Stack{{ID: stackID}}, nil); err != nil {
+		return pathToInlineBuilpack, err
+	}
+
+	shell := bp.Script.Shell
+	if shell == "" {
+		shell = "/bin/sh"
+	}
+
+	binBuild := fmt.Sprintf(`#!%s
```
dwillist, line 922: #! looks like it could cause problems on windows machines. And might be useful to have an integration test with one of these inline buildpacks just to show things do work in both envs.

jkutner: do you mean windows as the host machine, or windows containers (or both)?

dwillist: Windows Containers

jkutner: Should we have a different default shell for windows containers?

micahyoung: > Should we have a different default shell for windows containers?

Technically, Windows doesn't need an explicit shell chosen, nor a shebang line, just needs appropriate file suffixes (`detect.bat` / `build.bat`) since `cmd.Start()` will automatically call the all `.bat` files with `cmd.exe`. If the suffix is `.ps1`, they would be called with `powershell` (if it's present). 

Unfortunately though, the `#!` will be treated as invalid syntax, not as a comment (which are `:: ` or `REM ` in `.bat` files)

If it's possible to detect the container/daemon OS here, I think it's fine to assume the syntax is `.bat`. If it's not possible to detect, I'd also feel ok with as special case like `shell="cmd.exe"` triggering the Windows behavior (add `.bat` suffixes, remove shebang).

jkutner: @micahyoung  If I write both `bin/build` and `bin/build.bat` will the lifecycle just pick up the right one?

micahyoung: Yeah, that would work. I wrote a hybrid buildpack that does just that.

jkutner: I'll think about it. that would mean writing either an invalid `build` or `build.bat` in every case.

jkutner: I've updated this to create the hybrid buildpack (both `build` and `build.bat`)

jkutner:
@jromero most of the codecov warning are from error cases where I'm not checking some error from the file system (like creating a file). How should I handle that?

dfreilich:
@jkutner We disregard codecov/patch in most cases, as long as there are appropriate tests.

build.go
dfreilich: @jkutner I think that we would still need to remove the shebang in the windows case. https://github.com/buildpacks/pack/pull/1166/files#r634567908

build_test.go
```diff
@@ -1230,6 +1231,36 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 							{ID: "some-other-buildpack-id", Version: "some-other-buildpack-version"},
 						})
 					})
+
+					it("adds the buildpack form the project descriptor", func() {
```
dfreilich, line 1235:
```suggestion
					it("adds the buildpack from the project descriptor", func() {
```

new_buildpack.go
```diff
@@ -119,7 +88,50 @@ func createBinScript(path, name, contents string, c *Client) error {
 			return err
 		}
 
-		c.logger.Infof("    %s  bin/%s", style.Symbol("create"), name)
+		if c != nil {
+			c.logger.Infof("    %s  bin/%s", style.Symbol("create"), name)
```
dfreilich, line 92: When would the client be `nil`? Did you find this in your testing?

jkutner: It's `nil` when creating the temporary inline buildpack because we don't want to emit the log line. But when this is used for `pack buildpack new` it passes a Client.

build.go
jkutner: @dfreilich fixed in 81c66ff

dfreilich (approved):
Looks great for me! The UA you have in the initial post is perfect :D 

Can someone with access to Windows do UA for that? @micahyoung ?

aemengo:
> Looks great for me! The UA you have in the initial post is perfect :D
> 
> Can someone with access to Windows do UA for that? @micahyoung ?

I'll try to get this done.

aemengo (approved):

PR #1170, opened 2021-05-11
add gist action

dwillist:
Draft PR of what this could look like. Right now we just produce SVG gists. There is some work here to link thinks back onto the respective github issue.

### DIFF
--- a/.github/workflows/complexity.yml
+++ b/.github/workflows/complexity.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - 'release/**'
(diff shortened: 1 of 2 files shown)


dwillist:
content

dfreilich:
I'm assuming that it doesn't make sense to continue in this direction (we probably don't want to maintain a complexity action ourself). Perhaps we can look into using https://github.com/shoooooman/go-complexity-analysis-action if there's a continued need to display this, but I'll close it for now until there's a real ask.

PR #1173, opened 2021-05-11
Provide host port when starting the test registry

jromero:
## Summary

The latest version of Docker Desktop for Mac (>= 3.2.X) introduced a bug
where the `localhost` domain was not being resolved correctly when a bound
container port was not provided a host port. Providing a "0" host port
to continue using auto assignment resolves the issue.

#### Before

Screen Shot 2021-05-11 at 1 06 59 PM

#### After

Screen Shot 2021-05-11 at 1 08 41 PM


jromero:
Created an issue on https://github.com/docker/for-mac/issues/5678 to communicate bug upstream.

PR #1187, opened 2021-05-17
Bump to lifecycle v0.11.3

aemengo:
## Summary
Bump lifecycle to **v0.11.3**

## Documentation
    - [ ] Yes, see #___
    - [x] No

## Related
Resolves https://github.com/buildpacks/pack/pull/1169


rebase.go
```diff
@@ -78,7 +82,7 @@ func (c *Client) Rebase(ctx context.Context, opts RebaseOptions) error {
 	}
 
 	c.logger.Infof("Rebasing %s on run image %s", style.Symbol(appImage.Name()), style.Symbol(baseImage.Name()))
-	rebaser := &lifecycle.Rebaser{Logger: c.logger}
+	rebaser := &lifecycle.Rebaser{Logger: c.logger, PlatformAPI: api.Platform.Latest()}
```
aemengo, line 85: I don't really know what the correct move is here.

My rationale is that one usually get the platform API version from the builder, but there's no builder in this scenario.

jromero: My initial inclination was that it might have to do with the image being rebased but that was proven wrong when I looked at the [implementation](https://sourcegraph.com/github.com/buildpacks/lifecycle@a7428a55c2a14d8a37e84285b95dc63192e3264e/-/blob/rebaser.go#L91-94). It seems to only affect the output of the report.

That said, I don't think this should be `Latest`. I think **it should be hardcoded to a version that pack supports**. My fear is that this gets bumped when updating lifecycle version and _something_ unexpected changes. We should be conscious of all updates.

inspect_image.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"strings"
 
+	"github.com/buildpacks/lifecycle/platform"
+
```
jromero, line 8: Lint should be catching this but we should only have 3 import groups.

- stdlib
- dependencies
- local

IIRC, there's a setting in GoLand/IntelliJ to configure this.

🤔 , does running `make prepare-for-pr` automagically fix this?

inspect_image_test.go
```diff
@@ -7,12 +7,13 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/buildpacks/lifecycle/platform"
+
```
jromero, line 11: There should only be 3 import groups.

internal/commands/inspect_image_test.go
```diff
@@ -5,9 +5,9 @@ import (
 	"errors"
 	"testing"
 
-	"github.com/buildpacks/pack/internal/inspectimage"
+	"github.com/buildpacks/lifecycle/platform"
 
-	"github.com/buildpacks/lifecycle"
+	"github.com/buildpacks/pack/internal/inspectimage"
```
jromero, line 10: There should only be 3 import groups.

internal/inspectimage/writer/bom_json_test.go
```diff
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"testing"
 
+	"github.com/buildpacks/lifecycle/buildpack"
```
jromero, line 7: There should only be 3 import groups.

jromero (changes requested):

inspect_image.go
aemengo: it seems to have no effect here.

jromero:
Adding comment from [slack](https://buildpacks.slack.com/archives/CD61YAG69/p1621608483001600?thread_ts=1621516919.001400&cid=CD61YAG69):

> TLDR; It looks like an actual error from the branch changes not the worker.
> 
> Recreated the WCOW worker based on the assumption that it was the worker due to what looked like a low-level error. After recreation the errors are still present.
> 
> ERROR: saving image: failed to write image to the following tags: [10.32.101.130:65324/simple-layers-package-image-buildpack-kspcaxug: loading image “10.32.101.130:65324/simple-layers-package-image-buildpack-kspcaxug”. first error: embedded daemon response: re-exec error: exit status 1: output: hcsshim::ImportLayer - failed failed in Win32: The configuration registry database is corrupt. (0x3f1)]
> 
> Afterwards, I reran the tests on a different branch and the errors are not present. That leads me to believe that a change in this branch broke something.

aemengo:
@jromero Apologies, I did introduce failures on the **wcow** runner  with this PR.

I pulled in a commit from @micahyoung that should address these failures.

dfreilich (approved):
🥳

PR #1190, opened 2021-05-21
Adding --gid flag to override the user's group id

### LINKED ISSUE #961: Add -gid Flag to Pack to Support OpenShift

jjbustamante:
Signed-off-by: Juan Bustamante <jbustamante@vmware.com>

## Summary

This fix is using the lifecycle --gid flag to override the user's group id in the build and runtime image

## Output

- When the image is build using `--gid` flag

```
➜  pack git:(feature/add-gid-flag-to-support-openshift) ./out/pack build my-app-imag --path ../my-go-web --builder heroku/buildpacks:18 --gid=0

18: Pulling from heroku/buildpacks
Digest: sha256:998e6edd2c9a72ec28701ac95800b3ab5f58b557a447f12d798ab28d727d8bd3
Status: Image is up to date for heroku/buildpacks:18
18: Pulling from heroku/pack
Digest: sha256:57f1fe599095b485a0a5dd14306f14a781d1257bdfbf2fe3bdaf9b7608eed238
Status: Image is up to date for heroku/pack:18
===> DETECTING
======== Output: heroku/ruby@0.0.1 ========
no
err:  heroku/ruby@0.0.1 (1)
heroku/go       0.3.1
heroku/procfile 0.6.2
===> ANALYZING
Restoring metadata for "heroku/go:profile" from app image
Restoring metadata for "heroku/go:shim" from cache
===> RESTORING
Restoring data for "heroku/go:shim" from cache
===> BUILDING
-----> Fetching stdlib.sh.v8... done
----->
       Detected go modules via go.mod
----->
       Detected Module Name: examples.com/my-go-web-app
----->
 !!    The go.mod file for this project does not specify a Go version
 !!
 !!    Defaulting to go1.12.17
 !!
 !!    For more details see: https://devcenter.heroku.com/articles/go-apps-with-modules#build-configuration
 !!
-----> Using go1.12.17
-----> Determining packages to install

       Detected the following main packages to install:
       		examples.com/my-go-web-app

-----> Running: go install -v -tags heroku examples.com/my-go-web-app

       Installed the following binaries:
       		./bin/my-go-web-app

       Created a Procfile with the following entries:
       		web: bin/my-go-web-app

       If these entries look incomplete or incorrect please create a Procfile with the required entries.
       See https://devcenter.heroku.com/articles/procfile for more details about Procfiles

[INFO] Discovering process types
[INFO] Procfile declares types -> web
===> EXPORTING
Reusing layer 'heroku/go:profile'
Reusing 1/1 app layer(s)
Reusing layer 'launcher'
Reusing layer 'config'
Reusing layer 'process-types'
Adding label 'io.buildpacks.lifecycle.metadata'
Adding label 'io.buildpacks.build.metadata'
Adding label 'io.buildpacks.project.metadata'
Setting default process type 'web'
Saving my-app-imag...
*** Images (f43214b114de):
      my-app-imag
Adding cache layer 'heroku/go:shim'
Successfully built image my-app-imag
```

- When the --gid is a negative value, the following error message is shown
Screen Shot 2021-05-21 at 4 44 46 PM

#### Before

So far, the user's group id must be overridden using the `CNB_GROUP_ID` environment variable, in the image I used for testing the group id belongs to the same user. Check the following image

Screen Shot 2021-05-21 at 4 58 42 PM

#### After

Once the image is built and a container is running using that image, the following screenshot shows how the group for the workspace folder was configured with a different owner

Screen Shot 2021-05-21 at 4 47 26 PM

## Documentation

    - [x] No

## Related
Resolves #961


internal/commands/build.go
```diff
@@ -184,6 +191,7 @@ This option may set DOCKER_HOST environment variable for the build container if
 	cmd.Flags().BoolVar(&buildFlags.TrustBuilder, "trust-builder", false, "Trust the provided builder\nAll lifecycle phases will be run in a single container (if supported by the lifecycle).")
 	cmd.Flags().StringArrayVar(&buildFlags.Volumes, "volume", nil, "Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.\n- 'host path': Name of the volume or absolute directory path to mount.\n- 'target path': The path where the file or directory is available in the container.\n- 'options' (default \"ro\"): An optional comma separated list of mount options.\n    - \"ro\", volume contents are read-only.\n    - \"rw\", volume contents are readable and writeable.\n    - \"volume-opt=<key>=<value>\", can be specified more than once, takes a key-value pair consisting of the option name and its value."+multiValueHelp("volume"))
 	cmd.Flags().StringVar(&buildFlags.Workspace, "workspace", "", "Location at which to mount the app dir in the build image")
+	cmd.Flags().IntVar(&buildFlags.GID, "gid", 0, `Override GID of user's group in the stack's build and run images, it must be a value in the range of 0-2147483647`)
```
jromero, line 194:
```suggestion
	cmd.Flags().IntVar(&buildFlags.GID, "gid", 0, `Override GID of user's group in the stack's build and run images`)
```

This may be a difference in opinion but I don't think it's necessary to explain what the possible values are provided that rather basic information that is more dependant on the underlying usage.

build.go
```diff
@@ -122,6 +122,9 @@ type BuildOptions struct {
 	// Only trust builders from reputable sources.
 	TrustBuilder bool
 
+	// The build and stack user's group id must be override
+	OverrideGroupID bool
```
jromero, line 126: I'm not particularly a fan of having two fields that must be used in unison. I know we opted out of using `-1` to signify "don't override" from a user's perspective but I think it's perfectly fine to use `-1` internally. Then we don't have to worry about two state variables.

IOW, we can keep the current logic in `cmd` package where it looks for the presence of the `gid` flag but instead of using `OverrideGroupID`, it updates `GroupID` to `-1` to signify that we shouldn't override it on the image or provide it to the lifecycle. That should achieve the same behaviour right?

jromero (changes requested):

jromero:
@dwillist, can we try to squeeze this one in this release? I know it's fairly important for OpenShift users.

build.go
jjbustamante: @jromero Yeap, I agreed, we can keep the same logic using that approach. Good one!

internal/build/lifecycle_execution.go
```diff
@@ -181,6 +182,10 @@ func (l *LifecycleExecution) Create(ctx context.Context, publish bool, dockerHos
 		flags = append(flags, "-skip-restore")
 	}
 
+	if l.opts.GID >= 0 {
```
dfreilich, line 185: Could we pull this `0` inot a constant, so it's not a magic number? For all of the GID comparisons in this file

build_test.go
```diff
@@ -2191,6 +2191,44 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+						Workspace:       "app",
+						Builder:         defaultBuilderName,
+						Image:           "example.com/some/repo:tag",
+						GroupID:         -1,
+					}))
+					h.AssertEq(t, fakeLifecycle.Opts.GID, -1)
+				})
+			})
+
+			when("gid is greater than zero", func() {
```
dfreilich, line 2220: I kinda feel like this test alone is sufficient, given that there's no actual GID manipulating logic in the current implementation for `build.go`.  In general, testing a lot is something I'm a big fan of, but I don't feel like there's so much value added by the <=0 tests. What do you think?

internal/build/lifecycle_execution_test.go
```diff
@@ -714,6 +714,48 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 				})
 			})
 		})
+
+		when("override GID", func() {
+			when("override GID is provided", func() {
+				it("configures the phase with the expected arguments", func() {
+					verboseLifecycle := newTestLifecycleExec(t, true, func(options *build.LifecycleOptions) {
+						options.GID = 0
```
dfreilich, line 722:
```suggestion
						options.GID = 2
```

internal/build/lifecycle_execution_test.go
```diff
@@ -714,6 +714,48 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+					h.AssertNil(t, err)
+
+					lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
+					h.AssertNotEq(t, lastCallIndex, -1)
+
+					configProvider := fakePhaseFactory.NewCalledWithProvider[lastCallIndex]
+					h.AssertEq(t, configProvider.Name(), "creator")
+					h.AssertIncludeAllExpectedPatterns(t,
+						configProvider.ContainerConfig().Cmd,
+						[]string{"-gid"},
```
dfreilich, line 736: Also assert for the `gid` passed through?

internal/build/lifecycle_execution_test.go
```diff
@@ -714,6 +714,48 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+					)
+				})
+			})
+			when("override GID is not provided", func() {
```
dfreilich, line 740: I'm a bit confused by the differences between the previous test and this. Both seem to be doing the same thing, so what's happening here?

internal/commands/build.go
```diff
@@ -184,6 +190,7 @@ This option may set DOCKER_HOST environment variable for the build container if
 	cmd.Flags().BoolVar(&buildFlags.TrustBuilder, "trust-builder", false, "Trust the provided builder\nAll lifecycle phases will be run in a single container (if supported by the lifecycle).")
 	cmd.Flags().StringArrayVar(&buildFlags.Volumes, "volume", nil, "Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.\n- 'host path': Name of the volume or absolute directory path to mount.\n- 'target path': The path where the file or directory is available in the container.\n- 'options' (default \"ro\"): An optional comma separated list of mount options.\n    - \"ro\", volume contents are read-only.\n    - \"rw\", volume contents are readable and writeable.\n    - \"volume-opt=<key>=<value>\", can be specified more than once, takes a key-value pair consisting of the option name and its value."+multiValueHelp("volume"))
 	cmd.Flags().StringVar(&buildFlags.Workspace, "workspace", "", "Location at which to mount the app dir in the build image")
+	cmd.Flags().IntVar(&buildFlags.GID, "gid", 0, `Override GID of user's group in the stack's build and run images`)
```
dfreilich, line 193:
```suggestion
	cmd.Flags().IntVar(&buildFlags.GID, "gid", 0, `Override GID of user's group in the stack's build and run images. The provided value must be a positive number`)
```

dfreilich (commented):
Thanks for the PR, @jjbustamante ! 

This looks like a really neat PR. I was slightly confused by the tests in lifecycle_execution_test, so I'd appreciate a bit of clarification there, but overall great work!

build_test.go
jjbustamante: @dfreilich, yeah after removing the boolean attribute I also think these tests are not adding anything valuable. I will remove them

internal/build/lifecycle_execution.go
jjbustamante: Sure. @dfreilich

internal/build/lifecycle_execution_test.go
jjbustamante: Oh! you are right, but it wasn't the intention, let me fix it, the idea was to send a negative GID and verify the flag -gid was not added. My mistake after removing the boolean attribute

dfreilich (approved):
Great job!

# UA
## With `--gid 0`
image

## With `--gid 15`
image

You can see that the user was changed from `cnb` to `root` (in the case of `0`) and `kmem` in the case of `15`.

dfreilich:
@jjbustamante It looks like one of the tests is failing. Once you fix that, I can merge it in.

jjbustamante:
> @jjbustamante It looks like one of the tests is failing. Once you fix that, I can merge it in.

Oh! I will take a look after the working group!

dfreilich (approved):

PR #1191, opened 2021-05-24
Add a registry pull proxy

dwillist:

#### Before
All image pulls would come from the registry specified in their name Eg `index.docker.io/repo/name:latest` would pull from `index.docker.io`

#### After
You can specify a pull proxy that re-directs all image pulls. If passed the `--pull-proxy index.proxy.io/proxy_registry` all image pulls will be routed through this proxy instead of their original. 

## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

Resolves #___

### DIFF
--- a/build.go
+++ b/build.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"github.com/buildpacks/pack/internal/image"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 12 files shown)


dwillist:
Just to document the state of affairs in the PR. 

### Whats implemented so far.
The changes needed to be able to `pack build` with and pass a `--pull-proxy` flag so that all image pulls go through this proxy. 

In general the other pack commands will likely need to have a optional `--pull-proxy` flag added as well and need to have logic similar to [this](https://github.com/buildpacks/pack/blob/27de4edea31f1605287580f670388922cb6c65d5/build.go#L199-L201) so that image pulls are forced through the proxy.

Would recommend setting up [harbor](https://goharbor.io/docs/2.1.0/administration/configure-proxy-cache/) to test these other commands. 

Other commands:
- inspect image
- builder create
- builder inspect
- buildpack inspect
- buildpack package
- rebase

dfreilich:
Closing in favor of #1210

Issue #897, opened 2020-10-16
Configurable Previous Image


ekcasey:
### Description
`pack build` should accept am optional `--previous-image` flag that allows users to set the previous image to a particular tag reference, digest reference, or (when performing a daemon build) image ID.

This is useful If I want to rebuild from a particular previous version of my app. Currently I would need to tag the previous image with the tag to which I want to export to force this behavior. This is an extra and potentially undesirable step.

### Proposed solution
 Usage: `pack build --previous-image <prev-image> <image>`

#### Lifecycle Arguments
When the builder is trusted, `pack` should invoke `creator` with the `-previous-image` flag or `CNB_PREVIOUS_IMAGE` env var.
```
creator -previous-image <prev-image> <image>
```
When the builder is not truster, `pack` should pass the previous image to `analyzer`
```
analyzer <prev-image>
```
#### Edge Cases
When `pack build` is called with `--publish`, `<prev-image>` must be in the same image registry as `<image>` (`exporter` currently does not support previous images in a different registry).

Until we have a better solution for managing caches volumes this feature will only be useful for reusing `launch = true`, `cache = false` layers, because the previous cache is not necessarily provided.

### Additional context
- [ ] This feature should be documented somewhere

micahyoung:
This feels a bit like `docker build`'s [`--cache-from`](https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources) option, which really speeds up rebuilding Windows images.

samj1912:
does `pack build --cache-image` implement the required functionality for this? In which case should this be closed?

ekcasey:
@samj1912 some layer metadata comes from the previous app image, rather than from the cache image.

How these options interact might be a bit confusing for users but I think it is worth exposing this useful knobs now (maybe as an experimental feature?), and we can brainstorm a more cohesive UX for future iterations.

importhuman:
Hello! I'd like to work on this. Can you give me some pointers as to where I should start? I know some Go, and the basics of Docker and how to use the pack CLI.

jromero:
Hi @importhuman, very excited to hear that this is something you are interested in contributing.

Here are a few steps I can think off:

Prerequisites:
1. Read our DEVELOPMENT.md document and install all the [prerequisites](https://github.com/buildpacks/pack/blob/main/DEVELOPMENT.md#prerequisites).
1. Make sure you can run `make test` and all tests pass.

Development:
[...]

PR #1198, opened 2021-06-09
Add previous-image flag

importhuman:
Signed-off-by: Ujjwal Goyal <importujjwal@gmail.com>

## Summary

Add `previous-image` flag for #897 with an `Images` field. I have started with the changes required to the `analyze` phase, but am having trouble writing the test for it (passing when it should fail). 

Additional assistance required:
- cannot run `make format`, `make tidy` or `make prepare-for-pr`, getting the error `"./create_builder.go" contains more than 3 groups of imports`.
- observing some untracked files with the name format "internal/registry/registry177994269/" while making changes, what to do with them? 
- unclear about when to run `make prepare-for-pr`. Does "before creating a Pull Request" mean before adding and committing?

## Documentation

    - [ ] Yes, see #897 
   
## Related

Concerns #897 (Doesn't resolve yet)


aemengo:
Thanks so much for helping out!

> getting the error "./create_builder.go" contains more than 3 groups of imports.

Wild. I never see that error, but the linter is correct. The imports should be: **stdlib**, **external dependencies**, **internal dependencies**. You didn't introduce the error as part of this PR so I don't think it's worth your time, _in this PR_. See: https://github.com/buildpacks/pack/pull/1187#discussion_r634624030

> observing some untracked files with the name format "internal/registry/registry177994269/" while making changes, what to do with them?

Please just ignore them. We definitely need to cleanup, our cleanup logic there. 😅

> unclear about when to run make prepare-for-pr. Does "before creating a Pull Request" mean before adding and committing?

After adding and committing, but before _git push_.

build.go
```diff
@@ -159,6 +159,11 @@ type BuildOptions struct {
 
 	// User's group id used to build the image
 	GroupID int
+
+	// Slice of two images
+	// 1. A previous image to set to a particular tag reference, digest reference, or image ID;
+	// 2. The image to refer to
+	Images []string
```
aemengo, line 166: I'm a little concerned about the direction that this PR is headed.. The previous image flag is intended to have space for one `string`, not two. If you see https://github.com/buildpacks/pack/blob/5ea76488f9c16bca671dd417da1f311f423a7d8b/build.go#L72, you'll see where that `<image-name>` was _already_ being collected as part of the pack build command.

Happy to clarify, if I wasn't clear!

aemengo: Forgive me, I'm having trouble understanding.

If you're asking how do I know `PreviousImage` was passed in by a user? You could assert that the value is not equal to an empty string (`""`). 

If you wanna talk about these things in a more realtime way, we're all pretty active at: https://slack.buildpacks.io/

dfreilich (commented):
In terms of code, @importhuman , this looks superb. Fits with our coding styles, and is very clean and pristine. Good work! 

In terms of UX, I'm a bit concerned that people may not understand exactly what this option is (I would have difficulty explaining when to use this or not to use it). @ekcasey  or @natalieparellano , can we have some further explanation of how/why people would want to use this, which we can add to the flag description and make it a bit clearer?

importhuman:
@dfreilich Thanks! Trying to figure out the edge case now. Also, not working on the analyzer scenario in this PR, as suggested by @natalieparellano on slack.

I don't really understand how to use myself either, I've been going along what was given in the description and further discussions. 😅

acceptance/acceptance_test.go
```diff
@@ -1764,6 +1764,45 @@ func testAcceptance(
 						})
 					})
```
importhuman, line 1766: Couldn't figure out how to write the acceptance test. This is based off the cache-image test, but failing.

internal/build/lifecycle_execution_test.go
```diff
@@ -800,20 +800,95 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
...
+					h.AssertError(t, err, fmt.Sprintf("%s", err))
 				})
-				fakePhaseFactory := fakes.NewFakePhaseFactory()
-				err := lifecycle.Create(context.Background(), false, "", false, "test", "test", "test", fakeBuildCache, fakeLaunchCache, []string{}, []string{}, fakePhaseFactory)
-				h.AssertNil(t, err)
+			})
 
-				lastCallIndex := len(fakePhaseFactory.NewCalledWithProvider) - 1
-				h.AssertNotEq(t, lastCallIndex, -1)
+			// when("image is invalid", func() {
```
importhuman, line 859: Tried to add a unit test to check for invalid images, but couldn't figure out how to pass an invalid image name in the test as it gets validated. Tried passing tags like "%%%", tags with other invalid characters, tags with multiple slashes, etc.

internal/commands/build_test.go
```diff
@@ -709,6 +709,40 @@ builder = "my-builder"
...
+					it("succeeds", func() {
+						mockClient.EXPECT().
+							Build(gomock.Any(), EqBuildOptionsWithPreviousImage("index.docker.io/some/previous:latest")).
+							Return(nil)
+
+						command.SetArgs([]string{"--builder", "my-builder", "index.docker.io/some/image:latest", "--previous-image", "index.docker.io/some/previous:latest", "--publish"})
+						h.AssertNil(t, command.Execute())
+					})
+				})
+
```
importhuman, line 736: Couldn't figure out this test case, get the following error message:
`=== RUN   TestBuildCommand/Commands/#BuildCommand/previous-image_flag_is_provided/--publish_is_true/image_and_previous-image_are_not_in_same_registry/fails
    build.go:130: Unexpected call to *testmocks.MockPackClient.Build([context.Background { index.docker.io/some/image:latest my-builder     map[] map[]  true false false [] [] <nil> { []}  always . {{ []} {[] [] [] [] } map[]}   -1 example.io/some/previous-image}]) at /Users/ujjwal/coding/pack/internal/commands/build.go:130 because: there are no expected calls of the method "Build" for that receiver`

importhuman (commented):
Having some trouble with writing certain tests 😅

acceptance/acceptance_test.go
aemengo: Let's not bother with an acceptance test for this.

They're slower, require more resources, and are more likely to fail. We should reserve them for more critical code paths with more moving parts. Although this PR touches a few moving parts, we have plenty other acceptance tests that the reduce need for this one.

internal/build/lifecycle_execution_test.go
aemengo: Once again, I'd rather not bother with a test for this 😅 

A test to check that we're bubbling up the error would be of _low value_, imo. Such a test would be more valuable if it were _our logic_ deciding what a valid/invalid image tag is.

In this case, I'd be comfortable having faith in our third-party library having its own tests. And then saving ourselves the keystrokes for every error check.

internal/commands/build_test.go
aemengo: Mock assertions are always painful to read. In this case, it's complaining that your mock received a call, but your _test setup_ did not have the relevant `mockClient.EXPECT()` line.

Stepping back for a second. This unit test doesn't feel appropriate here. The logic for actually creating the error is way down in `internal/build/lifecycle_execution.go`. It feels like there's a decision to make about moving the test down or bringing the logic up...

I'd much prefer that you bring this logic (_and tests_) up. In fact, I'd prefer them to be at: **internal/commands/build.go**. That way we don't unnecessarily pull images, perform downloads, etc and then the action will fail anyway.

I'm glad that you tackled the edge case for us though 🙂

acceptance/acceptance_test.go
importhuman: @aemengo I did observe that it took a bit of time. I added the acceptance test because it was suggested to do so in the [issue](https://github.com/buildpacks/pack/issues/897#issuecomment-853059556). If @jromero says it's alright without this test then I agree that we can skip it.

internal/build/lifecycle_execution_test.go
importhuman: Cool, will remove.

internal/commands/build_test.go
importhuman: Again, I added the test here because it was mentioned in the [issue](https://github.com/buildpacks/pack/issues/897#issuecomment-853059556). 😅 

I believe I can add the logic to `internal/commands/build.go`, but I don't get how to move the tests. Aren't the tests for this file in `internal/commands/build_test.go`, which is the file we're talking about here? Or would the logic of checking for invalid previous-image passed remove the need for a test? 😅 

cc: @jromero 

But for the sake of better understanding, I'd like to understand the error message.
> Mock assertions are always painful to read. In this case, it's complaining that your mock received a call, but your test setup did not have the relevant mockClient.EXPECT() line.

When I have the `mockClient` line, I got the error "expected error but got nil". I tried it with various registry names, but got the same result. For example, with previous image name `example.io/some/previous-image` and image name `index.docker.io/some/image:latest`. Apologies if this information is insufficient.

Also, thank you so much for helping me with this PR, really appreciate it 😄

jromero: Just sent you a DM on slack if you want to do a bit of pair programing to get past this little hurdle.

importhuman:
Some tests are failing in the `TestLifecycleExecution` suite, but none of the failing tests are ones I've added. I updated my local branch with the latest changes in the main branch, could this be the reason?
cc: @jromero

importhuman:
@jromero How can I fix the tests failing at the "no test files" stage?  Other than this, the PR should be good to go. I've put back the image validation portion inside the condition as it was before, otherwise the tests were failing since not all of them were provided with images.

jromero (approved):

Issue #821, opened 2020-08-26
Support typical enterprise proxy and internal mirrors setup


rmohta:
### Description
It would be nice (and increase adoption) if pack cli can support typical enterprise proxy and mirror setup.

### Proposed solution
Without changing the toml files or re-building images, pack cli should be able to update the URL to download docker images used in buildpacks and stack settings. 
For example, http_proxy environment variables are inherited during the build process - but not everything would have to go via http_proxy. It's common for an enterprise to have internal mirrors using Artifactory or Nexus, so that engineers don't have to fiddle with proxy settings for different types of applications used across different tech stacks.

Similar to *--env*, if we have *--docker-mirror-registry* or *--docker-proxy-registry*, which would allow us to provide the prefix that can be used when downloading images.

### Describe alternatives you've considered
Re-build sample applications by adding our docker registry.

### Additional context
- [ ] This feature should be documented somewhere

jromero:
I thought I had replied to this... 

Images are typically declared including the hostname. A format such as `<hostname>/<repo>[(:<tag>|@<hash>)]`, for example `gcr.io/my-org/my-image:latest`.

If I'm understanding of how this would work with something like [Artifactory's virtual repos](https://www.jfrog.com/confluence/display/JFROG/Docker+Registry#DockerRegistry-VirtualDockerRepositories) or [Nexus's proxy](https://help.sonatype.com/repomanager3/formats/docker-registry/proxy-repository-for-docker)...

If a `--docker-mirror-registry` of `artifactory.my-company.org/docker-repo` was configured and a request for `gcr.io/my-org/my-image:latest` was made,  the actual request for the image should be `artifactory.my-company.org/docker-repo/my-org/my-image:latest`. The virtual registry would in turn search one or more registries based on how it's configured.

Ultimately we'd be removing the original `<hostname>` and replacing it with the `--docker-mirror-registry` value.

Does that sound right?

Additional references:
- https://help.sonatype.com/repomanager3/formats/docker-registry/pulling-images

jromero:
Another relevant conversation about being able to push to these proxies/mirrors. I suspect this may or may not work depending on the service (artifactory, nexus, harbor).


* https://github.com/goharbor/harbor/issues/220
* https://github.com/docker/distribution/issues/1218

rmohta:
[...]

PR #1210, opened 2021-06-16
Add registry mirror support for pack build

aemengo:
## Summary
Add support for a registry mirroring  during **pack build** command. The use-case being enterprise environments where public registry access is denied.

If pack's `config.toml` is configured like:
```toml
[registry-mirrors]
  "index.docker.io" = "nexus.mycompany.local/mirror"
```

Then requests from registries`paketobuildpacks/builder` will be turned into `nexus.mycompany.local/mirror/paketobuildpacks/builder`

This PR also adds the config subcommands to manipulate these entries to config.toml.

#### Before
```bash
$ pack build aemengo/hello
full: Pulling from paketobuildpacks/builder
Digest: sha256:2917564c2de52debd1df70d11c014b2708c8974e6e853804958825be84cfb859
Status: Image is up to date for paketobuildpacks/builder:full
...
```
#### After

```bash
$ pack build aemengo/hello
full: Pulling from mirror/paketobuildpacks/builder
Digest: sha256:2917564c2de52debd1df70d11c014b2708c8974e6e853804958825be84cfb859
Status: Image is up to date for nexus.mycompany.local/mirror/dockerhub-mirror/paketobuildpacks/builder:full
...
```

## Caveats
* The user entered `image-tag` is not modified. Not sure if this is desired behavior or not. 🤷🏾‍♂️

## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

Resolves https://github.com/buildpacks/pack/issues/821

### DIFF
--- a/internal/image/fetcher.go
+++ b/internal/image/fetcher.go
@@ -63,7 +63,7 @@ func NewFetcher(logger logging.Logger, docker client.CommonAPIClient, opts ...Fe
 var ErrNotFound = errors.New("not found")
 
 func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions) (imgutil.Image, error) {
-	name, err := pname.TranslateRegistry(name, f.registryMirrors)
+	name, err := pname.TranslateRegistry(name, f.registryMirrors, f.logger)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/image/fetcher.go
+++ b/internal/image/fetcher.go
@@ -19,13 +19,26 @@ import (
 
 	"github.com/buildpacks/pack/config"
 	ilogging "github.com/buildpacks/pack/internal/logging"
+	pname "github.com/buildpacks/pack/internal/name"
 	"github.com/buildpacks/pack/internal/style"
 	"github.com/buildpacks/pack/logging"
 )
 
+// FetcherOption is a type of function that mutate settings on the client.
+// Values in these functions are set through currying.
+type FetcherOption func(c *Fetcher)
+
+// WithRegistryMirrors supply your own mirrors for registry.
+func WithRegistryMirrors(registryMirrors map[string]string) FetcherOption {
+	return func(c *Fetcher) {
+		c.registryMirrors = registryMirrors
+	}
+}
+
 type Fetcher struct {
-	docker client.CommonAPIClient
-	logger logging.Logger
+	docker          client.CommonAPIClient
+	logger          logging.Logger
+	registryMirrors map[string]string
 }
 
 type FetchOptions struct {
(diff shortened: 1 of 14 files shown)


internal/image/fetcher.go
```diff
@@ -19,25 +19,49 @@ import (
...
+	return fetcher
 }
 
 var ErrNotFound = errors.New("not found")
 
 func (f *Fetcher) Fetch(ctx context.Context, name string, daemon bool, pullPolicy config.PullPolicy) (imgutil.Image, error) {
+	name, err := pname.TranslateRegistry(name, f.registryMirrors)
+	if err != nil {
+		return nil, err
+	}
```
aemengo, line 63: Just acknowledging that this line isn't tested. 😕

jromero:
The implemented proxy per host or base url is interesting. 

```toml
[registry-mirrors]
  "index.docker.io" = "nexus.mycompany.local/mirror"
  "gcr.io/some/path" = "nexus.mycompany.local/mirror"
```

I'm not sure where that requirement came from (I see it in the example config comment [here](https://github.com/buildpacks/pack/issues/821#issuecomment-785298898)) but mirrors are more commonly applied globally to any and all dependencies. I'd want to validate that this feature request does need a per host/base url type configuration. 


1. If it does, we'll need a wildcard proxy so it works across any and all images. 
    ```toml
    [registry-mirrors]
      "*" = "nexus.mycompany.local/mirror"
    ```

2. If there isn't a current requirement for per host/base url, I would remove it to reduce complexity.


Reference:

- maven < 2.0.5: [classic setup](https://www.jfrog.com/confluence/display/RT12/Configuring+Maven)
- maven >= 2.0.5: [`mirrorOf` wildcard](https://maven.apache.org/guides/mini/guide-mirror-settings.html#advanced-mirror-specification)
- npm: [default registry and scoped packages](https://www.jfrog.com/confluence/display/JFROG/npm+Registry#npmRegistry-SettingtheDefaultRegistry) ([scoped packages](https://docs.npmjs.com/cli/v7/using-npm/scope) are like namespaces - analogous to base url)

aemengo:
@jromero This flexibility for per host/base url came from my experience using [harbor as a proxy cache](https://goharbor.io/docs/2.1.0/administration/configure-proxy-cache/).

At the very least, clients using the proxy cache feature of Harbor cannot set a global mirror.

jromero:
TL;DR: we want both per base urls and wildcard (any).

That makes sense although it does appear to be a harbor specific limitation.

FWIW, the global mirror I was referring to is declared at the client side (pack in this case). It basically says, "for any image, I want you to go look it up at this registry" (ie. Harbor or Artifactory), then it's up to configuration in the registry to look up that image in configured remote (delegated) registries. In Artifactory's case you can setup multiple registries it can delegate/proxy so as a client you want "any" support so that it's up to the Artifactory to do the look up without having to manually map all base urls at the client side.

dfreilich (changes requested):
This is really super for my current use case, where I want to ensure all of our requests are proxied to Artifactory. Really happy to see this work @aemengo; superb as always. 

One very confusing aspect of it for me, in UA, was the fetcher doesn't immediately say that it's pulling from the mirror. I cancelled it a number of times, because I didn't think I had set my mirror appropriately, until I got to the `Status` line and saw that it did work appropriately. If you could make the image mirror explicit from the beginning, that would make the UX for me significantly better. 

```
$ out/pack build test-go -B paketobuildpacks/builder:tiny -p ../paketo-samples/go/mod
tiny: Pulling from paketobuildpacks/builder
b734f5039f3d: Pull complete
...
Digest: sha256:e6e5fcd7c7094ed28f45b98e1528303f0f3304696e6bd737537c838cdd8f463f
Status: Downloaded newer image for <Artifactory-mirror>/paketobuildpacks/builder:tiny
tiny-cnb: Pulling from paketobuildpacks/run
4fb75c639a6b: Pull complete
...
Digest: sha256:b643257f93f44aa4e77c1c746fea14f419e29a2123829c7a228bc7f1b633fe2a
Status: Downloaded newer image for <Artifactory Mirror>/paketobuildpacks/run:tiny-cnb
...
```

aemengo:
@dfreilich

> If we could change the `Pulling from` line to also reflect the registries mirror, that would be a slam-dunk from my end

It looks like we're not writing that line, it's being feed to us by the docker API. I don't know how you still feel about replacing it. To me, it's a bit unsavory.

Maybe a big yellow text `WARNING: using mirror <Artifactory-mirror>` is better?

dfreilich (approved):
Beautiful work. Thanks!

Issue #1197, opened 2021-06-08
`pack build` fails with untrusted builders on Apple Silicon


briandealwis:
### Summary

`pack build` from an Apple M1 will fail when using an untrusted builder.  With untrusted builders, the pack library will not use the builder's embedded `lifecycle` binary and instead attempt to run the lifecycle from a separate image:

https://github.com/buildpacks/pack/blob/e53adb9c9e55b33e830f2e92bc4741cc32dc24f7/build.go#L329-L343

Recent lifecycle images (e.g., `buildpacksio/lifecycle:0.11.1`) point to a manifest list with images for `linux/amd64` and `windows/amd64`, and so the attempt to fetch an image for `linux/arm64` fails.

I guess this is somewhat related to https://github.com/buildpacks/lifecycle/issues/435 except that it's actually useful to be able to use Docker Desktop for M1's preconfigured `binfmt_exec` emulation to build `linux/amd64` images.  Perhaps `pack` should attempt to pull down the lifecycle image for the same os/arch as the builder image.

---

##### Steps

On an Apple M1 device, try to build using the `gcr.io/buildpacks/builder:latest` builder, which is not considered a trusted builder.
1. `pack build --builder gcr.io/buildpacks/builder <image-name>`

##### Current behavior

```console
$ pack build --builder gcr.io/buildpacks/builder test
latest: Pulling from buildpacks/builder
Digest: sha256:5f109c6becf6ecd70d7201de44775f21902e1b027ee68eb982bd55787eb25165
Status: Downloaded newer image for gcr.io/buildpacks/builder:latest
v1: Pulling from buildpacks/gcp/run
Digest: sha256:13535b83b3cfb66571907c574bca1cc1852ac6cf5b3096d3e654cd2d95bbecd6
Status: Image is up to date for gcr.io/buildpacks/gcp/run:v1
0.11.1: Pulling from buildpacksio/lifecycle
ERROR: failed to build: fetching lifecycle image: no matching manifest for linux/arm64/v8 in the manifest list entries
```

##### Expected behavior

Hoped for the build to happen.

##### Workarounds

Indicate that the builder is trusted, so that `pack` will use the lifecycle embedded in the builder image.

---

##### pack info
```
$ pack version
0.18.1+git-b5c1a96.build-2373
```

##### docker info
```console
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
[...]

PR #1216, opened 2021-06-25
Use lifecycle image that matches builder platform

aemengo:
## Summary
This PR fetches a lifecycle image that is the same **platform** as the requested builder image, when this builder is untrusted.

This PR also includes a refactor the `image.Fetch()` method signature. This was done so that future modifications may not involve changes in so many files.

## Documentation
    - [ ] Yes, see #___
    - [x] No

## Caveats
_Full disclosure:_ I don't have an arm machine, so I've never seen this work 😬 

## Related

Resolves https://github.com/buildpacks/pack/issues/1197

### DIFF
--- a/internal/image/fetcher.go
+++ b/internal/image/fetcher.go
@@ -28,6 +28,14 @@ type Fetcher struct {
 	logger logging.Logger
 }
 
+type FetchOptions struct {
+	//TODO: would love to reverse (Daemon bool) to (Remote bool)
+	//      so that "false" becomes the default, when omitted
+	Daemon     bool
+	Platform   string
+	PullPolicy config.PullPolicy
+}
+
 func NewFetcher(logger logging.Logger, docker client.CommonAPIClient) *Fetcher {
 	return &Fetcher{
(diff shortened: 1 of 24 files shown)


internal/image/fetcher.go
```diff
@@ -28,6 +28,14 @@ type Fetcher struct {
 	logger logging.Logger
 }
 
+type FetchOptions struct {
+	//TODO: would love to reverse (Daemon bool) to (Remote bool)
+	//      so that "false" becomes the default, when omitted
```
jromero, line 33: We would rather not commit TODOs without an issue created and being referenced. This particular TODO seems a bit superfluous, not sure it's worth pursuing outside of this change. I'd be open to resolving this TODO as part of this PR or removing the TODO altogether. My 2c.

jromero (changes requested):
Aside from one critique this is a very nice clean PR. Thanks!

jromero (approved):

jromero:
🥂 Thank you for this fix!

PR #1222, opened 2021-07-01
Update registry helper to fallback to insecure registry network range

natalieparellano:
This may allow acceptance tests to run on a GitHub hosted runner

## Summary
imgutil and lifecycle acceptance tests use the registry helpers from imgutil and can run on a GitHub hosted runner. This change attempts to make that possible for pack.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___

### DIFF
--- a/testhelpers/registry.go
+++ b/testhelpers/registry.go
@@ -207,11 +207,14 @@ func waitForPortBinding(t *testing.T, containerID, portSpec string, duration tim
 func DockerHostname(t *testing.T) string {
 	dockerCli := dockerCli(t)
 
+	// try to use docker cli's daemon host
 	daemonHost := dockerCli.DaemonHost()
 	u, err := url.Parse(daemonHost)
 	if err != nil {
 		t.Fatalf("unable to parse URI client.DaemonHost: %s", err)
 	}
+
+	var daemonInfo dockertypes.Info
 
 	switch u.Scheme {
 	// DOCKER_HOST is usually remote so always use its hostname/IP
@@ -222,7 +225,7 @@ func DockerHostname(t *testing.T) string {
 	// if DOCKER_HOST is non-tcp, we assume that we are
 	// talking to the daemon over a local pipe.
 	default:
-		daemonInfo, err := dockerCli.Info(context.TODO())
+		daemonInfo, err = dockerCli.Info(context.TODO())
 		if err != nil {
 			t.Fatalf("unable to fetch client.DockerInfo: %s", err)
 		}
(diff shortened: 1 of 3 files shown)


natalieparellano:
This works in the sense that lifecycle acceptance tests are [passing](https://github.com/natalieparellano/lifecycle/actions/runs/991684390) on a public runner with one additional change to the workflow config: https://github.com/natalieparellano/lifecycle/blob/24f0227bce183a271818611cfade357c362aea5c/.github/workflows/build.yml#L51-L72

I thought it might be more advantageous to convert pack acceptance tests to use the imgutil in-memory helpers: https://github.com/buildpacks/imgutil/blob/main/testhelpers/docker_registry.go (used by imgutil & lifecycle) however pack is having trouble finding & running images pulled from this registry. cc @micahyoung for any thoughts!

testhelpers/registry.go
```diff
@@ -232,17 +235,60 @@ func DockerHostname(t *testing.T) string {
...
 
-		// Linux can use --network=host so always use "localhost"
+		for _, addr := range addrs {
+			var interfaceIP net.IP
+			switch interfaceAddr := addr.(type) {
+			case *net.IPAddr:
+				interfaceIP = interfaceAddr.IP
+			case *net.IPNet:
+				interfaceIP = interfaceAddr.IP
+			}
```
dfreilich, line 271: Why do we need to do this? What is it excluding?

dfreilich (commented):
I had also tried to get the `imgutil` Docker helpers working in pack, and couldn't get it done, but I also couldn't dedicate so much time to it. It would be a massive help if we could get that happening though, it would make things so much simpler for the project!

aemengo (changes requested):
It appears like this PR is trying to guess the correct `DOCKER_HOST` to use, based on what entries are in your `daemon.json` and what network interfaces you have configured.

This behavior makes sense for running our tests in CI _right now_, but there's no guarantee that other individuals running these tests locally have the same non-standard setup.

If it's necessary to configure the `DOCKER_HOST` other than:
* manipulating the `DOCKER_HOST` environment variable directly
* manipulating the `/etc/hosts` entry for `host.docker.internal`

then I'd much rather _be explicit_ and _be simple_, by introducing another environment variable or some other configuration method.

natalieparellano:
In the end, all changes to get pack acceptance running on GitHub's runners could be made in the lifecycle: https://github.com/buildpacks/lifecycle/pull/652/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R186-R215

Issue #761, opened 2020-07-22
Expose functionality to download buildpacks


jromero:
See comment below by requester @jabrown85

jabrown85:
@jromero I can't edit this, so you'll need to move it.

## Description
As a consumer of `pack` as a library, I want to be able use `pack`'s ephemeral builder functionality in a similar way on a platform. 

For instance, a user can run `pack build --buildpack mybpimage:v2 myimg --builder heroku/buildpacks` and `pack` will build an ephemeral image with the `mpbpimage` downloaded from dockerhub. There is a good bit of logic and decisions that `pack` makes when processing the `--buildpack argument`.  Those decisions include determining if what was provided is an image, relative folder, tar, buildpackage, parsing versioning, etc. As a platform, it would be nice to be able to leverage some of that logic via `pack` as a library. 

Aside: This example can also be expressed with a user using `package.toml`, which is maybe the more likely scenario.

## Proposed solution
`pack` can expose some of the core methods used to create ephemeral builders as primitives.

## Describe alternatives you've considered
Use `pack` itself - `pack` creates these builder images and controls the deletion of these builders as well. I'd like to use the core `pack` code to operate on a running image.

Re-implement buildpack parsing and downloading - We are afraid of drift in buildpack processing from `pack` itself. We would like a cnb user to be able to go from running `pack` to pushing that same code up to a platform without hitting differences in something as fundamental as defining which buildpacks and what order.

## Questions
Can `pack` expose these with or without `docker` being required? If you had a `tekton` pipeline with a binary referencing `pack` running in a container with user code, could the `pack` lib be executed _in the running image_ without needing docker? This would be useful for populating a `/cnb/buildpacks` volume that you could then carry on to the `detector` step.

jromero:
If this was part of the Go lib interface, in your example for usage inside of something like `tekton`, the user of this library would be compiling a program in Go to use in this workflow ahead of time. This issue is not requesting that such utility function be exposed via a utility CLI or from the pack CLI executable correct?

A simplified example of the interface:

```go
func DownloadBuildpack(buildpackURI string) (contents io.Reader, err error)
```

where:

- `buildpackURI` is implemented based on https://github.com/buildpacks/rfcs/blob/main/text/0037-buildpack-uris.md
- `contents` is a forward-only reader with the contents of the buildpacks in `tar` format

jabrown85:
Correct, not asking for a CLI executable or utility CLI. 

[...]

PR #1225, opened 2021-07-06
Expose functionality to download buildpacks

YousefHaggyHeroku:
## Summary
This PR aims to address https://github.com/buildpacks/pack/issues/761 by introducing a new `DownloadBuildpack` function. `DownloadBuildpack` allows consumers of `pack` as a library to use the existing buildpack URI parsing, downloading, and extracting logic without having to reimplement most of the work. One requirement for this request is to give consumers the option to not use docker; the existing `imageFetcher` struct seems to handle this concern.

*Q*: For this change, I also moved `blob` and `image` to become public facing packages so that consuming libraries can pass in their own imageFetchers and downloaders. Would it be simpler and more preferable to just keep those internal and create an `imageFetcher` if one isn't found? Something like:
```
if opts.imageFetcher == nil{
imageFetcher := image.NewFetcher(...)
}else{
imageFetcher := opts.ImageFetcher
}
```
Same question for the downloader. We could also both make the packages public and add that default creation logic to make life easier for consumers that don't want a custom fetcher while still keeping flexibility for those that do
## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related
Resolves https://github.com/buildpacks/pack/issues/761


create_builder.go
```diff
@@ -202,74 +203,107 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
...
-			}
+	// The base directory to use to resolve relative assets
+	RelativeBaseDir string
+
+	// The OS of the builder image
+	ImageOS string
+
+	Downloader   Downloader
+	Logger       logging.Logger
+	ImageFetcher ImageFetcher
```
aemengo, line 217: Thanks so much for this contribution!

Is it for your use-case that this `DownloadBuildpack(` method doesn't have a receiver on the `Client struct`? To me, it looks a little awkward passing such hefty abstractions through the `opts` parameter like this. Furthermore, it would address the following question you raised, since the `func NewClient(` factory method will in fact initialize an `ImageFetcher` and `Downloader` if they don't exist. Finally, it would avoid awkward lines [like this](https://github.com/buildpacks/pack/pull/1225/files#diff-3005b99fb5930f3734d223a0ea22956fd45c7a350f62801c78dd9c50ab2ab85fR250).

> Would it be simpler and more preferable to just keep those internal and create an imageFetcher if one isn't found?

I'd prefer to limit the scope of this PR and not expose those files (`downloader.go`, `fetcher.go`, etc), until someone requires that level of configurability.

Lastly, I think some tests are warranted for this method. Since this is now a public facing interface that anyone can use, and _reputations are on the line._ 😬

aemengo (changes requested):

create_builder.go
YousefHaggyHeroku: Thank you for your feedback! The only issue with having a receiver on `Client` is that the `NewClient` factory method creates a few unnecessary things for our use case like a `docker` client and a lifecycle executor. One requirement for our use case is to avoid using `docker` entirely. 

One I idea I had was to create a new Factory method, `NewSimpleClient` or perhaps `NewDockerlessClient`, that doesn't force create a docker client. What are your thoughts on this approach?

Good point about tests, I can definitely go through and add some once we are happy with the `opts` and the format of this method.

aemengo: While I normally wouldn't think twice about `NewSimpleClient` or `NewDockerlessClient` in more "private" code, I'd prefer not to introduce a different code path so early in a public facing interface.

Our docker client is configurable... What are your thoughts on the following? Just something that will satisfy the `interface` but won't actually do anything:

```go
// WithOutDockerClient supply a no-op docker client.
func WithOutDockerClient() ClientOption {
	return func(c *Client) {
		c.docker = &noopDockerClient{}
	}
}
```

🤔 I'm a little torn with having to create/generate a `dockerClient.CommonAPIClient` replacement _ourselves_. For the `lifecycleExecutor`, I have less reservations. I'm also completely for _your client-side code_ passing in these no-op types.

Happy to hear others' thoughts on this!

YousefHaggyHeroku: Ah interesting, I hadn't thought of passing a no-op docker client in. In that case, I feel like the existing `WithDockerClient` option suffices, and we can just pass in a no-op docker client in our consuming code. With your explanation, no changes to `NewClient` and no new `ClientOption`  makes the most sense to me, unless anyone else has other thoughts here.

YousefHaggyHeroku: I've made the discussed change to take add a Client receiver and moved `blob` back into internal. I kept `image` public so that we can at-least pass in `image.FetchOptions`.  If we don't want the entire `image` package to be public, I could split it into an internal package and external package with the types, but that doesn't sound ideal to me. Are there any strong reasons why we shouldn't just make the `image` package public now?

aemengo: No _strong_ reasons, from me. 🙂

jromero: If `image.FetchOptions` is the only reason for the move, I would suggest that we _try_ to find alternatives. The `Fetcher` would otherwise become a public interface that we'd need to be considerate about making changes to in the future.

YousefHaggyHeroku: 2 updates.
 - Added tests. Most of which are extracted from the existing `CreateBuilder` tests. I had to refactor things a bit to be more testable
-  I've moved image back to internal, instead just directly defining `Daemon` and `PullPolicy` fields. 

It probably seems odd that `mock_buildpack_downloader.go` is not in `testmocks`, but I can't put it into `testmocks` bc of a circular dependency issue.  Some of the test files are under the package `pack`, which requires `testmocks` which _would_ require `pack`.  Two probably neater approaches to fixing this so that the mock could be in `testmocks`:
 - The test files defined with `package pack` should instead be under `package pack_test`. This is actually a lot of work
 - Move `buildpack_downloader` to a new package. Also a small chunk of work: it currently uses ~3 private `pack` functions that are all in different files.

Those solutions might be more trouble than they are worth, im not sure 🤷

aemengo (approved):

jromero:
@YousefHaggyHeroku This looks great! There's a conflict now also need to get sign-offs on some of the commits. 😞

YousefHaggyHeroku:
> @YousefHaggyHeroku This looks great! There's a conflict now also need to get sign-offs on some of the commits. 😞

Just handled both, thanks!

jromero (approved):

PR #1229, opened 2021-07-12
Generate cyclonedx bom on release

samj1912:
Uses the cyclonedx go mod bom generator to generate a cdx bom for our projects

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -220,6 +220,13 @@ jobs:
           PACK_HELP="${PACK_HELP//$'\n'/'%0A'}"
           PACK_HELP="${PACK_HELP//$'\r'/'%0D'}"
           echo "::set-output name=help::$PACK_HELP"
+      - name: Generate SBOM JSON
+        uses: CycloneDX/gh-gomod-generate-sbom@v0.3.0
+        with:
+          json: true
+          output: pack-v${{ env.PACK_VERSION }}-bom.cdx.json
+          resolve-licenses: true
+          version: ^v0
       - name: Generate changelog
         uses: ./.github/workflows/actions/release-notes
         id: changelog


natalieparellano:
@samj1912 should we consider doing this for the lifecycle as well?

.github/workflows/build.yml
```diff
@@ -220,6 +220,13 @@ jobs:
           PACK_HELP="${PACK_HELP//$'\n'/'%0A'}"
           PACK_HELP="${PACK_HELP//$'\r'/'%0D'}"
           echo "::set-output name=help::$PACK_HELP"
+      - name: Generate SBOM JSON
+        uses: CycloneDX/gh-gomod-generate-sbom@v0.3.0
+        with:
+          json: true
+          output: bom.cdx.json
```
natalieparellano, line 227: Where would the bom be stored? Should it be attached to the release as an artifact?

samj1912: It is attached to the release as an artifact yes - see https://github.com/CycloneDX/cyclonedx-go/releases/tag/v0.2.1

samj1912:
> @samj1912 should we consider doing this for the lifecycle as well?

We probably should. Can make a PR.

.github/workflows/build.yml
natalieparellano: I'm confused... how do they accomplish this magic? Are you sure the release notes are referring to __your__ (the user's) CI, not just cyclonedx-go CI?

samj1912: Ah yes sorry, my bad, it looks like they use go releaser to attach the bom https://github.com/CycloneDX/cyclonedx-gomod/blob/0b09de5978d9120a2b1f03ad376e5e0eeaaf9c65/.goreleaser.yml#L54

jromero: If you follow the naming convention as our other artifacts `pack-v${{ env.PACK_VERSION }}-*`, it would be picked up by our release action on line 246. Can we make that change?

samj1912: Simple and convenient! Thanks for the tip!

jromero (approved):

nscuro:
Hi, I'm the maintainer of cyclonedx-gomod. I can reproduce this behavior with go 1.15.

What I've gathered so far:

* `io/fs` didn't exist prior to go 1.16
* `github.com/BurntSushi/toml@v0.4.1` requires go 1.16 according to its [`go.mod`](https://github.com/BurntSushi/toml/blob/v0.4.1/go.mod#L3)
* Code that imports `io/fs` in `github.com/BurntSushi/toml` [has](https://github.com/BurntSushi/toml/blob/v0.4.1/decode_go116.go#L1) the `+build go1.16` tag, meaning go < 1.16 *should* ignore those when building.

cyclonedx-gomod uses `go mod why -m -vendor` to force go to download required modules to its module cache and filter out unused modules. This fails with go 1.15:

```shell
root@f689cd4405a6:/usr/src/pack# go mod why -m -vendor github.com/BurntSushi/toml
github.com/buildpacks/pack imports
        github.com/BurntSushi/toml imports
        io/fs: package io/fs is not in GOROOT (/usr/local/go/src/io/fs)
```

`mod why` analyzes the imports in order to determine whether or not a given module is actually used. Apparently `mod why` in go <= 1.15 can't deal with nonexistent standard library packages (https://github.com/golang/go/issues/47152).

I'd recommend running cyclonedx-gomod in a context where go 1.16 is available. I fear I can't do much from the tool's side, as `mod why` is essential to its functionality.

An upcoming version will have [a new mode](https://github.com/CycloneDX/cyclonedx-gomod/tree/v1.0.0-alpha.2#app) that doesn't rely on `mod why` (testable with [v1.0.0-apha.2](https://github.com/CycloneDX/cyclonedx-gomod/releases/tag/v1.0.0-alpha.2)). This works with go 1.15:

```
$ ./cyclonedx-gomod-v1.0.0-alpha.2 app -json -output bom.json -licenses -main cmd/pack/main.go /usr/src/pack
```

Hope this helps. Let me know if I can assist with anything.

jromero:
Thank you @nscuro. That's very insightful. We should be able to address this on our end by using this the [`setup-go` action](https://github.com/actions/setup-go) with a higher go version.

PR #1235, opened 2021-07-20
Cleanup registry cache in tests

### LINKED ISSUE #1228: Unit tests should clean up better
### Description
If you run `make unit` a few times, files matching the pattern **internal/registry/registry*** will be created and left in your project working directory

### Proposed solution
These files shouldn't persist after test runs

### Additional context
``` shell
$ git status
# ...

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	internal/registry/registry027131156/
	internal/registry/registry028019158/
	internal/registry/registry029322477/
	internal/registry/registry030156948/
	internal/registry/registry036873021/
	internal/registry/registry036874482/
	internal/registry/registry040831238/
	internal/registry/registry042130857/
	internal/registry/registry050029389/
	internal/registry/registry060047820/
	internal/registry/registry063771539/
        # ...
```

jkutner:
Keep a reference to the registry cache dir, which can be used to cleanup in test. Fixes #1228

This only happens when we're testing with a an empty string (`""`) for the root of the registry cache, and the temp dir is actually created from within the production code (not the test code). Therefore, we need to keep a reference to the temp dir so it can be destroyed in `it.After`

#### Before

```
$ go test -run TestRegistryCache ./...
$ git status
...
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        internal/registry/registry087316852/
```

#### After

```
$ go test -run TestRegistryCache ./...
$ git status
...
nothing to commit, working tree clean
```

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1228

### DIFF
--- a/internal/registry/registry_cache_test.go
+++ b/internal/registry/registry_cache_test.go
@@ -217,7 +217,7 @@ func testRegistryCache(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			it.After(func() {
-				h.AssertNil(t, os.RemoveAll(filepath.Join(registryCache.Root, registryCache.RegistryDir)))
+				h.AssertNil(t, os.RemoveAll(registryCache.RegistryDir))
 			})
 
 			when("url is empty string", func() {
--- a/internal/registry/registry_cache_test.go
+++ b/internal/registry/registry_cache_test.go
@@ -191,7 +191,7 @@ func testRegistryCache(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			it.After(func() {
-				h.AssertNil(t, os.RemoveAll(filepath.Join(registryCache.Root, registryCache.RegistryDir)))
+				h.AssertNil(t, os.RemoveAll(registryCache.RegistryDir))
 			})
 		})
 	})
--- a/internal/registry/registry_cache_test.go
+++ b/internal/registry/registry_cache_test.go
@@ -189,6 +189,10 @@ func testRegistryCache(t *testing.T, when spec.G, it spec.S) {
 				err = registryCache.Refresh()
 				h.AssertError(t, err, "initializing")
 			})
+
+			it.After(func() {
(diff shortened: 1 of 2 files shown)


internal/registry/registry_cache_test.go
```diff
@@ -189,6 +189,10 @@ func testRegistryCache(t *testing.T, when spec.G, it spec.S) {
 				err = registryCache.Refresh()
 				h.AssertError(t, err, "initializing")
 			})
+
+			it.After(func() {
+				h.AssertNil(t, os.RemoveAll(filepath.Join(registryCache.Root, registryCache.RegistryDir)))
```
aemengo, line 194:
```suggestion
				h.AssertNil(t, os.RemoveAll(registryCache.RegistryDir))
```

I hate to be persnickety, but the `.RegistryDir` is already an absolute path. Having the `.Join` there is just a _tad_ more confusing. Enough for me to mention it.

internal/registry/registry_cache_test.go
```diff
@@ -212,6 +216,10 @@ func testRegistryCache(t *testing.T, when spec.G, it spec.S) {
 				h.AssertError(t, err, "creating registry cache")
 			})
 
+			it.After(func() {
+				h.AssertNil(t, os.RemoveAll(filepath.Join(registryCache.Root, registryCache.RegistryDir)))
```
aemengo, line 220:
```suggestion
				h.AssertNil(t, os.RemoveAll(registryCache.RegistryDir))
```

I hate to be persnickety, but the `.RegistryDir` is already an absolute path. Having the `.Join` there is just a _tad_ more confusing. Enough for me to mention it.

aemengo (changes requested):

aemengo (approved):

jromero (approved):

PR #1237, opened 2021-07-21
Refactor processBuildpacks to use BuildpackDownloader

YousefHaggyHeroku:
## Summary
Instead of maintaining the logic of locating / downloading build-packs from a URI, Registry, and Package twice, it might be better to instead refactor `processBuildpack` to using the new `BuildpackDownloader`. This change shouldn't affect any functionality
## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___


build_test.go
```diff
@@ -1136,7 +1139,7 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 					ClearCache: true,
 					Buildpacks: []string{"missing.bp@version"},
 				}),
-					"invalid buildpack string 'missing.bp@version'",
+					"downloading buildpack: error reading missing.bp@version: invalid locator: InvalidLocator",
```
YousefHaggyHeroku, line 1142: I guess there is a slight change in that users would see a different error message. Not sure how important it is that our messaging stays strictly the same.

aemengo (approved):
Thanks for going the extra mile! 🙇🏾

jromero (approved):

Issue #1127, opened 2021-04-04
RFC 140: Project Descriptor Domains


dfreilich:
### Description

Implement https://github.com/buildpacks/rfcs/pull/140

### Additional context
- [ ] This feature should be documented somewhere

YousefHaggyHeroku:
I'm working on this issue 👍

samj1912:
@buildpacks/platform-maintainers can we have more details on this issue on how we expect b/w compat with existing project descriptor files? Since this is a breaking change for app developers, this will impact a lot of users who may not be aware of CB internals.

YousefHaggyHeroku:
My thoughts were to do something like having a `ParseV1Descriptor` and `ParseV2Descriptor` in the pack code. If the "_" top level key exists, `ParseV2Descriptor`, else `ParseV1Descriptor`.  Curious to hear other thoughts

PR #1241, opened 2021-07-28
Add support for Project Descriptor 0.2

YousefHaggyHeroku:
## Summary

Implementing [this](https://github.com/buildpacks/spec/blob/extensions/project-descriptor/0.2/extensions/project-descriptor.md#iobuildpacksenvbuild-optional ) without breaking the old descriptor parsing. I'm not sure how to approach documentation with a change like this
## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

Resolves https://github.com/buildpacks/pack/issues/1127

### DIFF
--- a/build.go
+++ b/build.go
@@ -37,7 +37,7 @@ import (
 	"github.com/buildpacks/pack/internal/style"
 	"github.com/buildpacks/pack/logging"
 	"github.com/buildpacks/pack/pkg/archive"
-	"github.com/buildpacks/pack/project"
+	projectCommon "github.com/buildpacks/pack/project/common"
 )
 
 const (
@@ -151,7 +151,7 @@ type BuildOptions struct {
 	ProjectDescriptorBaseDir string
 
 	// ProjectDescriptor describes the project and any configuration specific to the project
-	ProjectDescriptor project.Descriptor
+	ProjectDescriptor projectCommon.Descriptor
 
 	// The lifecycle image that will be used for the analysis, restore and export phases
 	// when using an untrusted builder.
@@ -368,7 +368,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 	return c.logImageNameAndSha(ctx, opts.Publish, imageRef)
 }
 
-func getFileFilter(descriptor project.Descriptor) (func(string) bool, error) {
+func getFileFilter(descriptor projectCommon.Descriptor) (func(string) bool, error) {
 	if len(descriptor.Build.Exclude) > 0 {
 		excludes := ignore.CompileIgnoreLines(descriptor.Build.Exclude...)
 		return func(fileName string) bool {
@@ -920,7 +920,7 @@ func parseDigestFromImageID(id imgutil.Identifier) string {
 	return fmt.Sprintf("sha256:%s", digest)
 }
 
-func createInlineBuildpack(bp project.Buildpack, stackID string) (string, error) {
+func createInlineBuildpack(bp projectCommon.Buildpack, stackID string) (string, error) {
 	pathToInlineBuilpack, err := ioutil.TempDir("", "inline-cnb")
 	if err != nil {
 		return pathToInlineBuilpack, err
--- a/project/common/project.go
+++ b/project/common/project.go
@@ -1,0 +1,48 @@
+package common
+
+import (
+	"github.com/buildpacks/lifecycle/api"
+
+	"github.com/buildpacks/pack/internal/dist"
+)
+
+type Script struct {
+	API    string `toml:"api"`
+	Inline string `toml:"inline"`
+	Shell  string `toml:"shell"`
+}
+type Buildpack struct {
+	ID      string `toml:"id"`
+	Version string `toml:"version"`
+	URI     string `toml:"uri"`
+	Script  Script `toml:"script"`
+}
+
+type EnvVar struct {
+	Name  string `toml:"name"`
+	Value string `toml:"value"`
+}
+
+type Build struct {
+	Include    []string    `toml:"include"`
+	Exclude    []string    `toml:"exclude"`
+	Buildpacks []Buildpack `toml:"buildpacks"`
+	Env        []EnvVar    `toml:"env"`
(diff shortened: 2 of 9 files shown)


build.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/buildpacks/pack/internal/style"
 	"github.com/buildpacks/pack/logging"
 	"github.com/buildpacks/pack/pkg/archive"
-	"github.com/buildpacks/pack/project"
+	projectCommon "github.com/buildpacks/pack/project/common"
```
YousefHaggyHeroku, line 40: I wonder if there's a better name for this 🤷

aemengo: No strong feeling over here!

project/common/project.go
```diff
@@ -0,0 +1,48 @@
+package common
+
+import (
+	"github.com/buildpacks/lifecycle/api"
+
+	"github.com/buildpacks/pack/internal/dist"
+)
+
+type Script struct {
+	API    string `toml:"api"`
```
aemengo, line 10: These are internal only representations right? Meaning these won't be `Marshalled()`, `Unmarshalled()` to the filesystem?

If so, then these `toml:"..."` annotations won't be used.

project/project.go
```diff
@@ -1,70 +1,70 @@
...
-type Descriptor struct {
-	Project  Project                `toml:"project"`
-	Build    Build                  `toml:"build"`
-	Metadata map[string]interface{} `toml:"metadata"`
+type VersionDescriptor struct {
+	Project Project `toml:"_"`
 }
 
-func ReadProjectDescriptor(pathToFile string) (Descriptor, error) {
+func ReadProjectDescriptor(pathToFile string) (common.Descriptor, error) {
```
aemengo, line 30: I'm personally finding the self-updating structs a little awkward.

```suggestion
func ReadProjectDescriptor(pathToFile string) (common.Descriptor, error) {
	pathToFile = filepath.Clean(pathToFile)

	var versionDescriptor struct {
		Project struct {
			Version string `toml:"schema-version"`
		} `toml:"_"`
	}

	_, err := toml.DecodeFile(pathToFile, &versionDescriptor)
	if err != nil {
		return common.Descriptor{}, errors.Wrapf(err, "parsing schema version")
	}

	switch versionDescriptor.Project.Version {
	// _.schema-version does not exist in 0.1
	case "0.1", "":
		descriptor, err := v01.NewDescriptor(pathToFile)
		if err != nil {
			return common.Descriptor{}, err
		}

		return descriptor, validate(descriptor)
	case "0.2":
		descriptor, err := v02.NewDescriptor(pathToFile)
		if err != nil {
			return common.Descriptor{}, err
		}

		return descriptor, validate(descriptor)
	default:
		return common.Descriptor{}, fmt.Errorf("unknown project descriptor schema version %s", versionDescriptor.Project.Version)
	}
}
```

aemengo (commented):
This looks great! You're pretty good at this 🙂 

--

Also wanted to the highlight that the following windows CI errors aren't related to this PR.

```powershell
Setting up auth
  "C:\Program Files\Git\cmd\git.exe" config --local --name-only --get-regexp core\.sshCommand
  "C:\Program Files\Git\cmd\git.exe" submodule foreach --recursive "git config --local --name-only --get-regexp 'core\.sshCommand' && git config --local --unset-all 'core.sshCommand' || :"
  Error: fatal: No url found for submodule path 'internal/registry/registry139553254' in .gitmodules
  Error: The process 'C:\Program Files\Git\cmd\git.exe' failed with exit code 128
```

aemengo (approved):

project/project.go
YousefHaggyHeroku: I like this suggestion a lot! I think it looks a lot cleaner and less awkward. The only possible downside I could see is if we have lots of versions in the future, the switch statement could get a little large and redundant. What are your thoughts on the following:

```
var NewDescriptor = map[string]common.NewDescriptorType{
	"0.1": v01.NewDescriptor,
	"0.2": v02.NewDescriptor,
}

func ReadProjectDescriptor(pathToFile string) (common.Descriptor, error) {
	projectTomlContents, err := ioutil.ReadFile(filepath.Clean(pathToFile))
	if err != nil {
		return common.Descriptor{}, err
	}

	var versionDescriptor struct {
		Project struct {
			Version string `toml:"schema-version"`
		} `toml:"_"`
	}

	_, err = toml.Decode(string(projectTomlContents), &versionDescriptor)
	if err != nil {
		return common.Descriptor{}, errors.Wrapf(err, "parsing schema version")
	}

	version := versionDescriptor.Project.Version
	if version == "" {
		version = "0.1"
	}

	var descriptor common.Descriptor

	if _, ok := NewDescriptor[version]; !ok {
		return common.Descriptor{}, fmt.Errorf("unknown project descriptor schema version %s", version)
	}

	descriptor, err = NewDescriptor[version](pathToFile)
	if err != nil {
		return common.Descriptor{}, err
	}

	return descriptor, validate(descriptor)

}
```

I'm also totally happy to land on your suggestion; I do think it's the cleanest solution for the present. No strong feelings here, just wanted to put out one more thought 😄

project/common/project.go
jromero: I'm comfortable straddling both sides and treating `common` primarily as the internal representations with the (un)marshalling until there is a deviation between different versions.

project/project.go
aemengo: I don't know why I didn't get a notification. I'm so sorry for missing this! 🙇🏾 

Ultimately, I think it's a style preference which is why I have no problem with what you've presented. In my personal projects, large and redundant `switch` statements are very common; But I understand that it's not for everyone!

jromero: Since we slacked a bit here @YousefHaggyHeroku I'm going through the process of creating a polish PR for this so we can get this in ASAP. I'm like and therefore taking your latest suggestion into the follow up PR. I'll also take care of the merge conflicts for you.

YousefHaggy: Sounds good! Thanks for making that PR

PR #1274, opened 2021-09-04
Make pack golang API trust the suggested builders by default

### LINKED ISSUE #1254: Golang API users do not get proper trusted builder behavior
### Summary
Using the **pack golang API** does not have the same behavior as the pack CLI regarding trusted builders. In particular, trusted builders are not automatically detected as trusted. The ramification is that the _creator workflow_ does not run, when it should.

---

##### Example

```go

        // ...

	client, err := pack.NewClient()
	if err != nil {
		return nil, err
	}

	opts := pack.BuildOptions{
		Image:      "aemengo/test",
		Builder:    "heroku/buildpacks:18",
		AppPath:    appPath,
	}

	err = client.Build(ctx, opts)
	if err != nil {
		return nil, err
	}

        // ...
```

##### Expected behavior
In the above example, `heroku/buildpacks:18` is a trusted builder and the creator workflow would run _using the CLI_. But since `opts.TrustBuilder` defaults to `false` here, the same code path does not run. In the following output, the lifecycle gets downloaded when it shouldn't be

```shell
# $ ./run-program
2021/08/11 13:23:44.042847 DEBUG:  Pulling image index.docker.io/heroku/buildpacks:18
18: Pulling from heroku/buildpacks
Digest: sha256:447ce9f5bce9a75ccdde70766e5b7114254dad513cf30869ca4bc8bfebb5edf8
Status: Image is up to date for heroku/buildpacks:18
2021/08/11 13:23:44.473495 DEBUG:  Selected run image heroku/pack:18
2021/08/11 13:23:44.473508 DEBUG:  Pulling image heroku/pack:18
18: Pulling from heroku/pack
Digest: sha256:73dd079b11ee17a657278f348b86aac2efe4432d03b9c79c946afa59799a82d5
Status: Image is up to date for heroku/pack:18

# ...

2021/08/11 13:23:44.889620 DEBUG:  -> heroku/nodejs-npm@0.4.5
2021/08/11 13:23:45.172248 DEBUG:  Pulling image buildpacksio/lifecycle:0.11.4
0.11.4: Pulling from buildpacksio/lifecycle
Digest: sha256:7241fe488c54902bd0d2c95e664456d18d710bc2156383b50fc6786e0126218b
Status: Image is up to date for buildpacksio/lifecycle:0.11.4

# ...
```

---

##### pack info
[v0.20.0](https://github.com/buildpacks/pack/tree/v0.20.0)

jimil749:
Signed-off-by: Jimil Desai <jimildesai42@gmail.com>

## Summary

1. This PR changes the existing `TrustedBuilder` in `BuilderOption` struct from a boolean field to a func field. This change is done to have the Golang API trust the suggested builders by default.
2. This PR adds an extra check for suggested builders in the `pack/builder.go`. If the users, while calling `packClient.Build()` have the TrustedBuilder func configured, that value is used to trust the builder or not. Else by default the Build() function checks whether the provided builder is a suggested builder or not, if yes we trust builder or we don't.
3. In order to suffice the above behavior, I moved the existing `SuggestedBuilders` struct to the `internal/builder` package. (To avoid import cycles.) 

## Documentation

    - [x] Yes, IMO we need to update the changed struct field.
    - [ ] No

## Related

Resolves #1254


build.go
```diff
@@ -67,6 +67,8 @@ type LifecycleExecutor interface {
 	Execute(ctx context.Context, opts build.LifecycleOptions) error
 }
 
+type IsTrustedBuilder func() bool
```
jromero, line 70:
```suggestion
type IsTrustedBuilder func(string) bool
```

build.go
```diff
@@ -304,6 +306,21 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
...
+		}
+		for _, sugBuilder := range builder.SuggestedBuilders {
+			if opts.Builder == sugBuilder.Image {
+				opts.TrustBuilder = func() bool {
+					return true
+				}
+				break
+			}
+		}
+	}
```
jromero, line 322: I don't think this will work as you expect it to. What we're looking to do is delegate the decision to a function that looks at the `builder.SuggestedBuilders` list.

In go (pseudo) code:

```go
// define the default logic as a function
func IsSuggestedBuilderFunc = func(b string) bool {
  for _, suggestedBuilder := range builder.SuggestedBuilders {
      if b == suggestedBuilder {
          return true
      }
  }

  return false
}

...

// set it if it isn't set
if opts.TrustBuilder == nil {
    opts.TrustBuilder = IsSuggestedBuilderFunc
}

// use it
if lifecycleSupportsCreator && opts.TrustBuilder(opts.Builder) {
   ...
}
```

_For reference on the design pattern you can take a look at the [Strategy Pattern](https://springframework.guru/gang-of-four-design-patterns/strategy-pattern/) from Gang of Four._

jromero (changes requested):

build.go
jromero: Maybe an easier example of the pattern in Go: https://refactoring.guru/design-patterns/strategy/go/example

jimil749: Ahh gotcha! This is a lot simpler and easier to understand! :)

internal/commands/build.go
```diff
@@ -149,8 +149,10 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
 				DockerHost:        flags.DockerHost,
 				PullPolicy:        pullPolicy,
 				ClearCache:        flags.ClearCache,
-				TrustBuilder:      trustBuilder,
-				Buildpacks:        buildpacks,
+				TrustBuilder: func(string) bool {
+					return trustBuilder
```
jimil749, line 153: I have *not* actually moved the check, all I do in the CLI is to use the `trustBuilder` boolean to trust/untrust the builder.

jimil749:
@jromero you can have a look when free!

build.go
```diff
@@ -315,6 +317,20 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
...
 	}
 
+	IsSuggestedBuilderFunc := func(b string) bool {
+		for _, suggestedBuilder := range builder.SuggestedBuilders {
+			if b == suggestedBuilder.Image {
+				return true
+			}
+		}
+		return false
+	}
```
dfreilich, line 327: nit: Since this `func` doesn't depend on anything within the `Build` function, can you define it above?

dfreilich (approved):
Overall, this looks really great @jimil749! If you have a chance to fix that small nit I posted, great, and if not, I'll fix it and merge it in tomorrow. Thanks for the help!

jimil749:
Ahh, sorry @dfreilich! I did not notice your comment earlier, would have incorporated your suggestion had I seen it!

Issue #1203, opened 2021-06-14
Pack build should show terminal UI with image: and builder: info 


aemengo:
### Context
At this point of the `pack build` visual workflow, the pull and detect steps will have been completed and we will be in the *build step*. To avoid too much work in one PR, this issue is only about showing information about the **image** and **builder**. This is an opportunity to introduce the builder and its major components.

### Objective
**When** I run `pack build --interactive/-i <name>`
**And** the pull step completes successfully
**And** the detect step completes successfully
**Then** I should see a full terminal panel for the **image:** and **builder:** info

### Notes
- [Mockup](https://user-images.githubusercontent.com/4236888/121903113-f7b7be00-ccf5-11eb-8fb7-4ba1fa5bb37d.png)
   - _just those 2 sections, other sections will be done in other issues_
   - _just a starting point, the final design is up to you_
- RFC: https://github.com/buildpacks/rfcs/pull/160

aemengo:
This issue has been implemented and shipped.

Issue #1206, opened 2021-06-14
Pack build should depict completed build in terminal UI


aemengo:
### Context
At this point of the `pack build` visual workflow, the pull, detect, and build steps will have been completed. This issue is about depicting <ins>the result</ins> of a build within the terminal UI. This depiction consist of:
- **BUILD SUCCEEDED** / **BUILD FAILED** line at the end of the **logs** pane
- Checkmarks for every [metabuildpack](https://buildpacks.io/docs/concepts/components/buildpack/#meta-buildpack) that succeeded, during the build
- Visible **IMAGE IDs** for **image:** section of the terminal UI

### Objective
**When** I run `pack build --interactive/-i <name>`
**And** the pull step completes successfully
**And** the detect step completes successfully
**And** the build step completes successfully
**Then** I should see a full terminal panel depicting <ins>the result</ins> of my build

### Notes
- [Mockup](https://user-images.githubusercontent.com/4236888/119383754-bc474800-bc91-11eb-924e-91bb30d78820.png)
   - _user must enter Ctrl-C to exit terminal UI_
   - _just a starting point, the final design is up to you_
- RFC: https://github.com/buildpacks/rfcs/pull/160

aemengo:
This issue has been implemented and shipped.

Issue #1205, opened 2021-06-14
Pack build should show terminal UI with plan: info


aemengo:
### Context
At this point of the `pack build` visual workflow, the pull and detect steps will have been completed and we will be in the *build step*. To avoid too much work in one PR, this issue is only about showing information about the **plan**. The contents of a **plan** is the sequence of [metabuildpacks](https://buildpacks.io/docs/concepts/components/buildpack/#meta-buildpack) that will be invoked as part of the *build step*.

### Objective
**When** I run `pack build --interactive/-i <name>`
**And** the pull step completes successfully
**And** the detect step completes successfully
**Then** I should see a full terminal panel for the **plan:**

### Notes
- [Mockup](https://user-images.githubusercontent.com/4236888/121940012-a4f0fd00-cd1b-11eb-9541-313700ef098d.png)
   - _just a starting point, the final design is up to you_
- RFC: https://github.com/buildpacks/rfcs/pull/160

aemengo:
This issue has been implemented and shipped.

Issue #1204, opened 2021-06-14
Pack build should show terminal UI with log: info


aemengo:
### Context
At this point of the `pack build` visual workflow, the pull and detect steps will have been completed and we will be in the *build step*. To avoid too much work in one PR, this issue is only about showing information about the **logs**.

### Objective
**When** I run `pack build --interactive/-i <name>`
**And** the pull step completes successfully
**And** the detect step completes successfully
**Then** I should see a full terminal panel for the **logs:** info

### Notes
- [Mockup](https://user-images.githubusercontent.com/4236888/121939771-5c394400-cd1b-11eb-80fc-db7eccaf0186.png)
   - _the logs should be streaming in, as the build is occurring_
   - _just a starting point, the final design is up to you_
- RFC: https://github.com/buildpacks/rfcs/pull/160

smsufyian:
@aemengo Is there a way to assign the issue to myself , I want to work on this

aemengo:
@smsufyian Thanks so much for expressing interest! Writing that comment is as good as getting yourself assigned.

This issue is currently part of a track of work. I'm afraid that this issue isn't ready to work on: _as written, at this exact_ moment.

If you want to be involved with the track of work more generally, please consider my [first PR for it](https://github.com/buildpacks/pack/pull/1230) which introduces the general architecture. 

--

I could certainly use the help 😅. Feel free to comment. And, if you want, I can give you write access to my fork to make edits yourself.

aemengo:
This issue has been implemented and shipped.

PR #1278, opened 2021-09-07
Complete pack-interact workflow

aemengo:
## Summary
Add remaining functionality for pack-interact workflow

This PR kicks continues the following **user workflow**:

When a user runs a pack build like so: `pack build <image-name> --interactive`
Then the user will view the build process through a terminal UI screen
First the `Detecting...` page will show
Then the build "dashboard" page will show

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves:
https://github.com/buildpacks/pack/issues/1203
https://github.com/buildpacks/pack/issues/1204
https://github.com/buildpacks/pack/issues/1205
https://github.com/buildpacks/pack/issues/1206

### DIFF
--- a/internal/builder/builder.go
+++ b/internal/builder/builder.go
@@ -190,6 +190,11 @@ func (b *Builder) Order() dist.Order {
 	return b.order
 }
 
+// BaseImageName returns the name of the builder base image
+func (b *Builder) BaseImageName() string {
+	return b.baseImageName
+}
+
 // Name returns the name of the builder
 func (b *Builder) Name() string {
 	return b.image.Name()
--- a/internal/builder/builder_test.go
+++ b/internal/builder/builder_test.go
@@ -1326,6 +1326,12 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
 					h.AssertEq(t, bldr.GID(), 4321)
 				})
 			})
+
+			when("#BaseImageName", func() {
+				it("return name of base image", func() {
+					h.AssertEq(t, bldr.BaseImageName(), "base/image")
+				})
+			})
 		})
 	})
 }
--- a/internal/termui/dashboard.go
+++ b/internal/termui/dashboard.go
@@ -1,0 +1,143 @@
+package termui
+
+import (
+	"fmt"
+
+	"github.com/gdamore/tcell/v2"
+	"github.com/rivo/tview"
+
+	"github.com/buildpacks/pack/pkg/dist"
+)
+
+type Dashboard struct {
+	app         app
+	appTree     *tview.TreeView
+	builderTree *tview.TreeView
+	planList    *tview.List
+	logsView    *tview.TextView
+	logs        string
+}
+
+func NewDashboard(app app, appName string, bldr buildr, runImageName string, buildpackInfo []dist.BuildpackInfo) *Dashboard {
+	appTree, builderTree := initTrees(appName, bldr, runImageName)
+
+	planList, logsView := initDashboard(buildpackInfo)
+
+	d := &Dashboard{
+		app:         app,
+		appTree:     appTree,
+		builderTree: builderTree,
+		planList:    planList,
+		logsView:    logsView,
+	}
+
+	imagesView := tview.NewFlex().
+		SetDirection(tview.FlexRow).
+		AddItem(appTree, 0, 1, false).
+		AddItem(builderTree, 0, 1, true)
+
+	imagesView.
+		SetBorder(true).
+		SetTitleAlign(tview.AlignLeft).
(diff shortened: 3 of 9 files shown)


importhuman (commented):
Hey @aemengo, this looks insanely cool! Had a couple of questions:

- When starting the build, the initial steps show up in the terminal rather than the dashboard. Is this expected in certain cases? The mockup in the RFC opened up the dashboard immediately, hence my question. (I'd built the image with a Dockerfile before, so there's no pulling all the layers in the attached video)
- I didn't get the detect phase in my dashboard. The 1st video is building an already existing image (though without changes to source code), and the 2nd is building a new image (same name as earlier, but I'd deleted the image before building). 
- In which cases will the plan be shown? I didn't get anything in either case.
- It might be helpful to add a "Press control + c to exit dashboard" message after the build succeeds/fails, since this is meant to be beginner friendly, and also because some folks like me might not realize that the way to exit the dashboard is the same as stopping a running command. 

https://user-images.githubusercontent.com/69148722/135239664-5e946fc4-fca2-43f2-8d32-5715f8e018a2.mov


https://user-images.githubusercontent.com/69148722/135239705-6f7c696a-843f-4ab2-b53f-bef717affde1.mov

aemengo:
@importhuman 
> When starting the build, the initial steps show up in the terminal rather than the dashboard.

Yes, it's expected. Doing the mockup, exactly, was too hard 😕. For now I'm happy to compromise with how it is now.

> I didn't get the detect phase in my dashboard.

That was also intended. I thought it looked better that way because the `plan` is supposed to show the same information. But, I don't feel strongly about this at all.

> In which cases will the plan be shown?

This is definitely not intended. It's always supposed to show. Could be a bug.
The regex is supposed to match `(\S+\/\S+)\s+([\d\.]+)`. Maybe it's because of the lifecycle version? What does the `DETECT` phase logs say without `--interactive`?

> It might be helpful to add a "Press control + c to exit dashboard" message

I think this is a great idea 🙌🏾
If you have a _vision_ for how it should look, maybe you can draft the issue for it.

> this looks insanely cool!

Just you wait. It's gonna be _even cooler!_ 😁

importhuman:
Ah, now that I think about it the mockup would be before adding the feature itself.

> What does the DETECT phase logs say without --interactive?

```
===> DETECTING
3 of 6 buildpacks participating
google.go.runtime  0.9.1
google.go.build    0.9.0
google.utils.label 0.0.1
```

aemengo:
@importhuman Thanks for the providing the output, It was definitely a bug. I wasn't expecting the output to look like that. 
Should be: _Fixed in latest commit_ 👍🏾

dfreilich (approved):
In terms of code and experience, it's fabulous – thank you so much @aemengo! 
Four small UX nits, but they can definitely be added in a future PR: 
1. I thought that I would be able to do something, while in the interactive mode, by selecting ENTER, or would be able to go into a separate pane, but it seems like I wasn't able to. Is the intention, ultimately, that there would be actions given? 
2. As a result of not being able to change windows, once Build logs scrolled past, I was unable to see them by scrolling up in my terminal, though maybe that's my iterm settings
3. I think having a clear commands window, at the least saying Ctrl+c to close the interactive window, would be great
4. It wasn't initially intuitive that when I exited out of the interactive window, that the build was still continuing behind the scene. That's ok, but I do think signaling that to the user (maybe by printing a message out in the logger when they close the interactive window, that the build is still happening and they can exit out of it by pressing Ctrl+c), would make it much clearer for the users. 

What do you think of that? 

Regardless, this is really great work, and I'm happy to merge it in, once the merge conflicts are resolved

aemengo:
@dfreilich Thanks for reviewing this PR. I really appreciate the time and effort. 

> 1. I thought that I would be able to do something, while in the interactive mode, by selecting ENTER [...] Is the intention, ultimately, that there would be actions given?

Yes, this is the intention. I have a follow-up branch to this PR that will introduce a _"dive"_ panel but delineated by buildpack. Here's an example:

Screen Shot 2021-11-10 at 10 47 54 PM

> 2. As a result of not being able to change windows, once Build logs scrolled past, I was unable to see them by scrolling up in my terminal, though maybe that's my iterm settings

This will also be fixed in the follow-up branch mentioned above.

> 3. I think having a clear commands window, at the least saying Ctrl+c to close the interactive window, would be great

I share the sentiment, but I simply haven't come up with the UX yet. It will definitely come up in a subsequent PR though.

> 4. It wasn't initially intuitive that when I exited out of the interactive window, that the build was still continuing behind the scene. [...].

That's something I didn't think of.. Would definitely like to address this is in a subsequent PR.

PR #1282, opened 2021-09-13
Support for SSH connection

### LINKED ISSUE #1259: pack doesn't honor `DOCKER_HOST` using `ssh`
### Summary

`pack` doesn't honor `DOCKER_HOST` that uses `ssh`. 

### Reproduction

1.  Have docker running on a remote host that is accessible via ssh.
2. Optionally enable passwordless ssh access by setting an `authorized_keys` file on the remote host.
3. Run the following:

```
> set -x DOCKER_HOST 'ssh://ubuntu@192.168.1.130'  # chang the IP to the remote host you setup
> docker images  # confirm you see images from the remote host
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
ubuntu       focal     1318b700e415   3 weeks ago   72.8MB
> pack build maven   # fails, trying to connect over HTTP to the remote host, but the remote host uses SSH access
ERROR: failed to build: failed to fetch builder image 'index.docker.io/paketobuildpacks/builder:base': error during connect: Post "http://ubuntu%40192.168.1.130/v1.38/images/create?fromImage=paketobuildpacks%2Fbuilder&tag=base": dial tcp: lookup ubuntu@192.168.1.130: no such host
```

SSH is a convenient and secure way to access a remote docker host. It would be nice if this worked.

### Environment

```
> pack --version
0.20.0+git-66a4f32.build-2668

> docker version
Client:
 Cloud integration: 1.0.17
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.16.4
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:22 2021
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.8
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.6
  Git commit:       75249d8
  Built:            Fri Jul 30 19:52:33 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.9
  GitCommit:        e25210fe30a0a703442421b0f60afac609f950a3
 runc:
  Version:          1.0.1
  GitCommit:        v1.0.1-0-g4144b63
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

matejvasek:
## Summary
This allows for connection via SSH by setting `DOCKER_HOST`.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1259


matejvasek:
@dmikusa-pivotal please try this out.

You might need to turn off `SELinux` in the `podman` VM. Also you will need to set `--docker-host` flag (e.g. `--docker-host=unix:///var/run/docker.sock`).

cmd/cmd.go
```diff
@@ -133,7 +139,42 @@ func initConfig() (config.Config, string, error) {
...
-	client, err := pack.NewClient(pack.WithLogger(logger), pack.WithExperimental(cfg.Experimental), pack.WithRegistryMirrors(cfg.RegistryMirrors))
+	dockerClientOpts := []dockerClient.Opt{
+		dockerClient.WithVersion("1.38"),
+	}
+
+	usesSSH := false
+	if dockerHost, ok := os.LookupEnv("DOCKER_HOST"); ok {
+		if _url, err := url.Parse(dockerHost); err == nil && _url.Scheme == "ssh" {
+			usesSSH = true
+			dialer, err := sshdialer.CreateDialContext(_url, "", "")
```
matejvasek, line 150: Should I use some envvar (e.g. `DOCKER_HOST_SSH_IDENTITY`) and use it as args for `CreateDialContext()`?

matejvasek:
Work is still in progress. There are no prompts for password/passphrase.

matejvasek:
Why is CI using go 1.14?

cmd/cmd.go
```diff
@@ -133,10 +150,169 @@ func initConfig() (config.Config, string, error) {
...
+		dockerClient.WithVersion("1.38"),
+	}
+
+	usesSSH := false
+	if dockerHost, ok := os.LookupEnv("DOCKER_HOST"); ok {
+		if _url, err := url.Parse(dockerHost); err == nil && _url.Scheme == "ssh" {
+			usesSSH = true
+			credentialsConfig := sshdialer.Config{
+				Identity:           os.Getenv("DOCKER_HOST_SSH_IDENTITY"),
+				PassPhrase:         os.Getenv("DOCKER_HOST_SSH_IDENTITY_PASSPHRASE"),
```
matejvasek, line 176: How should we name these envvars?

matejvasek:
@dmikusa-pivotal please try this out.

matejvasek:
Also it would be great if somebody tried this on Windows.

cmd/cmd.go
```diff
@@ -133,10 +150,169 @@ func initConfig() (config.Config, string, error) {
...
+	}
+	client, err := pack.NewClient(pack.WithLogger(logger), pack.WithExperimental(cfg.Experimental), pack.WithRegistryMirrors(cfg.RegistryMirrors), pack.WithDockerClient(dc))
 	if err != nil {
 		return pack.Client{}, err
 	}
 
 	return *client, nil
 }
+
+func initDockerClient() (dockerClient.CommonAPIClient, error) {
```
matejvasek, line 165: Is this file right place for this function (and the functions that it refers to).

matejvasek:
@joshuawhite929 please try this out.

dmikusa-pivotal:
Sorry for the delay. It is working for me.

I checked out your code, `make build` and used `./out/pack` to do a pack build. I had `DOCKER_HOST` set to `ssh://user@my-vm:port`. It didn't require setting `--docker-host`. 

Looks good!

cmd/cmd.go
```diff
@@ -1,6 +1,23 @@
...
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+
+	"golang.org/x/crypto/ssh"
```
aemengo, line 15: Let's try to get these imports into 3 groups: **stdlib**, **remote**, **local**

cmd/cmd.go
```diff
@@ -133,10 +150,169 @@ func initConfig() (config.Config, string, error) {
...
+			}
+			dockerClientOpts = append(dockerClientOpts,
+				dockerClient.WithHTTPClient(httpClient),
+				dockerClient.WithHost("http://example.com/"),
+				dockerClient.WithDialContext(dialContext),
+			)
+		}
+	}
+	if !usesSSH {
+		dockerClientOpts = append(dockerClientOpts, dockerClient.FromEnv)
```
aemengo, line 210: How do we feel about `return nil if !usesSSH`? That way we could rely on the [fallback initializer](https://github.com/buildpacks/pack/blob/main/client.go#L164) and not have to duplicate these options? But then it would probably warrant a comment (_no strong feelings_)... 🤷🏾‍♂️

cmd/cmd.go
aemengo: Are these the same ones that Docker uses? If so, we should keep them

aemengo (commented):
_(fearfully)_ attempting to validate on Windows here. But not really sure how these VMs are set up to be able to communicate over SSH.

@dmikusa-pivotal, do you have some sort of ssh proxy forwarding traffic from 22 to a Docker daemon listening on a port? Maybe we can look at your setup together to speed this along.

dmikusa-pivotal:
@aemengo 

>@dmikusa-pivotal, do you have some sort of ssh proxy forwarding traffic from 22 to a Docker daemon listening on a port? Maybe we can look at your setup together to speed this along.

You need a VM with two things to make this work:

1. OpenSSH installed. Basically, you should be able to `ssh` into the VM. For a Linux VM, just install your distro's openssh server package.
2. Docker Daemon running. You can use the Docker instructions for installing Docker community. If you can `docker version` on the remote VM as the user you're planning to connect with over `ssh` then you should be set (you may also need to add the user you're connecting with to the docker group so it can access the daemon without needing sudo).

There's nothing specific you have to do to Docker to make it support this. It's basically functionality in the docker client & now pack to run docker commands remotely. 

Hope that helps!

aemengo:
@dmikusa-pivotal Thanks for the help. I followed your instructions but am still having trouble using an ssh connection on a windows machine. I am able to use the ssh client directly.

Please let me know if I'm doing something wrong. Happy to look at this with someone.

```powershell
PS> $env:DOCKER_HOST="ssh://aemengo@127.0.0.1"
PS> $env:DOCKER_HOST_SSH_IDENTITY="C:\Users\aemengo\Desktop\rdp\id_ed25519"
PS> pack build aemengo/hello -p ..\hello -B cnbs/sample-builder:dotnet-framework-1809
ERROR: failed to build: failed to fetch builder image 'index.docker.io/cnbs/sample-builder:dotnet-framework-1809': error during connect: Post "http://example.com%2F/v1.38/images/create?fromImage=cnbs%2Fsample-builder&tag=dotnet-framework-1809": ssh: rejected: connect failed (open failed)
```

dmikusa-pivotal:
@aemengo - FYI, I only tested with password-less login. I'm not sure if it works if you have to login over ssh. 

Are you able to `ssh aemengo@127.0.0.1` without any additional prompts? Password most notably, but also make sure you've trusted the host key.

Also, can you `docker images` and see images from the remote docker? Docker should also work over ssh. If that works but pack doesn't, then it would be something specific with pack that's failing.

matejvasek:
There should be prompt for password and for temporary trusting the server.

matejvasek:
so `ssh "ssh://aemengo@127.0.0.1" -i "C:\Users\aemengo\Desktop\rdp\id_ed25519"` does work?

matejvasek:
Does something like:
`ssh -NL localhost:2374:/var/run/docker.sock  "ssh://aemengo@127.0.0.1" -i "C:\Users\aemengo\Desktop\rdp\id_ed25519"`
work?

matejvasek:
also what Daniel said: try whether `docker` CLI works

aemengo:
```
PS> $env:DOCKER_HOST="ssh://non-existent-user@127.0.0.1"
PS> docker images
non-existent-user@127.0.0.1's password:
non-existent-user@127.0.0.1's password:
non-existent-user@127.0.0.1's password:
non-existent-user@127.0.0.1's password:
non-existent-user@127.0.0.1's password:
non-existent-user@127.0.0.1's password:
error during connect: Get http://docker/v1.39/images/json: command [ssh -l non-existent-user 127.0.0.1 -- docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr="Permission denied, please try again.\r\nPermission denied, please try again.\r\nReceived disconnect from 127.0.0.1 port 22:2: Too many authentication failures\r\nDisconnected from 127.0.0.1 port 22\r\n"

PS> $env:DOCKER_HOST="ssh://aemengo@127.0.0.1"
PS> docker images
REPOSITORY                                                        TAG                     IMAGE ID            CREATED             SIZE
lifecycle/acceptance/builder                                      latest                  0abf71bb6ad0        5 weeks ago         272MB
<none>                                                            <none>                  9177566a3d0a        5 weeks ago         272MB
lifecycle/acceptance/analyzer                                     latest                  942a1f62e9b8        5 weeks ago         272MB
...
```

If an environment is needed to test, I'm happy to assist.

cmd/cmd.go
```diff
@@ -133,10 +150,169 @@ func initConfig() (config.Config, string, error) {
...
+			httpClient := &http.Client{
+				// No tls
+				// No proxy
+				Transport: &http.Transport{
+					DialContext: dialContext,
+				},
+			}
+			dockerClientOpts = append(dockerClientOpts,
+				dockerClient.WithHTTPClient(httpClient),
+				dockerClient.WithHost("http://example.com/"),
```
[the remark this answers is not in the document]
matejvasek: I could use `http://dummy` or `http://placeholder`, wdyt?

cmd/cmd.go
```diff
@@ -133,10 +150,169 @@ func initConfig() (config.Config, string, error) {
...
+	usesSSH := false
+	if dockerHost, ok := os.LookupEnv("DOCKER_HOST"); ok {
+		if _url, err := url.Parse(dockerHost); err == nil && _url.Scheme == "ssh" {
+			usesSSH = true
+			credentialsConfig := sshdialer.Config{
+				Identity:           os.Getenv("DOCKER_HOST_SSH_IDENTITY"),
+				PassPhrase:         os.Getenv("DOCKER_HOST_SSH_IDENTITY_PASSPHRASE"),
+				PasswordCallback:   newPasswordCbk(),
+				PassPhraseCallback: newPassPhraseCbk(),
+				HostKeyCallback:    newHostKeyCbk(),
```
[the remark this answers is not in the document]
matejvasek: password can be incorporated in DOCKER_HOST URL

cmd/cmd.go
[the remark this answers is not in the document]
matejvasek: the host is not important as connection is provided by custom dialer

internal/sshdialer/ssh_dialer.go
```diff
@@ -0,0 +1,288 @@
+// NOTE: this code is based on "github.com/containers/podman/v3/pkg/bindings"
```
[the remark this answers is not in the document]
matejvasek: It's not verbatim copy I did some changes:
a) You are asked if you trust a server, in podman if server is not in `known_hosts` then it is trusted no question asked.
b) I changed order of crypo.algo. priorities, also I may do further changes like ordering it according local config if needed

Also I don't want to introduce another dependency for just few functions.

Most complexity is in tests that I added.

cmd/cmd.go
matejvasek: No they are not used by docker. It's my addition. With docker you need to put identity at default location with default name or add it to ssh-agent, you cannot specify identity file AFAIK.

cmd/cmd.go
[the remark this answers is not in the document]
matejvasek: > I'm also curious why passphrase isn't interactive
it is interactive you can set either PassPhrase or PassPhraseCallback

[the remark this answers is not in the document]
matejvasek: HostKeyCallback -- how would envvar for that looked like?

cmd/cmd.go
matejvasek: Good point, I'll try that.

matejvasek:
@aemengo and I found out why it's not working on Windows. The ssh is not capable of forwarding Windows pipe the same way it forwards unix sockets.

https://github.com/PowerShell/Win32-OpenSSH/issues/435

It doesn't matter whether you run `pack` on Windows or Linux, the OS of the ssh server is important.
You can run `pack` on both Windows and Linux but the server has to be unix machine.

matejvasek:
@dmikusa-pivotal I did some changes could you please re-test it?

matejvasek:
How can I rerun `Codecov`?

matejvasek:
> @matejvasek Don't know your magic, but things seems to be working well on my windows machine 👍🏾 Please give me a few more to finish reviewing.

I allowed use of standard `docker system dial-stdio` mechanism if the commands happens to be in the remote machine. If it's not it will use tunneling.

n1hility:
Nice work on buildpack ssh support @matejvasek

Issue #1208, opened 2021-06-16
Add `--security-opt` flag


matejvasek:
### Description
The `docker` CLI has flag `--security-opt` allowing to disable `SELinux` labeling.
This is useful when local `SELinux` policy is causing troubles with mounting docker socket.

### Proposed solution
Add the `--security-opt` flag so user can disable labeling for containers created by `pack` during build. Alternatively we could and some universal flag, something like `--other-docker-opts` where user could set some arbitrary `docker` flags. But that is probably overkill.

### Describe alternatives you've considered
Without this user must disable `SELinux` or create policy, however non-root user might not be able to do so.
In addition creating `SELinux` policy is not easy, and disabling `SELinux` is not nice.

matejvasek:
cc @cardil

jromero:
I feel dirty proposing it but how terrible would it be to disable SELinux labeling by default? Given that it's a very specific set of containers and operations there in I do wonder if it should be more of pack's "opinion" as oppose to propagating the control to the user and require them to  discover it.

cardil:
I also think that exposing another CLI option for pack isn't a good idea. In my opinion, it would be better to add the security opt of `label=disable` to containers that mount docker socket. That's the analyzer container, for sure. Maybe some others as well.

matejvasek:
> I feel dirty proposing it but how terrible would it be to disable SELinux labeling by default?

May feels dirty but I am not against.

jromero:
If there are no objections from @dfreilich, I'd be open to accepting a PR that adds the security opt `label=disable` as part of the `WithDaemonAccess ` option:

https://github.com/buildpacks/pack/blob/dc17e6486504896f32661d26a3a8c4a86b3f818c/internal/build/phase_config_provider.go#L126-L153

jromero:
@sclevine, as security compliance officer 😉, you might also have some thoughts here.

sclevine:
This seems reasonable for containers that need access to the daemon, given that they never execute untrusted buildpack/app code and are generally only used for local builds.

We should document that these containers have host access with privileges potentially equivalent to the daemon host process.

PR #1307, opened 2021-10-11
security opt label disable added

Shashankft9:
## Summary
fixes https://github.com/buildpacks/pack/issues/1208

- added security option `label=disabled` as mentioned here: https://github.com/buildpacks/pack/issues/1208#issuecomment-862469906
- not sure where this has to be documented as talked about in the issue mentioned above, please provide the pointers and I will create a PR for that


Shashankft9:
hey @dfreilich this is ready for another review I think

Issue #1289, opened 2021-09-17
[v0.21.0] Buildpack does not receive build-time env vars after first comma


arjun024:
### Summary

With v0.21, the `build` executable of the buildpack does not receive build-time env vars after first comma when running `pack build`

---

##### Steps

1. I'm using https://github.com/arjun024/print-cnb that simply prints out `MY_ENV` environment variable.

2.  v0.20.0 (Success)

```
$ pack --version
0.20.0+git-66a4f32.build-2668

$ pack build myimage --env MY_ENV=first,second -b ./print-cnb
...
...
===> BUILDING
[builder] MY_ENV=first,second
....
```

3. v0.21.0 (Failure)
```
$ /tmp/pack21 --version
0.21.0+git-a9c294a.build-2814

$ pack build myimage21 --env MY_ENV=first,second -b ./print-cnb
...
...
===> BUILDING
[builder] MY_ENV=first
....
```

### More Info

See Github Action failures associated with the  following Pull Request that updates pack cli on integration test suite run by Paketo Buildpacks 
https://github.com/paketo-buildpacks/node-run-script/pull/24

---

##### docker info
```
Client: Docker Engine - Community
 Version:           20.10.3
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        48d30b5
 Built:             Fri Jan 29 14:33:13 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:44:13 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

Related issue: https://github.com/buildpacks/pack/issues/1288

jromero:
Getting out a patch!

Resolved by e09e397bc45315fa537fded032e57e678525c0b4

Issue #1288, opened 2021-09-17
[v0.21.0] Unable to pass complex env to pack build


arjun024:
### Summary
With v0.21, `pack build` fails when passed a complex build time env variable
like `--env BP_GO_GENERATE_FLAGS='-run "^//go:generate sleep"'`

---

##### Steps

1. Clone and change directory to https://github.com/paketo-community/go-generate/blob/v0.0.5/integration/testdata/args_and_flags/main.go

2.  v0.20.0 (Success)

```
$ pack --version
0.20.0+git-66a4f32.build-2668

$ pack build goimage --builder paketobuildpacks/builder:full \
--env BP_GO_GENERATE_FLAGS='-run "^//go:generate sleep"'
...
...
Successfully built image goimage
```

3. v0.21.0 (Failure)
```
$ /tmp/pack21 --version
0.21.0+git-a9c294a.build-2814

$ /tmp/pack21 build goimage21 --builder paketobuildpacks/builder:full \
--env BP_GO_GENERATE_FLAGS='-run "^//go:generate sleep"'
ERROR: failed to parse env line 'BP_GO_GENERATE_FLAGS=-run "^//go:generate sleep"': invalid syntax between 0 and 48
```

### More Info

See Github Action failures associated with the 2 following Pull Requests that updates pack cli on integration test suite run by Paketo Buildpacks 
https://github.com/paketo-community/go-generate/pull/88
https://github.com/paketo-buildpacks/php-web/pull/200

---

##### docker info
```
Client: Docker Engine - Community
 Version:           20.10.3
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        48d30b5
 Built:             Fri Jan 29 14:33:13 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:44:13 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

[...]

PR #1323, opened 2021-11-11
Fix help text for StringArrayVar flags

vindvaki:
## Summary

`StringArrayVar` flags do not support comma-separated lists. They must instead be repeated once per value. [Background](https://gitlab.com/gitlab-org/cluster-integration/auto-build-image/-/issues/38#note_731233862).

## Output

For example

```
pack help build
```

has the following to say about `--volume`:

#### Before

```
      --volume stringArray          Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.
                                    - 'host path': Name of the volume or absolute directory path to mount.
                                    - 'target path': The path where the file or directory is available in the container.
                                    - 'options' (default "ro"): An optional comma separated list of mount options.
                                        - "ro", volume contents are read-only.
                                        - "rw", volume contents are readable and writeable.
                                        - "volume-opt=<key>=<value>", can be specified more than once, takes a key-value pair consisting of the option name and its value.
                                    Repeat for each volume in order, or supply once by comma-separated list
```

#### After

```
      --volume stringArray          Mount host volume into the build container, in the form '<host path>:<target path>[:<options>]'.
                                    - 'host path': Name of the volume or absolute directory path to mount.
                                    - 'target path': The path where the file or directory is available in the container.
                                    - 'options' (default "ro"): An optional comma separated list of mount options.
                                        - "ro", volume contents are read-only.
                                        - "rw", volume contents are readable and writeable.
                                        - "volume-opt=<key>=<value>", can be specified more than once, takes a key-value pair consisting of the option name and its value.
                                    Repeat for each volume in order (comma-separated lists not accepted)
```

## Documentation

    - [ ] Yes, see #___
    - [ ] No 

## Related

Resolves #___


dfreilich:
Related: #1212, #1288, #1289

dfreilich:
Thanks for the addition, @vindvaki ! I agree that the clarification is needed

dfreilich (approved):

PR #1326, opened 2021-11-15
Add file explorer to pack-interact workflow

aemengo:
## Summary
Adds a _dive_-style container layers explorer to `--interactive` mode, but delineated by buildpack

## Output
Screen Shot 2021-11-10 at 10 47 54 PM

## Documentation

    - [ ] Yes, see #___
    - [ ] No
    - [x] At some point

## Related

Resolves #___


internal/termui/detect.go
```diff
@@ -46,6 +46,10 @@ func (d *Detect) Stop() {
 	d.doneChan <- true
 }
 
+func (d *Detect) SetNodes(map[string]*tview.TreeNode) {
+	// not implemented
```
dfreilich, line 50: What will be happening here?

dfreilich (approved):
From a UX perspective, this is fabulous. I really appreciated how, even before you added in a toolbar for what commands are valid, that the colors of the buildpacks changed when you were able to select them. I thought that was a very nice signifier.

aemengo:
@dfreilich You still had good feedback about showing a keymap - at least for **Ctrl-C**. I haven't forgotten; it will be addressed in a subsequent PR.

internal/termui/detect.go
aemengo: Unfortunately that was me simply trying to satisfy the `type page interface`. That method is never intended to be implemented, but the _other page_ (Dashboard) does have it implemented.

Overall, I have serious misgivings the page orchestration. I planned to refactor but simply didn't come up with something I liked. If you'd prefer, I can spruce it up relatively soon.

```
type page interface {
	Handle(txt string)
	Stop()
	SetNodes(nodes map[string]*tview.TreeNode)
}
```

dfreilich: Hmmm. I think it's ok to merge it in without it spruced up, given how it's hidden within an experimental flag and is going to be worked on more, but maybe add that in in a clarifying comment?

aemengo: I added clarifying comments.

Issue #1319, opened 2021-11-08
ssh_dialer_test.go should use sclevine/spec reporter


aemengo:
### Description
The specs in **ssh_dialer_test.go** should use the sclevine/spec reporter; for consistency with the rest of the specs

aemengo:
Fixed by: https://github.com/buildpacks/pack/pull/1327

PR #1327, opened 2021-11-16
SSH test improvement: use Go SSH server not containerized OpenSSH 

### LINKED ISSUE #1325: ssh_dialer_test.go should use mock Go SSH server instead of container
### Description
The test right now requires Linux Containerization so test cannot be run on macOS.
It could run on Windows with Windows' containerization but that would require adding new Windows docker file,
and probably some more Windows specific code.

### Proposed solution
Replace containerized SSH testing server with SSH server in Go.

This would be also useful for porting the code back to `podman`.

### Additional context
See https://github.com/knative-sandbox/kn-plugin-func/blob/main/docker/docker_client_ssh_test.go#L62

matejvasek:
Resolves #1325


matejvasek:
@aemengo @jromero PTAL

aemengo:
@matejvasek Please consider the following PR: [matejvasek/pack/pull/44](https://github.com/matejvasek/pack/pull/44), which makes some amendments to `ssh_dialer_test.go`, to make it consistent with the rest of pack's spec. 

It would also address: https://github.com/buildpacks/pack/issues/1319

matejvasek:
@aemengo do you have any idea why the two windows test failed? Is it just a flake?

jromero:
@matejvasek just reran those. Appear like there was a temporary outage with codecov. All good now.

aemengo (approved):

PR #1336, opened 2021-11-24
Fix: overwrite existing buildpacks correctly 

### LINKED ISSUE #1322: `pack build --buildpack` doesn't overwrite existing buildpacks in the builder correctly
### Summary
When `pack build` is run using the `--buildpack` argument, and the provided buildpack ID and version match a buildpack already in the builder image, pack claims to overwrite the existing buildpack, however actually only overlays the imported buildpack on top of it, leaving any stray files intact.

ie: The resultant buildpack in the ephemeral builder used for the build ends up being a hybrid of both the old and new buildpack files.

This is presumably because there is no "remove existing buildpack" step in here, only "unpack tar directly into layer":
https://github.com/buildpacks/pack/blob/339b49e8c10978bb80a899fe8668741d9f730aba/internal/builder/builder.go#L409-L448

---

##### Steps

1. `git clone https://github.com/buildpacks/samples && cd samples`
2. `touch buildpacks/java-maven/buildpack-in-builder`
3. `pack builder create testcase --config builders/alpine/builder.toml`
4. `rm buildpacks/java-maven/buildpack-in-builder && touch buildpacks/java-maven/buildpack-local`
5. `echo -e '#!/usr/bin/env bash\nls -al "${CNB_BUILDPACK_DIR}"\nexit 1' > buildpacks/java-maven/bin/detect`
6. `pack build sample-app --builder testcase --buildpack buildpacks/java-maven/ --path apps/java-maven/ --verbose`

##### Current behavior
- Pack reports `buildpack samples/java-maven@0.0.1 already exists on builder and will be overwritten`
- However as can be seen in the `ls` output, files from the buildpack that was in the builder have been left behind (the file `buildpack-in-builder` should no longer exist).

```
$ pack build sample-app --builder testcase --buildpack buildpacks/java-maven/ --path apps/java-maven/ --verbose
...
Downloading buildpack from URI: file:///Users/emorley/src/samples/buildpacks/java-maven
Adding buildpack samples/java-maven version 0.0.1 to builder
Setting custom order
Creating builder with the following buildpacks:
-> samples/java-maven@0.0.1
-> samples/kotlin-gradle@0.0.1
-> samples/hello-universe@0.0.1
-> samples/hello-world@0.0.1
-> samples/hello-moon@0.0.1
-> samples/java-maven@0.0.1
buildpack samples/java-maven@0.0.1 already exists on builder and will be overwritten
  - existing diffID: sha256:ba3075c5689ac3a324105cd6690650f34decfd4536f167c05cb552944c850a08
  - new diffID: sha256:b036c35ff97ec7dff1f2ae1bc323085b056c0e4d55727be86234cdd1b2ea8e74
Adding buildpack samples/java-maven@0.0.1 (diffID=sha256:b036c35ff97ec7dff1f2ae1bc323085b056c0e4d55727be86234cdd1b2ea8e74)
...
[detector] ======== Output: samples/java-maven@0.0.1 ========
[detector] total 20
[detector] drwxr-xr-x    1 root     root          4096 Jan  1  1980 .
[detector] drwxr-xr-x    1 root     root          4096 Jan  1  1980 ..
[detector] -rw-r--r--    1 root     root           342 Jan  1  1980 README.md
[detector] drwxr-xr-x    1 root     root          4096 Jan  1  1980 bin
[detector] -rw-r--r--    1 root     root             0 Jan  1  1980 buildpack-in-builder
[detector] -rw-r--r--    1 root     root             0 Jan  1  1980 buildpack-local
[detector] -rw-r--r--    1 root     root           477 Jan  1  1980 buildpack.toml
```

##### Expected behavior
- Pack reports `buildpack samples/java-maven@0.0.1 already exists on builder and will be overwritten`
- The `ls` output shows that files from the buildpack that was in the builder have been removed (ie: the file `buildpack-in-builder` should no longer exist).

---

##### pack info
```
Pack:
  Version:  0.22.0+git-26d8c5c.build-2970
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.11.3

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6

Config:
  experimental = true

  [[trusted-builders]]
    name = "[REDACTED]"
```

##### docker info
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.3)
  compose: Docker Compose (Docker Inc., v2.0.0)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 23
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.10.47-linuxkit
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 7.773GiB
 Name: docker-desktop
 ID: REDACTED
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

fatiiates:
## Summary
According to the issue, the --buildpack flag was not correctly overwriting the buildpack. So I created a whiteout layer to fix this. When it is understood that the buildpacks have the same ID and version, a tarball is created with that directory whiteout and added between these layers. Since it is added as a child layer from the newly added buildpack, only those in the same directory in the lower layer will appear as whiteout.

## Output

Following the steps below, the files in the `ls` output should be completely new buildpack files passed as parameters in the --buildpack flag.

1. ```git clone https://github.com/buildpacks/samples && cd samples```  
2. ```touch buildpacks/java-maven/buildpack-in-builder```  
3. ```pack builder create testcase --config builders/alpine/builder.toml```  
4. ```rm buildpacks/java-maven/buildpack-in-builder && touch buildpacks/java-maven/buildpack-local```  
5. ```echo -e '#!/usr/bin/env bash\nls -al "${CNB_BUILDPACK_DIR}"\nexit 1' > buildpacks/java-maven/bin/detect```  
6. ```pack build sample-app --builder testcase --buildpack buildpacks/java-maven/ --path apps/java-maven/ --verbose```  

#### Before
before

#### After
after

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1322


aemengo:
Thanks so much for tackling this issue! A lot of us are on holiday, so please forgive us if you don't get a review until next week.

fatiiates:
It's okay @aemengo, have a nice holiday.

internal/builder/builder.go
```diff
@@ -417,6 +417,37 @@ func addBuildpacks(logger logging.Logger, tmpDir string, image imgutil.Image, ad
 			if existingBPInfo.LayerDiffID == diffID.String() {
 				logger.Debugf("Buildpack %s already exists on builder with same contents, skipping...", style.Symbol(bpInfo.FullName()))
 				continue
+			} else {
+
```
dfreilich, line 421: nit: remove extra line

internal/builder/builder.go
```diff
@@ -417,6 +417,37 @@ func addBuildpacks(logger logging.Logger, tmpDir string, image imgutil.Image, ad
...
+				bpWhiteoutsTmpDir := filepath.Join(tmpDir, strconv.Itoa(i)+"_whiteouts")
+				if err := os.MkdirAll(bpWhiteoutsTmpDir, os.ModePerm); err != nil {
+					return errors.Wrap(err, "creating buildpack whiteouts temp dir")
+				}
+
+				deletedMaps := map[string][]byte{
+					filepath.Join(
+						cnbDir,
+						"buildpacks",
```
dfreilich, line 430: Can you make that a file `var`, similar to `cnbDir`?

dfreilich (changes requested):
Thanks for the PR, @fatiiates! This definitely looks like it's on the right path. Is there any chance you could add a test for this case?

internal/builder/builder.go
fatiiates: I will deal with them ASAP.

internal/builder/builder.go
```diff
@@ -417,6 +417,37 @@ func addBuildpacks(logger logging.Logger, tmpDir string, image imgutil.Image, ad
...
+					)
+				}
+
+				if err := image.AddLayer(whiteoutsTarFile); err != nil {
+					return errors.Wrapf(err,
+						"adding whiteout layer tar for buildpack %s",
+						style.Symbol(bpInfo.FullName()),
+					)
+				}
+
```
dfreilich, line 450: nit: Can you remove this line?

dfreilich:
Hey @fatiiates  - for tests, we would be looking for a unit test that tests this functionality. I would look for inspiration at these current tests on that part of the code, which test the contents of the layers saved: https://github.com/buildpacks/pack/blob/4f04f34ecd57dfb268f60011b58376debfe01649/internal/builder/builder_test.go#L639-L752

Specifically, we should change it by adding a test which tries to add two buildpacks which have the same ID and version, but have different contents, and then verify that the contents of the 2nd buildpack, and not the 1st buildpack, are there. Does that make sense?

fatiiates:
Hey @dfreilich, thanks for the explanation. Everything is clearer now. And yes what you said makes perfect sense, that's how I'll do it.

internal/builder/builder.go
fatiiates: I changed that line, but I doubt I got it right :)

dfreilich (approved):
Beautiful! If you could just fix the DCO (you can see instructions for that by clicking on the link in the checks), we'll be happy to merge this in

dfreilich:
Merged in by #1362. Thank you for the help, @fatiiates ! 

Sorry for the hold-up with it, we had to figure out how to deal with some cases of a Windows image, which took me a bit of time.

fatiiates:
It's okay, I saw the other PR you opened but I couldn't help because I have final exams. Good news that the bug has been fixed! Congrats to both of us!

PR #1350, opened 2021-12-21
Remove `heroku/buildpacks:18` from trusted/suggested builders lists

edmorley:
## Summary
The `heroku/buildpacks:18` builder is based upon Heroku-18 / Ubuntu 18.04, which are due to enter their deprecation cycle at Heroku in the coming April/May.

In addition, we currently only run `heroku/buildpacks:20` on the Heroku platform, so whilst this builder does have some rudimentary CI coverage in our builder generation repository (https://github.com/heroku/pack-images), the older stack is unused and barely tested in practice. We've recently found a few internal developers who were inadvertently still using it, since they'd set it as the default builder some time ago and forgotten in the interim.

Lastly, we expected to release a new `heroku/buildpacks:22` builder based on Ubuntu-22 next year, at which point we'd have three builders in the suggested builders list, which seems excessive :-)

As such, the older stack has been removed from the trusted and suggested builders list. This is a breaking change for anyone using that image, since they will need to explicitly add it to their trusted builders list (so this will need to be in a new major Pack CLI release). However we expect this to affect very few users in practice.

## Output

See diff in output excerpt in tests.

## Documentation

    - [ ] Yes, see #___
    - [X] No


edmorley:
cc @hone @Malax

samj1912 (approved):

dfreilich (approved):
Thanks for the change! I'm sorry for the time it took to get to this.

PR #1353, opened 2022-01-05
fix: ssh-agent connection on Windows

matejvasek:


matejvasek:
@jromero @aemengo PTAL

aemengo:
@matejvasek have you already given this a look on a Windows machine? The project has some.

matejvasek:
@aemengo I have not, but colleague of mine @jrangelramos developed and tested this on some Windows VM.
See also https://github.com/knative-sandbox/kn-plugin-func/pull/733

jrangelramos:
+1 
I applied and tested similar fix on func. It works as expected.

dfreilich (approved):
Thanks for the help with this!

PR #1354, opened 2022-01-05
Revert "Bump github.com/docker/cli from 20.10.11+incompatible to 20.10.12+incompatible"

dfreilich:
Reverts buildpacks/pack#1343

It seems that this inadvertently broke the build, though I can't figure out why. I'll look into this further, but let's clear our pipelines first.


jromero:
I don't think this is necessary (the root cause of our woes). main is currently green without this change.

PR #1370, opened 2022-02-11
Set PACK_VERSION on Unix from GITHUB_REF_NAME

### LINKED ISSUE #1369: Pack 0.24.0 `--version` reports 0.0.0
### Summary
`pack --version` reports `0.0.0` in latest release.  This is set in the build time variable `PACK_VERSION`.  It appears that `PACK_VERSION` is set in [`.github/workflows/build.yml`](https://github.com/buildpacks/pack/blob/main/.github/workflows/build.yml#L57).

---

##### Steps
1. `go install github.com/buildpacks/pack/cmd/pack@latest`
    ```bash
    go: downloading github.com/buildpacks/pack v0.24.0
    go: downloading golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
    go: downloading github.com/BurntSushi/toml v1.0.0
    ....
    ```
2. ` ~/go/bin/pack --version`
   ```bash
    0.0.0
   ```

##### Current behavior
Reports version `0.0.0`

##### Expected behavior
I would expect this to report `0.24.0`

---

##### pack info
```bash
 ~/go/bin/pack report
Pack:
  Version:  0.0.0
  OS/Arch:  linux/amd64

Default Lifecycle Version:  0.13.3

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8

Config:
[[ REDACTED INTERAL CONFIG ]]
```

##### docker info

Not necessary in this report.

AidanDelaney:
Pull x.y.z out of releases/x.y.z in the GITHUB_REF_NAME variable

Signed-off-by: Aidan Delaney <adelaney21@bloomberg.net>

## Summary

Updates `build.yml` to correctly inject PACK_VERSION

#### Before

`pack --version` is `0.0.0`

#### After

`pack --version` is `0.0.24`

## Documentation

    - [ ] Yes, see #___
    - [X] No

## Related

Resolves #1369

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Derive pack version from branch name Unix
         if: runner.os != 'Windows'
         run: |
-          [[ $GITHUB_REF =~ ^refs\/heads\/release/(.*)$ ]] && version=${BASH_REMATCH[1]} || version=0.0.0
+          [[ $GITHUB_REF_NAME =~ ^release/(.*)$ ]] && version=${BASH_REMATCH[1]} || version=0.0.0
           echo "PACK_VERSION=${version}" >> $GITHUB_ENV
         shell: bash
       - name: Derive pack version from branch name Windows


AidanDelaney:
It _may_ be nicer to use `GITHUB_REF_NAME` over `GITHUB_REF`.  However, this PR does not fix any issue.

dfreilich:
Thanks for clarifying! This definitely can make sense; if you're going to make the change though, I think we should do it for all places in the file (it looks like there are 2 others. Could you change them as well?

jromero:
Per the [comment above](https://github.com/buildpacks/pack/pull/1370#issuecomment-1105328803), I'm closing this PR but feel free to reopen with the suggested modification if we'd like to move this change forward.

PR #1408, opened 2022-04-03
Set the Pack Homebrew formula version explicitly

edmorley:
Previously the Pack Homebrew formula did not set an explicit version, meaning that Homebrew attempts to detect it from the package URL. 

Unfortunately on ARM64, this results in the Pack version being detected as `64`, due to the additional `arm64` suffix in the package URL. This results in `pack upgrade` not detecting new pack CLI versions, so pack is never upgraded.

Now, the pack version is set explicitly, avoiding the issues with auto-detection. 

In addition, `version_scheme` has to be set, since otherwise Homebrew will not update the existing installations, as `64` would be treated as a newer version. See:
https://docs.brew.sh/Formula-Cookbook#version-scheme-changes

Lastly, whilst fixing this, I noticed a bug in the GitHub Actions workflow, whereby `tag_name` incorrectly had the `v` prefix removed before being passed to the GitHub releases API. This codepath is presumably unused most of the time (since it's only used when the workflow is manually triggered), which is why this had gone unnoticed.

Fixes buildpacks/homebrew-tap#20.

### DIFF
--- a/.github/workflows/delivery-homebrew.yml
+++ b/.github/workflows/delivery-homebrew.yml
@@ -29,11 +29,12 @@ jobs:
         with:
           script: |
             let payload = context.payload;
-            let tag = (payload.release && payload.release.tag_name) || (payload.inputs && payload.inputs.tag_name);
-            if (!tag) {
+            let tag_name = (payload.release && payload.release.tag_name) || (payload.inputs && payload.inputs.tag_name);
+            if (!tag_name) {
               throw "ERROR: unable to determine tag"
             }
-            let tag_name = tag.replace(/^v/, '');
+
+            core.setOutput("pack_version", tag_name.replace(/^v/, ''));
 
             var release = payload.release || await github.rest.repos.listReleases(context.repo)
                   .then(result => result.data.find(r => r.tag_name === tag_name))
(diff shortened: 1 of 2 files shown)


jromero (approved):

edmorley:
@jromero Thank you for reviewing/merging :-) I don't suppose you could trigger the workflow to update the generated files here?
https://github.com/buildpacks/homebrew-tap

jromero:
Gave it a [shot](https://github.com/buildpacks/pack/actions/runs/2102797755). Looks like we missed the reference to the outputs. Resolving...

jromero:
Resolved and triggered.

Results: https://github.com/buildpacks/homebrew-tap/blob/main/Formula/pack.rb

@edmorley want to confirm it all worked out?

edmorley:
@jromero Sorry I missed that - thank you for fixing & triggering the workflow.

Updates are working now :-)

```
$ brew update && brew upgrade
...
==> Upgrading 1 outdated package:
buildpacks/tap/pack 64 -> 0.24.1
==> Downloading https://github.com/buildpacks/pack/releases/download/v0.24.1/pack-v0.24.1-macos-arm64.tgz
==> Downloading from https://objects.githubusercontent.com/github-production-release-asset-2e65be/138544269/cd13ea4a-334d-492b-8b7c-dfd61237fc58?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-
######################################################################## 100.0%
==> Upgrading buildpacks/tap/pack
  64 -> 0.24.1

🍺  /opt/homebrew/Cellar/pack/0.24.1: 3 files, 18MB, built in 4 seconds

$ pack --version
0.24.1+git-3c7aff3.build-3213
```

PR #1418, opened 2022-04-11
Set SOURCE_DATE_EPOCH

### LINKED ISSUE #1281: Enable configuring the date used by reproducible builds
### Description

Presently when you build an image, all of the timestamps are zeroed out, which gives you a date like "41 years ago" for your images.

1. This is a confusion point for new users. Where does this date come from? Why is pack doing this? Why don't I see the actual date when I built my image?
2. It is inconvenient in a number of cases. Like if you are trying to compare two different image builds, there is no date by which to sort them. You need to keep track of the specific image hashes, not practical and not the way a human brain works.

### Proposed solution

Three suggestions:

1. Allow `pack build` to take as input the timestamp it should use. Instead of writing 0 as the timestamp, write this value.
2. By default, `pack build` will use the current timestamp and include that as a label on the image. 
3. Add a config setting for `pack` that users could customize if they want it to be zero by default or the current timestamp.

This would achieve a couple of things:

1. You're still consistently overwriting the timestamps on everything, which should achieve the same goal for reproducible builds.
2. The date on the image will appear to be the current time by default **which is what people expect**.
3. You can still reproduce an image build by taking the build timestamp from the label on the image you are trying to reproduce and passing that into `pack build`.
4. You can still use `0`. In fact, other platforms like kpack could ignore this and continue using zero as the default.

The current solution is where it always writes a 0 timestamp.

### Additional context
- [x] This feature should be documented somewhere. Yes. It would need to be documented in the pack documentation & reproducible build documentation.

Apologizes in advance if this is in the wrong repo. I wasn't entirely sure where to file this. Thanks!

natalieparellano:
pack should set SOURCE_DATE_EPOCH in exporter's env if --creation-time flag is provided

Also bumps default lifecycle version

#### Before

Image create time is always Jan 1, 1980

#### After

Image create time is configurable

## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

Resolves #1281

### DIFF
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -202,7 +202,7 @@ func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Co
 	cmd.Flags().StringVarP(&buildFlags.Builder, "builder", "B", cfg.DefaultBuilder, "Builder image")
 	cmd.Flags().StringVar(&buildFlags.CacheImage, "cache-image", "", `Cache build layers in remote registry. Requires --publish`)
 	cmd.Flags().BoolVar(&buildFlags.ClearCache, "clear-cache", false, "Clear image's associated cache before building")
-	cmd.Flags().StringVar(&buildFlags.DateTime, "creation-time", "", "Desired create time in the output image config. Accepted values are Unix timestamps (e.g., '1641013200'), or 'now'.")
+	cmd.Flags().StringVar(&buildFlags.DateTime, "creation-time", "", "Desired create time in the output image config. Accepted values are Unix timestamps (e.g., '1641013200'), or 'now'. Platform API version must be at least 0.9 to use this feature.")
 	cmd.Flags().StringVarP(&buildFlags.DescriptorPath, "descriptor", "d", "", "Path to the project descriptor file")
 	cmd.Flags().StringVarP(&buildFlags.DefaultProcessType, "default-process", "D", "", `Set the default process type. (default "web")`)
 	cmd.Flags().StringArrayVarP(&buildFlags.Env, "env", "e", []string{}, "Build-time environment variable, in the form 'VAR=VALUE' or 'VAR'.\nWhen using latter value-less form, value will be taken from current\n  environment at the time this command is executed.\nThis flag may be specified multiple times and will override\n  individual values defined by --env-file."+stringArrayHelp("env")+"\nNOTE: These are NOT available at image runtime.")
--- a/internal/commands/build.go
+++ b/internal/commands/build.go
@@ -4,7 +4,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/pkg/errors"
@@ -44,6 +46,7 @@ type BuildFlags struct {
 	GID                int
 	PreviousImage      string
 	SBOMDestinationDir string
+	DateTime           string
 }
 
 // Build an image from source code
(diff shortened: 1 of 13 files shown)


samj1912:
Should the flag instead be `--creation-time`?

natalieparellano:
Phew! Finally green...

internal/commands/build.go
```diff
@@ -44,6 +46,7 @@ type BuildFlags struct {
 	GID                int
 	PreviousImage      string
 	SBOMDestinationDir string
+	DateTime           string
```
jromero, line 49: This should probably match the flag a little bit better. `DateTime` is a little ambiguous.

```suggestion
	CreationTime           string
```

jromero (changes requested):
Two minor points:

- We should probably cary the term `CreationTime` across the entire code base instead of having an ambiguous `DateTime`
- Isn't this variable only respected on newer Platform API versions? If so, it seems like we should only be passing it to the container in those cases and warning the user if it's not applicable.

internal/commands/build.go
```diff
@@ -171,12 +179,30 @@ func Build(logger logging.Logger, cfg config.Config, packClient PackClient) *cob
...
+	return &parsedTime, nil
+}
+
 func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Config) {
 	cmd.Flags().StringVarP(&buildFlags.AppPath, "path", "p", "", "Path to app dir or zip-formatted file (defaults to current working directory)")
 	cmd.Flags().StringSliceVarP(&buildFlags.Buildpacks, "buildpack", "b", nil, "Buildpack to use. One of:\n  a buildpack by id and version in the form of '<buildpack>@<version>',\n  path to a buildpack directory (not supported on Windows),\n  path/URL to a buildpack .tar or .tgz file, or\n  a packaged buildpack image name in the form of '<hostname>/<repo>[:<tag>]'"+stringSliceHelp("buildpack"))
 	cmd.Flags().StringVarP(&buildFlags.Builder, "builder", "B", cfg.DefaultBuilder, "Builder image")
 	cmd.Flags().StringVar(&buildFlags.CacheImage, "cache-image", "", `Cache build layers in remote registry. Requires --publish`)
 	cmd.Flags().BoolVar(&buildFlags.ClearCache, "clear-cache", false, "Clear image's associated cache before building")
+	cmd.Flags().StringVar(&buildFlags.DateTime, "creation-time", "", "Desired create time in the output image config")
```
dfreilich, line 205: Could you add the valid options in the description of the flag?

internal/commands/build.go
dfreilich: Agree with @jromero , I think `DateTime` would be confusing in this context, where there could conceivably be multiple times. Being explicit that this is `CreationTime` would be great

dfreilich (changes requested):
Overall, as always, beautiful! just a few tweaks would be awesome

internal/commands/build.go
natalieparellano: > Isn't this variable only respected on newer Platform API versions?

This is interesting... I had a different memory from the [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0103-source-date-epoch.md) - I thought that while it would be spec'd in 0.9 (so that platforms can look in the spec and know how to use it), it would be respected in all platform api versions. However it looks like the [lifecycle implementation](https://github.com/buildpacks/lifecycle/blob/b6803be364429689a818a5d0a09db42bd21b9995/cmd/lifecycle/exporter.go#L442-L444) says otherwise. I'll change it.

dfreilich (approved):
Fantastic!

natalieparellano:
> The tests are green but I don't understand why - https://github.com/buildpacks/pack/runs/6009113466?check_suite_focus=true uses a lifecycle that doesn't support this feature. Investigating.

I found the issue - fixed in c23e970a (it was a bad assertion).

jromero (approved):

thedevelopnik:
Thank you so much! This is a long-awaited feature and I am so grateful it landed!

PR #1427, opened 2022-04-24
Add process specific working directories

### LINKED ISSUE #1422: Display process-specific working directories when available
### Acceptance Criteria

When a user runs `pack inspect <app-image>`
Then the output should include a new column under "Processes:" called "WORK DIR" for each process type listed

### Proposed Solution

1. Inspect the app image's configuration for a label `io.buildpacks.build.metadata`
2. For each process type in the contents look for `working-directory` property
3. Display that `working-directory` property in a `WORK DIR` column.

> NOTE: If label information is not present fallback to looking at the Config property `WorkingDir`:
> 
> ```
> docker inspect my-app | jq -r '.[0].Config'
> {
>   ...
>   "WorkingDir": "/workspace",
>   ...
>   "Labels": {
>     ...
>   }
> }
> ```

### References

- [x] https://github.com/buildpacks/spec/issues/216
- [x] https://github.com/buildpacks/spec/pull/290

imnitishng:
## Summary
- Show process' working directory from the `WorkingDirectory` param in metadata label
- If working directory from label is empty then use Image's inspect to figure out the working directory (stored as WorkingDir)
This PR depends on https://github.com/buildpacks/imgutil/pull/141, I did not find a way to expose the working directory so I had to go for this change, let me know if there's another way to handle the problem. 
Moreover, I had to update a bunch of dependencies to actually get the newly added `WorkingDirectory` param inside of `process`, which means `go.mod` and `go.sum` are modified.
I think we'll need to update the PR again with the new `imgutil` release if we decide to merge both PRs and go this route, if not then I'd be happy to work on the feedback.
Tests will be failing until we merge the PR for `imgutil` 

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related
https://github.com/buildpacks/pack/issues/1422

Resolves #1422


imnitishng:
Updating the PR soon, with passing test 😬

imnitishng:
I think we should be good to go once https://github.com/buildpacks/imgutil/pull/144 is merged

imnitishng:
Hey @jromero, I am having issues with the spacing for `WORK DIR`, for windows and linux. 
The spacing for both the platforms is a bit different, but the template is same. 
Should I create new template files `.txt` for windows and linux? Or should we modify the test in some other way?

jromero:
> Hey @jromero, I am having issues with the spacing for `WORK DIR`, for windows and linux. The spacing for both the platforms is a bit different, but the template is same. Should I create new template files `.txt` for windows and linux? Or should we modify the test in some other way? Once this is fixed, we should be good to go.

image

I can see the issue in Windows. It looks as though because `WORK DIR` comes after `ARGS` and on windows, args are pushing the column further to the right. 

Based on our conversation on `#buildpacks-platform`. I'd be comfortable removing the acceptance test for human-readable output comparison.

In otherwords, deleting these lines ([L976-L980](https://github.com/buildpacks/pack/blob/e4e8e588bd8bc2b3972abb8e730323ede8c5f86b/acceptance/acceptance_test.go#L976-L980), [L1627-L1631](https://github.com/buildpacks/pack/blob/e4e8e588bd8bc2b3972abb8e730323ede8c5f86b/acceptance/acceptance_test.go#L1627-L1631)) and any associated resources from `pack_fixtures`.

jromero (approved):

PR #1432, opened 2022-04-27
Fix: error of private registry in lifecycle phase

duanhongyi:
This PR is mainly used to solve the problem of private registry that cannot be verified during `pack build -- publish`.

The error is as follows:

```
[analyzer] ERROR: failed to : ensure registry read access to myprivate.registry.com/org/pack:bullseye-linux-amd64
ERROR: failed to build: executing lifecycle. This may be the result of using an untrusted builder: failed with status code: 1
```


jromero (approved):
This actually makes a lot of sense! Thank you for the fix!

Is this bug of high severity to your workflow or would you be okay waiting for the next natural release of pack (which may take a few weeks)?

PR #1437, opened 2022-05-08
Update error when bp order group is malformed for meta-buildpacks

### LINKED ISSUE #1384: `pack buildpack package` fails with unclear error message when missing version
### Summary
I am working bundling our individual buildpacks as meta-buildpacks and I forgot to include the `version` of buildpacks in my `order.groups`. This was a mistake on my part, but the error message from pack was not very helpful: 

```
ERROR: buildpack X references buildpack Y which is not present
```

There was no indication of the actual issue, which made it very difficult for me to track down.

##### Steps

1. Clone the [buildpack/samples](https://github.com/buildpacks/samples/) repo
2. Delete the `version` line in the `samples/hello-world`  `order.group` [here](https://github.com/buildpacks/samples/blob/main/buildpacks/hello-universe/buildpack.toml#L15)
3. Run `pack buildpack package my-buildpack --config ./package.toml`

##### Current behavior
The command fails with an error message like:

```
ERROR: buildpack samples/hello-universe@0.0.1 references buildpack samples/hello-world which is not present
```

##### Expected behavior

The error message should specifically indicate that the missing version is caused the failure

##### pack info

```
pack report
Pack:
  Version:  0.23.0+git-0db2c77.build-3056
  OS/Arch:  linux/amd64

Default Lifecycle Version:  0.13.1

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8

Config:

  [[trusted-builders]]
    name = "[REDACTED]"
```

imnitishng:
## Summary
The error for when pack encounters a malformed `buildpack.toml` order group entry referenced by `package.toml` wasn't very descriptive. This PR tries to update the methods used to fetch full name of a builpack complying with the documentation, and update the relevant error message.

I feel the change is quite small and kind of redundant 😬 , would be happy to rework this if there is something missing.
I avoided adding acceptance tests because that would only be possible by introducing new malformed `package.toml` and `buildpack.toml` file combination, but if we see value in them then I can write those as well.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1384

### DIFF
--- a/pkg/buildpack/builder.go
+++ b/pkg/buildpack/builder.go
@@ -269,15 +269,24 @@ func validateBuildpacks(mainBP Buildpack, depBPs []Buildpack) error {
 		bpd := bp.Descriptor()
 		for _, orderEntry := range bpd.Order {
 			for _, groupEntry := range orderEntry.Group {
-				if _, ok := depsWithRefs[groupEntry.BuildpackInfo.FullName()]; !ok {
+				bpFullName, err := groupEntry.BuildpackInfo.FullNameWithVersion()
+				if err != nil {
+					return errors.Wrapf(
+						err,
+						"buildpack %s must specify a version when referencing buildpack %s",
+						style.Symbol(bpd.Info.FullName()),
+						style.Symbol(bpFullName),
+					)
+				}
+				if _, ok := depsWithRefs[bpFullName]; !ok {
 					return errors.Errorf(
 						"buildpack %s references buildpack %s which is not present",
 						style.Symbol(bpd.Info.FullName()),
-						style.Symbol(groupEntry.FullName()),
+						style.Symbol(bpFullName),
 					)
 				}
 
-				depsWithRefs[groupEntry.BuildpackInfo.FullName()] = append(depsWithRefs[groupEntry.BuildpackInfo.FullName()], bpd.Info)
+				depsWithRefs[bpFullName] = append(depsWithRefs[bpFullName], bpd.Info)
 			}
 		}
 	}
--- a/pkg/buildpack/builder_test.go
+++ b/pkg/buildpack/builder_test.go
@@ -201,6 +201,41 @@ func testPackageBuilder(t *testing.T, when spec.G, it spec.S) {
 								h.AssertError(t, err, "buildpack 'bp.present.id@bp.present.version' references buildpack 'bp.missing.id@bp.missing.version' which is not present")
 							})
 						})
+
+						when("there is a referenced buildpack from dependency buildpack that does not have proper version defined", func() {
+							it("should error", func() {
+								mainBP, err := ifakes.NewFakeBuildpack(dist.BuildpackDescriptor{
+									API: api.MustParse("0.2"),
+									Info: dist.BuildpackInfo{
+										ID:      "bp.1.id",
+										Version: "bp.1.version",
+									},
+									Order: dist.Order{{
+										Group: []dist.BuildpackRef{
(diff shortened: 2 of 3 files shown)


pkg/buildpack/builder.go
```diff
@@ -268,15 +268,14 @@ func validateBuildpacks(mainBP Buildpack, depBPs []Buildpack) error {
...
 			for _, groupEntry := range orderEntry.Group {
-				if _, ok := depsWithRefs[groupEntry.BuildpackInfo.FullName()]; !ok {
+				if _, ok := depsWithRefs[groupEntry.BuildpackInfo.FullNameWithVersion()]; !ok {
 					return errors.Errorf(
-						"buildpack %s references buildpack %s which is not present",
-						style.Symbol(bpd.Info.FullName()),
-						style.Symbol(groupEntry.FullName()),
+						"Invalid buildpack %s defined in order group.",
+						style.Symbol(groupEntry.FullNameWithVersion()),
 					)
```
jromero, line 275: I feel like the new error message is a little too opaque. If the version was missing, all we would see is:

```
Invalid buildpack 'bp.missing.id@' defined in order group.
```


Based on our conversation [here](https://github.com/buildpacks/pack/issues/1384#issuecomment-1116129061), what if we had two errors:

1. the existing error for when a buildpack is genuinely missing
2. a new error for when the referenced buildpack is missing a version:
```
buildpack %s must specify a version when referencing buildpack %s.
```

jromero (changes requested):

pkg/buildpack/builder.go
imnitishng: Updated the error as discussed, added test as well.
Let me know if more work is required.

pkg/buildpack/builder.go
```diff
@@ -268,15 +268,23 @@ func validateBuildpacks(mainBP Buildpack, depBPs []Buildpack) error {
 		bpd := bp.Descriptor()
 		for _, orderEntry := range bpd.Order {
 			for _, groupEntry := range orderEntry.Group {
-				if _, ok := depsWithRefs[groupEntry.BuildpackInfo.FullName()]; !ok {
+				bpFullName, err := groupEntry.BuildpackInfo.FullNameWithVersion()
+				if err != nil {
```
jromero, line 272: It's a little strange that we're not using the `err` being returned from this method.

Could we figure out a way to incorporate the error message from the method into the message we are returning?

This would most likely change the exact verbiage we wanted but it would hopefully lead to a more detailed error message.

imnitishng: > It's a little strange that we're not using the err being returned from this method.

Completely agree, the idea was to get the error message that you [suggested](https://github.com/buildpacks/pack/pull/1437#discussion_r868451760).
I'll update this and find a way to use the error as well as keep the error message relevant.

imnitishng:
Updated error 
`ERROR: saving image: buildpack 'samples/hello-universe@0.0.1' must specify a version when referencing buildpack 'samples/hello-world': buildpack samples/hello-world does not have a version defined`
Let me know if any more changes are required, @jromero

jromero:
### Accepted

##### Before

```
pack buildpack package my-buildpack --config /tmp/samples/packages/hello-universe/package.toml
ERROR: saving image: buildpack samples/hello-universe@0.0.1 references buildpack samples/hello-world which is not present
```


##### After

```
./out/pack buildpack package my-buildpack --config /tmp/samples/packages/hello-universe/package.toml
ERROR: saving image: buildpack samples/hello-universe@0.0.1 must specify a version when referencing buildpack samples/hello-world: buildpack samples/hello-world does not have a version defined
```

PR #1439, opened 2022-05-11
bug: Update `getFileFilter` to include parent folders

### LINKED ISSUE #1419: include/exclude directives can lead to directories that are not writable at build time
When building an application using pack and specifying include and/or exclude, the permissions are not set correctly in some cases, and some directories might end up owned by root, which, in some cases, makes the build fail.

This is due to how docker works and can be reproduced by using docker solely.

I will focus on the `include` part of the equation here, but it is possible to trigger the same bug with `exclude`.

# Reproduction using pack

## Setup

In a directory, create a project:

```bash
mkdir -p testdir/testfile

cat << EOF > project.toml
[build]
# This triggers the bug
include = ["testdir/testfile"]

[[build.buildpacks]]
id = "testing/workspace-info"

  [build.buildpacks.script]
  api = "0.7"
  # The build will always fail, but we care about the output here
  inline = "ls -lahR && exit 1"
EOF
```

This technically means that `testdir` is ignored, and will not be copied, however, you want `testdir/testfile`, and thus, something will have to create
this `testdir`, to keep your hierarchy.

## Triggering the bug

Now, running `pack build --builder ${WHATEVER_BUILDER:paketobuildpacks/builder:tiny} test`, you will see the following:

```
===> BUILDING
.:
total 12K
drwsrwsrwt 3 cnb  cnb  4.0K Jan  1  1970 .
drwxr-xr-x 1 root root 4.0K Apr 12 21:16 ..
drwxr-sr-x 2 root root 4.0K Apr 12 21:16 testdir

./testdir:
total 8.0K
drwxr-sr-x 2 root root 4.0K Apr 12 21:16 .
drwsrwsrwt 3 cnb  cnb  4.0K Jan  1  1970 ..
-rw-r--r-- 1 cnb  cnb     0 Apr 12 21:16 testfile
```

## What did I expect?

**testdir** should be owned by `cnb:cnb`

# Triggering using Docker

This is not solely a problem with pack, it is rather how `docker cp` works.

Again, using an empty directory, we can do:

```bash
# Create the base setup
mkdir -p testdir
touch testdir/testfile

# Create a tar archive, including only testfile. This is the same as what pack does under the hood, when setting up the workspace
tar -czvf example.tar.gz testdir/testfile

# The archive will not contain `testdir`
tar tvf example.tar.gz

# Now let's copy it into docker, using docker cp
# Start a busybox instance in the background
docker run --rm -d --name test busybox sleep INFINITY

# And docker cp it
cat example.tar.gz | docker cp - test:/root

# We can see the same behavior as before
docker exec -it test ls -lahR /root/

# Finally, cleanup the leftovers
docker stop test
```

# Alternatives

In order not to hit this problem, this is what needs to be done:

```
include = ["testdir", "!testdir/*", "testdir/testfile"]
```

This is error prone, hard to debug when it goes wrong, and ultimately not intuitive for users.

# Fixing it

I am happy to provide a fix if we agree that this is a bug. However, I am not sure what is the best way of achieving this so I would need a bit of guidance.

My current idea would be to modify [WriteDirToTar](https://github.com/buildpacks/pack/blob/main/pkg/archive/archive.go#L167), and the corresponding entry for [zip](https://github.com/buildpacks/pack/blob/main/pkg/archive/archive.go#L252), to walk the path a first time, gather a list of everything needing copying, and then go over it a second time, inserting missing directories as needed.

Does that sound sound? A better way I would be missing?

Thanks!

imnitishng:
## Summary

- The `include` directive when specified with a file inside a
  folder `testdir/testfile` did not pass the parent folder to
  the builder image, which led to the folder being created by
  docker instead pack causing the permissions of the folder
  to be `cnb`.
- This PR fixes the include logic to get parent folders and
  add them to the file filter, so they get created via pack.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1419


imnitishng:
The PR focuses on `include` part of the issue for now.
Will add more unit/acceptance tests here.

imnitishng:
Hey @jromero,
I have tried to fix the `include` part of the issue based on https://github.com/buildpacks/pack/issues/1419#issuecomment-1121562242
The idea is to add parent folder entries to the `fileFilter` so the folders are included in the final build.
The approach works fine for cases when we know the directory name (for ex: `testdir/testfile.txt`, `testdir/abc/name.txt`, etc) but it would not work when we do not know the folder name as in Regex (for ex: `abc/*/name.txt`, `test/**/file.txt`, this would cause `abc/def/` to have `root` owner, since we did not add `["!abc/def/*", "abc/def/name.txt"]` to the filter).
I don't see a way where we could tweak the `getFileFilter` to include such cases, I'd like to know if there's a way I'm missing.
Moreover if we DO want to cover these cases then I guess we will have to modify [WriteDirToTar](https://github.com/buildpacks/pack/blob/main/pkg/archive/archive.go#L167) to achieve this, but I'm not sure if the added complexity would be worth.
For now I have added a check that reverts back to old logic on encountering a regex containing either of `*`.
The compatibility tests also fail due to the changes.

PR #1280, opened 2021-09-09
Embed Source Metadata in App

Haimantika:
## Summary
Provide project source metadata from local git information (#1137)

## Documentation

- Should this change be documented?
    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1137

PR #1444, opened 2022-05-17
Add support to embed source metadata from git

### LINKED ISSUE #1137: Provide project source metadata from local git information
### Description
As a buildpacks user, I'd like to have the source and version information of my project embedded as metadata in the final app image.

### Proposed solution

Based on the spec, we should be able to provide `<layers>/project-metadata.toml` in [this format](https://github.com/buildpacks/spec/blob/platform/0.6/platform.md#project-metadatatoml-toml) which would be translated into the [`io.buildpacks.project.metadata` label](https://github.com/buildpacks/spec/blob/platform/0.6/platform.md#iobuildpacksprojectmetadata-json) on the image.

This issue should be scoped to local git information. 

##### General case

```gherkin
Given I clone a project from git@github.com:buildpacks/pack.git
And `git rev-parse HEAD` returns "8348484564e6aa0699de0ee78c258e88240eb0b5"
And `git describe --tags --always` returns "v0.18.1-2-g83484845"
And `git branch --show-current` returns "main"
And `git remote -v | grep "fetch"` returns:
    "origin	git@github.com:buildpacks/pack.git (fetch)"
And I run `pack build my-app`
When I run the following command `docker inspect my-app | jq -r '.[0].Config.Labels["io.buildpacks.project.metadata"] | fromjson'`
Then I should see:
{
  "type": "git",
  "version": { "commit": "8348484564e6aa0699de0ee78c258e88240eb0b5", "describe": "v0.18.1-2-g83484845"},
  "metadata": {
    "refs": ["main"],
    "url": "git@github.com:buildpacks/pack.git"
  }
}
```

##### At a tag(s)

```gherkin
Given I clone a project from git@github.com:buildpacks/pack.git
And `git rev-parse HEAD` returns "8348484564e6aa0699de0ee78c258e88240eb0b5"
And `git describe --tags --always` returns "v0.18.1"
And `git branch --show-current` returns "main"
And `git tag --points-at HEAD` returns:
  v0.18.1
  v0.18.1-rc2
And `git remote -v | grep "fetch"` returns:
    "origin	git@github.com:buildpacks/pack.git (fetch)"
And I run `pack build my-app`
When I run the following command `docker inspect my-app | jq -r '.[0].Config.Labels["io.buildpacks.project.metadata"] | fromjson'`
Then I should see:
{
  "type": "git",
  "version": { "commit": "8348484564e6aa0699de0ee78c258e88240eb0b5", "describe": "v0.18.1"},
  "metadata": {
    "refs": ["main", "v0.18.1", "v0.18.1-rc2"],
    "url": "git@github.com:buildpacks/pack.git"
  }
}
```

### Additional context

This issue is part of an epic of work. The related issues may be see by filtering on [epic/source-metadata](https://github.com/buildpacks/pack/labels/epic%2Fsource-metadata).

imnitishng:
## Summary

Add support to add source metadata from git, to complete https://github.com/buildpacks/pack/pull/1280

- [x] Support current branch refs
- [x] Support current tag refs
- [x] Support `git describe --tags --always`
- [x] Support `git remote`
- [x] Add unit tests
- [ ] Add acceptance tests
- [x] Clean and organize code with proper error handling

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1137


pkg/client/metadata.go
```diff
@@ -0,0 +1,135 @@
...
+type TagInfo struct {
+	Name    string
+	Message string
+	Type    string
+	TagHash string
+	TagTime time.Time
+}
+
+func GitMetadata(appPath string) *platform.ProjectSource {
+	// appPath2 := "/Users/nitish/OSS/pack"
```
dfreilich, line 22: nit: remove

pkg/client/metadata.go
```diff
@@ -0,0 +1,135 @@
...
+	repo, err := git.PlainOpen(appPath)
+	if err != nil {
+		print("unable to open git repo")
```
dfreilich, line 25: Ultimately, I would remove the `print` statements here, and just return `nil` if there is an issue in these cases

pkg/client/metadata.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	headRef, err := repo.Head()
+	if err != nil {
+		print("unable to parse head")
```
dfreilich, line 29: nit: same as above

pkg/client/metadata.go
```diff
@@ -0,0 +1,135 @@
...
+
+	projectSource := &platform.ProjectSource{
+		Type: "git",
+		Version: map[string]interface{}{
+			"commit":   headRef.Hash().String(),
+			"describe": describe,
+		},
+		Metadata: map[string]interface{}{
+			"refs": refs,
+			"url":  "nitis",
```
dfreilich, line 44: nit: remove this

dfreilich (commented):
This looks like a really great start! 

Our `pkg/client` directory is already pretty full, and we seem to typically put our direct commands there. I would instead suggest putting the `metadata.go` file (and accompanying test) in the [pkg/project/v2](https://github.com/buildpacks/pack/tree/main/pkg/project/v02) directory). What do you think, @jromero ?

imnitishng:
There is a couple of things on git vs our implementation of the commands that I wanted to point out here 
-  [`git describe --always`](https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---always) returns the short hash of a commit, whose length could vary based on the number of commits (to be unique) in the repository. So for now I went with returning the full commit hash.
    - In the same way full hash is used to generate the tag in the case if the tag exists in a different branch from `HEAD`, [more details on how this tag is generated by git](https://git-scm.com/docs/git-describe#_examples). So git would output `v1.0.4-14-g2414721` but this implementation would output `v1.0.4-14-g2414721234l23wer323sas58sd8k`.
        - The prefix `-g` here denotes that the tag was made using [`git` SCM](https://git-scm.com/docs/git-describe#_examples), since we are only dealing with git repositories here for now, could we omit this? 
- In case of multiple tags `git describe --tags --always` priortizes annotated tags over unannotated and if there are multiple tags of the same type then they are sorted based on creation date for annotated tags and alphabetical order for unannotated tags, this behaviour was observed by me during testing, hence I have implemented it in the logic.

I am not too sure how to write acceptance tests for this, (are they even required or not?) but if we decide on writing it then we'll probably have to create some dummy files and folders that act as our git repository to test on. Let me know if there is another approach available.

imnitishng:
I might also need some pointers on the failing test cases, not sure what is happening and it's also taking a long time for it to complete.

jromero:
> I am not too sure how to write acceptance tests for this, (are they even required or not?) but if we decide on writing it then we'll probably have to create some dummy files and folders that act as our git repository to test on. Let me know if there is another approach available.

I'm going to merge this as is given the code coverage at the unit level. I'm working on refactoring the acceptance test suite to be easier to work with. As part of that work, I'll see if we can test this set of functionality as well.

jromero (approved):

Issue #1412, opened 2022-04-06
xenial delivery workflow is failing


jromero:
### Description
Ubuntu 16.04 (xenial) runners have been removed making our delivery workflows for [linux fail](https://github.com/buildpacks/pack/actions/runs/2082877271).

> This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.

### Proposed solution
Remove xenial support.

Alternatively, setup our own instance on [Equinix metal](https://metal.equinix.com/developers/docs/operating-systems/supported/#ubuntu).

jromero:
@buildpacks/platform-maintainers what are our thoughts on continuing to support Ubuntu 16.04 releases?

dfreilich:
The question is what our support requirements are, and how much we want to go past them. It's hard to say, if we don't have metrics on usage, and don't know what will happen. 

I would rather not set up our own instance on Equinix - we already have trouble with the Windows runners, and adding a standalone one just for packaging seems suspect. 

If those are the only two options, then I would vote for removing xenial support, unless we get pushback from the community (according to https://wiki.ubuntu.com/Releases, while it isn't EOL, it is at the end of standard support, and as an open source community, I don't feel like we are obligated to provide full support until EOL unless our community shows that it's invested in helping us do that). 

The question is whether there's another option. It is `go`, so we definitely shouldn't need to build it on `xenial` to create a compatible binary; do we need a xenial machine to push the ppa? Can we do it otherwise? And overall, I wonder whether we should consider moving to a tool like goreleaser to manage our pretty complicated release process/matrix.

jromero:
> The question is whether there's another option.

I did bring this up on another [similar request](https://github.com/buildpacks/pack/issues/1400#issuecomment-1090326214).

> maybe there's a way to improve our workflow to better support many more possible releases.

Maybe we can use docker images to produce and publish the PPAs. We already do this for [_testing_ our PPA installation](https://github.com/buildpacks/pack/blob/8921b1b1b91be49a4d6141f2b12ff016820632ba/.github/workflows/deliver-ubuntu.yml#L98-L120) so I can't foresee any challenges.

> do we need a xenial machine to push the ppa?

IIRC the answer is yes for our current method.

> I wonder whether we should consider moving to a tool like goreleaser.

Don't have a lot of knowledge there but if it's possible, I would definitely be in favor of it.

jromero:
Unfortunately, it seems like we won't be able to get the xenial delivery pipeline back up for this release.

[...]

PR #1456, opened 2022-06-14
Delivery ubuntu improvements

### LINKED ISSUE #1455: Add Ubuntu 22.02 LTS (jammy) to delivery pipeline
@jromero Ubuntu 22.02 LTS (jammy) has been available for a month, it would be nice to see pack support the current LTS :+1:

_Originally posted by @flexiondotorg in https://github.com/buildpacks/pack/issues/1400#issuecomment-1138432246_

jromero:
## Summary

* #1455
* Update minimum go version to 1.16 due to:
    * go fails to build with `io/fs: malformed module path "io/fs": missing dot in first path element` which is imported from `BurntSushi/toml`. This issue needs to be investigated further since it should be gated via [build tags](https://github.com/BurntSushi/toml/blob/7eb955fc8b927eea24fa3db743d1954adff04db3/decode_go116.go#L1-L2). (this [issue](https://github.com/BurntSushi/toml/issues/340) may be related)
    * Will continue to investigate as part of https://github.com/buildpacks/pack/issues/1412 
* Use docker images for PPA creation.
  * This would remove our dependancy on GitHub runners (discussed [here](https://github.com/buildpacks/pack/issues/1412#issuecomment-1104226306)).

Resolves #1455


.github/workflows/delivery-ubuntu.yml
```diff
@@ -9,35 +9,33 @@ on:
...
           include:
-            - target: xenial
-              runner: ubuntu-16.04
-              image: docker://ubuntu:16.04
-              i386-go-version: 1.13
             - target: bionic
-              runner: ubuntu-18.04
-              image: docker://ubuntu:18.04
-              i386-go-version: 1.15
+              go-dep-entry: golang (>=1.16)
```
dfreilich, line 20: It seems like this is the same all 3 times, why do we need this to be part of the matrix?

jromero: You are right in that they are the same at the moment. There are two reasons why I left it.

1. If we want to fix xenial, it would be a different entry (something like golang-1.13).
2. As OS versions get older, the PPA packages may move from golang to golang-{{version}} based on the PPA maintainers discretion.

Happy to remove it from the matrix until it's necessary if you think it causes confusion.

jromero: Simplified the pipeline which simplifies this as well. Please re-review.

.github/workflows/delivery-ubuntu.yml
```diff
@@ -0,0 +1,101 @@
...
+              GO_DEP_LIB_PATH: /usr/lib/go
+              HOMEPAGE: "https://buildpacks.io"
+              PACK_VERSION: ${{ steps.version.outputs.result }}
+              REPO: "buildpacks/pack"
+              UBUNTU_VERSION:  ${{ matrix.target }}
+
+            ###
+            # NOTE: 'uses' does not suport interpolation so we have to manually define the
+            # following steps per variant.
+            ###
```
dfreilich, line 68: Great point, thanks for clarifying!

dfreilich (approved):

Issue #1468, opened 2022-06-23
`pack builder create` should support extensions


natalieparellano:
* builder.toml can contain `[[extensions]]`
  * referenced extensions should be present in the builder image at `/cnb/extensions/<extension-id>/<extension-version>`
* builder.toml can contain `[[order-extensions]]`

PR #1478, opened 2022-07-05
`pack builder create` should support extensions

natalieparellano:
## Summary
If a builder.toml contains `[extensions]` and `[order-extensions]`:
* See https://github.com/buildpacks/spec/pull/316 for detailed requirements

## Documentation

    - [x] Yes, see https://github.com/buildpacks/docs/issues/501
    - [ ] No

## Related

Resolves https://github.com/buildpacks/pack/issues/1468


pkg/buildpack/buildpack.go
```diff
@@ -17,73 +17,110 @@ import (
...
-type buildpack struct {
-	descriptor dist.BuildpackDescriptor
+type Descriptor interface {
+	EnsureStackSupport(stackID string, providedMixins []string, validateRunStageMixins bool) error
+	EscapedID() string
+	Kind() string
+	ModuleAPI() *api.Version
+	ModuleInfo() dist.BuildpackInfo
+	ModuleOrder() dist.Order
+	ModuleStacks() []dist.Stack
```
jromero, line 37: Are we able to remove the `Module` prefix to the getter methods? They fill a tad awkward.

jromero (commented):
Initial feedback:

I like the term module overall for the generic shared concept. There are still a few references where “buildpacks” are used when talking about extensions though. I think they need to be fully resolved.

ie.
- BuildpackLayers
- `/pkg/buildpack` package

pkg/buildpack/buildpack.go
natalieparellano: Updated in bd37213a

natalieparellano:
Still to be done:
* fix use of deprecated strings.Title
* test filling in extension version from the provided order for extensions
* Ensure `optional = true` isn’t redundantly printed in the order for extensions

natalieparellano:
I'm going to mark this as ready for review as I am pretty confident that once the WCOW worker comes back online, Codecov will be happy.

I'm really sorry about all the changed files and (due to package rename) undecipherable diffs. I'm happy to pair on the code review if that would be helpful.

jromero:
@natalieparellano I've updated the description to define blockers. We'll want the specs released before we can merge this PR. Let me know if that's a concern.

jromero:
As discussed in WG today, we can work around needing to wait for the spec releases if we were to:

1. gate this functionality behind "experimental"
2. remove the statement that platform API 0.10 is supported

jromero (dismissed):
@natalieparellano correct me if I'm wrong, but we're still waiting on this to be updated based on the previous comment.

natalieparellano:
@jromero you are correct - apologies for letting this one go dormant. I'll circle back to it soon...

natalieparellano:
@jromero I made the changes requested. Assuming this passes CI could you PTAL? 🙏🏼

natalieparellano:
It seems that something about the changes make layer diff IDs non-deterministic on Windows. At least, asserting on the contents of `io.buildpacks.extension.layers` is failing, and [this action run](https://github.com/buildpacks/pack/actions/runs/3085654321) shows that layer diff IDs can be [predicted](https://github.com/buildpacks/pack/pull/1516/files) with the code on `main`. I'm a bit surprised that no other tests are failing... will need to dig into it more.

natalieparellano:
Also need to fix compat suite

.github/workflows/build.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           if (Test-Path C:\ProgramData\docker\config\daemon.json) {
             $config=(Get-Content C:\ProgramData\docker\config\daemon.json | ConvertFrom-json)
           }
-          $config."insecure-registries" = @("$IPAddress/32")
+          $config | Add-Member -Force -Name "insecure-registries" -value @("$IPAddress/32") -MemberType NoteProperty
```
natalieparellano, line 107: Fixes unrelated issue; the existing Powershell was too brittle.

pkg/buildpack/buildpack.go
```diff
@@ -100,31 +137,36 @@ func (b *distBlob) Open() (io.ReadCloser, error) {
 	return b.openFn(), nil
 }
 
-func toDistTar(tw archive.TarWriter, bpd dist.BuildpackDescriptor, blob Blob) error {
+func toDistTar(tw archive.TarWriter, descriptor Descriptor, blob Blob) error {
 	ts := archive.NormalizedDateTime
 
+	parentDir := dist.BuildpacksDir
+	if descriptor.Kind() == "extension" {
```
jkutner, line 144: Should we make `"extension"` a const somewhere? I see it used a few times

jkutner (approved):

pkg/buildpack/buildpack.go
natalieparellano: Updated in [a257564](https://github.com/buildpacks/pack/pull/1478/commits/a257564cf8b45a62ca93ce90e011025d9c4c05c8)

jjbustamante (approved):
It looks good to me

PR #1479, opened 2022-07-06
Fix project.toml build env to match spec

jkutner:
## Summary

Change the parsing of `project.toml`'s `io.buildpack.build.env` to match the [project descriptor spec](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksbuildenv-optional)

https://cloud-native.slack.com/archives/C033DV9EBDF/p1657116095577129

#### Before

```toml
[[io.buildpacks.env.build]]
name = "JAVA_OPTS"
value = "-Xmx300m"
```

#### After

```toml
[[io.buildpacks.build.env]]
name = "JAVA_OPTS"
value = "-Xmx300m"
```

## Documentation

    - [ ] Yes, see #___
    - [x] No


jromero:
Although I agree with this change, this may be considered a breaking change. I'd want to take a deeper look into the reasoning for the change, it's potential impact and mitigation strategies.

samj1912:
As discussed in the WG we cannot make this change directly since this is an end-user breaking change and the current implementation is mimicked by other platforms like kpack as well. If we do change it, we should support both the current and the future version of this key.

jromero:
@jkutner Can you update this PR with backward compatibility support?

I would envision that we would warn about the misuse of the invalid namespace and that the support may be removed at any point in the future.

@samj1912 was there a timeframe for support of the invalid key?

samj1912:
@jromero I would prefer if we soften that warning. Since this is an end user change that we inadvertently implemented in both pack and kpack I would like us to keep this key around at least until project descriptor 0.3.

jkutner:
@jromero @samj1912 i've updated this to include backwards compat

samj1912 (approved):

jkutner:
@jromero can you give this a look-see?

jromero (approved):
LGTM!

PR #1493, opened 2022-07-28
CI: Fix setup-go on windows-lcow runner

jromero:
This PR temporarily addresses an issue in the setup-go action that prevents downloading new go versions on Windows. 

These changes can be reverted after https://github.com/actions/setup-go/issues/241 is resolved.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,9 @@ jobs:
           }
         shell: powershell
       - name: Set up go
-        uses: actions/setup-go@v3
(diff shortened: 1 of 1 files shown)


jromero:
Closing this PR since just by the simple fact of putting up this PR with the patched version downloaded the necessary go version on the current custom runner. It is now being pulled from cache on the `main` pipelines. 

This PR provided a workaround that resolved the current issue. The hope is that https://github.com/actions/setup-go/pull/250 will be merged and released by the time either a) we use a different version or b) we rebuild our custom runner.

PR #1497, opened 2022-08-10
cache: Add support for bind cache

### LINKED ISSUE #1473: [cache-flag] Support build bind cache
# Summary

Add support for bind-based build cache to `--cache` flag.

```bash
pack build --cache 'type=build;format=bind;source=<path-to-dir>'
```

#### Relative path

```gherkin
Given the current working directory contains source code
And configured builder can build source code
And I am using the Docker Desktop
When I execute 'pack build myorg/myapp:latest --cache "type=build;format=bind;name=./build-cache"'
And build finishes successfully
Then I should now have a directory in my current working directory named 'build-cache'
And directory 'build-cache' should contain the cached build layers
```

#### Absolute path

```gherkin
Given the current working directory contains source code
And configured builder can build source code
And I am using the Docker Desktop
When I execute 'pack build myorg/myapp:latest --cache "type=build;format=bind;name=/tmp/build-cache"'
And build finishes successfully
Then I should now have a directory named 'build-cache' in '/tmp'
And directory '/tmp/build-cache' should contain the cached build layers
```

### LINKED ISSUE #1474: [cache-flag] Support launch bind cache
# Summary

Add support for bind-based launch cache to `--cache` flag.

```bash
pack build --cache 'type=launch;format=bind;source=<path-to-dir>'
```

#### Relative path

```gherkin
Given the current working directory contains source code
And configured builder can build source code
And I am using the Docker Desktop
When I execute 'pack build myorg/myapp:latest --cache "type=launch;format=bind;name=./build-cache"'
And build finishes successfully
Then I should now have a directory in my current working directory named 'launch-cache'
And directory 'launch-cache' should contain the cached launch layers
```

#### Absolute path

```gherkin
Given the current working directory contains source code
And configured builder can build source code
And I am using the Docker Desktop
When I execute 'pack build myorg/myapp:latest --cache "type=launch;format=bind;name=/tmp/launch-cache"'
And build finishes successfully
Then I should now have a directory named 'launch-cache' in '/tmp'
And directory '/tmp/launch-cache' should contain the cached launch layers
```

imnitishng:
## Summary
Add bind cache support to pack

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1474
Resolves #1473


internal/build/lifecycle_execution.go
```diff
@@ -135,11 +135,16 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
...
+		case cache.CacheVolume:
+			buildCache = cache.NewVolumeCache(l.opts.Image, l.opts.Cache.Build, "build", l.docker)
+		case cache.CacheBind:
+			buildCache = cache.NewBindCache(l.opts.Cache.Build, l.docker)
+		}
 	}
 
 	l.logger.Debugf("Using build cache volume %s", style.Symbol(buildCache.Name()))
-	if l.opts.ClearCache {
+	if l.opts.ClearCache && l.opts.Cache.Build.Format != cache.CacheBind {
```
imnitishng, line 147: I have not allowed clear cache support for bind caches due to security reasons, let me know if this needs to be changed.

jromero: If the user provides a very specific directory to use as a cache, why would we want to make an exception in this case but not in a named volume? I think the unexpected difference in behavior is more problematic than deleting files within a targeted directory.

internal/build/lifecycle_execution.go
```diff
@@ -135,11 +135,16 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
...
-		buildCache = cache.NewVolumeCache(l.opts.Image, l.opts.Cache.Build, "build", l.docker)
+		switch l.opts.Cache.Build.Format {
+		case cache.CacheVolume:
+			buildCache = cache.NewVolumeCache(l.opts.Image, l.opts.Cache.Build, "build", l.docker)
+		case cache.CacheBind:
+			buildCache = cache.NewBindCache(l.opts.Cache.Build, l.docker)
+		}
 	}
 
 	l.logger.Debugf("Using build cache volume %s", style.Symbol(buildCache.Name()))
```
jromero, line 146: Could we update this log message to say `volume` or `directory` depending on bind or volume?

jromero (commented):
Overall this PR is 💯. I'd like to discuss the behavior of ClearCache.

Maybe @dfreilich has some opinions here as well.

internal/build/lifecycle_execution.go
imnitishng: > If the user provides a very specific directory to use as a cache, why would we want to make an exception in this case but not in a named volume?

I feel like if user does something like `--cache="type=build;format=bind;name=."` alongwith `--clear-cache`, then we'd be cleaning up the current dir. How do we tackle this?

jromero: That's an interesting callout. I would agree that it _could_ be dangerous. What if we completely restricted the combination of options instead of bypassing the functionality? IOW, what do you think about failing early if the user provides any `bind` format with `--clear-cache`.

We can provide a helpful message like "'--clear-cache' is not supported when using 'bind' cache. You may instead delete the cache directory manually.".

The only drawback is that it would prevent --clear-cache when using a mix of any other cache format.

---

Alternatively, what if we had a nested directory with an explicit name for bind? That way at least it's less likely that we could delete unexpected content. 

```
.
└── <user-defined>/
    └── <type>-cache/
```

For example, when the user provides:
`--cache="type=build;format=bind;name=./my-cache/"`

we create the parent directory (if missing), but additionally create a child directory `<type>-cache` (ie. `build-cache`) such as:

```
.
└── my-cache/
    └── build-cache/
```

then when the users provides the same flag `--cache="type=build;format=bind;name=./my-cache/"` plus `--clear-cache` we only delete the `my-cache/build-cache` directory.

WDYT?

jromero (approved):
@imnitishng, It looks like I made a mistake in the original issues I created for this feature. As per the RFC, the "name" option of the "bind" format should have been "source". I went ahead and made the changes along with other minor polishing items. You can review the changes in [this commit](https://github.com/buildpacks/pack/pull/1497/commits/b80b0c43a989aad7980c4b3ebd1523d3ed5bd7c0).

Overall, 💯 implementation. Thank you!

imnitishng:
Hey I just noticed the same, thanks for the cleanup!

PR #1510, opened 2022-09-10
chore: change lib handling npipes

matejvasek:
## Summary

Use Microsoft's library to handle npipe.
Replace `gopkg.in/natefinch/npipe.v2` => `github.com/Microsoft/go-winio`.


jromero (approved):
Nice! LGTM!

PR #1527, opened 2022-10-12
Support for `pack build` with extensions (platform 0.10)

### LINKED ISSUE #1469: pack should support platform 0.10
Changes needed:

* Declare platform API 0.10 as supported
* Set `CNB_EXPERIMENTAL_MODE=warn` in the lifecycle's execution environment
* Invoke the restorer with the -build-image flag and cache volume mounted at /kaniko (when there are extensions)
* Invoke the extender binary with the builder image digest reference as the first argument and cache volume mounted at /kaniko (when there are extensions)
* Invoke the exporter as usual
* `pack inspect <image with 0.9 buildpacks>` should succeed (with existing codebase it shows `ERROR: unable to find image 'hello-extensions' locally or remotely`)

Out of scope (need separate issues):
* The `--extension` equivalent of `--buildpack`
* Support for extensions in project.toml
* `pack build` with extensions and `--cache-image` (adds complexity as it requires managing both a cache image and a cache volume; kaniko always needs a volume)
* `pack build` with extensions and a local (only) builder (because kaniko needs a manifest for the builder image, the builder must be pushed to a registry; would require a re-implementation of the extender in pack that uses the docker daemon instead of kaniko, or an `extender` lifecycle binary that can use the daemon socket)

natalieparellano:
When extensions-phase-1 is merged, we can repoint the branch to main.

## Summary
`pack build` should support extensions

## Output
See docs PR which uses this branch: https://github.com/buildpacks/docs/pull/526

## Documentation

    - [x] Yes, see https://github.com/buildpacks/docs/issues/501
    - [ ] No

## Related

Resolves #1469

### DIFF
--- a/acceptance/testdata/pack_fixtures/report_output.txt
+++ b/acceptance/testdata/pack_fixtures/report_output.txt
@@ -4,7 +4,7 @@ Pack:
 
 Default Lifecycle Version:  0.14.1
 
-Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9
+Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.10
 
 Config:
   default-builder-image = "{{ .DefaultBuilder }}"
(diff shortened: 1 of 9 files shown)


natalieparellano:
Superseded by https://github.com/buildpacks/pack/pull/1547

PR #1522, opened 2022-10-02
Try upgrading Powershell cmd

dfreilich:
Signed-off-by: David Freilich <david.freilich@appsflyer.com>

## Documentation

- Should this change be documented?
    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___

PR #1539, opened 2022-11-02
Refactor lifecycle execution and tests

natalieparellano:
## Summary
- Don't pass anything in `l.opts` as function arguments (makes things simpler)
- Remove duplication in tests by moving configurable inputs to top-level vars

Hopefully this should make this part of the code base a bit easier to understand and make changes to.

## Documentation

    - [ ] Yes, see #___
    - [x] No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,7 +104,7 @@ jobs:
           if (Test-Path C:\ProgramData\docker\config\daemon.json) {
             $config=(Get-Content C:\ProgramData\docker\config\daemon.json | ConvertFrom-json)
           }
-          $config."insecure-registries" = @("$IPAddress/32")
+          $config | Add-Member -Force -Name "insecure-registries" -value @("$IPAddress/32") -MemberType NoteProperty
           ConvertTo-json $config | Out-File -Encoding ASCII C:\ProgramData\docker\config\daemon.json
 
           Restart-Service docker
--- a/internal/build/lifecycle_execution_test.go
+++ b/internal/build/lifecycle_execution_test.go
@@ -1143,6 +1143,31 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 					)
 					h.AssertSliceContains(t, configProvider.HostConfig().Binds, expectedBinds...)
 				})
+
+				when("override GID", func() {
+					when("override GID is provided", func() {
+						lifecycleOps = append(lifecycleOps, func(options *build.LifecycleOptions) {
+							options.GID = 2
+						})
+
+						it("configures the phase with the expected arguments", func() {
+							h.AssertIncludeAllExpectedPatterns(t,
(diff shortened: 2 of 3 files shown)


natalieparellano:
The diff is quite large 😞 but it should be pretty viewable given that it's mostly code deletions

lifecycle_execution_test.go is still crying out for improvement (could be made a table test potentially) but this is a step in the right direction

internal/build/lifecycle_execution_test.go
```diff
@@ -1719,62 +1146,20 @@ func testLifecycleExecution(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 
-		when("override GID", func() {
```
natalieparellano, line 1722: Un-nested one context

natalieparellano:
I think I uncovered a bug as part of this work: 

When `gid` is provided all other flags are overridden: https://github.com/buildpacks/pack/blob/495ef2a4ef38f02996434cd894ef50b2d8dc1ecd/internal/build/lifecycle_execution.go#L409-L411

but I think it makes sense to fix that separately so as to not leave any doubt that no logic was changed

jjbustamante:
> I think I uncovered bugs as part of this work:
> 
> * For analyze phase, when `gid` is provided all other flags are overridden: https://github.com/buildpacks/pack/blob/495ef2a4ef38f02996434cd894ef50b2d8dc1ecd/internal/build/lifecycle_execution.go#L409-L411
> * For analyze phase, `-cache-dir <cache>` is still provided when using a cache image: https://github.com/buildpacks/pack/blob/495ef2a4ef38f02996434cd894ef50b2d8dc1ecd/internal/build/lifecycle_execution.go#L381
>    and https://github.com/buildpacks/pack/blob/495ef2a4ef38f02996434cd894ef50b2d8dc1ecd/internal/build/lifecycle_execution.go#L401
> 
> but I think it makes sense to fix them separately so as to not leave any doubt that no logic was changed

I agree, we can add coverage later

jjbustamante (approved):
I ❤️  this refactoring! Just fix the test

.github/workflows/build.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           if (Test-Path C:\ProgramData\docker\config\daemon.json) {
             $config=(Get-Content C:\ProgramData\docker\config\daemon.json | ConvertFrom-json)
           }
-          $config."insecure-registries" = @("$IPAddress/32")
+          $config | Add-Member -Force -Name "insecure-registries" -value @("$IPAddress/32") -MemberType NoteProperty
```
dfreilich, line 107: Thank you for doing this! I had tried, and then dropped the ball with this: https://github.com/buildpacks/pack/pull/1522

dfreilich (approved):
Great work! I really like the simplification of the call api

dfreilich (approved):

PR #1547, opened 2022-11-08
Add support for Platform 0.10

### LINKED ISSUE #1469: pack should support platform 0.10

natalieparellano:
## Summary
This PR builds off of https://github.com/buildpacks/pack/pull/1478 so will probably be easier to review when that is merged

Supersedes https://github.com/buildpacks/pack/pull/1527 (but keep that one open until we release `pack` with these changes as the docs PR points to https://github.com/buildpacks/pack/pull/1527)

Still a draft because:
* We need to update go.mod and the default lifecycle to 0.15.1 when released
* We need to add unit tests for `pack inspect` for platform API >= 0.10

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1469


CODEOWNERS
```diff
@@ -1,2 +1 @@
-* @buildpacks/platform-maintainers
-* @buildpacks/toc
+* @buildpacks/platform-maintainers @buildpacks/toc
```
natalieparellano, line 1: 😢 I did something wrong

natalieparellano:
This is ready for review but needs to be re-pointed when `extensions-phase-1` is merged as it is based off of that PR.

jjbustamante (approved):
LGTM ❤️

dfreilich (approved):
Looks really good! Overall, I think it could've used a bit more clarification in the code about why there were so many specific `kaniko` changes, but I assume that's covered in RFC.

PR #1601, opened 2023-01-19
add pack extension --help

### LINKED ISSUE #1470: pack should support `pack extension *` commands
`pack extension --help` should show (analogously to `pack buildpack --help`):

```
  inspect     Show information about an extension
  package     Package an extension in OCI format. <- made a separate issue, please see https://github.com/buildpacks/pack/issues/1489
  new         Creates basic scaffolding of an extension.
  pull        Pull an extension from a registry and store it locally
  register    Register an extension to a registry
  yank        Yank an extension from a registry
```

We should probably make a separate issue for each one of these commands, as they could be implemented separately.

Edit: Made a separate issue (https://github.com/buildpacks/pack/issues/1489) for `pack extension package`.

itsdarshankumar:
## Summary
Added command to show output for "pack extension --help"

## Output
image

image

#### Before
No such command exists before

## Related

Resolves #1470

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -76,6 +76,7 @@ func NewPackCommand(logger ConfigurableLogger) (*cobra.Command, error) {
 	rootCmd.AddCommand(commands.Build(logger, cfg, packClient))
 	rootCmd.AddCommand(commands.NewBuilderCommand(logger, cfg, packClient))
 	rootCmd.AddCommand(commands.NewBuildpackCommand(logger, cfg, packClient, buildpackage.NewConfigReader()))
+	rootCmd.AddCommand(commands.NewExtensionCommand(logger, cfg, packClient, buildpackage.NewConfigReader()))
 	rootCmd.AddCommand(commands.NewConfigCommand(logger, cfg, cfgPath, packClient))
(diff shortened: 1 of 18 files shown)


cmd/cmd.go
```diff
@@ -76,6 +76,7 @@ func NewPackCommand(logger ConfigurableLogger) (*cobra.Command, error) {
 	rootCmd.AddCommand(commands.Build(logger, cfg, packClient))
 	rootCmd.AddCommand(commands.NewBuilderCommand(logger, cfg, packClient))
 	rootCmd.AddCommand(commands.NewBuildpackCommand(logger, cfg, packClient, buildpackage.NewConfigReader()))
+	rootCmd.AddCommand(commands.NewExtensionCommand(logger, cfg, packClient, buildpackage.NewConfigReader()))
```
jjbustamante, line 79: I wonder if this line should be move inside the statement `if cfg.Experimental` on line 100. I believe the extensions is actually an experimental feature.

jjbustamante (changes requested):
Great job @itsdarshankumar. 

A part from my doubt with the location of NewExtensionCommand and the fact that I think the PR needs to rebase against the main branch. Every thing looks good to me!

Issue #1091, opened 2021-02-24
Detect the right version of platform api to use 


samj1912:
### Description

With the PR #1088 we allow users to provide a custom lifecycle image name. Previously we used to have pack select the appropriate lifecycle version based on the version it supported and the version used by the builder.

With this change we need to figure out the right platform version which is compatible with the lifecycle, the builder and pack 

See @ekcasey 's comment at https://github.com/buildpacks/pack/pull/1088#issuecomment-785178471

### Proposed solution

pack should pick the highest platform version compatible with the lifecycle image, the builder and pack. 

### Additional context
- [ ] This feature should be documented somewhere

natalieparellano:
The latest buildpacksio/lifecycle has the following metadata:

```
"Labels": {
        "io.buildpacks.builder.metadata": "{\"lifecycle\":{\"version\":\"0.10.2\"},\"api\":{\"buildpack\":\"0.2\",\"platform\":\"0.3\"}}",
        "io.buildpacks.lifecycle.apis": "{\"buildpack\":{\"deprecated\":[],\"supported\":[\"0.2\",\"0.3\",\"0.4\",\"0.5\"]},\"platform\":{\"deprecated\":[],\"supported\":[\"0.3\",\"0.4\",\"0.5\"]}}",
        "io.buildpacks.lifecycle.version": "0.10.2"
      }
```

`io.buildpacks.builder.metadata` is the "legacy label" and `io.buildpacks.lifecycle.apis` is updated with the latest supported apis. However it looks like `pack` doesn't currently know or care about `io.buildpacks.lifecycle.apis`.

jromero:
@natalieparellano It looks like `io.buildpacks.lifecycle.apis` is up and coming from the [builder 0.1 spec](https://github.com/buildpacks/spec/blob/extensions/builder/0.1/extensions/builder.md#iobuildpackslifecycleapis-json) release so we don't technically have it yet. Good call out though.

Until that is released and more public builders move to that spec, we'll need to rely on `io.buildpacks.builder.metadata`

jromero:
@natalieparellano  Is there a contract/spec provided for lifecycle images? I just now realize that you were talking about the lifecycle image which is interesting because AFAIK, that isn't spec'd making it more difficult for us to set expectations on the lifecycle image provided to pack.

natalieparellano:
@jromero apologies for the slow reply. AFAIK lifecycle image isn't spec'd.

jromero:
[...]

PR #1607, opened 2023-01-23
if lifecycle image is supplied, consider its apis

joe-kimmel-vmw:
## Summary
if we're injecting a custom lifecycle image into an untrusted builder, when we select the platform api version we'll consider the intersection between the builder, the lifecycle, and this instance of pack when choosing which version of the api to speak.

because this is a rare case, the default is that there is no lifecycle api list, which is treated as no constraints.

#### Before
technically you could've used a lifecycle that didn't support the highest common version supported by both pack and the builder, but in practice i think this never happened.

#### After
Now it really can't ever happen.

## Documentation

    - [ ] Yes, see #___
    - [ x] No  -- imo this is a very niche case of interest primarily to lifecycle developers, but by all means let me know if i'm misunderstanding the use cases and/or the documentation needs of these valuable members of our community.

## Related

Resolves https://github.com/buildpacks/pack/issues/1091

### DIFF
--- a/internal/build/lifecycle_execution.go
+++ b/internal/build/lifecycle_execution.go
@@ -38,10 +38,10 @@ type LifecycleExecution struct {
 }
 
 func NewLifecycleExecution(logger logging.Logger, docker client.CommonAPIClient, opts LifecycleOptions) (*LifecycleExecution, error) {
-	latestSupportedPlatformAPI, err := findLatestSupported(append(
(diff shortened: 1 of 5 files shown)


natalieparellano (approved):
Looks great, amazing work @joe-kimmel-vmw! Thank you for the contribution 😸

jkutner (approved):

PR #1616, opened 2023-01-31
Extension-Inspect

### LINKED ISSUE #1606: Support `pack extension inspect`
### Description
`pack extension inspect` should provide similar functionality to `pack buildpack inspect`, but for extensions.

itsdarshankumar:
## Summary
Provides Functionality to inspect an extension

## Related

Resolves #1606


internal/commands/inspect_extension.go
```diff
@@ -0,0 +1,112 @@
...
+    {{- else }}
+      {{- range $mixinIndex, $mixin := $stack.Mixins }}
+      {{ $mixin }}
+      {{- end }}
+    {{- end }}
+  {{- else }}
+      (omitted)
+  {{- end }}
+{{- end }}
+
```
natalieparellano, line 32: I think we can remove this - extensions aren't "stack aware" (there is no `[[stacks]]` in extension.toml).

internal/commands/inspect_extension.go
```diff
@@ -0,0 +1,112 @@
...
+Extensions:
+{{ .Extensions }}
+
+Detection Order:
+{{- if ne .Order "" }}
+{{ .Order }}
+{{- else }}
+  (none)
+{{ end }}
```
natalieparellano, line 41: Similar to the above, there is no `[[order]]` in extension.toml, so this could be removed as well

internal/commands/inspect_extension.go
```diff
@@ -0,0 +1,112 @@
...
+      {{- range $mixinIndex, $mixin := $stack.Mixins }}
+      {{ $mixin }}
+      {{- end }}
+    {{- end }}
+  {{- else }}
+      (omitted)
+  {{- end }}
+{{- end }}
+
+Extensions:
```
natalieparellano, line 33:
```suggestion
Extension:
```
(there is only ever one extension)

internal/commands/extension_inspect_test.go
```diff
@@ -0,0 +1,663 @@
...
+				}).Return(complexInfo, nil)
+			})
+
+			it("displays detection order to specified depth", func() {
+				command.SetArgs([]string{"/other/path/to/test/extension", "-d", "2"})
+				assert.Nil(command.Execute())
+
+				assert.AssertTrimmedContains(outBuf.String(), depthExtensionOutputSection)
+			})
+		})
```
natalieparellano, line 578: This can be removed, `--depth` is only useful if there is an order

internal/commands/extension_inspect_test.go
```diff
@@ -0,0 +1,663 @@
...
+			}).Return(simpleInfo, nil)
+		})
+
+		it("displays all mixins", func() {
+			command.SetArgs([]string{"/another/path/to/test/extension", "-v"})
+			assert.Nil(command.Execute())
+
+			assert.AssertTrimmedContains(outBuf.String(), simpleExtensionMixinOutputSection)
+		})
+	})
```
natalieparellano, line 597: This can also be removed, since there are no stacks, there are no mixins - so we don't need `--verbose`

internal/registry/registry_cache.go
```diff
@@ -47,6 +47,11 @@ type Entry struct {
 	Buildpacks []Buildpack `json:"buildpacks"`
 }
 
+// ExtensionEntry is a list of extensions stored in a registry
+type ExtensionEntry struct {
+	Extensions []Buildpack `json:"extensions"`
+}
+
```
natalieparellano, line 54: We might be able to reduce some duplication by renaming `BuildpackEntry.Buildpacks` to `ModuleEntry.Modules` and re-using the same struct

internal/registry/registry_cache.go
```diff
@@ -115,6 +120,47 @@ func (r *Cache) LocateBuildpack(bp string) (Buildpack, error) {
 	return Buildpack{}, fmt.Errorf("no entries for buildpack: %s", bp)
 }
 
+// LocateBuildpack stored in registry
+func (r *Cache) LocateExtension(ex string) (Buildpack, error) {
```
natalieparellano, line 124: If we do the above suggestion we could make a private helper like `locateModule(kind string, ex string)` with different invocations for buildpack and extension (see for example builder.go `addModules`)

internal/registry/registry_cache.go
```diff
@@ -354,3 +400,38 @@ func (r *Cache) readEntry(ns, name string) (Entry, error) {
 
 	return entry, nil
 }
+
+func (r *Cache) readExtensionEntry(ns, name string) (ExtensionEntry, error) {
```
natalieparellano, line 404: If we do https://github.com/buildpacks/pack/pull/1616/files/48d602bc8cbf9db935a77784fb48fe2186c78bd6..8b9ed1f73ea83b9dbe2ad3aedf08cc84fd4a039d#r1094913127 this could be a shared helper across buildpacks and extensions (passing in `kind` as suggested in https://github.com/buildpacks/pack/pull/1616/files/48d602bc8cbf9db935a77784fb48fe2186c78bd6..8b9ed1f73ea83b9dbe2ad3aedf08cc84fd4a039d#r1094914669)

internal/commands/extension_inspect.go
```diff
@@ -24,18 +28,39 @@ func ExtensionInspect(logger logging.Logger, cfg config.Config, client PackClien
...
 			if registry == "" {
-				// fix registry for extension
+				registry = cfg.DefaultRegistryName
 			}
 
 			return extensionInspect(logger, extensionName, registry, flags, cfg, client)
 		}),
 	}
-	// flags will be added here
+	cmd.Flags().IntVarP(&flags.Depth, "depth", "d", -1, "Max depth to display for Detection Order.\nOmission of this flag or values < 0 will display the entire tree.")
```
natalieparellano, line 37: Ah, here is one more that could be removed (since there is no order)

internal/commands/extension_inspect.go
```diff
@@ -24,18 +28,39 @@ func ExtensionInspect(logger logging.Logger, cfg config.Config, client PackClien
...
+	cmd.Flags().StringVarP(&flags.Registry, "registry", "r", "", "buildpack registry that may be searched")
```
natalieparellano, line 38:
```suggestion
	cmd.Flags().StringVarP(&flags.Registry, "registry", "r", "", "extension registry that may be searched")
```
(or maybe just omit it for now, since there is no extension registry :))

internal/commands/extension_inspect.go
```diff
@@ -24,18 +28,39 @@ func ExtensionInspect(logger logging.Logger, cfg config.Config, client PackClien
...
+	logger.Infof("Inspecting extension: %s\n", style.Symbol(extensionName))
+
+	inspectedExtensionsOutput, err := inspectAllExtensions(
+		pack,
+		flags,
+		client.InspectExtensionOptions{
+			ExtensionName: extensionName,
+			Daemon:        true,
+			Registry:      registryName,
+		},
```
natalieparellano, line 54: Does this need to be provided twice?

internal/commands/inspect_extension.go
```diff
@@ -0,0 +1,112 @@
...
+		return "", joinErrors(errArray)
+	}
+	return buf.String(), nil
+}
+
+func inspectExtensionOutput(info *client.ExtensionInfo, prefix string, flags ExtensionInspectFlags) (output []byte, err error) {
+	tpl := template.Must(template.New("inspect-extension").Parse(inspectExtensionTemplate))
+	exOutput, err := buildpacksOutput(info.Extensions)
+	if err != nil {
+		return []byte{}, fmt.Errorf("error writing buildpack output: %q", err)
```
natalieparellano, line 79:
```suggestion
		return []byte{}, fmt.Errorf("error writing extension output: %q", err)
```

internal/commands/inspect_extension.go
```diff
@@ -0,0 +1,112 @@
...
+		}{
+			Location:   prefix,
+			Metadata:   info.ExtensionMetadata,
+			ListMixins: flags.Verbose,
+			Extensions: exOutput,
+			Order:      orderOutput,
+		})
+
+		if err != nil {
+			return []byte{}, fmt.Errorf("error templating buildpack output template: %q", err)
```
natalieparellano, line 102:
```suggestion
			return []byte{}, fmt.Errorf("error templating extension output template: %q", err)
```

natalieparellano:
Hi @itsdarshankumar I left a few more comments - I think when we remove the concepts of "order" and "stacks" from an extension the code will be a lot simpler. Hopefully it's not too much trouble - LMK if you have any questions about it.

internal/commands/extension_inspect.go
itsdarshankumar: Yes, making all the changes, today everything will be done ☺️

internal/commands/extension_inspect.go
itsdarshankumar: Yes I've made the changes locally, didn't push yet as wrapping up all the work at once, previous pushes was just to clean up the git history due to merges

pkg/client/inspect_extension.go
```diff
@@ -0,0 +1,75 @@
...
+	if _, err := dist.GetLabel(img, dist.ExtensionLayersLabel, &layerMd); err != nil {
+		return dist.ModuleLayers{}, fmt.Errorf("unable to get image label %s: %q", dist.BuildpackLayersLabel, err)
+	}
+
+	return layerMd, nil
+}
+
+func extractExtension(layerMd dist.ModuleLayers) dist.ModuleInfo {
+	result := dist.ModuleInfo{}
+	extensionSet := map[*dist.ModuleInfo]bool{}
```
natalieparellano, line 57: I think this function could be made simpler by removing this map. I believe it's used in `extractBuildpacks` so that we can do `sort.Slice` on the result, but since we're always just returning one element we don't need to sort. We could maybe do a check on the length of the `layerMd` provided and if there is more than one, throw an error.

natalieparellano (commented):
Beautiful work @itsdarshankumar - I left one small suggestion for simplifying `extractExtension` but overall this LGTM.

internal/commands/extension_inspect_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+					mockClient.EXPECT().InspectExtension(client.InspectExtensionOptions{
+						ExtensionName: "only-remote-test/extension",
+						Daemon:        false,
+					}).Return(info, nil)
+
+					mockClient.EXPECT().InspectExtension(client.InspectExtensionOptions{
+						ExtensionName: "only-remote-test/extension",
+						Daemon:        true,
+					}).Return(nil, errors.Wrap(image.ErrNotFound, "remote image not found!"))
```
jjbustamante, line 147: should be "local image not found!" ?

jjbustamante (approved):
Awesome @itsdarshankumar! 
I just added one small typo I think

natalieparellano (approved):

natalieparellano:
@buildpacks/toc could you please take a look?

jkutner (approved):

PR #1622, opened 2023-02-06
cleanup: narrow docker interface

matejvasek:
## Summary

Narrow used docker interface by defining interface that is a subset of `CommonAPIClient` and only contains functions that we really need.

### DIFF
--- a/internal/build/container_ops.go
+++ b/internal/build/container_ops.go
@@ -12,7 +12,6 @@ import (
 	"github.com/buildpacks/lifecycle/platform"
 	"github.com/docker/docker/api/types"
 	dcontainer "github.com/docker/docker/api/types/container"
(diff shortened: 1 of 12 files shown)


dfreilich (approved):
Good point! Definitely a best practice, appreciate the help in adding this @matejvasek 

Was this something you had bumped into internally, requiring this fix, or was it just a general code cleanup?

matejvasek:
@dfreilich I use `pack` as a library and I am passing custom implementation of docker client (e.g. custom implementation utilising `podman`, or implementation that logs calls for debugging). 

The whole `CommonAPIClient` is rather big and I end up with implementing load of method that I don't even use.
Most of the code is auto-generated, still why generate code that is not used.

PR #1634, opened 2023-02-15
perf: parallelize builder's module processing

matejvasek:
## Summary

Performance: parallelised some operation for `pack builder create`.
On my machine this bring 4-5x speed up when building paketo base builder.


natalieparellano (approved):
This is a huge improvement... thanks for the PR @matejvasek

jkutner (approved):

PR #1651, opened 2023-03-01
pack build support --extension

### LINKED ISSUE #1551: pack build should support `--extension`
### Description
Today you can run `pack build --buildpack` to dynamically add buildpacks to the builder and modify the order. It should be possible to add extensions in the same way.

### Proposed solution
* `--extension`s should always come first
* There is a separate order for extensions (`order-extensions` in order.toml) so we'll need to take that into account
* ~This should be blocked on https://github.com/buildpacks/pack/issues/1470 (support for `pack extension package`) as `--extension` can accept a package input~ - we can make a separate issue for extension packages

### Additional context

edithwuly:
## Summary

pack build support --extension.

#### Before
Error: unknown flag: --extension

#### After
...
Adding extension samples/tree version 0.0.1 to builder
...
Successfully built image sample-app

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1551


pkg/client/build.go
```diff
@@ -939,7 +953,34 @@ func prependBuildpackToOrder(order dist.Order, bpInfo dist.ModuleInfo) (newOrder
...
+
+	orderExtensions = dist.Order{{Group: []dist.ModuleRef{}}}
+	for _, ex := range declaredExs {
+		locatorType, err := buildpack.GetLocatorType(ex, relativeBaseDir, builderExs)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		switch locatorType {
+		case buildpack.RegistryLocator:
```
natalieparellano, line 968: A note, in case it helps others, `buildpack.FromBuilderLocator` is deprecated now that we have pre- and post- buildpacks. We may want `--pre-extension` and `--post-extension` (like exist for buildpacks) but that should be considered out-of-scope for this issue.

natalieparellano (commented):
@edithwuly awesome work! This is really great to see. The code looks sound, we just need to add some tests in order to verify. You might take a look at the tests for `--buildpack` here: https://github.com/buildpacks/pack/blob/deb588f276c9d355a5813ce589d88b80d9c4a4b6/pkg/client/build_test.go#L684

I don't think all of the tests in this block are valid for extensions, probably the most important ones are:
* https://github.com/buildpacks/pack/blob/deb588f276c9d355a5813ce589d88b80d9c4a4b6/pkg/client/build_test.go#L693
* https://github.com/buildpacks/pack/blob/deb588f276c9d355a5813ce589d88b80d9c4a4b6/pkg/client/build_test.go#L720
* https://github.com/buildpacks/pack/blob/deb588f276c9d355a5813ce589d88b80d9c4a4b6/pkg/client/build_test.go#L1073
This may also benefit from an acceptance test, like we have [here](https://github.com/buildpacks/pack/blob/deb588f276c9d355a5813ce589d88b80d9c4a4b6/acceptance/acceptance_test.go#L1325) - however even experienced contributors will agree that modifying the pack acceptance tests can be painful, so we could definitely help you with that or work to minimize the scope for this PR.

edithwuly:
Thanks @natalieparellano for your review! I'm glad to know my code is satisfying. I'm working on the most important unit tests and will patch this PR as soon as completed.

edithwuly:
Hi @natalieparellano! I try to add two unit tests for `--extension`. Please help me review. Thanks！

But I have some trouble on the `"all extensions are added to ephemeral builder"` test. I think it is used to test whether child extension can be added to builder with meta extension as a whole layer. And it uses the `extractPackaged()` function, in which it seems that `extractExtensions()` function has not been ready. https://github.com/buildpacks/pack/blob/558ae9a3c7e51f1e4e15aaa54bf90e9ca7f164a8/pkg/buildpack/downloader.go#L201

pkg/client/build.go
edithwuly: I have added `case buildpack.FromBuilderLocator` in processExtensions()` funtion. Please help me review, thanks!

internal/commands/build.go
```diff
@@ -218,6 +221,7 @@ func parseTime(providedTime string) (*time.Time, error) {
 func buildCommandFlags(cmd *cobra.Command, buildFlags *BuildFlags, cfg config.Config) {
 	cmd.Flags().StringVarP(&buildFlags.AppPath, "path", "p", "", "Path to app dir or zip-formatted file (defaults to current working directory)")
 	cmd.Flags().StringSliceVarP(&buildFlags.Buildpacks, "buildpack", "b", nil, "Buildpack to use. One of:\n  a buildpack by id and version in the form of '<buildpack>@<version>',\n  path to a buildpack directory (not supported on Windows),\n  path/URL to a buildpack .tar or .tgz file, or\n  a packaged buildpack image name in the form of '<hostname>/<repo>[:<tag>]'"+stringSliceHelp("buildpack"))
+	cmd.Flags().StringSliceVarP(&buildFlags.Extensions, "extension", "", nil, "Extension to use. One of:\n  a extension by id and version in the form of '<extension>@<version>',\n  path to a extension directory (not supported on Windows),\n  path/URL to a extension .tar or .tgz file, or\n  a packaged extension image name in the form of '<hostname>/<repo>[:<tag>]'"+stringSliceHelp("extension"))
```
natalieparellano, line 224:
```suggestion
	cmd.Flags().StringSliceVarP(&buildFlags.Extensions, "extension", "", nil, "Extension to use. One of:\n  an extension by id and version in the form of '<extension>@<version>',\n  path to an extension directory (not supported on Windows),\n  path/URL to an extension .tar or .tgz file, or\n  a packaged extension image name in the form of '<hostname>/<repo>[:<tag>]'"+stringSliceHelp("extension"))
```

pkg/client/build.go
```diff
@@ -135,6 +135,11 @@ type BuildOptions struct {
 	// share both an ID and Version with a buildpack on the builder.
 	Buildpacks []string
 
+	// List of extension images or archives to add to a builder.
+	// These extensions may overwrite those on the builder if they
+	// share both an ID and Version with a extension on the builder.
```
natalieparellano, line 140:
```suggestion
	// share both an ID and Version with an extension on the builder.
```

pkg/client/build.go
```diff
@@ -1063,7 +1077,51 @@ func prependBuildpackToOrder(order dist.Order, bpInfo dist.ModuleInfo) (newOrder
...
+			case len(orderExtensions) > 1:
+				// This should only ever be possible if they are using from=builder twice which we don't allow
+				return nil, nil, errors.New("extensions from builder can only be defined once")
+			default:
+				newOrderExtensions := dist.Order{}
+				groupToAdd := orderExtensions[0].Group
+				for _, bOrderEntry := range builderOrder {
+					newEntry := dist.OrderEntry{Group: append(groupToAdd, bOrderEntry.Group...)}
+					newOrderExtensions = append(newOrderExtensions, newEntry)
+				}
```
natalieparellano, line 1107:
```suggestion
			return nil, nil, errors.New("from builder is not supported for extensions")
```
Apologies, my comment above was a bit unclear. I think we just want to error in this case. 

(We'd also need to delete the three lines below - 1108-1110 currently - but I can't include them in my suggestion because of Codecov's comment)

pkg/client/build.go
```diff
@@ -1081,6 +1139,16 @@ func (c *Client) createEphemeralBuilder(rawBuilderImage imgutil.Image, env map[s
 		bldr.SetOrder(order)
 	}
 
+	for _, ex := range extensions {
+		exInfo := ex.Descriptor().Info()
+		c.logger.Debugf("Adding extension %s version %s to builder", style.Symbol(exInfo.ID), style.Symbol(exInfo.Version))
+		bldr.AddExtension(ex)
+	}
+	if len(orderExtensions) > 0 && len(orderExtensions[0].Group) > 0 {
+		c.logger.Debug("Setting custom order")
```
natalieparellano, line 1148:
```suggestion
		c.logger.Debug("Setting custom order for extensions")
```

pkg/client/build_test.go
```diff
@@ -1708,6 +1710,92 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
...
+
+  [[order-extensions.group]]
+    id = "extension.1.id"
+    version = "extension.1.version"
+`)
+				})
+			})
+		})
+
+		//TODO: "all buildpacks are added to ephemeral builder" test after extractPackaged() is completed.
```
natalieparellano, line 1797: Ah - looking at your comment [here](https://github.com/buildpacks/pack/pull/1651#issuecomment-1465596855), I think we can exclude this one, as there is no such thing as a "meta extension".

natalieparellano:
@edithwuly any update on this one? This one just requires a couple of updates (and would make acceptance tests for https://github.com/buildpacks/pack/pull/1698 a LOT easier 🙃 )

pkg/client/build_test.go
edithwuly: Ohhh, great!

edithwuly:
Yes, sorry about the delay. I have merged your suggestions, removed the rest three lines and merged the buildpacks:main. Thank you very much! @natalieparellano

pacostas:
I tested this PR using the [extension](https://github.com/nodeshift/ubi-nodejs-extension) we've been working on, and it seems to be working properly. We're really excited about getting this PR landed :)

jkutner (approved):

PR #1652, opened 2023-03-01
Removes tee so that we can see if CI reporting failures

### LINKED ISSUE #1653: There are test broken but we didn't notice on CI
### Summary
After merging this [PR](https://github.com/buildpacks/pack/pull/1585) the CI started not reporting when the tests are failing, as a consequence, we've been merging pull requests into the main branch without noticing the tests are broken

### Reproduction

Just run `make test` in your local and you will notice the tests are broken

jjbustamante:
## Summary

This PR just removes the `tee` from the Makefile so the CI can fail when the tests are broken.

As we can see the tests are actually broken but I do not intent to fix them on this PR, I will do it on a different PR

Resolves #1653

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,16 +105,16 @@ unit: GOTESTFLAGS:=$(GOTESTFLAGS) --tags=example
 endif
 unit: out
 	@echo "> Running unit/integration tests..."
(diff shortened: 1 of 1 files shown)


dfreilich (approved):
Good catch @jjbustamante !

PR #96, opened 2019-02-05
Fixes typo in README.md

poy:
fixes #95

PR #1660, opened 2023-03-04
Implement RFC-0096 Remove Stacks & Mixins: phase 1

jkutner:
## Summary

Phase of [RFC #0096](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md) - Remove Stacks & Mixins implementation.

#### Before

Support `[stack]` in `builder.toml, and write `/cnb/stack.toml` in builder.

#### After

Support both `[stack]` and `[run]`/`[build]` tables. If both are present, `[run]`/`[build]` take precedent. Stack ID will still be accepted. 

A `/cnb/stack.toml` will be added to the builder if a Stack ID is provided. A `/cnb/run.toml` may also be added to the builder if `[run]`/`[build]` are definied.

## Documentation

    - [x] Yes, see #___
    - [ ] No

## Related

Resolves #___


internal/builder/builder_test.go
```diff
@@ -240,15 +240,15 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
 				it.Before(func() {
 					h.AssertNil(t, baseImage.SetEnv("CNB_USER_ID", "1234"))
 					h.AssertNil(t, baseImage.SetEnv("CNB_GROUP_ID", "4321"))
 				})
 
-				it("returns an error", func() {
+				it("does not return an error", func() {
 					_, err := builder.New(baseImage, "some/builder")
-					h.AssertError(t, err, "image 'base/image' missing label 'io.buildpacks.stack.id'")
+					h.AssertNilE(t, err)
```
natalieparellano, line 251: Why don't we error here?

internal/build/container_ops.go
```diff
@@ -228,6 +228,37 @@ func WriteStackToml(dstPath string, stack builder.StackMetadata, os string) Cont
...
 
+// WriteRunToml writes a `run.toml` based on the RunConfig provided to the destination path.
+func WriteRunToml(dstPath string, runImages []builder.RunImageMetadata, os string) ContainerOperation {
+	return func(ctrClient DockerClient, ctx context.Context, containerID string, stdout, stderr io.Writer) error {
+		buf := &bytes.Buffer{}
+		err := toml.NewEncoder(buf).Encode(builder.RunImages{
+			Images: runImages,
+		})
+		if err != nil {
+			return errors.Wrap(err, "marshaling stack metadata")
```
natalieparellano, line 239:
```suggestion
			return errors.Wrap(err, "marshaling run metadata")
```

natalieparellano (approved):
Looks good @jkutner!

internal/builder/builder_test.go
natalieparellano: Ah, makes sense - I guess what threw me was the addition of `and run image` to the test description. I thought of a builder that is missing a run image reference.

hone (approved):

dfreilich (approved):

PR #1667, opened 2023-03-07
Disable AUR git publishing because we got revoked

jkutner:
## Summary

Disable the AUR git repo because we were publishing too frequently and got revoked:

https://aur.archlinux.org/packages/pack-cli-git#comment-897395

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___

### DIFF
--- a/.github/workflows/delivery-archlinux-git.yml
+++ b/.github/workflows/delivery-archlinux-git.yml
@@ -1,9 +1,9 @@
 name: delivery / archlinux / git
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


hone (approved):
:(

hone:
We should turn this off until we fix abusing AUR.

dfreilich:
That's crazy! I don't recall seeing any notifications of that, we should make sure that the maintainer email gets any messages from the site

PR #1669, opened 2023-03-09
Remove replacments for docker and buildkit. Upgrade deps

jkutner:
## Summary

It's not clear if we still need these dep replacements. I do know [a lot of the ecosystem is stuck on docker v20.10.21](https://github.com/helm/helm/blob/main/go.mod#L65). I'm not sure how this plays with lifecycle either.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


jkutner:
@natalieparellano no i don't think so. that doesn't remove the `replace`

PR #1690, opened 2023-03-29
fix: Registry cache for Azure DevOps

diznq:
## Summary
Go Git doesn't work with Azure DevOps due to unsupported protocol: https://github.com/go-git/go-git/issues/64

## Output
If `dev.azure.com/` URL is used for Git, `git clone` is used instead of Go Git library, the downside is `git` must be installed locally

### DIFF
--- a/internal/registry/registry_cache.go
+++ b/internal/registry/registry_cache.go
@@ -8,8 +8,10 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/go-git/go-git/v5"
@@ -168,6 +170,7 @@ func (r *Cache) Initialize() error {
 
 // CreateCache creates the cache on the filesystem
 func (r *Cache) CreateCache() error {
+	var repository *git.Repository
 	r.logger.Debugf("Creating registry cache for %s/%s", r.url.Host, r.url.Path)
 
 	registryDir, err := os.MkdirTemp(filepath.Dir(r.Root), "registry")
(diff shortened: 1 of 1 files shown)


internal/registry/registry_cache.go
```diff
@@ -177,11 +180,23 @@ func (r *Cache) CreateCache() error {
 
 	r.RegistryDir = registryDir
 
-	repository, err := git.PlainClone(r.RegistryDir, false, &git.CloneOptions{
-		URL: r.url.String(),
-	})
-	if err != nil {
-		return errors.Wrap(err, "cloning remote registry")
+	if strings.Contains(r.url.String(), "dev.azure.com/") {
```
jkutner, line 183: Could this be instead:
```suggestion
	if r.url.Host == "dev.azure.com" {
```

internal/registry/registry_cache.go
```diff
@@ -177,11 +180,23 @@ func (r *Cache) CreateCache() error {
...
+		err = exec.Command("git", "clone", r.url.String(), r.RegistryDir).Run()
+		if err != nil {
+			return errors.Wrap(err, "cloning remote registry")
```
jkutner, line 186:
```suggestion
			return errors.Wrap(err, "cloning remote registry using native git")
```

jkutner:
I think it's ok to put a dependency on `git`. We just need to make sure we fail in a way that makes it clear. We might need a better error message

internal/registry/registry_cache.go
```diff
@@ -177,11 +180,23 @@ func (r *Cache) CreateCache() error {
...
+		repository, err = git.PlainOpen(r.RegistryDir)
+		if err != nil {
+			return errors.Wrap(err, "opening remote registry clone")
+		}
+	} else {
+		repository, err = git.PlainClone(r.RegistryDir, false, &git.CloneOptions{
+			URL: r.url.String(),
+		})
+		if err != nil {
+			return errors.Wrap(err, "cloning remote registry")
```
cubxxw, line 198: Could such error handling have been more perfect?

cubxxw (approved):

cubxxw:
Strange, did you make mistakes with the rebase, why are there multiple commits?

diznq:
I just signed off the comments raised here but realized strings import wasn't removed and Go is strict about unused stuff

PR #1717, opened 2023-04-08
Bump to latest lifecycle release

AidanDelaney:
Default to 0.16.1


jkutner:
@joe-kimmel-vmw how do we need to coordinate this with your update to lifecycle?

jkutner:
Oh, nevermind, this doesn't update the `go.mod`, just the default version

natalieparellano:
Not exactly superseded by https://github.com/buildpacks/pack/pull/1730, but we should bump to lifecycle 0.17.0 when that is available. This will probably be the last thing to close out the milestone

natalieparellano:
I'm going to put up another PR when there is a lifecycle 0.17.0-rc.2 to bump the default download (and also the library) so that we could include it in the next pre- or rc- version.

natalieparellano:
https://github.com/buildpacks/pack/pull/1826 brings the lifecycle up to 0.17.0-rc.3 but we should leave this issue open in the milestone so that we remember to update to lifecycle 0.17.0 (final released) before cutting 0.30.0

PR #1735, opened 2023-04-25
Wait for non-running state to prevent concurrency

### LINKED ISSUE #1732: Infinite waiting time when building
### Summary
Pack is waiting for the builder container to be created but the container stays in `Created` state.

```
CONTAINER ID   IMAGE                                            COMMAND                  CREATED         STATUS    PORTS     NAMES
10a1cdda7669   pack.local/builder/776e637a696363757074:latest   "/cnb/lifecycle/crea…"   4 seconds ago   Created             hopeful_northcutt
```

### Reproduction
```go
import (
	pack "github.com/buildpacks/pack/pkg/client"
	docker "github.com/docker/docker/client"
        "github.com/buildpacks/pack/pkg/logging"
)
...

dockerClient, err := docker.NewClientWithOpts(docker.FromEnv, docker.WithAPIVersionNegotiation())
if err != nil {
	return nil, fmt.Errorf("could not connect to Docker: %w", err)
}

packClient, err := pack.NewClient(pack.WithDockerClient(dockerClient))
if err != nil {
	return nil, fmt.Errorf("could not create pack client: %w", err)
}

err = packClient.Build(t.Ctx, pack.BuildOptions{
	AppPath:      ".",
	Builder:      "paketobuildpacks/builder:base",
	Image:        "user/image:latest",
	TrustBuilder: func(string) bool { return true },
})
if err != nil {
	return nil, fmt.Errorf("could not build: %w", err)
}
```

##### Current behavior
Pack is waiting indefinitely for `/v1.41/containers/10a1cdda7669382a1ff660a47e4f3858f55754ceda4715fe70add4c2fe5c3ac0/wait?condition=next-exit` because the container stays in a `Created` state.

##### Expected behavior
The container should (I think?) start to continue the build process.
I tested this theory by doing `docker start 10a1cdda7669` which resumed the process well.

---

##### pack info
Commit 1190b6c521288f7fe32b04201dcea9b18b126609

##### docker info
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.10.4)
  compose: Docker Compose (Docker Inc., v2.17.2)
  dev: Docker Dev Environments (Docker Inc., v0.1.0)
  extension: Manages Docker extensions (Docker Inc., v0.2.19)
  init: Creates Docker-related starter files for your project (Docker Inc., v0.1.0-beta.2)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.25.0)
  scout: Command line tool for Docker Scout (Docker Inc., v0.9.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 46
 Server Version: 20.10.24
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2456e983eb9e37e47538f59ea18f2043c9a73640
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.49-linuxkit
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 7.675GiB
 Name: docker-desktop
 ID: LFZE:MO2Q:VNT5:CWXX:KVY4:TWCP:SLTL:MIDJ:ZIXC:CPYF:CQTP:GRYJ
 Debug Mode: false
 No Proxy: hubproxy.docker.internal
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  0.0.0.0:8080
  127.0.0.0/8
 Live Restore Enabled: false
 ```

quantumsheep:
## Summary
During container creation, pack was using the Docker <1.30 container wait way of doing. This made building waits undefinitely in some cases.

#### Before
`ContainerWait` is blocking.

#### After
`ContainerWait` is not blocking and we use the channels to wait for the states we wants.

## Related
Resolves #1732

### DIFF
--- a/internal/container/run.go
+++ b/internal/container/run.go
@@ -19,8 +19,32 @@ type DockerClient interface {
 	ContainerStart(ctx context.Context, container string, options types.ContainerStartOptions) error
 }
 
+func ContainerWaitWrapper(ctx context.Context, docker DockerClient, container string, condition dcontainer.WaitCondition) (<-chan dcontainer.WaitResponse, <-chan error) {
+	bodyChan := make(chan dcontainer.WaitResponse)
+	errChan := make(chan error)
+
+	go func() {
+		defer close(bodyChan)
+		defer close(errChan)
+
+		waitBodyChan, waitErrChan := docker.ContainerWait(ctx, container, dcontainer.WaitConditionNextExit)
+		for {
+			select {
+			case body := <-waitBodyChan:
+				bodyChan <- body
+				return
+			case err := <-waitErrChan:
+				errChan <- err
+				return
+			}
+		}
+	}()
+
+	return bodyChan, errChan
+}
+
 func RunWithHandler(ctx context.Context, docker DockerClient, ctrID string, handler Handler) error {
-	bodyChan, errChan := docker.ContainerWait(ctx, ctrID, dcontainer.WaitConditionNotRunning)
+	bodyChan, errChan := ContainerWaitWrapper(ctx, docker, ctrID, dcontainer.WaitConditionNextExit)
 
 	resp, err := docker.ContainerAttach(ctx, ctrID, types.ContainerAttachOptions{
 		Stream: true,
--- a/internal/container/run.go
+++ b/internal/container/run.go
@@ -20,7 +20,7 @@ type DockerClient interface {
 }
 
 func RunWithHandler(ctx context.Context, docker DockerClient, ctrID string, handler Handler) error {
-	bodyChan, errChan := docker.ContainerWait(ctx, ctrID, dcontainer.WaitConditionNextExit)
+	bodyChan, errChan := docker.ContainerWait(ctx, ctrID, dcontainer.WaitConditionNotRunning)
 
 	resp, err := docker.ContainerAttach(ctx, ctrID, types.ContainerAttachOptions{
 		Stream: true,
--- a/testhelpers/testhelpers.go
+++ b/testhelpers/testhelpers.go
@@ -24,7 +24,7 @@ import (
 	"github.com/buildpacks/imgutil/fakes"
 
 	dockertypes "github.com/docker/docker/api/types"
-	"github.com/docker/docker/api/types/container"
+	dcontainer "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/pkg/stdcopy"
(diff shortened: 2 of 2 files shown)


quantumsheep:
Issue fixed, it was an error on my side, not on buildpack's side

quantumsheep:
@natalieparellano Ok so... it's effectively an issue from buildpacks, if the request takes some time to be performed, the state is already `created` so it blocks infinitely

quantumsheep:
After investigating a little bit more, it seems like buildpacks was conceived to work with Docker < 1.30, at that time they used to not block on `ContainerWait` but instead returning two channels which can be used to wait for the container to be ready. That might be what is used by buildpacks.

joe-kimmel-vmw:
Hi @quantumsheep -  thanks for iterating on this and explaining what's going on!

I looked into this together with @natalieparellano today.  What you did makes sense, but it's basically shimming us backwards into an older api style.  It looks like in the newer api style, we could do without your Wrapper function if we called `start`, `attach`, and then `wait` (instead of calling wait first)?   I'm basing this guess on the [sdk examples](https://docs.docker.com/engine/api/sdk/examples/) provided by docker.  

Would you check whether we can also succeed in your use-case without the wrapper you wrote, by re-ordering our docker API calls?   If it works, this would be preferable from our perspective since it would update our usage of docker to the more modern paradigm, rather than shimming for backwards compatibility.

Thank you!

natalieparellano:
I'll defer to the `pack` maintainers, but IMO editing the `Handler` signature should be acceptable - it's already in an internal package so the surface area should be small.

quantumsheep:
> editing the `Handler` signature should be acceptable - it's already in an internal package so the surface area should be small.

I really insist on the fact that it might not work because a container can exit before we use `ContainerWait` if we're not fast enough. (I tested it 👀)

joe-kimmel-vmw:
Thanks @quantumsheep  - it sounds like the solution in this PR is the one to go with!  Really appreciate you doing the extra legwork to verify.

jkutner (approved):

PR #1736, opened 2023-04-25
Create benchmark.yml and add end-to-end benchmark

edithwuly:
## Summary
Create git action to keep runing benchmarks and graphing the results each time push on main branch. Benchmark for `pack build --buildpack` is added as well.

## Output
The graphs for benchmark results will be shown in [https://buildpack.github.io/pack/dev/bench/](url).

#### Before
Brank branch `gh-pages` is required for saving benchmark result data. index.html and data.js will be created automatically. `alert-threshold` and `alert-comment-cc-users` field should be decided.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,57 @@
+name: benchmark
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  benchmark:
+    name: Run Go benchmark example
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up go
+        uses: actions/setup-go@v4
+        with:
+          go-version: "1.19"
+      - name: Set up go env
+        run: |
+          echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+        shell: bash
+      - name: Run benchmark
+        run: |
+             mkdir out || (exit 0)
+             go test -bench=. -benchtime=1s ./benchmarks/... -tags=benchmarks | tee ./out/benchmark.txt
+
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: Go Benchmark
+          tool: 'go'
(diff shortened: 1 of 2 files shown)


benchmarks/build_test.go
```diff
@@ -21,10 +21,12 @@ import (
...
 var (
-	baseImg     = "some-org/" + h.RandString(10)
-	trustedImg  = baseImg + "-trusted-"
-	builder     = "cnbs/sample-builder:bionic"
-	mockAppPath = filepath.Join("..", "acceptance", "testdata", "mock_app")
+	baseImg             = "some-org/" + h.RandString(10)
+	trustedImg          = baseImg + "-trusted-"
+	builder             = "cnbs/sample-builder:bionic"
+	mockAppPath         = filepath.Join("..", "acceptance", "testdata", "mock_app")
+	paketoBuilder       = "paketobuildpacks/builder:base"
```
natalieparellano, line 28: Could we make this configurable with an env var or something? It would be nice to make it easy for contributors to test this out with builders that they care about.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,43 @@
...
+          name: Go Benchmark
+          tool: 'go'
+          output-file-path: ./out/benchmark.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '200%'
+          comment-on-alert: true
+          fail-on-alert: true
+          alert-comment-cc-users: '@edithwuly'
```
natalieparellano, line 43: Eventually we'll probably want this to be `'@buildpacks/platform-maintainers'` but this will be helpful to reduce noise while we're iterating on the benchmark suite

edithwuly: Ok, so I just leave `@edithwuly` here temporarily.

benchmarks/build_test.go
edithwuly: Sure, I have created `setEnv()` to set these values. If value is not set in the environment, default values will be assigned. `setEnv()` will be invoked at first of `BenchmarkBuild()`. Does this treatment make sense?

PR #1744, opened 2023-05-01
inspect image: add one unit test

joe-kimmel-vmw:


pkg/client/inspect_image_test.go
```diff
@@ -214,6 +214,23 @@ func testInspectImage(t *testing.T, when spec.G, it spec.S) {
 					h.AssertEq(t, infoWithExtension.StackID, "test.stack.id")
 				})
 
+				it("something stg run image", func() {
```
natalieparellano, line 217:
```suggestion
				it("returns the stack from runImage.Image if set", func() {
```

dfreilich: LGTM once this suggestion is merged

PR #1745, opened 2023-05-03
Update Paketo stack & builder references to Jammy

dmikusa:
## Summary

The Paketo Bionic stack is going away at the end of May (it will be out of support upstream then). This PR replaces all references to the Bionic stack with the new Jammy stack.

In addition, it adds references for the new static stack & builder. The static stack is even smaller than the tiny stack and is suitable for completely static binaries like those you can build with Go or Rust.


dfreilich (approved):
LGTM! Thanks @dmikusa 😄

PR #1757, opened 2023-05-10
Add support for extensions in pack Buildpack downloader

### LINKED ISSUE #1684: pack buildpack downloader should support extension packages
### Summary
Once https://github.com/buildpacks/pack/pull/1661 is merged the buildpack downloader should handle extension packages instead of returning nil [here](https://github.com/buildpacks/pack/blob/9e9b7ba1f197656a83c4daf7ee8e6d9e6648a6d9/pkg/buildpack/downloader.go#L201).

We should probably add an acceptance test that uses extension packages in a builder to ensure that things are working end-to-end.

itsdarshankumar:
## Summary
Adds feature to add packaged extensions while creating a builder

## Output
image
image
image

#### Before
Used to throw SIGSEV

## Related

Resolves #1684


natalieparellano:
Thank you @itsdarshankumar - this looks great! Would it be possible to add a test case to pkg/buildpack/downloader_test.go? Something along the lines of:
```
		when("package image lives in docker registry", func() {
			when("image key is provided", func()
				it("should succeed", func() {
```
The other tests in this context (for buildpacks) would probably be redundant for extensions.

itsdarshankumar:
> Thank you @itsdarshankumar - this looks great! Would it be possible to add a test case to pkg/buildpack/downloader_test.go? Something along the lines of:
> 
> ```
> 		when("package image lives in docker registry", func() {
> 			when("image key is provided", func()
> 				it("should succeed", func() {
> ```
> 
> The other tests in this context (for buildpacks) would probably be redundant for extensions.

sure, I'll add the test cases

natalieparellano (approved):
Awesome work @itsdarshankumar! IMO Codecov is being too picky... this will really smooth out the experience of using extensions for end users!

jjbustamante (approved):
Awesome @itsdarshankumar !!!

jkutner (approved):

PR #1772, opened 2023-05-19
Improve the `flatten-exclude` flag behavior when a composite buildpack is provided

jjbustamante:
## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___

### DIFF
--- a/internal/builder/builder.go
+++ b/internal/builder/builder.go
@@ -66,23 +66,22 @@ const (
 
 // Builder represents a pack builder, used to build images
 type Builder struct {
-	baseImageName            string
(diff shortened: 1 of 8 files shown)


jjbustamante:
Moving to milestone 0.31.0, I think we can release the flatten feature in 0.30.0 and after that work in the improvements

jjbustamante:
Closing this PR because the `flatten-exclude` flag is not [required](https://github.com/buildpacks/rfcs/pull/290#issuecomment-1660543031) anymore.

PR #1773, opened 2023-05-22
change additional buildpack to java

edithwuly:
## Summary
Change additional buildpack for benchmark to java.

## Output
The time spent on adding additional buildpack for `pack build` will be more apparent.

#### Before
Old additional buildpack is too small to show the difference after some optimizations are made.

#### After
Java buildpack is much bigger. Hopefully, we can see obvious drop in benchmark graphs.

## Documentation

- Should this change be documented?
    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___

PR #1777, opened 2023-05-23
Remove history for "additional buildpack"

natalieparellano:
The previous benchmarks used a buildpack that was small in size, hence any performance improvements would be hard to spot.

https://github.com/buildpacks/pack/pull/1773 will switch the buildpack to a large buildpack, thus making the benchmark more useful.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


natalieparellano:
I guess we can close this PR. We merged #1773 which would've caused a one-time huge performance hit, but https://buildpacks.github.io/pack/dev/bench/ only displays the last few commits anyway

PR #1781, opened 2023-05-25
Avoid writing duplicate tar files for flattened buildpacks

natalieparellano:
When `pack build --buildpack <flattened buildpack>` is invoked, `pack` doesn't know the buildpack is flattened, so it treats it like `pack build --buildpack bp/1 --buildpack bp2`,
resulting in unnecessary tar files being created
and duplicate layers being added to the ephemeral builder, which can exhaust disk space depending on the environment.

This optimization avoids writing the tar file more than once by returning a no-op read closer
when the diffID has already been queued for downloading.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


natalieparellano:
Closing in favor of https://github.com/buildpacks/pack/pull/1787

PR #1783, opened 2023-05-25
Flatten buildpack package is move to be experimental

### LINKED ISSUE #1595: Layer compression/flattening
### Description

When you build more complicated buildpacks and then pull many of those together into a builder, you can quickly hit the layer limit for a container image.

### Proposed solution

There are some seemingly natural points where pack could compress or flatten layers for the images it generates. For example, composite/meta buildpacks. When you are creating a builder or buildpack that consists of other composite/meta buildpacks pack could first flatten all of the layers from the composite buildpack and then create the new image with just a single layer for that composite/meta buildpack (this is opposed to the current behavior which inlines all of the layers from the composite/meta buildpack).

We might be able to hack this together with something like the following but I haven't tried it.

1. `pack buildpack package` a composite buildpack
2. somehow compress/flatten the layers on the image generated in step 1.
3. repeat for all buildpacks in the builder
4. `pack builder create` the builder

I worry that this will create problems with metadata though. 🤔 

### Additional context

This is a fairly urgent need for Paketo. We have hit the layer limit in our Paketo builders and are struggling to add new functionality because we can't add any more layers.

jjbustamante:
## Summary
After our working group meeting today (05/25/2023) some concerns were raised with breaking the [distribution spec ](https://github.com/buildpacks/spec/blob/distribution/0.2/distribution.md#buildpackage) flattening a buildpack package. This pull request is wrapping the functionality to be under the *experimetal* mode, it means, users will need to enable the experimental mode in order to see the new flags. Also, a warning message is display to end users. The idea is to do not block the feature for being released but still have the open conversation with the community in how the spec is being broke.

#### Before

`--flatten`, `--depth` and `flatten-exclude` flags were added to be available with the `pack buildpack package` command.

#### After
The flags were hidden if experimental mode is not enabled

Screenshot 2023-05-25 at 5 32 35 PM

They will be only available when experimental mode is enabled
Screenshot 2023-05-25 at 5 32 44 PM

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1595

### DIFF
--- a/internal/commands/buildpack_package.go
+++ b/internal/commands/buildpack_package.go
@@ -54,7 +54,7 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, packager Buildpa
 			"and they can be included in the configs used in `pack builder create` and `pack buildpack package`. For more " +
 			"on how to package a buildpack, see: https://buildpacks.io/docs/buildpack-author-guide/package-a-buildpack/.",
 		RunE: logError(logger, func(cmd *cobra.Command, args []string) error {
-			if err := validateBuildpackPackageFlags(&flags); err != nil {
+			if err := validateBuildpackPackageFlags(cfg, &flags); err != nil {
 				return err
 			}
 
@@ -96,6 +96,10 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, packager Buildpa
 					logger.Warnf("%s is not a valid extension for a packaged buildpack. Packaged buildpacks must have a %s extension", style.Symbol(ext), style.Symbol(client.CNBExtension))
 				}
 			}
+			if flags.Flatten {
+				logger.Warn("Creating a flattened Buildpack package could break the distribution specification. Please use it with caution.")
+			}
+
 			if err := packager.PackageBuildpack(cmd.Context(), client.PackageBuildpackOptions{
 				RelativeBaseDir: relativeBaseDir,
 				Name:            name,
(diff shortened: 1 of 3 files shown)


internal/commands/buildpack_package.go
```diff
@@ -134,22 +138,33 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, packager Buildpa
...
 		return errors.Errorf("--config and --path cannot be used together. Please specify the relative path to the Buildpack directory in the package config file.")
 	}
 
-	if p.Flatten && len(p.FlattenExclude) > 0 {
-		for _, exclude := range p.FlattenExclude {
-			if strings.Count(exclude, "@") != 1 {
-				return errors.Errorf("invalid format %s; please use '<buildpack-id>@<buildpack-version>' to exclude buildpack from flattening", exclude)
+	if p.Flatten {
+		if !cfg.Experimental {
+			return client.NewExperimentError("Flattening a buildpack package currently experimental.")
```
natalieparellano, line 160:
```suggestion
			return client.NewExperimentError("Flattening a buildpack package is currently experimental.")
```

internal/commands/buildpack_package.go
```diff
@@ -96,6 +96,10 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, packager Buildpa
 					logger.Warnf("%s is not a valid extension for a packaged buildpack. Packaged buildpacks must have a %s extension", style.Symbol(ext), style.Symbol(client.CNBExtension))
 				}
 			}
+			if flags.Flatten {
+				logger.Warn("Creating a flattened Buildpack package could break the distribution specification. Please use it with caution.")
```
natalieparellano, line 100:
```suggestion
				logger.Warn("Flattening a buildpack package could break the distribution specification. Please use it with caution.")
```
(for consistency with error message)

natalieparellano (approved):
Looks good to me! I left some nits...

jjbustamante:
@dfreilich , @hone , @jkutner 

Can I have some eyes or this PR merge?

internal/commands/buildpack_package.go
```diff
@@ -134,22 +138,33 @@ func BuildpackPackage(logger logging.Logger, cfg config.Config, packager Buildpa
...
 		return errors.Errorf("--config and --path cannot be used together. Please specify the relative path to the Buildpack directory in the package config file.")
 	}
 
-	if p.Flatten && len(p.FlattenExclude) > 0 {
-		for _, exclude := range p.FlattenExclude {
-			if strings.Count(exclude, "@") != 1 {
-				return errors.Errorf("invalid format %s; please use '<buildpack-id>@<buildpack-version>' to exclude buildpack from flattening", exclude)
+	if p.Flatten {
+		if !cfg.Experimental {
+			return client.NewExperimentError("Flattening a buildpack package is currently experimental.")
```
dfreilich, line 160: Maybe add in how they can enable experimental features, for users who don't know how

dfreilich (changes requested):
Small nit, but I'll approve it right when it's in ☺️

jjbustamante:
> Small nit, but I'll approve it right when it's in ☺️

@dfreilich thanks for your feedback. We already have a *TIP* message to guide users, it is shown like this:

Screenshot 2023-06-15 at 8 37 47 AM

internal/commands/buildpack_package.go
jjbustamante: We already have a message set [here](https://github.com/buildpacks/pack/blob/52902b093e04f3c37479def536cfaa54cc24ce17/internal/commands/commands.go#L77)

PR #1691, opened 2023-03-29
Add ` --flatten`, `--depth` and `flatten-exclude` flags to `pack builder create` and `pack buildpack package` command

jjbustamante:
## Summary

This PR adds new flags to:
 -  `pack builder create`
 -  `pack buildpack package`

The flags will try to flatten the layers in the final image.

for both commands the following flags were added:

```bash
      --depth int                   Max depth to flatten compose buildpacks layers.
                                    Omission of this flag or values < 0 will flatten the entire tree. (default -1)
      --flatten                     Flatten the composite buildpacks layers
  -e, --flatten-exclude strings     Buildpack to exclude for being flatten. Use buildpack by id and version in the form of '<buildpack>@<version>'
```

These flags can be used with the following combinations:
- `--flatten` only OR with `--depth` < 0: it will flatten **ALL** the buildpacks layers into 1 final layer, because the default value for `--depth` is < 0
- `--flatten` and `--depth` = 1: it will flatten only the top level of composites to be flattened (i.e. composites directly referenced from either the builder or directly referenced from a composite if building a buildpack)
- `--flatten` and `--depth` = 2: it will flatten the top 2 levels of composites buildpacks 

And so on 

In case the users want to **EXCLUDE** some buildpacks for being flatten, they can use the `--flatten-exclude` flag in combination with the previous one to exclude any buildpack they want. The buildpack *MUST* be specify using the format `flatten-exclude '<buildpack>@<version>'`

### Example

For testing the feature I used the buildpacks defined in our sample [repository](https://github.com/buildpacks/samples/tree/main/buildpacks) and a `builder.toml` with the following content:

```toml
# Buildpacks to include in builder
[[buildpacks]]
uri = "samples/buildpacks/hello-processes"

[[buildpacks]]
# Packaged buildpacks to include in builder;
# the "hello-universe" package contains the "hello-world" and "hello-moon" buildpacks
uri = "docker://cnbs/sample-package:hello-universe"

# Order used for detection
[[order]]
    # This buildpack will display build-time information (as a dependency)
    [[order.group]]
    id = "samples/hello-world"
    version = "0.0.1"

    # This buildpack will display build-time information (as a dependant)
    [[order.group]]
    id = "samples/hello-moon"
    version = "0.0.1"

    # This buildpack will create a process type "sys-info" to display runtime information
    [[order.group]]
    id = "samples/hello-processes"
    version = "0.0.1"

# Stack that will be used by the builder
[stack]
[...]

PR #1787, opened 2023-05-30
Explode flattened buildpacks when `pack build`

### LINKED ISSUE #1798: Fixups for flatten feature on pack 0.30.0
### Summary

We are planning to release a new experimental feature to flatten builder layers and buildpack package. This feature was requested by the following issue #1595. Since pack [0.30.0-pre2 release](https://github.com/buildpacks/pack/releases) a version of the feature was released. 
This issue is meant to capture all the bugs to be fixed before we can ship the next pack.

TODO:

 - [ ] --flatten-exclude should exclude from flattening (vs separately flatten)
 - [ ] --flatten-include

---

- [x] https://github.com/buildpacks/pack/pull/1783 - `pack buildpack package --flatten` must be released as experimental
- [x] https://github.com/buildpacks/pack/pull/1781 - ` disk issue bug`
- [ ]  https://github.com/buildpacks/pack/pull/1787 - `pack build --buildpack <flattened> bug`

jjbustamante:
## Summary
After the pack `0.30.0-pre2` was released some users reported some troubles when trying to use the new `flatten` flag added to the `pack builder create` and `pack buildpack package` commands. [See](https://cloud-native.slack.com/archives/C0331B61A1Y/p1684957284723219) slack thread

1. When `pack build app --buildpack` was used with a flattened buildpack on a GHA runner it was causing the runner to run out of disk space
2. When `pack build app --buildpack` was used with a flattened buildpack that is also present into the builder we want to use, pack was not able to deduplicate those buildpacks correctly.

This PR is trying to fix both issues.

#### Before

- *Issue 1*, when a buildpack is flattened we are saving in the same layer N number of buildpacks, during `pack build` an ephemeral builder is created, in this process we read each buildpack layer tar to disk in a temp folder and then create this builder (that is deleted at the end of the command execution) adding those tars to it. Because each flattened layer contains more than 1 buildpack, the iteration over the layers creates a file on disk with all the flattened buildpacks. that is the reason for the process to consume more hard drive space.

- *Issue 2*, when pack is creating the ephemeral builder and some buildpacks are pass through `-b` flag, pack calculates the hash of the tar file inside the provided buildpack and then compares this value against the hash of the same buildpack in the builder, because the provided buildpack is flattened (there are more than 1 buildpack in the layer) we need to split those buildpacks into individual ones.  

#### After

- *To fix issue 1*: This PR adds some logic to keep track of the layers DiffIds when a buildpack is fetched, if the same layer DiffID is found in more than 1 buildpack then only 1 of them returns the content of the layer tar (with all the other buildpacks inside of it) and the other returns an empty tar. In this way we only use the same hard drive space when the layer tar is written into disk. 

- *To fix issue 2*: This PR adds some logic when a flattened buildpack is fetched, Pack will try to write different tar files for each buildpack flattened inside a layer and after that the calculation of the hash to execute the deduplication logic remains the same.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1798


internal/builder/builder.go
```diff
@@ -560,57 +567,13 @@ func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata) e
...
-			}
-			lids[i] <- modInfo{
-				info:     info,
-				layerTar: layerTar,
-				diffID:   diffID,
-			}
-		}(i, module)
-	}
-
-	for i, module := range additionalModules {
```
jjbustamante, line 608: I move this logic to a new method `explodeBuildModules` to handle the parallelism and the logic to organize the BuildModules when they are flattened

pkg/buildpack/buildpack.go
```diff
@@ -29,6 +32,7 @@ type BuildModule interface {
 	// timestamp and root UID/GID).
 	Open() (io.ReadCloser, error)
 	Descriptor() Descriptor
+	ContainsFlattenedModules() bool
```
jjbustamante, line 35: This method is added to optimize when a BuildModule is going to be write into a tar, we need to know if there are more than one buildpack inside of it. Initially I was opening the tar and reading its content

pkg/buildpack/buildpack.go
```diff
@@ -48,22 +52,48 @@ type Blob interface {
 	Open() (io.ReadCloser, error)
 }
 
+type BlobOption func(*blobOption) error
```
jjbustamante, line 55: Adding an option when the Blob is created was the best way I found to modify the behavior without affecting a bunch of places in the code. Only where is required we can create flattened blobs

internal/builder/builder.go
```diff
@@ -1175,11 +1132,102 @@ func (b *Builder) whiteoutLayer(tmpDir string, i int, bpInfo dist.ModuleInfo) (s
...
+		}
+		for _, moduleTar := range mi.moduleTars {
+			// it could be an individual Buildpack or flattened Buildpack that writes an empty tar on disk
+			eBM := moduleWithDiffID{tarPath: moduleTar.Path()}
+			diffID, err := dist.LayerDiffID(moduleTar.Path())
+			if err != nil {
+				errs = append(errs, errors.Wrapf(err, "calculating layer diffID for path %s", moduleTar.Path()))
+				continue
+			}
+			if diffID.String() == "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" {
```
joe-kimmel-vmw, line 1185: why is it this exact hash?  is this a magic value that should be turned into a named constant?  do we use it anywhere else?

internal/builder/builder.go
```diff
@@ -1175,11 +1132,102 @@ func (b *Builder) whiteoutLayer(tmpDir string, i int, bpInfo dist.ModuleInfo) (s
...
 	}
 	sort.Strings(keys)
 	return keys
 }
+
+// splitBuildModules takes the given Build Modules and concurrently read their internal tar files. In
+// case a flattened module is found, it will split each flattened buildpack into an individual module and skip all the empty ones
+// returns an explodedBuildModule array with the Build Module information but also the diffId and the tar file on disk
+func splitBuildModules(kind, tmpDir string, additionalModules []buildpack.BuildModule, logger logging.Logger) ([]moduleWithDiffID, []error) {
+	lids := make([]chan modInfo, len(additionalModules))
```
joe-kimmel-vmw, line 1148: naming nit  -- there might be a convention i don't know, but  the name "lids" sounds either like it's a collection of the tops to jars or containers (lid) or like it's an ID for an L-word (L. IDs),  but neither of those suggests "lists of channels" to me? 

 e.g. would `modInfoChans` or even  be an ok compromise on the spectrum of short names vs. informative names?

pkg/buildpack/buildpack.go
```diff
@@ -336,3 +493,82 @@ func Set(exclude []string) map[string]struct{} {
...
+		module, err := newModuleTar(m.rootPath, id, version)
+		if err != nil {
+			return moduleTar{}, err
+		}
+		m.modules[key] = module
+	}
+	return m.modules[key], nil
+}
+
+func (m *moduleTarCollection) getModules() []ModuleTar {
```
joe-kimmel-vmw, line 556: this method seems awkward -  Can `m.modules` be `m.Modules` so we don't need this method at all?  Is there a case for converting here between reference and value types (but aren't maps always reference types)?

joe-kimmel-vmw (commented):
I only have cosmetic / nit level comments,  I don't think I know enough to critique your actual approach.  so for whatever it's worth, here you go. :)

pkg/buildpack/buildpack_test.go
```diff
@@ -548,6 +563,172 @@ version = "1.2.3"
...
+		it.After(func() {
+			err := os.RemoveAll(tmpDir)
+			if runtime.GOOS != "windows" {
+				// avoid "The process cannot access the file because it is being used by another process"
+				// error on Windows
+				h.AssertNil(t, err)
+			}
+		})
+
+		when("BuildModule contains N flattened buildpack", func() {
```
natalieparellano, line 587:
```suggestion
		when("BuildModule contains N flattened buildpacks", func() {
```

pkg/buildpack/buildpack_test.go
```diff
@@ -548,6 +563,172 @@ version = "1.2.3"
...
+				tarPaths, err := buildpack.ToNLayerTar(tmpDir, bp, logger)
+				h.AssertNil(t, err)
+				h.AssertEq(t, len(tarPaths), 2)
+				for _, tarPath := range tarPaths {
+					h.AssertOnTarEntry(t, tarPath.Path(), fmt.Sprintf("/cnb/buildpacks/%s", tarPath.Info().ID),
+						h.IsDirectory(),
+					)
+					h.AssertOnTarEntry(t, tarPath.Path(), fmt.Sprintf("/cnb/buildpacks/%s/%s", tarPath.Info().ID, tarPath.Info().Version),
+						h.IsDirectory(),
+					)
```
natalieparellano, line 647: Something about this test is a little off... I was going to suggest making the expectation a bit more robust by calculating the expected "buildpack-XXX-id" and "buildpack-XXX-version-1" from the loop index, but it didn't work. Then when I added `fmt.Println("tarPath.Info().ID:", tarPath.Info().ID)` to the beginning of the loop I saw 
```
tarPath.Info().ID: buildpack-2-id
tarPath.Info().ID: buildpack-2-id
```
The same ID prints out twice...

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,133 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "opening blob")
+	}
+	defer modReader.Close()
+
+	tarCollection := newModuleTarCollection(dest)
+	rootFolderHeaders := map[string]*tar.Header{}
+
+	// the build module __could__ have more than one buildpack on it,
+	// if we detect multiple buildpacks, explode them
```
natalieparellano, line 374:
```suggestion
```
Comment is outdated, we definitely have more than one module if we get here :)

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,133 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+
+func parseBpIDAndVersion(hdr *tar.Header) (id, version string) {
+	// splitting "/cnb/buildpacks/{ID}/{version}/*" returns
+	// [0] = "" -> first element is empty
+	// [1] = "cnb"
+	// [2] = "buildpacks"
+	// [3] = "{ID}"
+	// [4] = "{version}"
+	// ...
+	parts := strings.Split(hdr.Name, "/")
```
natalieparellano, line 460: Is this true even for Windows buildpack packages?

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,133 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+					// If the reader was empty, we need to write an empty tar on disk
+					return handleSingleOrEmptyModule(dest, module)
+				}
+				break
+			}
+			return nil, errors.Wrapf(err, "failed to read next data '%s'", header.Name)
+		}
+
+		id, version := parseBpIDAndVersion(header)
+		if !matchIDVersion(module.Descriptor(), id, version) {
```
natalieparellano, line 391: Can we add a comment here about how this could happen?

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,133 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+	errs := tarCollection.close()
+	if len(errs) > 0 {
+		return nil, errors.New("closing files")
+	}
+
+	return tarCollection.getModules(), nil
+}
+
+func matchIDVersion(desc Descriptor, id, version string) bool {
+	sameID := id == desc.EscapedID() || id == desc.Info().ID
```
natalieparellano, line 445: I'm curious to know why we have to check both escaped and not escaped. Shouldn't we always see the escaped id when reading files from the package?

pkg/buildpack/buildpack.go
```diff
@@ -336,3 +493,82 @@ func Set(exclude []string) map[string]struct{} {
...
+		}
+		m.modules[key] = module
+	}
+	return m.modules[key], nil
+}
+
+func (m *moduleTarCollection) getModules() []ModuleTar {
+	var modulesTar []ModuleTar
+	for _, v := range m.modules {
+		vv := &v // TODO check this, its a hack for lint
```
natalieparellano, line 559: This is why I was seeing the odd behavior in https://github.com/buildpacks/pack/pull/1787/files#r1232634690 - `v` has the same address throughout the loop. To fix, you could do something like:
```
for _, v := range m.modules {
  v := v
  vv := &v
  ...
```
But I think there might be a better way to solve it

internal/builder/builder.go
jjbustamante: Oh yeah, I forgot to create the constant, the value is the hash of a empty tar file

internal/builder/builder.go
jjbustamante: Sorry, this one was just a copy from a previous method.

pkg/buildpack/buildpack.go
jjbustamante: The idea here is:

acceptance/acceptance_test.go
```diff
@@ -2253,6 +2253,88 @@ include = [ "*.jar", "media/mountain.jpg", "/media/person.png", ]
...
+						assertImage.HasLengthLayers(flattenedPackageName, 1)
+					})
+
+					it.After(func() {
+						assert.Nil(os.RemoveAll(tmpDir))
+						imageManager.CleanupImages(flattenedPackageName)
+					})
+
+					when("--flatten", func() {
+						it("creates the package as a single layer and then use it to run pack build -b", func() {
```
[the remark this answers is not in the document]
jjbustamante: Wait, maybe I just need to add more asserts to this one. 

The setup for this test has the following configuration

1. *The Builder*

Screenshot 2023-06-20 at 12 19 21 PM

it contains the *simple_layer* buildpack as individual layer

2. *The flattened buildpack*

Screenshot 2023-06-20 at 12 18 33 PM

It contains also the *simple_layer* buildpack

pkg/buildpack/buildpack_test.go
```diff
@@ -571,3 +737,58 @@ type readerBlob struct {
...
+				found = true
+			}
+		}
+		h.AssertTrue(t, found)
+	}
+
+	for _, moduleTar := range actual {
+		h.AssertOnTarEntry(t, moduleTar.Path(), fmt.Sprintf("/cnb/buildpacks/%s", moduleTar.Info().ID),
+			h.IsDirectory(),
+		)
```
natalieparellano, line 777:
```suggestion
					id := idx + 1
					h.AssertOnTarEntry(t, tarPath.Path(), fmt.Sprintf("/cnb/buildpacks/buildpack-%d-id", id),
						h.IsDirectory(),
					)
```

`moduleTar` is an output of the function being tested, and as such, it isn't safe to use in the expectation. `moduleTar.Info().ID` was hiding the bug mentioned earlier, as it was returning `2` twice.

jjbustamante: yeah, I wanted to check first the expected ID and Version in line 764. The comparison is made using the fixed expected values

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,131 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
 	return layerTar, nil
 }
 
+func ToNLayerTar(dest string, module BuildModule, logger logging.Logger) ([]ModuleTar, error) {
```
natalieparellano, line 359: Maybe it's just me, but personally I found this function a bit hard to read and understand. The need to keep track of `rootFolderHeaders` and then `continue` within the loop was a bit confusing. What about something like:

```
func ToNLayerTar(dest string, module BuildModule) ([]ModuleTar, error) {
	if !module.ContainsFlattenedModules() {
		return handleSingleOrEmptyModule(dest, module)
	}

	modReader, err := module.Open()
	if err != nil {
		return nil, errors.Wrap(err, "opening blob")
	}
	defer modReader.Close()

	tarCollection := newModuleTarCollection(dest)
	tr := tar.NewReader(modReader)

	// read the first header
	header, err := tr.Next()
	if err != nil {
		if err == io.EOF {
			return handleSingleOrEmptyModule(dest, module)
		}
		return nil, fmt.Errorf("failed to read first header '%s': %w", header.Name, err)
	}

	// the original version should be blank because the first header should look like "/cnb/buildpacks/<buildpack-id>"
	origID, origVersion := parseBpIDAndVersion(header)
	if origVersion != "" {
		return nil, fmt.Errorf("first header '%s' contained unexpected version", header.Name)
	}

	if err := toNLayerTar(origID, origVersion, header, tr, tarCollection); err != nil {
		return nil, err
	}

	errs := tarCollection.close()
	if len(errs) > 0 {
		return nil, errors.New("closing files")
	}

	return tarCollection.getModules(), nil
}
```

That uses a recursive function:
```
func toNLayerTar(origID, origVersion string, firstHeader *tar.Header, tr *tar.Reader, tc *moduleTarCollection) error {
	toWrite := []*tar.Header{firstHeader}
	if origVersion == "" { // TODO: test flattened module that contains buildpacks with same ID but different versions
		// the first header only contains the id - e.g., /cnb/buildpacks/<buildpack-id>,
		// read the next header to get the version
		secondHeader, err := tr.Next()
		if err != nil {
			return err
		}
		nextID, nextVersion := parseBpIDAndVersion(secondHeader)
		if nextID != origID || nextVersion == "" {
			return fmt.Errorf("second header '%s' contained unexpected id or missing version", secondHeader.Name)
		}
		origVersion = nextVersion
		toWrite = append(toWrite, secondHeader)
	} else {
		// the first header contains id and version - e.g., /cnb/buildpacks/<buildpack-id>/<buildpack-version>,
		// we need to write the parent header - e.g., /cnb/buildpacks/<buildpack-id>
		realFirstHeader := firstHeader
		realFirstHeader.Name = filepath.Dir(firstHeader.Name) // TODO: check this logic
		toWrite = append([]*tar.Header{realFirstHeader}, toWrite...)
	}
	mt, err := tc.get(origID, origVersion)
	if err != nil {
		return err
	}
	for _, h := range toWrite {
		if err := mt.writer.WriteHeader(h); err != nil {
			return fmt.Errorf("failed to write header '%s': %w", h.Name, err)
		}
	}
	// write the rest of the package
	var header *tar.Header
	for {
		header, err = tr.Next()
		if err != nil {
			if err == io.EOF {
				return nil
			}
			return err
		}
		nextID, nextVersion := parseBpIDAndVersion(header)
		if nextID != origID || nextVersion != origVersion {
			// we found a new module, recurse
			return toNLayerTar(nextID, nextVersion, header, tr, tc)
		}

		err = mt.writer.WriteHeader(header)
		if err != nil {
			return errors.Wrapf(err, "failed to write header for '%s'", header.Name)
		}

		buf, err := io.ReadAll(tr)
		if err != nil {
			return errors.Wrapf(err, "failed to read contents of '%s'", header.Name)
		}

		_, err = mt.writer.Write(buf)
		if err != nil {
			return errors.Wrapf(err, "failed to write contents to '%s'", header.Name)
		}
	}
}
```

pkg/buildpack/package.go
```diff
@@ -155,6 +204,27 @@ func extractExtensions(pkg Package) (mainExt BuildModule, err error) {
 	return mainExt, nil
 }
 
+type void struct{}
+
+var member void
+
+func flattenedBuildpacksDiffIDs(layers dist.ModuleLayers) map[string]struct{} {
+	var seenDiffID = make(map[string]struct{})
+	var flattenedDiffIDs = make(map[string]struct{})
```
natalieparellano, line 213: Why do we need two maps?

pkg/buildpack/package.go
```diff
@@ -57,6 +58,38 @@ func extractBuildpacks(pkg Package) (mainBP BuildModule, depBPs []BuildModule, e
...
+	//      "stacks": [
+	//        {
+	//          "id": "io.buildpacks.samples.stacks.jammy"
+	//        },
+	//        {
+	//          "id": "io.buildpacks.samples.stacks.alpine"
+	//        },
+	//        {
+	//          "id": "io.buildpacks.stacks.jammy"
+	//        },
```
natalieparellano, line 76:
```suggestion
```
(makes the comment more succinct, while conveying the same information)

pkg/buildpack/buildpack.go
jjbustamante: I think I added that one because some unit tests, let me try removing it.

pkg/buildpack/package.go
jjbustamante: I added a new method `ContainsFlattenedModules() bool` to the BuildModule interface. 

Let's suppose we have a. flattened buildpack with buildpacks [A,B,C] inside of it.

- seenDiffID: tracks the second time you find the same DiffID, this is useful to return the empty reader. In the example, it will tell us to return an empty reader for B and C for example
- flattenedDiffIDs: tracks all the build modules associated with a particular DiffID, basically to being able to return *true* with the new method added. In the example, it will tell me [A,B,C] are flattened. 

Basically is for keep track of all the buildpacks associated with a DiffID.

pkg/buildpack/buildpack.go
jjbustamante: Suppose we have the following Build Module array = [ BM1, BM2 ] those are flattened so it means, when I open them and read their content I will find:
- 1 empty tarball 
- 1 tarball containing 
   -  /cnb/buildpacks/module-1/version-1 
   - /cnb/buildpacks/module-2/version-2

When we are iterating over the array [BM1, BM2].
**id, version** could be: (module-1, version-1) or (module-2,version-2)
**module** could be also BM1 or BM2. depending on which header we are reading from the tarball, those values could be the same or not, when they are not the same, it means *we are reading a flattened module from the tarball*

pkg/buildpack/buildpack.go
jjbustamante: Well, the code works fine, all the tests and acceptance tests are passing.

pkg/buildpack/buildpack.go
```diff
@@ -336,3 +491,83 @@ func Set(exclude []string) map[string]struct{} {
...
+		}
+		m.modules[key] = module
+	}
+	return m.modules[key], nil
+}
+
+func (m *moduleTarCollection) getModules() []ModuleTar {
+	var modulesTar []ModuleTar
+	for _, v := range m.modules {
+		v := v
```
natalieparellano, line 557: Alternatively if you wanted to avoid this inefficiency you could change the `moduleTar` methods to be declared like `func (t moduleTar) Info() dist.ModuleInfo` (removing the pointer receiver)

pkg/buildpack/package.go
natalieparellano: Oic... what about `var seenDiffID = make(map[string]bool` where `bool` is "is flattened" (or not)?

internal/builder/builder.go
```diff
@@ -1175,11 +1134,103 @@ func (b *Builder) whiteoutLayer(tmpDir string, i int, bpInfo dist.ModuleInfo) (s
...
+	// are included in a different Build Module that was split
+	for i, module := range additionalModules {
+		mi := <-modInfoChans[i]
+		if mi.err != nil {
+			errs = append(errs, mi.err)
+			continue
+		}
+		for _, moduleTar := range mi.moduleTars {
+			// it could be an individual Buildpack or flattened Buildpack that writes an empty tar on disk
+			eBM := moduleWithDiffID{tarPath: moduleTar.Path()}
```
natalieparellano, line 1182: What does `eBM` mean? 🙃

internal/builder/builder.go
```diff
@@ -1175,11 +1134,103 @@ func (b *Builder) whiteoutLayer(tmpDir string, i int, bpInfo dist.ModuleInfo) (s
...
+				errs = append(errs, errors.Wrapf(err, "calculating layer diffID for path %s", moduleTar.Path()))
+				continue
+			}
+			if diffID.String() == emptyTarDiffID {
+				logger.Debugf("%s %s is a component of a flattened buildpack that will be added elsewhere, skipping...", istrings.Title(kind), style.Symbol(moduleTar.Info().FullName()))
+				continue // we don't need to keep empty tars
+			}
+			eBM.diffID = diffID.String()
+			if moduleTar.Info().FullName() == fmt.Sprintf("%s@%s", module.Descriptor().EscapedID(), module.Descriptor().Info().Version) ||
+				moduleTar.Info().FullName() == module.Descriptor().Info().FullName() {
```
natalieparellano, line 1194: A `namesMatch` helper function (or something) might make this more readable

pkg/buildpack/buildpack.go
natalieparellano: Sounds good to me :) I think this extra bit of context will be helpful to future readers of this code.

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,133 @@
...
+
+	tarCollection := newModuleTarCollection(dest)
+	rootFolderHeaders := map[string]*tar.Header{}
+
+	// the build module __could__ have more than one buildpack on it,
+	// if we detect multiple buildpacks, explode them
+	tr := tar.NewReader(modReader)
+	headersProcessed := 0
+	for {
+		header, err := tr.Next()
```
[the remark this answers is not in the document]
natalieparellano: We should address this. The linked code suggestion looks pretty easy to implement:

```
func unzipGood(f string) {
	r, _ := zip.OpenReader(f)
	for _, f := range r.File {
		p, _ := filepath.Abs(f.Name)
		// GOOD: Check that path does not contain ".." before using it
		if !strings.Contains(f.Name, "..") {
			ioutil.WriteFile(p, []byte("present"), 0666)
		}
	}
}
```

pkg/buildpack/buildpack.go
natalieparellano: I had (in order) the same thought as Joe, then the same thought as Juan... maybe `moduleTars()` would be slightly clearer but IDK

acceptance/acceptance_test.go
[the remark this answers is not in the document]
natalieparellano: I think the failure case is when the buildpack that already exists is non-flattened (with two or more layers) and then we try to drop a flattened version on top of that builder. That is the case where we will randomly replace one of the layers and fail to replace the other layer, resulting in something like `embedded daemon response: Error processing tar file(duplicates of file paths not supported)`

pkg/buildpack/buildpack.go
natalieparellano: This code ^^ seems clearer to me, but I wrote it 🙃 @joe-kimmel-vmw do you have any opinions about which one is friendlier to a newcomer?

pkg/buildpack/buildpack.go
natalieparellano: Or you could make the map `map[string]*moduleTar` ...that might be the most straightforward. But honestly I don't think it matters too much.

internal/builder/builder.go
```diff
@@ -1175,11 +1132,102 @@ func (b *Builder) whiteoutLayer(tmpDir string, i int, bpInfo dist.ModuleInfo) (s
...
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
 	return keys
 }
+
+// splitBuildModules takes the given Build Modules and concurrently read their internal tar files. In
+// case a flattened module is found, it will split each flattened buildpack into an individual module and skip all the empty ones
+// returns an explodedBuildModule array with the Build Module information but also the diffId and the tar file on disk
+func splitBuildModules(kind, tmpDir string, additionalModules []buildpack.BuildModule, logger logging.Logger) ([]moduleWithDiffID, []error) {
```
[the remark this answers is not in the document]
jjbustamante: Yeah! you nailed it with that! I think that change fix this issue! thanks!

pkg/buildpack/buildpack.go
jjbustamante: Adding the test case you mentioned in the `TODO: test flattened module that contains buildpacks with same ID but different versions
the first header only contains the id - e.g., /cnb/buildpacks/<buildpack-id>,
read the next header to get the version`


I am getting the error:

```bash
buildpack_test.go:776: Expected nil: could not find entry path '/cnb/buildpacks/buildpack-1-id/buildpack-1-version-2': not exist
```

But I am also having an error with the previous code, so I will take a look

internal/builder/builder.go
jjbustamante: explodedBuildModule

pkg/buildpack/package.go
jjbustamante: mmm, lets see. 
`var seenDiffID = make(map[string]bool)` the logic will be something like:

```golang
if _, ok := seenDiffID[bpInfo.LayerDiffID]; ok {
  // second time we see the same DiffID
  seenDiffID[bpInfo.LayerDiffID] = true
} else {
  // first time we see the DiffID, we don't know if it is flattened
  seenDiffID[bpInfo.LayerDiffID] = false
}

// some where out of the loop we will need to iterate over seenDiffID to just get those keys that are true (flattened) 
// or returned the map[string]bool and change the logic to check the key and the value right?

```

internal/builder/builder.go
natalieparellano: How about `explodedMod` or just `exploded`?

jjbustamante: I like `explodedMod`

natalieparellano:
@jjbustamante one last follow up from #1691 would be to unskip the Windows tests that were pended (or fail gracefully if flatten is not supported on Windows). They are this one: https://github.com/buildpacks/pack/pull/1691/files#diff-d427bceabb3bc661df60689acac5f3dacce1e91fdb5cf5e5e9d8fe8ac33919dbR47 and this one: https://github.com/buildpacks/pack/pull/1691/files#diff-2c747e1652f70c82a8f571462e63b90a1fb6e248ee09572e1b80f6e2c74f4bc2R735

I think the latter is more important as that is a test that was running and passing before https://github.com/buildpacks/pack/pull/1691. We should ensure there isn't a regression from pack 0.29.0 -> 0.30.0

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +356,146 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "opening blob")
+	}
+	defer modReader.Close()
+
+	tarCollection := newModuleTarCollection(dest)
+	tr := tar.NewReader(modReader)
+
+	// read the first header
+	header, err := tr.Next()
```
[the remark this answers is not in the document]
natalieparellano: We should handle this...

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +355,173 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+}
+
+func toNLayerTar(origID, origVersion string, firstHeader *tar.Header, tr *tar.Reader, tc *moduleTarCollection) error {
+	toWrite := []*tar.Header{firstHeader}
+	if runtime.GOOS == "windows" && !strings.Contains(firstHeader.Name, "/cnb/buildpacks/") {
+		// On windows we can see the following headers before finding */cnb/buildpacks/<buildpack-id>/*
+		// Files
+		// Hives
+		// Files/cnb
+		// Files/cnb/buildpacks
```
natalieparellano, line 406: Do we need to add these files back on Windows in the case that the first header contains `/cnb/buildpacks/<buildpack-id>/<buildpack-version>`?

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +355,173 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+		for {
+			next, err := tr.Next()
+			if err != nil {
+				return err
+			}
+			if !strings.Contains(next.Name, "/cnb/buildpacks/") {
+				toWrite = append(toWrite, next)
+			} else {
+				origID, origVersion = parseBpIDAndVersion(next)
+				tc.buffer = toWrite
```
natalieparellano, line 416: Why do we need `buffer` to live on the tar collection instead of being a local variable? It is only used in this function...

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +355,173 @@ func ToLayerTar(dest string, module BuildModule) (string, error) {
...
+				toWrite = []*tar.Header{next}
```
natalieparellano, line 417: What happened to `firstHeader` in this case? It seems that it is lost...

pkg/buildpack/buildpack.go
natalieparellano: This function might be easier to understand if we handle the Files/Hives before recursing

pkg/buildpack/buildpack.go
```diff
@@ -48,22 +51,48 @@ type Blob interface {
 	Open() (io.ReadCloser, error)
 }
 
+type BlobOption func(*blobOption) error
+
+type blobOption struct {
+	flattened bool
+}
+
+func Flattened() BlobOption {
```
natalieparellano, line 60: Do we ever actually use this option? I only see it being called in tests...

pkg/buildpack/buildpack.go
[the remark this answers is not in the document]
jkutner: oh yea. I've been victim to this before...

pkg/buildpack/buildpack.go
natalieparellano: I guess in this case the `firstHeader` was `Files` so we are throwing it away... but we might need to add it back (see here: https://github.com/buildpacks/pack/pull/1787#discussion_r1243827884)

pkg/buildpack/buildpack.go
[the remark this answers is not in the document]
natalieparellano: According to @joe-kimmel-vmw this MAY be addressed in go 1.20. That would be nice...

pkg/buildpack/buildpack.go
jjbustamante: After merging the suggestions branch, we are not. I removed it

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +327,183 @@
...
+	tarCollection := newModuleTarCollection(dest)
+	tr := tar.NewReader(modReader)
+
+	var (
+		header     *tar.Header
+		forWindows bool
+	)
+
+	for {
+		header, err = tr.Next()
```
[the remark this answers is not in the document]
jjbustamante: Handled at line 353

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +327,183 @@
...
+
+	return tarCollection.moduleTars(), nil
+}
+
+func toNLayerTar(origID, origVersion string, firstHeader *tar.Header, tr *tar.Reader, tc *moduleTarCollection, forWindows bool) error {
+	toWrite := []*tar.Header{firstHeader}
+	if origVersion == "" {
+		// the first header only contains the id - e.g., /cnb/buildpacks/<buildpack-id>,
+		// read the next header to get the version
+		secondHeader, err := tr.Next()
```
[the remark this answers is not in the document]
jjbustamante: handled at line 397

pkg/buildpack/buildpack.go
```diff
@@ -326,6 +327,183 @@
...
+	}
+	for _, h := range toWrite {
+		if err := mt.writer.WriteHeader(h); err != nil {
+			return fmt.Errorf("failed to write header '%s': %w", h.Name, err)
+		}
+	}
+	// write the rest of the package
+	var header *tar.Header
+	for {
+		header, err = tr.Next()
```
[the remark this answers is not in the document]
jjbustamante: handled at line 435

jjbustamante:
@natalieparellano @joe-kimmel-vmw @jkutner I will love if you can review the PR again

pkg/buildpack/buildpack.go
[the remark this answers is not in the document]
natalieparellano: Perhaps we don't need to do it as part of this PR (because we use tar readers all over) but it would be cleaner to wrap `tar.NewReader` with our own reader that sanitizes the path as part of calling `tr.Next()`

natalieparellano (approved):
Looks good to me :) Thanks for all of your hard work in getting this over the finish line @jjbustamante

joe-kimmel-vmw:
IMO this is in good shape.  I did make one tiny suggestion in this PR (set to merge into this branch): https://github.com/buildpacks/pack/pull/1829/files

jkutner (approved):

PR #1788, opened 2023-05-30
Fix: deriving stack.toml from the new run image information

natalieparellano:
...we should write the file with the correct schema

Fixes bug identified during Working Group demo


jjbustamante (approved):
Sweet!

dfreilich (approved):
image

PR #1789, opened 2023-05-30
Stack fix & create new volume cache for kaniko instead of re-using build cache

natalieparellano:
This removes the unnecessary restriction that the build cache must be a volume cache in order to use extensions.


pkg/cache/cache_opts.go
```diff
@@ -139,22 +140,21 @@ func sanitize(c *CacheOpts) error {
...
+	)
+	if c.Build.Format == CacheBind {
+		if resolvedPath, err = filepath.Abs(c.Build.Source); err != nil {
+			return errors.Wrap(err, "resolve absolute path")
+		}
+		c.Build.Source = filepath.Join(resolvedPath, "build-cache")
+	}
+	if c.Launch.Format == CacheBind {
+		if resolvedPath, err = filepath.Abs(c.Launch.Source); err != nil {
+			return errors.Wrap(err, "resolve absolute path")
```
natalieparellano, line 155: Totally unrelated change... I can pull it out of this PR if desired. Looking through the code it seems the current implementation unnecessarily skips resolving the launch cache when the build cache is a bind.

jjbustamante:
@natalieparellano

I can see you have the following error message during the test execution:

` ERROR: failed to read builder image: failed to create cache directory: mkdir /kaniko/cache: permission denied`

I have something similar in my flattened fixes PR, do you have any thoughts on root cause for that?

natalieparellano:
Thanks for the review @jjbustamante! This looks to be an issue in the lifecycle - I put up https://github.com/buildpacks/lifecycle/pull/1117

natalieparellano:
I've confirmed manually that lifecycle built from `main` together with the fix from https://github.com/buildpacks/pack/pull/1795 should cause the test that's currently failing to go green; when lifecycle 0.17.0-rc.1 is out, we can update this PR to point to the lifecycle release candidate in `DefaultLifecycleVersion`

jkutner (approved):

PR #1791, opened 2023-06-02
add support for pack to extend base images using daemon

### LINKED ISSUE #1623: [RFC #0105] pack should support build image extension with Docker
RFC: https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md

### Description

`pack` enables build image extension by calling the `extender`. However, this is sub-optimal. While enabling users to play with the feature, it is not performant, using `kaniko` under the hood.

### Proposed solution

As `pack` already has access to a daemon, it can apply the dockerfiles directly, saving the extended build base image in the daemon. Thus it will not need to use the `extender`.

### Additional context

In the build image case, having this feature will allow us to drop the requirement that the image being extended is published to a registry (because we are not using kaniko, we don't need a manifest).

### Acceptance criteria

`pack build` with image extensions should complete N ms faster compared to `pack build` with image extensions now.

itsdarshankumar:
## Summary

As pack already has access to a daemon, it can apply the dockerfiles directly, saving the extended build base image in the daemon which gives us great performance benefits. Thus it will not need to use the extender phase of lifecycle. Additionally it will drop the requirement that the image being extended must be published to a registry.

## Output

image

image

## Related

Resolves #1623


internal/build/helper.go
```diff
@@ -0,0 +1,90 @@
...
+	fmt.Println("extensions.Extensions", extensions.Extensions)
+	return nil
+}
+
+func readExtensionsGroup(path string) ([]buildpack.GroupElement, error) {
+	var group buildpack.Group
+	_, err := toml.DecodeFile(filepath.Join(path, "group.toml"), &group)
+	for e := range group.GroupExtensions {
+		group.GroupExtensions[e].Extension = true
+		group.GroupExtensions[e].Optional = true
```
natalieparellano, line 83: Nice :) We should really offer better helpers for this in the lifecycle...

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +722,50 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+	if err != nil {
+		return fmt.Errorf("getting %s.Dockerfiles: %w", DockerfileKindRun, err)
+	}
+	fmt.Println("Dockerfiles: ", dockerfiles)
+	fmt.Println("extend: ", dockerfiles[1].Extend)
+	for _, dockerfile := range dockerfiles {
+		if dockerfile.Extend {
+			fmt.Println("dockerfile: ", dockerfile)
+			fmt.Println("dockerfile.Path dir: ", filepath.Dir(dockerfile.Path))
+			buildContext := archive.ReadDirAsTar(filepath.Dir(dockerfile.Path), "/", 0, 0, -1, true, false, defaultFilterFunc)
```
natalieparellano, line 741: I'd need to double check, but I believe the build context when extending with kaniko is the app directory

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +722,50 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+			fmt.Println("buildContext: ", buildContext)
+			buildArguments := map[string]*string{}
+			if dockerfile.WithBase == "" {
+				buildArguments["base_image"] = currentRunImage
+			}
+			buildOptions := types.ImageBuildOptions{
+				Context:    buildContext,
+				Dockerfile: "Dockerfile",
+				Tags:       []string{"run-image"},
```
natalieparellano, line 750: We probably need to come up with some unique identifier here

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +722,50 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+				Remove:     true,
+				BuildArgs:  buildArguments,
```
natalieparellano, line 752: Build arguments should additionally include a UUID, user_id, group_id, and any arguments that were defined by the extension in `extend-config.toml`. If it's possible to re-use lifecycle methods to construct these, that would be awesome :)

jjbustamante:
Moving this one to milestone 0.31.0

internal/build/phase_config_provider.go
```diff
@@ -244,6 +244,12 @@ func WithRoot() PhaseConfigProviderOperation {
 	}
 }
 
+func WithoutPrivilege() PhaseConfigProviderOperation {
+	return func(provider *PhaseConfigProvider) {
+		provider.ctrConf.User = "1000:1001"
```
natalieparellano, line 249: Rather than hard-code the uid/gid we can get this information from the builder image

internal/build/lifecycle_execution.go
```diff
@@ -228,10 +234,25 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
...
-			group.Go(func() error {
-				l.logger.Info(style.Step("EXTENDING (BUILD)"))
-				return l.ExtendBuild(ctx, buildCache, phaseFactory)
-			})
+			if l.opts.Publish {
+				group.Go(func() error {
+					l.logger.Info(style.Step("EXTENDING (BUILD)"))
+					return l.ExtendBuild(ctx, buildCache, phaseFactory)
+				})
+			} else {
```
natalieparellano, line 242:
```suggestion
```
We should be able to remove this block. We can extend the build image using the daemon even when we are publishing the app image to a registry

internal/build/lifecycle_execution.go
```diff
@@ -245,14 +266,25 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
...
+				start = time.Now()
+				group.Go(func() error {
+					l.logger.Info(style.Step("EXTENDING (RUN) BY DAEMON"))
+					defer func() {
+						duration := time.Since(start)
+						fmt.Println("Execution time:", duration)
+					}()
+					return l.ExtendRunByDaemon(ctx, group, &currentRunImage)
+				})
+			}
```
natalieparellano, line 286:
```suggestion
			group.Go(func() error {
				l.logger.Info(style.Step("EXTENDING (RUN)"))
				return l.ExtendRun(ctx, buildCache, phaseFactory)
			})
```
(as we saw worse performance using the daemon to extend the run image due to the time spent saving the extended layers out of the daemon)

natalieparellano: We should add a test for this... or maybe we could just rework the existing tests for `ExtendBuild` to cover the daemon case, since we probably won't use kaniko ever to extend the build image

internal/build/lifecycle_execution.go
```diff
@@ -638,6 +674,8 @@ func (l *LifecycleExecution) Build(ctx context.Context, phaseFactory PhaseFactor
 		WithNetwork(l.opts.Network),
 		WithBinds(l.opts.Volumes...),
 		WithFlags(flags...),
+		If((!l.opts.Publish && l.hasExtensionsForBuild()), WithImage("newbuilder-image:latest")),
```
natalieparellano, line 677: Perhaps something like `<original-builder-image-tag>-extended-randString(10)`

internal/build/lifecycle_execution.go
```diff
@@ -733,6 +898,12 @@ func (l *LifecycleExecution) Export(ctx context.Context, buildCache, launchCache
 		}
 	}
 
+	extDirEnv := NullOp()
+	if !l.opts.Publish {
+		l.logger.Debug("export for extend by daemon")
+		extDirEnv = WithEnv("CNB_EXTENDED_DIR=" + filepath.Join("/", "extended-new"))
+	}
+
```
natalieparellano, line 906:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -769,6 +940,7 @@ func (l *LifecycleExecution) Export(ctx context.Context, buildCache, launchCache
 		),
 		WithArgs(append([]string{l.opts.Image.String()}, l.opts.AdditionalTags...)...),
 		WithRoot(),
+		WithBinds(filepath.Join(l.tmpDir, "extended-new") + ":/extended-new"),
```
natalieparellano, line 943:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -785,6 +957,7 @@ func (l *LifecycleExecution) Export(ctx context.Context, buildCache, launchCache
 			CopyOut(l.opts.Termui.ReadLayers, l.mountPaths.layersDir(), l.mountPaths.appDir()))),
 		epochEnv,
 		expEnv,
+		extDirEnv,
```
natalieparellano, line 960:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
 	defer extend.Cleanup()
 	return extend.Run(ctx)
 }
+func (l *LifecycleExecution) ExtendBuildByDaemon(ctx context.Context) error {
+	extendtime := time.Now()
```
natalieparellano, line 748: As we get closer to completion we could remove some of the profiling stuff in favor of descriptive log messages

internal/build/helper.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+const (
+	DockerfileKindBuild = "build"
+	DockerfileKindRun   = "run"
+)
+
+type Extensions struct {
+	Extensions []buildpack.GroupElement
+}
```
natalieparellano, line 30: Can we re-use structs from the lifecycle?

internal/build/helper.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	for _, groupEl := range groupExt {
+		if err = cmd.VerifyBuildpackAPI(groupEl.Kind(), groupEl.String(), groupEl.API, logger); err != nil {
+			return err
+		}
+	}
+	extensions.Extensions = groupExt
+	fmt.Println("extensions.Extensions", extensions.Extensions)
+	return nil
+}
```
natalieparellano, line 83: It's too bad that the lifecycle doesn't expose a more usable library method for this... if possible I'd be open to improving this in the lifecycle vs copying logic over here

internal/build/helper.go
```diff
@@ -0,0 +1,187 @@
...
+	dockerfilePath := filepath.Join(path, kind, escapeID(extID), "Dockerfile")
+	if _, err := os.Stat(dockerfilePath); err != nil {
+		return nil, nil
+	}
+	return &buildpack.DockerfileInfo{
+		ExtensionID: extID,
+		Kind:        kind,
+		Path:        dockerfilePath,
+	}, nil
+}
```
natalieparellano, line 65: Could we make `func (e *Extender) dockerfileFor(kind, extID string)` in lifecycle/extender.go a static and public method to avoid duplicating this logic?

internal/build/helper.go
```diff
@@ -0,0 +1,187 @@
...
+		group.GroupExtensions[e].Optional = true
+	}
+	return group.GroupExtensions, err
+}
+
+func escapeID(id string) string {
+	return strings.ReplaceAll(id, "/", "_")
+}
+
+func SaveLayers(group *errgroup.Group, image v1.Image, origTopLayerHash string, dest string) (*time.Duration, error) {
```
natalieparellano, line 99: This (to the end of the file) can all probably be deleted if we're not doing run image extension, right?

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+	l.logger.Debugf("extensions.SetExtensions for build took %s", time.Since(time1))
+	time2 := time.Now()
+	dockerfiles, err := extensions.DockerFiles(DockerfileKindBuild, l.tmpDir, l.logger)
+	if err != nil {
+		return fmt.Errorf("getting %s.Dockerfiles: %w", DockerfileKindBuild, err)
+	}
+	l.logger.Debugf("extensions.DockerFiles for build took %s", time.Since(time2))
+	dockerapplytime := time.Now()
+	for _, dockerfile := range dockerfiles {
+		buildContext := archive.ReadDirAsTar(filepath.Dir(dockerfile.Path), "/", 0, 0, -1, true, false, defaultFilterFunc)
```
natalieparellano, line 763: The build context should be the application workspace, see here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/internal/extend/kaniko/dockerfile_applier.go#L88

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+		buildArguments := map[string]*string{}
+		if dockerfile.WithBase == "" {
+			buildArguments["base_image"] = &builderImageName
+		}
```
natalieparellano, line 767: Actually build.Dockerfiles can't change the base image, so we should be good here

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+	time2 := time.Now()
+	dockerfiles, err := extensions.DockerFiles(DockerfileKindBuild, l.tmpDir, l.logger)
+	if err != nil {
+		return fmt.Errorf("getting %s.Dockerfiles: %w", DockerfileKindBuild, err)
+	}
+	l.logger.Debugf("extensions.DockerFiles for build took %s", time.Since(time2))
+	dockerapplytime := time.Now()
+	for _, dockerfile := range dockerfiles {
+		buildContext := archive.ReadDirAsTar(filepath.Dir(dockerfile.Path), "/", 0, 0, -1, true, false, defaultFilterFunc)
+		buildArguments := map[string]*string{}
```
natalieparellano, line 764: We need to build these up from the contents of extend-config.toml, see here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/extender.go#L313-L317 and here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/extender.go#L428-L432

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+		if dockerfile.WithBase == "" {
+			buildArguments["base_image"] = &builderImageName
+		}
+		buildOptions := types.ImageBuildOptions{
+			Context:    buildContext,
+			Dockerfile: "Dockerfile",
+			Tags:       []string{"newbuilder-image"},
+			Remove:     true,
```
natalieparellano, line 772: What does this do?

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+		response, err := l.docker.ImageBuild(ctx, buildContext, buildOptions)
+		if err != nil {
+			return err
+		}
+		defer response.Body.Close()
+		_, err = io.Copy(os.Stdout, response.Body)
+		if err != nil {
+			return err
+		}
+		l.logger.Debugf("build response for the extend: %v", response)
```
natalieparellano, line 784: Can we also log warnings if the extension leaves the user as root and/or if the final user is different from the original user? See here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/extender.go#L354 and here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/extender.go#L361

internal/build/lifecycle_execution.go
```diff
@@ -706,6 +744,133 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
 	defer extend.Cleanup()
 	return extend.Run(ctx)
 }
+func (l *LifecycleExecution) ExtendBuildByDaemon(ctx context.Context) error {
```
natalieparellano, line 747: We'll need to write some unit tests for this method, which will involve using a mock for the daemon. This is done in the lifecycle (but mocking the kaniko "dockerfile applier") here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/extender_test.go#L74 and making assertions here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/extender_test.go#L233-L256

natalieparellano (commented):
@itsdarshankumar really awesome progress with this! I left some comments and suggestions. I think the next big unit of work will be to drive this out with unit tests.

natalieparellano:
Here is where we require that the builder (and run image) exist in a registry if image extensions are used: https://github.com/buildpacks/pack/blob/f1619bfa853596852a21bea5b9208ebd537c9f9c/pkg/client/build.go#L462-L464

I think this can go away when we extend the builder image with the daemon. We still need the run image to exist in a registry, but in the case that the run image was switched by extensions it will be pulled here: https://github.com/buildpacks/pack/blob/f1619bfa853596852a21bea5b9208ebd537c9f9c/internal/build/lifecycle_execution.go#L265-L267

There is an edge case where if the run image was NOT switched by extensions (but we are extending the original run image) and the remote copy does not match the daemon copy, the lifecycle will fail to find it by digest reference when initializing the app image here: https://github.com/buildpacks/lifecycle/blob/e06b69b5acb14e441659178b13de4a38e7a2e612/cmd/lifecycle/exporter.go#L238C25-L238C36 ...we might check what is the error message when this situation is encountered (so that we can make it more helpful) OR we could go ahead and pull the run image after the restore phase if extensions were used. I think there are arguments for either approach, so let's discuss further & bring it up with other interested parties.

internal/build/lifecycle_execution.go
natalieparellano: Rather than stream in the app directory I think we want to make it available as a volume, along with the layers dir and platform dir

internal/build/extend_build_test.go
```diff
@@ -0,0 +1,138 @@
...
+			opts.ClearCache = providedClearCache
+			opts.Publish = providedPublish
+		}
+
+		lifecycleOps = []func(*build.LifecycleOptions){configureDefaultTestLifecycle}
+
+	})
+
+	when("Extend Build Image By Docker", func() {
+		it("should extend build image using 1 extension", func() {
```
natalieparellano, line 68: Looks good :) let's also test 2 extensions if possible. Ideally we'd want to show that the state from the first extension is persisted to the second extension. This might be a bit tough with our mocks, but we could at least show that the image reference is threaded through.

internal/build/lifecycle_execution.go
itsdarshankumar: fixed in the latest commit

internal/build/lifecycle_execution.go
itsdarshankumar: added the tests :)

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/extend_build_test.go
itsdarshankumar: done in the latest commit

internal/build/helper.go
```diff
@@ -0,0 +1,187 @@
...
+	for i := range groupExt {
+		groupExt[i].Extension = true
+	}
+	for _, groupEl := range groupExt {
+		if err = cmd.VerifyBuildpackAPI(groupEl.Kind(), groupEl.String(), groupEl.API, logger); err != nil {
+			return err
+		}
+	}
+	extensions.Extensions = groupExt
+	fmt.Println("extensions.Extensions", extensions.Extensions)
```
natalieparellano, line 81: We should remove this or use the logger instead

internal/build/lifecycle_execution.go
natalieparellano: We also have `LifecycleExecution.layersVolume` ...I will get back to you about the platform directory :)

internal/build/phase_config_provider.go
natalieparellano: Update: maybe we can just get rid of this operation. `CNB_USER_ID` should still be set in the builder env, and the lifecycle will drop privileges to that user before running the build command.

natalieparellano: Or we could use `l.opts.Builder.UID()` etc

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/phase_config_provider.go
itsdarshankumar: done

internal/build/extend_build_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/heroku/color"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	"github.com/buildpacks/pack/pkg/archive"
+	"github.com/buildpacks/pack/pkg/logging"
+	h "github.com/buildpacks/pack/testhelpers"
+
+	mockdocker "github.com/buildpacks/pack/internal/build/mockdocker"
```
natalieparellano, line 28:
```suggestion
	"bytes"
	"context"
	"io"
	"path/filepath"
	"strconv"
	"strings"
	"testing"

	"github.com/apex/log"
	"github.com/docker/docker/api/types"
	"github.com/golang/mock/gomock"
	"github.com/heroku/color"
	"github.com/sclevine/spec"
	"github.com/sclevine/spec/report"

	"github.com/buildpacks/pack/pkg/archive"
	"github.com/buildpacks/pack/pkg/logging"
	h "github.com/buildpacks/pack/testhelpers"
	"github.com/buildpacks/pack/internal/build"
	"github.com/buildpacks/pack/internal/build/fakes"
	mockdocker "github.com/buildpacks/pack/internal/build/mockdocker"
```
This groups the imports by go standard library, external, and internal - you might need to additionally sort them

internal/build/helper.go
natalieparellano: Let's remove it for now...

internal/build/lifecycle_execution.go
```diff
@@ -3,24 +3,31 @@ package build
...
 	"context"
 	"fmt"
+	"io"
 	"math/rand"
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
+
+	"github.com/BurntSushi/toml"
```
natalieparellano, line 13: I couldn't leave a code suggestion as the lines didn't change, but could we group the imports here as well?

internal/build/lifecycle_execution.go
```diff
@@ -228,10 +235,24 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
 
 		group, _ := errgroup.WithContext(context.TODO())
 		if l.platformAPI.AtLeast("0.10") && l.hasExtensionsForBuild() {
-			group.Go(func() error {
-				l.logger.Info(style.Step("EXTENDING (BUILD)"))
-				return l.ExtendBuild(ctx, buildCache, phaseFactory)
-			})
+			if false {
```
natalieparellano, line 238: Can we leave a code comment here that we are planning to deprecate the older method?

internal/build/lifecycle_execution.go
natalieparellano: Did we add the warnings anywhere?

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +745,152 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+		l.logger.Debugf("build response for the extend: %v", response)
+		builderImageName = l.opts.BuilderImage + "-extended"
+	}
+	l.logger.Debugf("docker apply time: %v", time.Since(dockerapplytime))
+	l.logger.Debugf("Build extend time: %v", time.Since(extendtime))
+
+	return nil
+}
+
+func (l *LifecycleExecution) ExtendRunByDaemon(ctx context.Context, group *errgroup.Group, currentRunImage *string) error {
```
natalieparellano, line 811: This can be deleted

internal/build/lifecycle_execution.go
natalieparellano: This still needs to be removed I think

internal/build/lifecycle_execution.go
natalieparellano: Same here

internal/build/lifecycle_execution.go
natalieparellano: Same here

internal/build/phase_config_provider.go
```diff
@@ -244,6 +245,12 @@ func WithRoot() PhaseConfigProviderOperation {
 	}
 }
 
+func WithoutPrivilege(uid int, gid int) PhaseConfigProviderOperation {
```
natalieparellano, line 248:
```suggestion
func WithUser(uid int, gid int) PhaseConfigProviderOperation {
```
More descriptive and accurate, I think

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +745,152 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+			buildArguments[arg.Name] = &arg.Value
+		}
+		buildOptions := types.ImageBuildOptions{
+			Context:    buildContext,
+			Dockerfile: "Dockerfile",
+			Tags:       []string{extendedBuilderImageName},
+			Remove:     true,
+			BuildArgs:  buildArguments,
+		}
+		fmt.Println("buildContext: ", buildContext)
```
natalieparellano, line 791:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +745,152 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+		response, err := l.docker.ImageBuild(ctx, buildContext, buildOptions)
+		if err != nil {
+			return err
+		}
+		defer response.Body.Close()
+		fmt.Println("build response for the extend: ", response.Body)
```
natalieparellano, line 797:
```suggestion
```

internal/build/helper.go
```diff
@@ -0,0 +1,224 @@
...
```
natalieparellano, line 118:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -707,6 +745,152 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, buildCache Cache, ph
...
+	dockerapplytime := time.Now()
+	for _, dockerfile := range dockerfiles {
+		dockerfile.Args = append([]Arg{
+			{Name: argBuildID, Value: uuid.New().String()},
+			{Name: argUserID, Value: strconv.Itoa(l.opts.Builder.UID())},
+			{Name: argGroupID, Value: strconv.Itoa(l.opts.Builder.GID())},
+		}, dockerfile.Args...)
+		buildContext := archive.ReadDirAsTar(filepath.Dir(dockerfile.Info.Path), "/", 0, 0, -1, true, false, defaultFilterFunc)
+		buildArguments := map[string]*string{}
+		fmt.Println("builderImageName: ", builderImageName)
```
natalieparellano, line 778:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -245,14 +266,25 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
...
 			}
 		}
-
+		var start time.Time
 		if l.platformAPI.AtLeast("0.12") && l.hasExtensionsForRun() {
-			group.Go(func() error {
-				l.logger.Info(style.Step("EXTENDING (RUN)"))
-				return l.ExtendRun(ctx, buildCache, phaseFactory)
-			})
+			if l.opts.Publish {
```
natalieparellano, line 271:
```suggestion
```

internal/build/lifecycle_execution.go
natalieparellano, line 286:
```suggestion
```

natalieparellano (commented):
Thanks for adding the tests @itsdarshankumar :) I added a few comments to hopefully clean it up a little before we can undraft the PR

internal/build/lifecycle_execution.go
itsdarshankumar: not yet, then we need to inspect the image as well

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/helper.go
itsdarshankumar: yes, it was for run-extension, removed it now

internal/build/extend_build_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+		it("should extend build image using multiple extension", func() {
+			// set tmp directory
+			tmpDir = "./testdata/fake-tmp/build-extension/multi"
+			lifecycle = getTestLifecycleExec(t, true, tmpDir, mockDockerClient, lifecycleOps...)
+			mockResponse := types.ImageBuildResponse{
+				Body:   io.NopCloser(strings.NewReader("mock-build-response-body")),
+				OSType: "linux",
+			}
+			mockDockerClient.EXPECT().ImageBuild(gomock.Any(), gomock.Any(), gomock.Any()).Return(mockResponse, nil).Times(2)
```
natalieparellano, line 113: Ideally here we'd also want to check that the options provided to the second Dockerfile are correct. `Dockerfile` and `BuildArgs` seem to be the most important to test.

internal/build/lifecycle_execution.go
```diff
@@ -256,9 +261,17 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
 
 		group, _ := errgroup.WithContext(context.TODO())
 		if l.platformAPI.AtLeast("0.10") && l.hasExtensionsForBuild() {
+			/*
+				[RFC #0105] - As decided, Pack should support build image extension with Docker #1623. We removed the previous implementation that was using kaniko in the extend lifecycle phase and shifted the implementation to use docker daemon to extend the build Image. As pack already has access to a daemon, it can apply the dockerfiles directly, saving the extended build base image in the daemon. Thus it will not need to use the extender phase of lifecycle. Additionally it dropped the requirement that the image being extended must be published to a registry. This implementation resulted us to have build Extension Improved by 87.3578% wrt kaniko implementation with caching and 20.5567% wrt kaniko implementation without caching.
```
natalieparellano, line 265: Nice comment :)

internal/build/extend_build_test.go
```diff
@@ -0,0 +1,160 @@
...
+			opts.Publish = providedPublish
+		}
+
+		lifecycleOps = []func(*build.LifecycleOptions){configureDefaultTestLifecycle}
+	})
+
+	when("Extend Build Image By Docker", func() {
+		it("should extend build image using 1 extension", func() {
+			// set tmp directory
+			tmpDir = "./testdata/fake-tmp/build-extension/single"
```
natalieparellano, line 76: nit: `filepath.Join` would be safer here for portability. But, we should probably also just skip these tests on Windows as extensions aren't supported there.

internal/build/lifecycle_execution.go
```diff
@@ -669,6 +683,8 @@ func (l *LifecycleExecution) Build(ctx context.Context, phaseFactory PhaseFactor
 		WithNetwork(l.opts.Network),
 		WithBinds(l.opts.Volumes...),
 		WithFlags(flags...),
+		If((!l.opts.Publish && l.hasExtensionsForBuild()), WithImage(l.opts.BuilderImage+"-extended")),
+		If((!l.opts.Publish && l.hasExtensionsForBuild()), WithUser(l.opts.Builder.UID(), l.opts.Builder.GID())),
```
natalieparellano, line 687: Aren't we removing the publish restriction?

internal/build/lifecycle_execution.go
```diff
@@ -720,6 +740,67 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, kanikoCache Cache, p
...
+	* Reading the dockerfiles that were generated during the `generate` phase and also parsing the Arguments given by the user from
+	`extend-config.toml`.
+	* Using ImageBuild method of docker API client to extend the Image and save it to the daemon.
+	* Invoking Build phase of lifecycle by creating a container from the extended Image and dropping the privileges.
+
+*/
+
+func (l *LifecycleExecution) ExtendBuildByDaemon(ctx context.Context) error {
+	builderImageName := l.opts.BuilderImage
+	extendedBuilderImageName := l.opts.BuilderImage + "-extended"
```
natalieparellano, line 761: I wonder if it might be better to use a UUID here, to avoid tagging over another image that might already be called `<builder>-extended`. However that might cause the number of images to in the daemon to balloon... I think we can leave it for now unless others feel strongly about it.

internal/build/lifecycle_execution.go
```diff
@@ -720,6 +740,67 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, kanikoCache Cache, p
...
+	defaultFilterFunc := func(file string) bool { return true }
```
natalieparellano, line 762: nit: can we make this an anonymous function, or move it closer to where it's being used, so that it's easier for the reader?

internal/build/lifecycle_execution.go
```diff
@@ -720,6 +740,67 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, kanikoCache Cache, p
...
+	extensions.SetExtensions(l.tmpDir, l.logger)
+	dockerfiles, err := extensions.DockerFiles(DockerfileKindBuild, l.tmpDir, l.logger)
+	if err != nil {
+		return fmt.Errorf("getting %s.Dockerfiles: %w", DockerfileKindBuild, err)
+	}
+	for _, dockerfile := range dockerfiles {
+		dockerfile.Args = append([]Arg{
+			{Name: argBuildID, Value: uuid.New().String()},
+			{Name: argUserID, Value: strconv.Itoa(l.opts.Builder.UID())},
+			{Name: argGroupID, Value: strconv.Itoa(l.opts.Builder.GID())},
```
natalieparellano, line 773: Technically, a Dockerfile could change the user on the extended image, and we want to persist that user for the next Dockerfile. Can you get the config for the immediate extended image within this loop?

internal/build/lifecycle_execution.go
```diff
@@ -720,6 +740,67 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, kanikoCache Cache, p
...
+		}, dockerfile.Args...)
+		buildContext := archive.ReadDirAsTar(filepath.Dir(dockerfile.Info.Path), "/", 0, 0, -1, true, false, defaultFilterFunc)
+		buildArguments := map[string]*string{}
+		buildArguments["base_image"] = &builderImageName
+		for i := range dockerfile.Args {
+			arg := &dockerfile.Args[i]
+			buildArguments[arg.Name] = &arg.Value
+		}
```
natalieparellano, line 781:
```suggestion
		for _, arg := range dockerfile.Args {
			buildArguments[arg.Name] = &arg.Value
		}
```
This might read better

natalieparellano:
Given that we're relying on mocks here, it might be prudent to add an acceptance test along the lines of https://github.com/buildpacks/pack/blob/c38f7da1b43d5b142d1e870f3d17da93b454c64f/acceptance/acceptance_test.go#L855

This test relies on the "read env" extension detecting - we could update https://github.com/buildpacks/pack/blob/c38f7da1b43d5b142d1e870f3d17da93b454c64f/acceptance/testdata/mock_buildpacks/0.2/read-env-extension/bin/generate to print out the build args to show at a minimum that that's working

internal/build/lifecycle_execution.go
```diff
@@ -720,6 +740,67 @@ func (l *LifecycleExecution) ExtendRun(ctx context.Context, kanikoCache Cache, p
...
+		buildOptions := types.ImageBuildOptions{
+			Context:    buildContext,
+			Dockerfile: "Dockerfile",
```
natalieparellano, line 784: Must the dockerfile reside within the build context? For example,

`$  docker build -t test -f ./Dockerfile ../other-dir/`
`$  docker build -t test -f ../other-dir/Dockerfile .`

both work when running the CLI.

natalieparellano: Ah nevermind, it appears that `docker build` copies the Dockerfile into the context :)

internal/build/lifecycle_execution.go
itsdarshankumar: then we need to inspect the image!

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/lifecycle_execution.go
itsdarshankumar: oops, forgot to change here, done

internal/build/extend_build_test.go
itsdarshankumar: done

internal/build/lifecycle_execution.go
itsdarshankumar: no, actually i did this bcs ideally we shouldn't export a loop variable's pointer so made a copy of the loop variable

internal/build/lifecycle_execution.go
itsdarshankumar: done

internal/build/helper.go
itsdarshankumar: done

internal/build/extend_build_test.go
jjbustamante: @itsdarshankumar please take a look at this suggestion, I think is very valid to make test useful

internal/build/helper.go
```diff
@@ -0,0 +1,158 @@
+package build
```
dlion, line 1: Can we have some tests for this file and these helper functions?

itsdarshankumar: yess sure:)

internal/build/lifecycle_execution.go
itsdarshankumar: will add in the help!

internal/build/extend_build_test.go
itsdarshankumar: yes eyes on it :)

internal/build/lifecycle_execution.go
natalieparellano: I think that makes sense - we do this in the lifecycle here: https://github.com/buildpacks/lifecycle/blob/aeb177ca468cc814ab47fac0cb878497495b5819/extender.go#L329

internal/build/lifecycle_execution.go
natalieparellano: I think that makes sense - we do this in the lifecycle here: https://github.com/buildpacks/lifecycle/blob/aeb177ca468cc814ab47fac0cb878497495b5819/extender.go#L329

internal/build/extend_build_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/golang/mock/gomock"
+	"github.com/heroku/color"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	"github.com/buildpacks/pack/pkg/logging"
+	h "github.com/buildpacks/pack/testhelpers"
+
+	mockdocker "github.com/buildpacks/pack/internal/build/mockdocker"
```
natalieparellano, line 30: nit: can we group imports here?

internal/build/helper.go
```diff
@@ -0,0 +1,158 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/buildpacks/lifecycle/buildpack"
+	"github.com/buildpacks/lifecycle/cmd"
+
+	"github.com/buildpacks/pack/pkg/archive"
+
+	"github.com/buildpacks/pack/pkg/logging"
```
natalieparellano, line 18:
```suggestion
	"github.com/buildpacks/pack/pkg/archive"
	"github.com/buildpacks/pack/pkg/logging"
```

internal/build/lifecycle_execution.go
```diff
@@ -3,16 +3,20 @@ package build
...
 	"context"
 	"fmt"
+	"io"
 	"math/rand"
 	"os"
 	"path/filepath"
 	"strconv"
 
 	"github.com/BurntSushi/toml"
+
```
natalieparellano, line 13:
```suggestion
```

internal/build/lifecycle_execution.go
```diff
@@ -433,6 +444,8 @@ func (l *LifecycleExecution) Detect(ctx context.Context, phaseFactory PhaseFacto
 			CopyOutToMaybe(filepath.Join(l.mountPaths.layersDir(), "analyzed.toml"), l.tmpDir))),
 		If(l.hasExtensions(), WithPostContainerRunOperations(
 			CopyOutToMaybe(filepath.Join(l.mountPaths.layersDir(), "generated", "build"), l.tmpDir))),
+		If(l.hasExtensions(), WithPostContainerRunOperations(
+			CopyOutToMaybe(filepath.Join(l.mountPaths.layersDir(), "group.toml"), l.tmpDir))),
```
natalieparellano, line 448: I'm not able to leave a code suggestion, but lines 483-495 could become something like 

```
	kanikoCacheBindOp := NullOp()
	if l.platformAPI.AtLeast("0.12") {
		if l.runImageChanged() || l.hasExtensionsForRun() {
			registryImages = append(registryImages, l.runImageAfterExtensions())
		}
		if l.hasExtensionsForRun() {
			kanikoCacheBindOp = WithBinds(fmt.Sprintf("%s:%s", kanikoCache.Name(), l.mountPaths.kanikoCacheDir()))
		}
	}
```

We no longer need to provide `-build-image` to the restorer, because we don't need to fetch any information about it to provide to kaniko (since we're not using kaniko). This eliminates the requirement that the builder image must reside in a registry.

internal/build/lifecycle_execution.go
natalieparellano: I'm not totally sure where this would go... maybe a warning when we begin to extend the builder?

natalieparellano:
Some early user acceptance:

When I run a build with multiple builder image extensions:
✅ Each build.Dockerfile is processed
✅ build args from extend-config.toml are provided to each build.Dockerfile
✅ The intermediate extended image from application of the first build.Dockerfile is used as the base image for applying the second build.Dockerfile
❗ If the first build.Dockerfile mutates the user, the user_id and group_id build args are updated for the second build.Dockerfile
❗ If an intermediate build.Dockerfile left the user as root, we see a warning
❗ If the final build.Dockerfile left the user as root, we see a warning
✅ Second and later builds use caching
❗ The context is /workspace and we see the contents of /workspace
  * When I run pwd from a build.Dockerfile I see `/layers` and when I run `ls /workspace` I see empty

internal/build/helper.go
itsdarshankumar: hey @dlion added the tests, sorry for getting late!

internal/build/extend_build_test.go
natalieparellano: To be fixed in https://github.com/itsdarshankumar/pack/pull/1

internal/build/lifecycle_execution.go
natalieparellano: To be fixed in https://github.com/itsdarshankumar/pack/pull/1

internal/build/lifecycle_execution.go
natalieparellano: To be fixed in https://github.com/itsdarshankumar/pack/pull/1

internal/build/lifecycle_execution.go
natalieparellano: Is this still outstanding?

pkg/archive/archive.go
```diff
@@ -319,6 +319,33 @@ func WriteZipToTar(tw TarWriter, srcZip, basePath string, uid, gid int, mode int
 	return nil
 }
 
+func WriteFileToTar(tw TarWriter, srcFile, destPath string, uid, gid int, mode int64, normalizeModTime bool) error {
```
natalieparellano, line 322: Is it possible to write a test for this?

internal/build/lifecycle_execution.go
natalieparellano: Is this still outstanding? Some kind of messaging would surely help.

internal/build/helper.go
```diff
@@ -0,0 +1,158 @@
...
+func (dockerfile *DockerfileInfo) CreateBuildContext(path string) (io.Reader, error) {
+	defaultFilterFunc := func(file string) bool { return true }
+	buf := new(bytes.Buffer)
+	tarWriter := tar.NewWriter(buf)
+	var completeErr error
+
+	defer func() {
+		if err := tarWriter.Close(); err != nil {
+			fmt.Println("Error closing tar writer:", err)
+			completeErr = archive.AggregateError(completeErr, err)
```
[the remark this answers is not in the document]
itsdarshankumar: We need to call the `archive.AggregateError` so that any error during the execution of the deferred function (`tarWriter.Close()`) will affect the final value of completeErr and gives the error in the return statement.

internal/build/helper.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (dockerfile *DockerfileInfo) CreateBuildContext(path string) (io.Reader, error) {
+	defaultFilterFunc := func(file string) bool { return true }
+	buf := new(bytes.Buffer)
+	tarWriter := tar.NewWriter(buf)
+	var completeErr error
+
+	defer func() {
+		if err := tarWriter.Close(); err != nil {
+			fmt.Println("Error closing tar writer:", err)
```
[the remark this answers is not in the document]
itsdarshankumar: done!

pkg/archive/archive.go
itsdarshankumar: done!

internal/build/lifecycle_execution.go
itsdarshankumar: args for `ImageBuildOptions` method of docker to make the image

internal/build/extend_build_test.go
itsdarshankumar: done!!

internal/build/lifecycle_execution.go
itsdarshankumar: done!

internal/build/lifecycle_execution.go
itsdarshankumar: added earlier!

internal/build/extend_build_test.go
itsdarshankumar: done!!

internal/build/helper.go
itsdarshankumar: done!!

internal/build/lifecycle_execution.go
itsdarshankumar: added!

internal/build/lifecycle_execution.go
itsdarshankumar: done!!
added a warning right after the start of phase

itsdarshankumar: > I think we should document this in some place, maybe in the command help?

also added this too! would be surely helpful

internal/build/lifecycle_execution.go
itsdarshankumar: done in latest commit

itsdarshankumar: done!

internal/build/lifecycle_execution.go
itsdarshankumar: done!

itsdarshankumar:
> mediate extended image from a

4,5,6 done!

natalieparellano:
@itsdarshankumar thank you for making all the updates to your branch. I tried to re-run the test mentioned in https://github.com/buildpacks/pack/pull/1791#issuecomment-1714446276, but I'm still getting the error: `[extender (build)] {"errorDetail":{"message":"COPY failed: file not found in build context or excluded by .dockerignore: stat mvnw: file does not exist"},"error":"COPY failed: file not found in build context or excluded by .dockerignore: stat mvnw: file does not exist"}`. When I rerun the command with released pack, the build succeeds.

Before we proceed further, it might be prudent to check with @loewenstein @c0d1ngm0nk3y and others to see if there is still appetite for this change, given forthcoming changes to the extend phase outlined in https://github.com/buildpacks/spec/pull/384. To summarize, with https://github.com/buildpacks/pack/pull/1791 we'd use the daemon to extend __builder__ images only (effectively chaining `docker build` together using the generated `build.Dockerfile`s). This gives us a nice performance boost but introduces code complexity because we cannot extend run images this way as it makes performance worse (due to time spent saving extended layers from the daemon). If we merge this PR the changes from https://github.com/buildpacks/spec/pull/384 might be more complicated to implement. Another possible consideration is that a future integration with BuildKit (if we ever get there) might make relying on the daemon for image extension a bit simpler. Please opine!

loewenstein:
I will leave a thorough technical comment up to @pbusko @c0d1ngm0nk3y @modulo11 @phil9909 - but I am wondering how a `pack` optimisation could make a lifecycle change harder?

natalieparellano:
@loewenstein it wouldn't make the lifecycle change harder - but, some things would need to be re-implemented in pack since we aren't relying on the lifecycle for build image extension.

loewenstein:
> @loewenstein it wouldn't make the lifecycle change harder - but, some things would need to be re-implemented in pack since we aren't relying on the lifecycle for build image extension.

FWIW https://github.com/buildpacks/lifecycle/pull/1276 is said lifecycle implementation ;)

natalieparellano:
@itsdarshankumar thank you for your work here. Given the evolution of the "Dockerfiles" feature and the introduction of configurable context directories (https://github.com/buildpacks/spec/pull/384), it may no longer by prudent to have a separate implementation of build image extension that uses the daemon, as this would require changes to be implemented twice (once in the lifecycle for run image extension and non-pack platforms, and once in pack for build image extension). Also having build image extension (uses daemon) and run image extension (uses lifecycle) behave differently in pack might be confusing, especially for newer contributors.

However the work in this PR is still extremely valuable as it gives us an indication of the performance benefits we might see in relying on the daemon for image extension vs kaniko. I suggest we keep on eye on potential upcoming changes to the daemon (such as https://github.com/moby/moby/issues/44369) that might run image extension with the daemon more performant. If build image extension and run image extension could both use the daemon, that would make things simpler in pack. That doesn't solve the need for changes to be implemented twice (lifecycle, pack) but we can still learn from this PR how to make that easier, if we decide the performance improvement is worth the effort.

PR #1795, opened 2023-06-07
only add the stack arg to lifecycle for older platform APIs

joe-kimmel-vmw:
## Summary
only adds the stack arg to lifecycle for older platform APIs

## Output
hard to say but it sure doesn't add the stack arg when invoking lifecyle if the platform API is older

#### Before

it would add the stack arg to lifecycle indiscriminately 

#### After
only adds the stack arg to lifecycle for older platform APIs

## Documentation

    - [ ] Yes, see #___
    - [ x ] No

## Related

Resolves #___


jjbustamante:
Is it possible to add a test case?

natalieparellano:
I'm gonna pull this into my feature branch & add a test

PR #1801, opened 2023-06-14
Fix log message when buildpack doesn't satisfy target constraints for builder

natalieparellano:
To recreate: 

`pack builder create test-log --config ./some-builder.toml`

Where `some-builder.toml` reference base images that are linux/amd64 and also references a buildpack that has `[[targets]]\n os = "windows"` in buildpack.toml.

For the [lifecycle](https://github.com/buildpacks/lifecycle/pull/1122/files#diff-20ea3c405db84cf96fba3c11413e94a80b149499efd186494fedbdfb12a5feb4R268-R269) we decided that printing as JSON would be clearer than any custom `String()` method we could provide.

#### Before

Error message looks like: `ERROR: validating buildpacks: buildpack samples/hello-extensions@0.0.1 does not support target: (linux amd64, @)`

#### After

Error message looks like: `ERROR: validating buildpacks: unable to satisfy target os/arch constraints; build image: {"os":"linux","arch":"amd64","distribution":{}}, buildpack samples/hello-extensions@0.0.1: [{"os":"windows","arch":""}]`


dfreilich (approved):
I like the maybe 🤣

dfreilich:
I'll probably need to approve again after you fix thx pipeline

jkutner (approved):

natalieparellano:
Ah well, I fixed the test expectation. This should go green now :)

jjbustamante (approved):

PR #1803, opened 2023-06-15
Fix small issue with deprecated BOM display

natalieparellano:
Extensions don't output legacy BOMs so including them in the BOM display is a bit misleading.


jjbustamante (approved):
It is always nice to remove lines of code! :)

jkutner (approved):

PR #1810, opened 2023-06-23
If includeRoot by default the permission value is always set to 0777

### LINKED ISSUE #1800: Fix from https://github.com/buildpacks/pack/pull/1672 is breaking change
### Summary
This was discussed in Working Group and summarized in Slack [here](https://cloud-native.slack.com/archives/C0331B61A1Y/p1686236551802739) and [here](https://cloud-native.slack.com/archives/C0331B61A1Y/p1685728684812679)

At build time, we copy the application (source) directory into the build container using `archive.WriteDirToTar`. https://github.com/buildpacks/pack/pull/1672 fixes weird permissions on this directory, but it's also a breaking change: 
* if the build-time user and the runtime user are different (as recommended by the spec), and 
* the host umask at build-time is less permissive resulting in e.g., `0755` permissions for newly created directories, and 
* the runtime user tries to write to the application directory, THEN

applications can fail to start. This can be remedied (e.g., by setting the `umask` at build time to something like `0002`) but this could be a breaking change for some users of `pack`.

Proposed fix:
* Partially revert https://github.com/buildpacks/pack/pull/1672 so that we don't create the weird permissions when `includeRoot` is provided to `archive.WriteDirToTar`, but instead of using default permissions for newly created directories on the host, simply use `0777` for the application directory as the least breaking change. See linked threads for other fixes discussed and why we feel that this one is the "best"

dlion:
## Summary
This PR solves the issue https://github.com/buildpacks/pack/issues/1800
The logic behind it is:
Whenever we pass `mode = -1` and `includeRoot = true` to the method `archive.WriteDirToTar` it sets the permissions to `0777`.
One thing that is worth noticing is that we *ALWAYS* pass `mode=-1` and `includeRoot=true` which implies we always set `0777` for the rootDir
Codebase reference:
*  `container_ops.createReader`
* `lifecycle_execution.Create:370`
* `lifecycle_execution.Detect:430`

#### Before

```
drwxr-xr-x    2 123      456 (.*) some-vol
```

#### After

```
drwxrwxrwx    2 123      456 (.*) some-vol
```

## Documentation

    - [ ] Yes, see #___
    - [X] No

## Related

Resolves #1800

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -174,10 +174,13 @@ func WriteDirToTar(tw TarWriter, srcDir, basePath string, uid, gid int, mode int
 			Name:     basePath,
 			Mode:     mode,
 		}
-		if rootHeader.Mode == -1 {
-			rootHeader.Mode = int64(fs.ModePerm &^ Umask)
+
+		if mode == -1 {
+			mode = int64(fs.ModePerm)
 		}
+
 		finalizeHeader(rootHeader, uid, gid, mode, normalizeModTime)
+
 		if err := tw.WriteHeader(rootHeader); err != nil {
 			return err
 		}
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -169,19 +169,9 @@ func ReadTarEntry(rc io.Reader, entryPath string) (*tar.Header, []byte, error) {
 // contents will be placed. The includeRoot param sets the permissions and metadata on the root file.
 func WriteDirToTar(tw TarWriter, srcDir, basePath string, uid, gid int, mode int64, normalizeModTime, includeRoot bool, fileFilter func(string) bool) error {
 	if includeRoot {
-		rootHeader := &tar.Header{
-			Typeflag: tar.TypeDir,
-			Name:     basePath,
-			Mode:     mode,
-		}
-
-		if mode == -1 {
-			mode = int64(fs.ModePerm)
-		}
-
-		finalizeHeader(rootHeader, uid, gid, mode, normalizeModTime)
-
-		if err := tw.WriteHeader(rootHeader); err != nil {
+		mode = modePermIfNegativeMode(mode)
+		err := writeRootHeader(tw, basePath, mode, uid, gid, normalizeModTime)
+		if err != nil {
 			return err
 		}
 	}
@@ -197,31 +187,9 @@ func WriteDirToTar(tw TarWriter, srcDir, basePath string, uid, gid int, mode int
 				return nil
 			}
 		}
+
 		if err != nil {
 			return err
-		}
-
-		if fi.Mode()&os.ModeSocket != 0 {
-			return nil
-		}
-
-		var header *tar.Header
-		if fi.Mode()&os.ModeSymlink != 0 {
-			target, err := os.Readlink(file)
-			if err != nil {
-				return err
-			}
-
-			// Ensure that symlinks have Linux link names, independent of source OS
-			header, err = tar.FileInfoHeader(fi, filepath.ToSlash(target))
-			if err != nil {
-				return err
-			}
-		} else {
-			header, err = tar.FileInfoHeader(fi, fi.Name())
-			if err != nil {
-				return err
-			}
 		}
 
 		if relPath == "" {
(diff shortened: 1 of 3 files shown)


pkg/archive/archive_test.go
```diff
@@ -214,14 +213,14 @@ func testArchive(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 
-		when("includeRoot is true", func() {
-			it("sets metadata on base dest file", func() {
+		when("mode is set to 0755", func() {
```
natalieparellano, line 216: Added test, right?

pkg/archive/archive.go
```diff
@@ -169,16 +169,9 @@ func ReadTarEntry(rc io.Reader, entryPath string) (*tar.Header, []byte, error) {
...
-			Typeflag: tar.TypeDir,
-			Name:     basePath,
-			Mode:     mode,
-		}
-		if rootHeader.Mode == -1 {
-			rootHeader.Mode = int64(fs.ModePerm &^ Umask)
-		}
-		finalizeHeader(rootHeader, uid, gid, mode, normalizeModTime)
-		if err := tw.WriteHeader(rootHeader); err != nil {
+		mode = modePermIfNegativeMode(mode)
```
natalieparellano, line 172:
```suggestion
		mode := modePermIfNegativeMode(mode)
```
(otherwise `mode` will be 0777 for child entries also... I think)

pkg/archive/archive_test.go
```diff
@@ -256,12 +259,34 @@ func testArchive(t *testing.T, when spec.G, it spec.S) {
...
+
+					tr := tar.NewReader(file)
+
+					verify := h.NewTarVerifier(t, tr, 1234, 2345)
+					verify.NextDirectory("/nested/dir/dir-in-archive", 0777)
+					verify.NextFile("/nested/dir/dir-in-archive/some-file.txt", "some-content", 0777)
+					verify.NextDirectory("/nested/dir/dir-in-archive/sub-dir", 0777)
+					if runtime.GOOS != "windows" {
+						verify.NextSymLink("/nested/dir/dir-in-archive/sub-dir/link-file", "../some-file.txt")
+					}
```
natalieparellano, line 288: We should keep the old assertions, as the child entries should still have their original permissions, it's just the root directory that should get 0777

pkg/archive/archive.go
```diff
@@ -231,14 +202,28 @@ func WriteDirToTar(tw TarWriter, srcDir, basePath string, uid, gid int, mode int
 			return nil
 		}
 
-		header.Name = filepath.ToSlash(filepath.Join(basePath, relPath))
-		finalizeHeader(header, uid, gid, mode, normalizeModTime)
+		if hasModeSocket(fi) != 0 {
```
natalieparellano, line 205: Nice use of helper functions to make the code more readable :)

natalieparellano (commented):
Thanks for the PR @dlion! I added a few comments...

pkg/archive/archive_test.go
dlion: Yeah, We had just the test for the 0777 but since we were defaulting to it when `includeRoot` I also wanted to prove/document through tests that we are also capable to set the mod to `0755` if `includeRoot` is false and we pass a different mode than `0777` and `-1`

pkg/archive/archive_test.go
dlion: I have to change the implementation because I assumed we wanted to give the children those permissions too.
Thank you, Natalie, I'll fix it 😁

dlion: Now it should have been fixed :)

pkg/archive/archive.go
dlion: as in the other comment, I assumed we wanted to do differently, fixed it, thanks! 😄

dlion:
I just fixed it 😄 
Thank you for the feedback @natalieparellano ! 🚀

pkg/archive/archive_test.go
```diff
@@ -233,18 +232,22 @@ func testArchive(t *testing.T, when spec.G, it spec.S) {
...
+				if runtime.GOOS != "windows" {
+					verify.NextSymLink("/nested/dir/dir-in-archive/sub-dir/link-file", "../some-file.txt")
+				}
 			})
 		})
 
-		when("mode is set to -1", func() {
-			it("writes a tar to the dest dir with preexisting file mode", func() {
+		when("includeRoot is true", func() {
+			it("sets metadata on base dest file", func() {
```
natalieparellano, line 244:
```suggestion
			it("writes a tar to the root dir with the provided mode", func() {
```

natalieparellano (approved):
Looks good to me! Thanks for the PR @dlion!

natalieparellano:
@dfreilich @hone @jkutner could you give this one a review?

PR #1815, opened 2023-06-28
Fix: when running restore, provide creds for run image if there are r…

natalieparellano:
…un image extensions

Otherwise the restorer could fail to find the run image, resulting in unexpected behavior

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


natalieparellano:
Will undraft once tests are added

natalieparellano:
See https://github.com/buildpacks/lifecycle/pull/1134#issuecomment-1614781077 for further evidence that this fix does what it is supposed to

jjbustamante (approved):
Nice.

joe-kimmel-vmw (approved):
LGTM

jkutner (approved):

PR #1822, opened 2023-07-03
Merge the same TOML file writing logic

ESWZY:
## Summary

Seeing that there is the same TOML file writing logic here, for maintainability considerations, they can be merged together.

## Output

No change.

## Documentation

    - [x] No


internal/build/container_ops.go
```diff
@@ -207,92 +207,60 @@ func findMount(info types.ContainerJSON, dst string) (types.MountPoint, error) {
...
 	}
 }
 
 // WriteStackToml writes a `stack.toml` based on the StackMetadata provided to the destination path.
 func WriteStackToml(dstPath string, stack builder.StackMetadata, os string) ContainerOperation {
 	return func(ctrClient DockerClient, ctx context.Context, containerID string, stdout, stderr io.Writer) error {
-		buf := &bytes.Buffer{}
-		err := toml.NewEncoder(buf).Encode(stack)
-		if err != nil {
-			return errors.Wrap(err, "marshaling stack metadata")
```
joe-kimmel-vmw, line 245: imo the loss of specific error messages here and above is undesirable, but we could fix this by passing in one additional parameter to `writeToml` something like `dataDescriptor string` and it could then be given the value `stack metadata` on line 253 below,  and appropriate values elsewhere.   What do you think?

internal/build/container_ops.go
```diff
@@ -207,92 +207,60 @@ func findMount(info types.ContainerJSON, dst string) (types.MountPoint, error) {
...
+		dirName := paths.WindowsDir(dstPath)
+		return copyDirWindows(ctx, ctrClient, containerID, reader, dirName, stdout, stderr)
+	}
+
+	return ctrClient.CopyToContainer(ctx, containerID, "/", reader, types.CopyToContainerOptions{})
+}
 
-		return ctrClient.CopyToContainer(ctx, containerID, "/", reader, types.CopyToContainerOptions{})
+// WriteToml writes a `data.toml` for test only.
+func WriteToml(dstPath string, data interface{}, os string) ContainerOperation {
```
joe-kimmel-vmw, line 237: I think rather than introduce a new public API function only for the tests, it would be better to test one of the existing public functions

joe-kimmel-vmw (commented):
Thanks for contributing @ESWZY ! 

It looks like your intention is to "DRY" up this part of the code by merging redundant codepaths?   I think that's a fine idea overall though I'll defer to the actual pack maintainers' judgement.   

I've made a couple of suggestions.  I think as a project we really do want our error messages to be as helpful and informative as possible,  so maintaining the existing specificity of the error messages is probably desired.

ESWZY:
> 

Thanks for your generous suggestions, I will consider it!

joe-kimmel-vmw (approved):
LGTM  but it'll have to wait for a pack maintainer review :)

jkutner (approved):

PR #1826, opened 2023-07-05
Bump lifecycle library version

natalieparellano:


internal/build/lifecycle_execution.go
```diff
@@ -860,11 +851,11 @@ func (l *LifecycleExecution) hasExtensionsForRun() bool {
 }
 
 func (l *LifecycleExecution) runImageAfterExtensions() string {
-	var amd analyzedMD
+	var amd files.Analyzed
 	if _, err := toml.DecodeFile(filepath.Join(l.tmpDir, "analyzed.toml"), &amd); err != nil {
 		return l.opts.RunImage
 	}
-	if amd.RunImage == nil {
+	if amd.RunImage == nil || amd.RunImage.Image == "" {
```
natalieparellano, line 858: It looks like `github.com/BurntSushi/toml` had a bug before where empty structs without an `omitempty` annotation weren't getting written out to TOML files, so in our fixture file `run-image` was not present and `amd.RunImage` actually was `nil`. This change fixes the test no matter which version of the library we use.

cc @jkutner

acceptance/acceptance_test.go
```diff
@@ -2643,6 +2643,7 @@ include = [ "*.jar", "media/mountain.jpg", "/media/person.png", ]
 								"rebase", repoName,
 								"--run-image", runAfter,
 								"--pull-policy", "never",
+								"--force",
```
natalieparellano, line 2646: Worth noting, the need for `--force` is a breaking change. We should call this out in the release notes.

joe-kimmel-vmw (approved):
LGTM

internal/build/lifecycle_execution.go
jkutner: nice! thanks

jkutner (approved):

PR #1852, opened 2023-07-31
Add -daemon to restorer for newer platform

natalieparellano:
See https://github.com/buildpacks/lifecycle/pull/1168 and https://github.com/buildpacks/spec/pull/347#issuecomment-1658451549


natalieparellano:
The acceptance tests are going to fail until there is a new lifecycle RC

jkutner (approved):

PR #1866, opened 2023-08-09
Fix the rebasable flag logic in case of missing label

dlion:
## Summary
With this PR I fixed the problem with the `rebasable` label being set as a `false` if the `io.buildpacks.rebasable` didn't exist.

#### Before

> Rebasable: false

#### After

> Rebasable: true

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


natalieparellano (approved):
Thanks @dlion!!!

Maintainers, could we please pull this into 0.30.0?

natalieparellano:
I think we could override Codecov in this case as multiple new tests were added!

jkutner (approved):

PR #1874, opened 2023-08-24
Remove legacy beta release message

dlion:
## Summary
This PR remove the legacy beta warning message from the release page

#### Before

> This is a beta release of the Cloud Native Buildpack local CLI. This platform implementation should be relatively stable and reliable, but breaking changes in the underlying [specification](https://github.com/buildpack/spec) may be implemented without notice.

## Documentation

    - [ ] Yes, see #___
    - [X] No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -280,8 +280,6 @@ jobs:
           prerelease: true
           files: pack-v${{ env.PACK_VERSION }}-*
           body: |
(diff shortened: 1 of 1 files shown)


natalieparellano (approved):
@dlion thanks for this!

jjbustamante (approved):
It seems good to me, the message is very old and it can confuse users.

jkutner (approved):

Issue #1886, opened 2023-09-05
"pack builder create" runs out of memory when using pack 0.30


bruthphorce:
### Summary

We are running a `pack builder create` in our CICD pipeline on runners that have 7GB of memory available. One of the buildpacks included in the builder is 4GB in size (because of GPU drivers necessary to be in there). Up until pack 0.29, this was working fine.

However, when testing the upgrade to pack 0.30, the `pack builder create` command fails with exit code 137. The last logs that are written:
```
Creating builder with the following buildpacks:
-> <all expected buildpacks are listed>
```
The process crashes approximately 30 seconds after this.

When we run the same process locally we initially didn't encounter any problems with pack 0.30, it seems to work fine given that enough memory is available. As a rough guess, I would say the process fails once the size of the buildpacks exceeds the available memory divided by 2.

---

##### Steps

I have reproduced the issue as follows:
- Create colima instance with limited amount of memory (2GB)
- If necessary, download the Linux distribution of both pack 0.29 and pack 0.30 so that it's available inside Colima
- Clone the buildpacks sample repo as described here https://buildpacks.io/docs/operator-guide/create-a-builder/
- Add a new buildpack in this repo:
Create directory: `mkdir buildpacks/memory-test`
Define include-files: `vi buildpacks/memory-test/buildpack.toml`
```
api = "0.7"

[buildpack]
  id = "memory-test"
  version = "0.0.1"
[metadata]
  include-files = [
    "testfile"]

[[stacks]]
  id = "*"
```
Create sample large file: `dd if=/dev/zero of=buildpacks/memory-test/testfile bs=1G  count=1`

Then running the following command works fine using pack 0.29:
`pack builder create my-builder:jammy --config ./builder.toml`
```
jammy: Pulling from cnbs/sample-stack-run
Digest: sha256:e0a2065681ca908be7d612e71eed30963ccd6560f210798f3939cbc8a58daefe
Status: Image is up to date for cnbs/sample-stack-run:jammy
jammy: Pulling from cnbs/sample-stack-build
Digest: sha256:2f7566720a7c32a0ef574415952573fb2b203af740802ab09c6e9fac455f6763
Status: Image is up to date for cnbs/sample-stack-build:jammy
Successfully created builder image my-builder:jammy
Tip: Run pack build <image-name> --builder my-builder:jammy to use this builder
```

But using pack 0.30 it fails:
```
jammy: Pulling from cnbs/sample-stack-run
[...]

PR #1896, opened 2023-09-11
Hardlinks are dereferenced in generated archives

### LINKED ISSUE #1286: Hardlinks are dereferenced in generated archives increasing buildpack and app image sizes
### Summary
If a local buildpack contains files that are hardlinks of each other, the hard links are not preserved in the `.cnb` file or OCI image created by `pack buildpack package`, and instead the file contents is duplicated, increasing the size of the buildpack.

This affects compiled buildpacks that use a single hard-linked binary for detect/build (where the binary determines which mode it is being run in, based on its filename; such as in [libcnb.rs](https://github.com/Malax/libcnb.rs/blob/c502c73b0b586acd3cd45e39947b015dcd60f088/src/runtime.rs#L64-L84) and the [Go libcnb](https://github.com/buildpacks/libcnb/blob/d893d2a8defa3a61a6eb22d20b6fbe018f50ff9d/main.go#L45-L53)). 

This contrasts to using Docker natively, which preserves hardlinks both in the build context, and in layers created by `RUN`  commands.

---

##### Steps
1. `pack buildpack new test-buildpack`
2. `dd if=/dev/urandom of=test-buildpack/large-file count=$((100 * 1024)) bs=1024` (a 100MB file)
3. `ln test-buildpack/large-file test-buildpack/large-file2` (a hard link)
4. `pack buildpack package --verbose --path test-buildpack/ --format file test-buildpack.cnb`
5. `pack buildpack package --verbose --path test-buildpack/ --format image test-buildpack`
6. `ls -lh test-buildpack.cnb`
7. `docker images | rg test-buildpack`

##### Current behavior
Both the `test-buildpack.cnb` file and the `test-buildpack` OCI image are ~200MB each, since the hardlink wasn't preserved.

For the `.cnb` file:
```
$ ls -lh test-buildpack.cnb
-rw-r--r-- 1 emorley staff 201M Sep 16 16:00 test-buildpack.cnb
```

For the OCI image:
```
$ docker images | rg test-buildpack
test-buildpack                            latest    0f41579f403c   41 years ago   210MB
```

##### Expected behavior
Both the `test-buildpack.cnb` file and the `test-buildpack` OCI image are only ~100MB in size each, since the content of the hardlinked file is only included once.

For the `.cnb` file:
```
$ ls -lh test-buildpack.cnb
-rw-r--r-- 1 emorley staff 101M Sep 16 16:00 test-buildpack.cnb
```

For the OCI image:
```
$ docker images | rg test-buildpack
test-buildpack                            latest    0f41579f403c   41 years ago   105MB
```

### Additional Context
Compare this to the Docker behaviour...

```Dockerfile
FROM ubuntu:20.04
COPY . /
```

1. Create `Dockerfile` with above contents
2. `dd if=/dev/urandom of=large-file count=$((100 * 1024)) bs=1024`
3. `ln large-file large-file2`
4. `docker build -t test-docker .`
5. `docker images | rg '(ubuntu|test-docker)'`

As can be seen from the logs, only 100MB is transferred in the build context, and the resultant image is also only 100MB larger than the base Ubuntu 20.04 image:

```
$ docker build -t test-docker .
[+] Building 0.8s (7/7) FINISHED                                                                                                                                                                     
 => [internal] load build definition from Dockerfile                                                                                                                                            0.0s
 => => transferring dockerfile: 69B                                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                               0.0s
 => => transferring context: 2B                                                                                                                                                                 0.0s
 => [internal] load metadata for docker.io/library/ubuntu:20.04                                                                                                                                 0.0s
 => [internal] load build context                                                                                                                                                               0.0s
 => => transferring context: 232B                                                                                                                                                               0.0s
 => CACHED [1/2] FROM docker.io/library/ubuntu:20.04                                                                                                                                            0.0s
 => [2/2] COPY . /                                                                                                                                                                              0.2s
 => exporting to image                                                                                                                                                                          0.5s
 => => exporting layers                                                                                                                                                                         0.5s
 => => writing image sha256:488b0bd9f029e11c15fcdb30b40ec6257f3967dbe63a5c525bc3b8415935d2bb                                                                                                    0.0s
 => => naming to docker.io/library/test-docker                                                                                                                                                  0.0s

$ docker images | rg '(ubuntu|test-docker)'
test-docker                               latest    488b0bd9f029   4 seconds ago        178MB
ubuntu                                    20.04     fb52e22af1b0   2 weeks ago          72.8MB
```

---

##### pack info
```
$ pack report
Pack:
  Version:  0.20.0+git-66a4f32.build-2668
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.11.3

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6

Config:
  experimental = true
  
  [[trusted-builders]]
    name = "[REDACTED]"
```

##### docker info
```
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.6.1-docker)
  compose: Docker Compose (Docker Inc., v2.0.0-rc.3)
  scan: Docker Scan (Docker Inc., v0.8.0)

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 27
 Server Version: 20.10.8
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e25210fe30a0a703442421b0f60afac609f950a3
 runc version: v1.0.1-0-g4144b63
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 5.10.47-linuxkit
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 7.773GiB
 Name: docker-desktop
 ID: [REDACTED]
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

jjbustamante:
## Summary
This PR adds a fix to handle hardlinks included into a buildpack when `pack buildpack package` is executed. 

#### Before

```bash
> docker images | rg test-buildpack
test-buildpack                   latest                 210fe8ffb454   43 years ago    210MB
```

`large-file` and `large-file2` references the same file with a hardlink, we can notice the size of the buildpack image is bigger than expected because the hardlink is lost during the image creation.

Using [dive](https://github.com/wagoodman/dive), we can see:

Screenshot 2023-09-11 at 2 48 35 PM

#### After

```bash
> docker images | rg test-buildpack
test-buildpack                latest              db42087024b4   43 years ago    105MB
```

After the fix is applied, the final image size is as expected not duplicating the files

Using [dive](https://github.com/wagoodman/dive), we can see:

Screenshot 2023-09-11 at 2 51 18 PM

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1286


pkg/buildpack/buildpack_tar_writer.go
```diff
@@ -107,12 +107,7 @@ func (b *BuildModuleWriter) writeBuildModuleToTar(tw archive.TarWriter, module B
 			return errors.Wrapf(err, "failed to write header for '%s'", header.Name)
 		}
 
-		buf, err := io.ReadAll(tr)
-		if err != nil {
-			return errors.Wrapf(err, "failed to read contents of '%s'", header.Name)
-		}
-
-		_, err = tw.Write(buf)
+		_, err = io.Copy(tw, tr)
```
jjbustamante, line 110: I noticed this line could throw the same error reported [here](https://github.com/buildpacks/pack/issues/1886)

pkg/buildpack/buildpack_test.go
```diff
@@ -1002,3 +1037,11 @@ func assertWindowsBuildpacksToTar(t *testing.T, actual []buildpack.ModuleTar, ex
 		h.AssertTrue(t, found)
 	}
 }
+
+type blobDir struct {
+	path string
+}
+
+func (b *blobDir) Open() (io.ReadCloser, error) {
+	return archive.ReadDirAsTar(b.path, ".", 0, 0, -1, true, false, nil), nil
```
jjbustamante, line 1046: I need a `Blob` which source is a folder, so basically I took the same method call from [here](https://github.com/buildpacks/pack/blob/main/pkg/blob/blob.go#L33)

jjbustamante: My fault, there was already a constructor!

natalieparellano (commented):
@jjbustamante it's cool to see this taking shape. I left a couple comments in case it's helpful

pkg/archive/archive_windows.go
```diff
@@ -0,0 +1,13 @@
+package archive
+
+import (
+	"os"
+)
+
+func hasHardlinks(fi os.FileInfo) bool {
+	return false
```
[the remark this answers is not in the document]
jjbustamante: Honestly, I didn't spend a lot of time trying to check on to implement it on Windows, in fact, I am not sure if Windows has the same concept of a hard link.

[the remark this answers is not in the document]
natalieparellano: I think it does - https://learn.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions for example

pkg/buildpack/buildpack_test.go
```diff
@@ -511,6 +512,38 @@ version = "1.2.3"
...
+				// create a hard link
+				err := os.Link(filepath.Join(bpRootFolder, "original-file"), filepath.Join(bpRootFolder, "original-file-2"))
+				h.AssertNil(t, err)
+			})
+
+			it.After(func() {
+				os.RemoveAll(filepath.Join(bpRootFolder, "original-file-2"))
+			})
+
+			it.Focus("hardlink is preserved in the output tar file", func() {
```
natalieparellano, line 530:
```suggestion
			it("hardlink is preserved in the output tar file", func() {
```

testhelpers/tar_assertions.go
```diff
@@ -113,6 +136,19 @@ func SymlinksTo(expectedTarget string) TarEntryAssertion {
 	}
 }
 
+func HardLinks() TarEntriesAssertion {
+	return func(t *testing.T, header1 *tar.Header, _ []byte, header2 *tar.Header, _ []byte) {
+		t.Helper()
+		if header1.Typeflag != tar.TypeLink && header2.Typeflag != tar.TypeLink {
```
natalieparellano, line 142: I'm slightly confused - this code updates the header type only for 2nd (and later) instances of the file:

```
		if processedPath, ok := hardLinkFiles[inode]; ok {
			header.Typeflag = tar.TypeLink
			header.Linkname = processedPath
			header.Size = 0
		} else {
			hardLinkFiles[inode] = header.Name
		}
```

How do both headers get type flag `tar.TypeLink`?

pkg/buildpack/buildpack_test.go
```diff
@@ -511,6 +512,38 @@ version = "1.2.3"
...
+				)
+				h.AssertNil(t, err)
+
+				tarPath := writeBlobToFile(bp)
+				defer os.Remove(tarPath)
+
+				h.AssertOnTarEntries(t, tarPath,
+					"/cnb/buildpacks/bp.one/1.2.3/original-file",
+					"/cnb/buildpacks/bp.one/1.2.3/original-file-2",
+					h.HardLinks(),
```
natalieparellano, line 543:
```suggestion
					h.AreEquivalentHardLinks(),
```
(nit) this more descriptive name could help the reader

natalieparellano (commented):
Nice work @jjbustamante - I left one comment/question about the test assertion, but this seems just about ready. Hats off to you for figuring out the Windows stuff.

testhelpers/tar_assertions.go
jjbustamante: Actually I am checking that at least one of them is a hardlink, but I don't know which one is going to be pick up first

PR #1897, opened 2023-09-11
Fix target validation when buildpack fails to declare field

natalieparellano:
Optional fields are treated as wildcards when not specified


jjbustamante (approved):
LGTM

jkutner (approved):

PR #1910, opened 2023-09-19
update: use DOCKER_HOST as the default

### LINKED ISSUE #1338: Better defaults for `--docker-host`.
### Description
When running `pack` against docker socket at non-standard location using the `DOCKER_HOST` environment variable,
the user also has to set the `--docker-host` flag so export container knows how to access the daemon.

Right now it defaults to `unix://var/run/docker.sock` which is not always right.

The value of the flag can be determined if docker host is either `unix socket` or `ssh`.
For `unix socket` the mapping is identity.
For `ssh` we can acutally ask remote what is the value of the `DOCKER_HOST` there.
For `TCP` there is no way that I know of; user simply must know what to enter.

till:
## Summary

Make `--docker-host=inherit` the default behavior.

#### Before

The default was to use a variant of /var/run/docker.sock (depending on OS/environment).

#### After

The default is to check `DOCKER_HOST` and fall back to the old default before.

## Documentation

    - [x] Yes, see #1338
    - [ ] No

## Related

Resolves #1338


matejvasek:
Could we default to `DOCKER_HOST` conditionally?

* With `ssh` never default.
* With `tcp` default only when host in not localhost (`localhost`, `127.0.0.1`, `::1`) and  `DOCKER_CERT_PATH`,`DOCKER_TLS_VERIFY` are not set.
* With `unix` default only on Linux, do not default on macOS or Windows.

This would be more backward compatible.

till:
> Could we default to `DOCKER_HOST` conditionally?
> 
> * With `ssh` never default.
> * With `tcp` default only when host in not localhost (`localhost`, `127.0.0.1`, `::1`) and  `DOCKER_CERT_PATH`,`DOCKER_TLS_VERIFY` are not set.

Do you mean, it's not used when the `DOCKER_HOST` contains `ssh://`, same for `tcp://` (along with the vars)?

matejvasek:
I mean for `ssh` the DOCKER_HOST should not be used ever. For `TCP` it should be used as long as the hostname != localhost and TLS is not employed. For `unix` socket it should be used only on Linux.

matejvasek:
This will break things for people using `ssh`.

till:
> I mean for `ssh` the DOCKER_HOST should not be used ever. For `TCP` it should be used as long as the hostname != localhost and TLS is not employed. For `unix` socket it should be used only on Linux.

Oh, you mean — employ a validator for what's "in" `DOCKER_HOST`?

till:
So as a summary, these are the cases you would like to address — where a value from `DOCKER_HOST` is _ignored_:

- OS != "linux" and `DOCKER_HOST=unix://...`
- `DOCKER_HOST=tcp://localhost` (any variant of it, such as `127.0.0.1` or `::1`)
- `DOCKER_HOST=tcp://...`and the vars `DOCKER_CERT_PATH` and `DOCKER_TLS_VERIFY` not empty

What happens then? Empty it in code and let the current logic pick up the OS-specific binds for the socket? Add a log line?

natalieparellano:
Any further updates here?

till:
Not yet, I can add a commit some time to address the tls settings. Is there anything else that needs to be addressed?

natalieparellano:
@till we discussed this in Slack here: https://cloud-native.slack.com/archives/C0331B61A1Y/p1725391083581559?thread_ts=1724932960.525119&cid=C0331B61A1Y

I'm not sure if we have a clear way forward with this. Do you have any thoughts about it?

PR #1913, opened 2023-09-19
Update acceptance fixtures to use newer Buildpack API versions

natalieparellano:
Lifecycle 0.18.0 and above will only support Buildpack API versions 0.7 and above


jkutner (approved):

acceptance/acceptance_test.go
```diff
@@ -794,23 +794,20 @@ func testAcceptance(
...
-						assertImage.HasLabelWithData(builderName, "io.buildpacks.extension.layers", `{"read/env":{"read-env-version":{"api":"0.9","layerDiffID":"`+extReadEnvDiffID+`","name":"Read Env Extension"}},"simple/layers":{"simple-layers-version":{"api":"0.2","layerDiffID":"`+extSimpleLayersDiffID+`","name":"Simple Layers Extension"}}}`)
-						assertImage.HasLabelWithData(builderName, "io.buildpacks.buildpack.order-extensions", `[{"group":[{"id":"read/env","version":"read-env-version"},{"id":"simple/layers","version":"simple-layers-version"}]}]`)
-						// buildpacks
-						assertImage.HasLabelWithData(builderName, "io.buildpacks.buildpack.layers", `{"read/env":{"read-env-version":{"api":"0.2","stacks":[{"id":"pack.test.stack"}],"layerDiffID":"`+bpReadEnvDiffID+`","name":"Read Env Buildpack"}},"simple/layers":{"simple-layers-version":{"api":"0.2","stacks":[{"id":"pack.test.stack"}],"layerDiffID":"`+bpSimpleLayersDiffID+`","name":"Simple Layers Buildpack"}}}`)
-						assertImage.HasLabelWithData(builderName, "io.buildpacks.buildpack.order", `[{"group":[{"id":"read/env","version":"read-env-version","optional":true},{"id":"simple/layers","version":"simple-layers-version","optional":true}]}]`)
```
natalieparellano, line 813: This does make our assertions weaker, but I'd argue it's okay because:
* It's unclear how much value we get from these assertions in the first place
* We're keeping the assertions for Linux containers
* Keeping the Windows assertions requires contributors to either have a Windows box OR they must let the tests fail not once, but twice in CI in order to grab the correct diffIDs ...this seems like an undue burden

Issue #1632, opened 2023-02-13
[RFC #0109] `pack builder create` should accept builder env config


natalieparellano:
### Description
As per [RFC 0109](https://github.com/buildpacks/rfcs/blob/main/text/0109-build-config.md), files in `/cnb/build-config/env` will be translated to buildpack environment variables by the lifecycle.

It would be nice if builder.toml exposed a way to create these files.

### Proposed solution

Add a table to builder.toml:

```
[[build.env]]
name = "<string>"
value = "<string>"
suffix = "<string>"
delim = "<string>"
```

that will be translated to files in `/cnb/build-config/env` by `pack` during `pack builder create`.

### Describe alternatives you've considered
`[[build.env]]` mirrors `[[io.buildpacks.build.env]]` in project descriptor but we might also call it `[[env]]`

[Proposed changes]([[build.env]]) to the platform spec would introduce `[[run.image]]` to builder.toml.

### Additional context
- [x] This feature should be documented somewhere

This may benefit from a sub-team RFC.

cc @samj1912 @AidanDelaney

natalieparellano:
Tagging as "good first issue" as the implementation will borrow heavily from what already exists for translating `[[io.buildpacks.build.env]]` in project descriptor to files in `<platform>/env`. But this will probably require at least some mentorship from an experienced contributor to make this pleasant for someone who is totally new to the project.

bhavyastar:
Hey @natalieparellano, I want to work on this issue. I am also looking to participate in LFX Mentorship so it will be a good mentorship issue for me so that I will also get familiar with the project. Can you please assign me?

jjbustamante:
@bhavyastar I just assigned to you #1615, do you prefer to work on that one first?

natalieparellano:
Apologies, this issue is actually blocked on the discussion that will establish the desired schema for builder.toml (and whether an RFC is necessary). Hopefully this could be sorted soon, in the meantime it's probably best to start in on another issue :)

jjbustamante:
We still need to discuss about the schema for builder.toml should look it, let's bring this up in the next working group meeting

jjbustamante:
- @joe-kimmel-vmw was working on adding the [builder.toml](https://github.com/buildpacks/spec/blob/ac752ecb405d6e4b34e346087abd24e143da5a01/distribution.md#buildertoml-toml) to the Spec
[...]

PR #1921, opened 2023-09-25
added targets flag for buildpack new cli

### LINKED ISSUE #1918: `pack buildpack new` should accept `--targets`
### Description

`pack buildpack new` accepts `--stacks` but stacks are deprecated. It should accept `--targets` and print a warning when `--stacks` are used.

### Proposed solution

From `github.com/buildpacks/lifecycle/buildpack` package - the structure of targets is a **list** of 

```
type TargetMetadata struct {
	OS          string     `json:"os" toml:"os"`
	Arch        string     `json:"arch" toml:"arch"`
	ArchVariant string     `json:"arch-variant,omitempty" toml:"arch-variant"`
	Distros     []OSDistro `json:"distros,omitempty" toml:"distros"`
}
```

```
type OSDistro struct {
	Name    string `json:"name" toml:"name"`
	Version string `json:"version" toml:"version"`
}
```

For each target, all fields are optional (but at least one should be specified).

### Additional context
- [X] This feature should be documented somewhere - we should update our use of `pack buildpack new` in the docs to include targets

WYGIN:
## Summary

- added new flag `--targets` to `pack buildpack new` with shorthand `-t`
- now this new flag `--targets` will take args in format [os][/arch][/archvariant]:[distroName@version@another-version];[another-distro@v1@v2]
- mark `--stacks` flag deprecated
- removed the default value of stacks, so `id` of stacks no more exists in file `buildpack.toml`
- remove `[[stacks]]` from `buildpack.toml` file
- will add a default `--target` of current device when `target` flag is not specified 
- will show an error on unknown target is added from cli (only for [os] [arch] [archVariant]) but still adds those targets to final output of `buildpack.toml` file

#### Before
` pack buildpack new test`
```toml
api = "0.8"
WithWindowsBuild = false
WithLinuxBuild = false

[buildpack]
  id = "test"
  version = "1.0.0"

[[stacks]]
  id = "io.buildpacks.stacks.jammy"

```

#### After
`pack buildpack new test -t "linux/arm/v6" -t "linux/amd64" -t "windows/amd64:windows-nano@10.0.19041.1415" -t "linux/arm/v6:ubuntu@14.04" -t "linux/arm/v6:ubuntu@16.04" -t "linux/arm/v6:ubuntu@16.01@16.02;debian@10.10@8.06"`
```toml
api = "0.8"
WithWindowsBuild = false
WithLinuxBuild = false

[buildpack]
  id = "test"
  version = "1.0.0"

[[targets]]
  os = "linux"
  arch = "arm"
  variant = "v6"

[[targets]]
  os = "linux"
  arch = "amd64"

[[targets]]
  os = "windows"
  arch = "amd64"

  [[targets.distributions]]
    name = "windows-nano"
    versions = ["10.0.19041.1415"]

[[targets]]
  os = "linux"
  arch = "arm"
  variant = "v6"

  [[targets.distributions]]
    name = "ubuntu"
    versions = ["14.04"]

[[targets]]
  os = "linux"
  arch = "arm"
  variant = "v6"

  [[targets.distributions]]
    name = "ubuntu"
    versions = ["16.04"]

[[targets]]
  os = "linux"
  arch = "arm"
  variant = "v6"

  [[targets.distributions]]
    name = "ubuntu"
    versions = ["16.01", "16.02"]

  [[targets.distributions]]
    name = "debian"
    versions = ["10.10", "8.06"]

```
## Documentation

    - [x] Yes, see [# doc for `--targets` flag](https://github.com/WYGIN/buildpacks-pack/blob/90c2195d1493590e106a4403b9d2b15f7b73bd26/internal/commands/buildpack_new.go#L130)
    - [ ] No

## Related

Resolves #1918


internal/commands/buildpack_new.go
```diff
@@ -85,6 +99,14 @@ func BuildpackNew(logger logging.Logger, creator BuildpackCreator) *cobra.Comman
 	cmd.Flags().StringVarP(&flags.Path, "path", "p", "", "Path to generate the buildpack")
 	cmd.Flags().StringVarP(&flags.Version, "version", "V", "1.0.0", "Version of the generated buildpack")
 	cmd.Flags().StringSliceVarP(&flags.Stacks, "stacks", "s", []string{"io.buildpacks.stacks.jammy"}, "Stack(s) this buildpack will be compatible with"+stringSliceHelp("stack"))
+	cmd.Flags().MarkDeprecated("stacks", "stacks is deprecated in the favor of `targets`")
+	cmd.Flags().Var(&flags.Targets, "targets",
```
jjbustamante, line 103: What we've done in other parts of the code is to define a flag that receives `[]string` and then we parse each individual string to convert it into a struct.

For example, if you add

```go
type BuildpackNewFlags struct {
    //  ... other fields	
    Targets  []string
}
```

Then you can have a method to parse the targets, in this method, you can `strings.split` by `:` to divide the target on:
1.  `os/arch/variant`
   -  Here you can `strings.split` by `/`
2. `name@version` 
  - Here you can `strings.split` by ';' (based on the new format suggested by you)
      - `string.split` by '@' 
After having all the pieces you can instantiate your `Target` structure.

I think this can simplify all the logic you have to create the `Target`

WYGIN:
@jjbustamante i am unsure why the tests are failing & i tried to run tests on main branch & undoing all changes i made in my current branch, is these tests are failing due to my changes ?

jjbustamante:
> @jjbustamante i am unsure why the tests are failing & i tried to run tests on main branch & undoing all changes i made in my current branch, but still getting same result :( are these tests are failing due to my changes ?

Let me check, I tried yesterday to check the logs but I was not able to.

jjbustamante:
Yes, You are getting an error like this:

Screenshot 2023-09-27 at 9 37 05 AM

```bash
buildpack_new.go:112: Unexpected call to *testmocks.MockPackClient.NewBuildpack([context.Background {0.8 /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000g
n/T/build-test163248382/some-cnb example/some-cnb 1.0.0 [] [{   []}]}]) 
at /Users/runner/work/pack/pack/internal/commands/buildpack_new.go:112 because:
expected call at /Users/runner/work/pack/pack/internal/commands/buildpack_new_test.go:58 doesn't match the argument at index 1.

Got: {0.8 /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/build-test163248382/some-cnb example/some-cnb 1.0.0 [] [{   []}]} (client.NewBuildpackO
ptions)

 Want: is equal to {0.8 /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/build-test163248382/some-cnb example/some-cnb 1.0.0 [{io.buildpacks.stacks
.jammy []}] [{linux arm v6 [{ubuntu [14.04 16.04]}]}]} (client.NewBuildpackOptions)
```

It seems to be because of the new argument, the `Got` expectation doesn't know about the new platform values. I think you may need to re-create the Mock.

jjbustamante:
I think I know what is going on, if you take a look at your change in [internal/commands/buildpack_new_test.go](https://github.com/buildpacks/pack/pull/1921/files#diff-f51bc37c8d2ae5443008386b25498b4ea07f3d3e6eff94edbe55bcd685edec01) you said:

- I am expecting the pack client to be called with these arguments, and you added the `Targets` stuff

But in the line 78

```golang
path := filepath.Join(tmpDir, "some-cnb")
command.SetArgs([]string{"--path", path, "example/some-cnb"})
```
Where you are actually setting the flags to the `pack buildpack package` command, you are not setting the `--targets` flag there. You should set the values you are expecting to process there.
 
Screenshot 2023-09-27 at 9 53 36 AM

A suggestion, do not modify the current tests just add a new test case for the new scenario.

```golang
when("BuildpackNew#Execute", func() {
	it("uses the args to generate artifacts", func() { 
              // This is the current existing test
        }   
})
```

You could modify it in this way:

```golang
when("BuildpackNew#Execute", func() {
	it("uses the args to generate artifacts", func() { 
              // This is the current existing test
        }   
        when("targets is specified", func() { 
              it("uses the targets to generate artifacts", func(){ 
                 // Copy here copy from the previous test, and modify it as you need. 
                 // Also any refactor is valid to avoid code duplication
              }
        }
})
```

WYGIN:
@jjbustamante @natalieparellano pr is ready for review, please let me know if there there is something to fix

internal/commands/buildpack_new.go
```diff
@@ -66,11 +71,46 @@ func BuildpackNew(logger logging.Logger, creator BuildpackCreator) *cobra.Comman
...
+				if err != nil {
+					return err
+				}
+				targets = append(targets, dist.Target{
+					OS:            os,
+					Arch:          arch,
+					ArchVariant:   variant,
+					Distributions: distroMap,
+				})
+			}
```
jjbustamante, line 106: I will suggest a quick refactor, the reason behind that is because your code to parse the platform input could be reuse in other commands on pack where are working to use the pattern. 😄 

I will suggest to create a `internal/target` package with two files `parse.go` and `platform.go`. 

In `parse.go` move

```golang 
func ParseTargets(targetsInput []string, logger logging.Logger) ([]dist.Target, error) {
   // move all this logic from L74 to L106
}

func getTarget(t string) ([]string, []string, []string, error) { }

func getSliceAt[T interface{}](slice []T, index int) (T, error) { }

```

In `platform.go` move
 
```golang
var GOOSArch = map[string][]string{ }             // I think this can be internal to the package
var GOArchVariant = map[string][]string{ }      // also, can be rename to lower case to be internal to the package

func getPlatform(t []string) (string, string, string, error) { }

func isOS(os string) bool { }

func supportsArch(os string, arch string) bool { }

func supportsVariant(arch string, variant string) bool { }

```


Then you can replace with something like:

```golang
	targets, err := target.ParseTargets(flags.Targets, logger)
	if err != nil {
		return err
	}
```

If you want, you can also simplify your test cases, because now you can focus on your function `ParseTargets`. I want to suggest the function to return the parsed strings, but then I noticed we have the `dist.Distributions` and make things complicated, so it is fine to return the `dist.Target` object.

WYGIN: done :)

jjbustamante:
@WYGIN Amazing work!

I compiled your branch and played with it. I have a little doubt with one scenario, @natalieparellano  maybe you can help me to validate the expected behavior.

I ran the [example](https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/building-blocks-cnb/) from our docs

```bash
>  out/pack buildpack new examples/ruby \
    --api 0.8 \
    --path ruby-buildpack \
    --version 0.0.1 \
    --stacks io.buildpacks.samples.stacks.jammy
Flag --stacks has been deprecated, prefer `--targets` instead: https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md
    create  buildpack.toml
    create  bin/build
    create  bin/detect
Successfully created examples/ruby
```
The structure was created, but checking the `buildpack.toml`

```bash
> cat ruby-buildpack/buildpack.toml
api = "0.8"
WithWindowsBuild = false
WithLinuxBuild = false

[buildpack]
  id = "examples/ruby"
  version = "0.0.1"

[[stacks]]
  id = "io.buildpacks.samples.stacks.jammy"

[[targets]]
  os = "darwin"
  arch = "amd64"
```

We have `stacks` and `targets`, and I think we *can't* have [both](https://github.com/buildpacks/pack/blob/main/pkg/buildpack/buildpack.go#L283-L291)

WYGIN:
> We have `stacks` and `targets`, and I think we _can't_ have [both](https://github.com/buildpacks/pack/blob/main/pkg/buildpack/buildpack.go#L283-L291)

if i am not wrong it is part of build time but i think we can add a warning for `pack buildpack new` when both `stacks` and `targets` specified
> @jkutner Should `WithWindowsBuild` and `WithLinuxBuild` be writing down into the buildpack.toml? I didn't see those fields specify in the RFC and I am wondering if this is a bug, and those are just implementation details.
i think `WithWindowsBuild` and `WithLinuxBuild` are used with `stacks` flag & having it part of `buildpack.toml` file might not be needed when `--stacks` flag is not used

i am also thinking about adding cli `auto completion` when defining `targets` but it might not be very useful if users only target few platforms, so i would love to add autocompletion if it improves `UX` of the user

WYGIN:
@jjbustamante i would like to know if there are any changes still required, i am willing to work on this issue(#1632 ) if i am done with this pr

jjbustamante:
> 
> if i am not wrong it is part of build time but i think we can add a warning for `pack buildpack new` when both `stacks` and `targets` specified
> 

Yes, the validation is executed during build time, but, what I am totally happy with is the following:

- `pack buildpack new` command is meant to create a folder structure for a new buildpack, but, if we create a scaffolding that fails later with other command, it will confuse the end-user. I want to avoid creating a wrong structure. I am going to add a code suggestion for that case.

internal/commands/buildpack_new_test.go
```diff
@@ -82,5 +84,122 @@ func testBuildpackNewCommand(t *testing.T, when spec.G, it spec.S) {
...
+					mockClient.EXPECT().NewBuildpack(gomock.Any(), client.NewBuildpackOptions{
+						API:     "0.8",
+						ID:      "example/stacks",
+						Path:    filepath.Join(tmpDir, "stacks"),
+						Version: "1.0.0",
+						Stacks: []dist.Stack{{
+							ID:     "io.buildpacks.stacks.jammy",
+							Mixins: []string{},
+						}},
+						Targets: targets,
```
jjbustamante, line 189: Based on my comment [here](https://github.com/buildpacks/pack/pull/1921#issuecomment-1743100281), I will remove this `Targets: targets`. 
In this case, the test is going to fail, but we will fix it

internal/commands/buildpack_new.go
```diff
@@ -66,11 +70,20 @@ func BuildpackNew(logger logging.Logger, creator BuildpackCreator) *cobra.Comman
 				})
 			}
 
+			targets, warn, err := target.ParseTargets(flags.Targets)
```
jjbustamante, line 73: I will add the following to fix the test from my [comment](https://github.com/buildpacks/pack/pull/1921/files#r1342753791)

```golang
var targets []dist.Target
if len(stacks) == 0 {
	targets, err = target.ParseTargets(flags.Targets, logger)
	if err != nil {
		return err
	}
}
```

Also, I will suggest to change the signature of `target.ParseTargets` to:

```golang
func ParseTargets(t []string, logger logging.Logger) (targets []dist.Target, err error) 
```

You can warn the messages according and avoid the new *warn.Warn* structure

internal/commands/buildpack_new_test.go
WYGIN: done :)

jjbustamante (approved):
@WYGIN thank you so much for this PR!! ❤️ , it is going to be very valuable for our end goal for supporting multi-arch!

Great job!

jkutner (approved):

PR #1933, opened 2023-10-12
Ensure the downloaded os/arch always matches the expected os/arch

natalieparellano:
## Summary

Previously, we didn't provide a platform, so if the downloaded thing (run image or buildpack/extension) is multi-arch we would get whatever is the default. 

Now, we ensure:
* the run image platform matches the builder for `pack build` and previous image for `pack rebase`
* the buildpack/extension platform matches the builder for `pack build` and `pack builder create`


natalieparellano:
I updated this PR a bit to encompass more things. It might be easier to review as two separate commits

jjbustamante (approved):
LGTM

jkutner (approved):

PR #1935, opened 2023-10-17
Dependency Updates

jkutner:
## Summary
Dependency Updates
## Output
N/A

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___


edmorley:
Dependabot has just added support for grouped updates, in case that would help reduce toil?
https://github.blog/changelog/2023-08-24-grouped-version-updates-for-dependabot-are-generally-available/

hone (approved):

Issue #1722, opened 2023-04-10
Add command `pack manifest inspect` to display a manifest list


jjbustamante:
### Context

We've been working on the idea to enable multi-arch support on CNB. A mentorship [project](https://mentorship.lfx.linuxfoundation.org/project/ee387e1b-de4e-4c1e-9bef-0239a2e9ca40) was created on LFX and this feature is part of the efforts to build the Minimal Viable Product (MVP) during the program duration.

More information can be found:
- [Notes](https://docs.google.com/document/d/1nrMmz8ZFOdEwzlzQYS963BfSqCvfoyhbk5FvExueuM4/edit?usp=sharing)
- Meeting [recording](https://www.youtube.com/watch?v=B-rg5SQQ7tA)

### Additional context
- [ ] This feature should be documented somewhere

nitishfy:
Hi, I'd like to take this up!

jjbustamante:
Hi @nitishfy! Sorry, we already have a PR open for this one

Issue #1721, opened 2023-04-10
Add command `pack manifest remove` to remove a manifest list


jjbustamante:
### Context

We've been working on the idea to add multi-arch support on CNB. A mentorship [project](https://mentorship.lfx.linuxfoundation.org/project/ee387e1b-de4e-4c1e-9bef-0239a2e9ca40) was created on LFX and this feature is part of the efforts to build the Minimal Viable Product (MVP) during the program duration.

More information can be found:
- [Notes](https://docs.google.com/document/d/1nrMmz8ZFOdEwzlzQYS963BfSqCvfoyhbk5FvExueuM4/edit?usp=sharing)
- Meeting [recording](https://www.youtube.com/watch?v=B-rg5SQQ7tA)

### Description

Delete one or more manifest lists from local storage

### Proposed solution

A new feature on Pack can be implemented to help our users. The feature is inspired by the similar functionality in [docker](https://docs.docker.com/engine/reference/commandline/manifest/) or [podman](https://docs.podman.io/en/v3.2.0/markdown/podman-manifest-create.1.html)

```bash
Usage:  pack manifest remove [manifest-list] [manifest-list...]

Delete one or more manifest lists from local storage

```

### Additional context
- [ ] This feature should be documented somewhere

Issue #1720, opened 2023-04-10
Add command `pack manifest push` to push an manifest list to a remote registry


jjbustamante:
### Context

We've been working on the idea to add multi-arch support on CNB. A mentorship [project](https://mentorship.lfx.linuxfoundation.org/project/ee387e1b-de4e-4c1e-9bef-0239a2e9ca40) was created on LFX and this feature is part of the efforts to build the Minimal Viable Product (MVP) during the program duration.

More information can be found:
- [Notes](https://docs.google.com/document/d/1nrMmz8ZFOdEwzlzQYS963BfSqCvfoyhbk5FvExueuM4/edit?usp=sharing)
- Meeting [recording](https://www.youtube.com/watch?v=B-rg5SQQ7tA)

### Description

When a user already has a manifest list saved in the docker daemon, they require to push that manifest list (Image Index) to a remote registry. 

### Proposed solution

A new feature on Pack can be implemented to help our users. The feature is inspired by the similar functionality in [docker](https://docs.docker.com/engine/reference/commandline/manifest/) or [podman](https://docs.podman.io/en/v3.2.0/markdown/podman-manifest-create.1.html)

```bash
pack manifest push [OPTIONS] <manifest-list> 

Push a manifest list to a repository

  options:
          --format      Manifest list type (oci or v2s2) to use when pushing the list (default is v2s2)
          --insecure   Allow push to an insecure registry
     -p --purge       Delete the manifest list or image index from local storage if pushing succeeds.
```

### Additional context
- [ ] This feature should be documented somewhere

Issue #1719, opened 2023-04-10
Add command `pack manifest add` to add an image to a manifest list


jjbustamante:
### Context

We've been working on the idea to add multi-arch support on CNB. A mentorship [project](https://mentorship.lfx.linuxfoundation.org/project/ee387e1b-de4e-4c1e-9bef-0239a2e9ca40) was created on LFX and this feature is part of the efforts to build the Minimal Viable Product (MVP) during the program duration.

More information can be found:
- [Notes](https://docs.google.com/document/d/1nrMmz8ZFOdEwzlzQYS963BfSqCvfoyhbk5FvExueuM4/edit?usp=sharing)
- Meeting [recording](https://www.youtube.com/watch?v=B-rg5SQQ7tA)

### Description

When a manifest list already exists in the docker daemon or in a registry, sometimes the user expects to update it *by adding* a new image to the manifest list.

### Proposed solution

A new feature on Pack can be implemented to help our users. The feature is inspired by the similar functionality in [docker](https://docs.docker.com/engine/reference/commandline/manifest/) or [podman](https://docs.podman.io/en/v3.2.0/markdown/podman-manifest-create.1.html)

```bash
pack manifest add  [OPTIONS] <manifest-list> <manifest> 
  options:
   --all         If the image which should be added to the list or index is itself a list or index, add all of the contents to the local list
   --arch     Set the architecture 
   --os         Set the operating system
   --variant  Set the architecture variant

```

### Additional context
- [ ] This feature should be documented somewhere

PR #1967, opened 2023-11-09
Add a new manifest command to support multi-architecture builds

WYGIN:
## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves [Add command pack manifest create to create a manifest list](https://github.com/buildpacks/pack/issues/1678)
Resolves [It would be nice to support easily creating a manifest list packages and builders](https://github.com/buildpacks/pack/issues/1460)
Resolves [Add command pack manifest inspect to display a manifest list](https://github.com/buildpacks/pack/issues/1722)
Resolves [Add command pack manifest push to push an manifest list to a remote registry](https://github.com/buildpacks/pack/issues/1720)
Resolves [Add command pack manifest remove to remove a manifest list](https://github.com/buildpacks/pack/issues/1721)
Resolves [Add command pack manifest add to add an image to a manifest list](https://github.com/buildpacks/pack/issues/1719)
Resolves [Add command pack manifest annotate to update information on manifest list](https://github.com/buildpacks/pack/issues/1718)


WYGIN:
@jjbustamante is there any possible way to build buildpacks into images directly without cli flags from `pack build <image-name> [flags]` so that any local imageIndex is pushed to the registry the local images to pushed into the registry without any action from users, for now i am going to notify user about the local image that need to be pushed to registry

internal/commands/manifest_create.go
```diff
@@ -0,0 +1,92 @@
...
+	cmdFlags.StringVar(&flags.arch, "arch", "", "If any of the specified images is a list/index, choose the one for `arch`")
+	if err := cmdFlags.MarkHidden("arch"); err != nil {
+		panic(fmt.Sprintf("error marking --arch as hidden: %v", err))
+	}
+	cmdFlags.BoolVar(&flags.insecure, "insecure", false, "Allow publishing to insecure registry")
+	if err := cmdFlags.MarkHidden("insecure"); err != nil {
+		panic(fmt.Sprintf("error marking insecure as hidden: %v", err))
+	}
+	cmdFlags.BoolVar(&flags.publish, "publish", false, "Publish to registry")
+	cmdFlags.BoolVar(&flags.all, "all", false, "Add all of the list's images if the images to add are lists/index")
```
jjbustamante, line 83: We actually leave the complexity of dealing with recursive Image Indexes because no all registries handle that correctly, it is possible according to the OCI Image spec, but in practice it is not very useful.

internal/commands/manifest_create.go
```diff
@@ -0,0 +1,92 @@
...
+	cmdFlags.BoolVar(&flags.amend, "amend", false, "Modify an existing list/index if one with the desired name already exists")
```
jjbustamante, line 84: I am not sure if I want to use the same *create* command to update an existing image index, I think it could be another command, if we actually need it.

jjbustamante (commented):
A few comments in the `manifest create commands flags`

jjbustamante:
> @jjbustamante is there any possible way to build buildpacks into images directly without cli flags from `pack build <image-name> [flags]` 

I think I am not understanding what you want to do here, `pack buildpack package` will create a Buildpack OCI image that could be saved into the daemon, registry or in a file on disk with the .cnb extension (this is actually OCI layout format).

We do want to add capabilities to pack to handle the creation of Image Indexes from the beginning, but we are trying to define the behavior with the following [RFC](https://github.com/buildpacks/rfcs/pull/295). 

In this particular work, we just want to provide end-users with some commands to create the Image Index assuming they already have OCI images created. Check this [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0124-pack-manifest-list-commands.md)

PR #1969, opened 2023-11-09
Retrying logic when pulling an image and platform doesn't match

### LINKED ISSUE #1968: platform mismatch during pack build / pack builder create
### Summary
When running `pack build` 0.32.0 wih the paketo builder/buildpack images, we see the following error:

```
fetching image: image with reference gcr.io/paketo-buildpacks/dotnet-core:0.41.0 was found but does not match the specified platform: wanted linux/amd64, actual: linux
```

This is true for all our images - it's nothing specific to the dotnet-core buildpack.

---

### Reproduction

`pack build` or `pack builder create` with the paketo builders.

e.g. pick a [Paketo Samples](https://github.com/paketo-buildpacks/samples) app and try to `pack build` it.

jjbustamante:
## Summary

The Paketo team is trying to create `arm64` builders and Buildpacks, but they are doing iteratively, because of that, they are in a situation where they need some mixed support for `arm64` and `amd64` users. With the latest pack release 0.32.0 we tried to fix an issue where users running M1 laptops and trying to build a `amd64` application image ended up with an application image having `amd64` code on the top of a `arm64` run image. see this [PR](https://github.com/buildpacks/pack/pull/1933)

The problem was, we added a `platform` configuration (os/arch) when pulling Buildpacks and unfortunately this is not valid because most of the Buildpack ecosystem doesn't specify *architecture* for their buildpacks.

As a consequence, running `pack build` or `pack builder create` or any command that is pulling Builpacks throws an error, I added a retry logic to pull again, without platform (previous behavior), when an error is found with the message `does not match the specified platform` 

#### Before

```bash
ERROR: failed to add buildpacks to builder: downloading buildpack: extracting from registry docker://gcr.io/paketo-buildpacks/dotnet-core: fetching image: image with reference gcr.io/paketo-buildpacks/dotnet-core:latest was found but does not match the specified platform: wanted linux/amd64, actual: linux
```

#### After

`pack` will silently retry

```bash
latest: Pulling from paketo-buildpacks/dotnet-core
Digest: sha256:bf28b7631c222794f01e637c2de0d097f39079c0aa6c8fbdaefeef239d404219
Status: Image is up to date for gcr.io/paketo-buildpacks/dotnet-core:latest
latest: Pulling from paketo-buildpacks/dotnet-core
Digest: sha256:bf28b7631c222794f01e637c2de0d097f39079c0aa6c8fbdaefeef239d404219
```

Screenshot 2023-11-14 at 10 52 52 AM

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1968

### DIFF
--- a/pkg/image/fetcher.go
+++ b/pkg/image/fetcher.go
@@ -82,7 +82,6 @@ func NewFetcher(logger logging.Logger, docker DockerClient, opts ...FetcherOptio
 }
 
 var ErrNotFound = errors.New("not found")
-var ErrPlatformNotMatch = errors.New("platform does not match")
 
 func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions) (imgutil.Image, error) {
 	name, err := pname.TranslateRegistry(name, f.registryMirrors, f.logger)
@@ -110,10 +109,12 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions) 
 	}
 
 	f.logger.Debugf("Pulling image %s", style.Symbol(name))
-	err = f.pullImage(ctx, name, options.Platform)
-	if errors.Is(err, ErrPlatformNotMatch) {
-		f.logger.Info(err.Error())
-		err = f.pullImage(ctx, name, "")
+	if err = f.pullImage(ctx, name, options.Platform); err != nil {
+		// sample error from docker engine:
+		// image with reference <image> was found but does not match the specified platform: wanted linux/amd64, actual: linux
+		if strings.Contains(err.Error(), "does not match the specified platform") {
+			err = f.pullImage(ctx, name, "")
+		}
 	}
 	if err != nil && !errors.Is(err, ErrNotFound) {
 		return nil, err
@@ -197,11 +198,6 @@ func (f *Fetcher) pullImage(ctx context.Context, imageID string, platform string
 
 	err = jsonmessage.DisplayJSONMessagesStream(rc, &colorizedWriter{writer}, termFd, isTerm, nil)
 	if err != nil {
-		// sample error from docker engine:
-		// image with reference <image> was found but does not match the specified platform: wanted linux/amd64, actual: linux
-		if strings.Contains(err.Error(), "does not match the specified platform") {
-			return errors.Wrap(ErrPlatformNotMatch, err.Error())
-		}
 		return err
 	}
 
--- a/pkg/image/fetcher.go
+++ b/pkg/image/fetcher.go
@@ -82,6 +82,7 @@ func NewFetcher(logger logging.Logger, docker DockerClient, opts ...FetcherOptio
 }
 
 var ErrNotFound = errors.New("not found")
+var ErrPlatformNotMatch = errors.New("platform does not match")
 
 func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions) (imgutil.Image, error) {
 	name, err := pname.TranslateRegistry(name, f.registryMirrors, f.logger)
@@ -110,6 +111,10 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions) 
 
 	f.logger.Debugf("Pulling image %s", style.Symbol(name))
 	err = f.pullImage(ctx, name, options.Platform)
+	if errors.Is(err, ErrPlatformNotMatch) {
+		f.logger.Info(err.Error())
+		err = f.pullImage(ctx, name, "")
+	}
 	if err != nil && !errors.Is(err, ErrNotFound) {
 		return nil, err
 	}
@@ -192,6 +197,11 @@ func (f *Fetcher) pullImage(ctx context.Context, imageID string, platform string
 
 	err = jsonmessage.DisplayJSONMessagesStream(rc, &colorizedWriter{writer}, termFd, isTerm, nil)
 	if err != nil {
+		// sample error from docker engine:
+		// image with reference <image> was found but does not match the specified platform: wanted linux/amd64, actual: linux
+		if strings.Contains(err.Error(), "does not match the specified platform") {
+			return errors.Wrap(ErrPlatformNotMatch, err.Error())
+		}
 		return err
 	}
 
--- a/pkg/image/fetcher_test.go
+++ b/pkg/image/fetcher_test.go
@@ -6,7 +6,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
+
+	"github.com/buildpacks/imgutil"
 
 	"github.com/buildpacks/imgutil/local"
 	"github.com/buildpacks/imgutil/remote"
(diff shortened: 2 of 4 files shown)


jjbustamante:
cc @natalieparellano

pkg/client/build.go
```diff
@@ -1037,14 +1037,13 @@ func (c *Client) fetchBuildpack(ctx context.Context, bp string, relativeBaseDir
...
 
-		builderArch, err := builderImage.Architecture()
 		if err != nil {
 			return nil, nil, errors.Wrapf(err, "getting builder architecture")
 		}
 		downloadOptions := buildpack.DownloadOptions{
 			RegistryName:    registry,
 			ImageOS:         builderOS,
-			Platform:        fmt.Sprintf("%s/%s", builderOS, builderArch),
+			Platform:        builderOS,
```
natalieparellano, line 1046: The `was found but does not match the specified platform` error is coming from the daemon (see [here](https://github.com/moby/moby/blob/5bff90cea2a77446163019e00016f5e952f591c1/daemon/images/image.go#L234)). Prior to #1933 we never set any platform (see [here](https://github.com/buildpacks/pack/pull/1933/files#diff-a0c732d7543d7450a4886f604425cdd9195ca3da77338f9a0f4b6980a6ad5eb6L328)) so I think the platform provided [here](https://github.com/buildpacks/pack/blob/b12c9b3a0a8b3e9e0df590eaf901af44e5c82f06/pkg/image/fetcher.go#L181) was an empty string. Rather than override the expected platform, I wonder if we could inspect the error and retry mismatched platform with empty string.

pkg/image/fetcher_test.go
```diff
@@ -213,6 +216,20 @@ func testFetcher(t *testing.T, when spec.G, it spec.S) {
 						_, err := imageFetcher.Fetch(context.TODO(), repoName, image.FetchOptions{Daemon: true, PullPolicy: image.PullAlways, Platform: "some-unsupported-platform"})
 						h.AssertError(t, err, "unknown operating system or architecture")
 					})
+
+					when("remote platform does not match", func() {
+						it.Before(func() {
+							h.SkipIf(t, runtime.GOOS != "linux", "Skipped on non-linux")
```
jjbustamante, line 222: I didn't know how to avoid the error `image operating system windows cannot be used on this platform` when running tests on LCOW

pkg/client/build.go
jjbustamante: I changed the implementation to retry on failure

pkg/image/fetcher.go
```diff
@@ -109,7 +110,17 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
 	}
 
 	f.logger.Debugf("Pulling image %s", style.Symbol(name))
-	err = f.pullImage(ctx, name, options.Platform)
+	platform := options.Platform
+	for i := 0; i <= 1; i++ {
```
c0d1ngm0nk3y, line 114: The for loop is not really needed, ist it? I think it overcomplicates things. How about something like:

```
err = f.pullImage(ctx, name, platform)

if(errors.Is(err, ErrPlatformNotMatch)) {
  err = f.pullImage(ctx, name, "")
}
```

pkg/image/fetcher.go
```diff
@@ -109,7 +110,17 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
+		err = f.pullImage(ctx, name, platform)
+		if err != nil && errors.Is(err, ErrPlatformNotMatch) {
```
c0d1ngm0nk3y, line 116:
```suggestion
		if errors.Is(err, ErrPlatformNotMatch) {
```

pkg/image/fetcher.go
```diff
@@ -109,7 +110,17 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
+			f.logger.Warnf("%s, retrying without platform", err.Error())
```
c0d1ngm0nk3y, line 117: But since version `0.32` `pack build` seems to be broken always showing `does not match the specified platform: wanted linux/amd64, actual: linux`

Wouldn't that mean that there will always be a warning? For every buildpack. Is that the desired UX?

c0d1ngm0nk3y:
> The problem was, we added a platform configuration (os/arch) when pulling Buildpacks and unfortunately this is not valid because most of the Buildpack ecosystem doesn't specify architecture for their buildpacks.

I wonder if it would not be better to do 3 steps:

1. Try (os/arch)
2. Try (os)
3. Try anything

With that, there could be an `info` before 2. and a warning before 3. WDYT?

pkg/image/fetcher.go
natalieparellano: > Is that the desired UX?

Probably not. We could make it an info message, or just omit it.

natalieparellano: > With that, there could be an info before 2. and a warning before 3. WDYT?

This makes sense to me

jjbustamante: I added the **warning** because we are working to improve the buildpack package creation and set the `os/arch`.
I was expecting in short term, we will start having buildpacks with the `os/arch` defined and the warning will be not be shown.

pkg/image/fetcher.go
jjbustamante: Thank you @colincasey

pkg/image/fetcher.go
jjbustamante: For now, I just removed the **warning** message, because adding `Try (os)` will imply to parse the given platform, we still have work to do on this, but I think falling back to the default behavior is good to fix the bug.

pkg/image/fetcher_test.go
jjbustamante: done

pkg/image/fetcher.go
```diff
@@ -192,6 +197,11 @@ func (f *Fetcher) pullImage(ctx context.Context, imageID string, platform string
 
 	err = jsonmessage.DisplayJSONMessagesStream(rc, &colorizedWriter{writer}, termFd, isTerm, nil)
 	if err != nil {
+		// sample error from docker engine:
+		// image with reference <image> was found but does not match the specified platform: wanted linux/amd64, actual: linux
+		if strings.Contains(err.Error(), "does not match the specified platform") {
```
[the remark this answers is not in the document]
natalieparellano: I agree... but, the type of error returned is `*jsonmessage.JSONError` 😕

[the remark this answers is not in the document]
natalieparellano: The alternative I guess would be to always retry

natalieparellano:
I added an acceptance test in 5451248a and cleaned up the error wrapping so that we're passing less strings around.

I played around with retrying with platform`linux` but I still see the `wanted linux/amd64` error and I'm not sure why - perhaps Docker is filling in the arch when platform is non-nil? For now, platform empty seems like a good fallback option since it always worked before.

WDYT @jjbustamante @c0d1ngm0nk3y ?

c0d1ngm0nk3y:
> I added an acceptance test...

Great. Because for use, this is issue is quite nasty, so this should help prevent such things in future.

c0d1ngm0nk3y (approved):

jjbustamante:
@jkutner , @hone  Can I get some approvals here? I can't approve my own pull requests 😄

jkutner (approved):

Issue #1976, opened 2023-11-14
`pack inspect` does not provide `ARGS` for npm project


kotapeter:
### Summary

I can see the default web process command during the build time but the `pack inspect` does not show the same result. The `ARGS` is empty.

<img width="390" alt="Screenshot 2023-11-14 at 16 57 57" src="https://github.com/buildpacks/pack/assets/3911387/fd9bddcd-15c9-4c5c-8be4-cf9b81498a8d">
<img width="568" alt="Screenshot 2023-11-14 at 16 58 03" src="https://github.com/buildpacks/pack/assets/3911387/d2d03666-077a-4f16-b45d-c094c590b082">

---

##### Steps

1. `git clone https://github.com/kotapeter/pack-inspect-bug.git`
2. `cd pack-inspect-bug`
3. `pack build test --builder heroku/builder:22 --path=.`
4. `pack inspect test`

##### Current behavior

`ARGS` is empty

##### Expected behavior

`ARGS` should display `start`

---

##### pack info

```
Pack:
  Version:  0.29.0+git-95c8060.build-4209
  OS/Arch:  darwin/arm64

Default Lifecycle Version:  0.15.2

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.10

Config:
  experimental = true
  
  [[trusted-builders]]
    name = "[REDACTED]"
```

##### docker info

```
Client:
 Version:    24.0.6
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2-desktop.5
    Path:     /Users/kotapeter/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.22.0-desktop.2
    Path:     /Users/kotapeter/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/kotapeter/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.20
    Path:     /Users/kotapeter/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.8
    Path:     /Users/kotapeter/.docker/cli-plugins/docker-init
[...]

PR #1977, opened 2023-11-15
fix: display `Command` remaining parts if `Args` is `nil`

kotapeter:
## Summary

Since lifecycle@0.15.0 the `Command` is an array that can contain args too. We just pick the first element of `Command.Entries` so the remaining parts are gone.

#### Before
Screenshot 2023-11-15 at 08 30 13

#### After
Screenshot 2023-11-15 at 08 30 07

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves https://github.com/buildpacks/pack/issues/1976


internal/inspectimage/info_display.go
```diff
@@ -187,12 +187,18 @@ func convertToDisplay(proc launch.Process, isDefault bool) ProcessDisplay {
 	case false:
 		shell = "bash"
 	}
+	var argsToUse []string
+	if proc.Args != nil {
+		argsToUse = proc.Args
+	} else {
+		argsToUse = proc.Command.Entries[1:]
```
natalieparellano, line 194: Should we check the length here? To avoid index out of range?

natalieparellano:
Cross-posting from the issue: https://github.com/buildpacks/pack/issues/1976#issuecomment-1812605046

internal/inspectimage/info_display.go
kotapeter: Thank you @natalieparellano 🙇 I've just added a length check.

internal/inspectimage/info_display.go
```diff
@@ -187,12 +187,18 @@ func convertToDisplay(proc launch.Process, isDefault bool) ProcessDisplay {
 	case false:
 		shell = "bash"
 	}
+	var argsToUse []string
+	if proc.Args != nil {
+		argsToUse = proc.Args
+	} else if len(proc.Command.Entries) > 1 {
+		argsToUse = proc.Command.Entries[1:]
+	}
```
natalieparellano, line 195: This is still going to drop args on the floor if the process has both non-overridable and overridable args. We could concatenate them, but then the end-user has no way of knowing what will happen if command-line args are provided (which args are going to be overridden, if any?).

If you'd be amenable to it, I think the most correct way to solve this would be something like:

1) Rename `ProcessDisplay.Args` to `ProcessDisplay.AlwaysArgs` or `ProcessDisplay.NonOverridableArgs`
2) Add `ProcessDisplay.OverridableArgs`
3) Populate `ProcessDisplay` as follows:
* If env `CNB_PLATFORM_API` on the image is >= 0.11 AND `proc.BuildpackID` has Buildpack API >= 0.9:
  * `ProcessDisplay.AlwaysArgs` is `proc.Command.Entries[1:]`
  * `ProcessDisplay.OverridableArgs` is `proc.Args`
* If  env `CNB_PLATFORM_API` on the image is < 0.10 OR `proc.BuildpackID` has Buildpack API < 0.9:
  * `ProcessDisplay.AlwaysArgs` is `proc.Args` (`proc.Command` will only ever have one entry)
  * `ProcessDisplay.OverridableArgs` is empty
* If env `CNB_PLATFORM_API` on the image is = 0.10, we kind of don't know how to display the process because we are missing the buildpack ID and hence we don't know the buildpack API. 
  * If `proc.Command.Entries[1:]` exist, then the buildpack must be newer, so we can treat it like the first case ^^
  * Otherwise, `ProcessDisplay.AlwaysArgs` is `proc.Args` and we throw up a warning that they might actually be overridable

We'd also have to make some changes to where we render the display (see where we call `NewInfoDisplay`). This is maybe more work than you bargained for, and if so, that's totally understandable. In the short-term, something like:
```suggestion
	if len(proc.Command.Entries) > 1 {
		argsToUse = proc.Command.Entries[1:]
	}
	argsToUse = append(argsToUse, proc.Args...)
```
will at least capture all the information, while leaving the end-user in the dark about which args may be overridden.

natalieparellano: @jjbustamante @kotapeter any opposition to merging this PR with the suggested code change, and creating a separate issue to more precisely describe overridable arguments? The current (broken) state could be pretty confusing.

jjbustamante (approved):

PR #1983, opened 2023-11-17
Allow the build to proceed with /cnb/lifecycle/creator when the order contains extensions

natalieparellano:
(Note that this will have no effect until there is a lifecycle 0.19.0 - but, we can prepare for it now)

The creator will error if the detected group contains extensions, but if no extensions detect then the build can proceed and potentially succeed.

This avoids end-user headaches when popular builders contain extensions (see https://github.com/buildpacks/community/discussions/244) but the extensions don't detect by default (and hence using the creator is just fine).

###

Related lifecycle PR: https://github.com/buildpacks/lifecycle/pull/1246


internal/build/lifecycle_execution.go
```diff
@@ -281,7 +281,7 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
 		return l.Export(ctx, buildCache, launchCache, kanikoCache, phaseFactory)
 	}
 
-	if l.platformAPI.AtLeast("0.10") && l.hasExtensions() {
+	if l.platformAPI.AtLeast("0.10") && l.hasExtensions() && !l.opts.UseCreatorWithExtensions {
 		return errors.New("builder has an order for extensions which is not supported when using the creator")
```
jjbustamante, line 285: I wonder if we can improve the error message to give some guidance to end-users on what's going on or what they can do to fix it.. maybe a link to some documentation?

jjbustamante (approved):
LGTM

PR #2017, opened 2023-12-28
add support to override UID for container

### LINKED ISSUE #1347: Override UID for container
`pack` supports a `--gid` flag, but why does it not support a `--uid` flag?

I would like to override the default UID that's used  because in docker rootless mode, the mappings are different (the 0 uid maps to the user running the docker daemon).

Parthiba-Hazra:
Here in the pr I simply add a flag --uid to override uid for container and keep the implementation same as gid, as a new contributor to buildpack project I might be missing some points .. A swift review or guidance would greatly assist me in diving into this project.

Resolves #1347

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,7 @@ require (
 	golang.org/x/mod v0.14.0
 	golang.org/x/oauth2 v0.15.0
 	golang.org/x/sync v0.5.0
+	golang.org/x/sys v0.15.0
 	golang.org/x/term v0.15.0
 	golang.org/x/text v0.14.0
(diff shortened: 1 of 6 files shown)


natalieparellano:
@Parthiba-Hazra - thanks a lot for this. Is it possible to add a test in lifecycle_execution_test.go? If you search for `override GID` you will find the ones for the group ID.

natalieparellano (approved):

jjbustamante (approved):
@Parthiba-Hazra great work! thank you!

PR #2022, opened 2024-01-11
Update run image metadata when --run-image is provided

### LINKED ISSUE #1964: Pack build command not picking the run image
### Summary

Just having one question about pack command: if somebody could help, it would be really helpful, or if this is not the right channel, please guide me on how to reach the right people to get it addressed.
When I'm running the pack build command with the --run-image argument, the pack command does not pick a specific run image version based on my command instruction.
After building an image, when I'm doing the pack inspect command with the app image, I can only see the latest run image. Can you please help me on this?

---
`pack inspect app-1.0-release
Inspecting image: app-1.0-release

REMOTE:
(not present)

LOCAL:

Stack: io.buildpacks.stacks.jammy

Base Image:
  Reference: 945a5e4052d9b9b024f69bf89f8cf8bd72bc727882ab0b78df9592b37119a31c
  Top Layer: sha256:be9eb0a7d50b0438c62376a29b5557f85496a55ae83d41a49f934160ffa09421

Run Images:
  index.docker.io/paketobuildpacks/run-jammy-base:latest`
##### Steps

1. passed run-image flag inside the pack command
2. updated /cnb/buildpacks/run.toml run image version and created custom image and pushed to personal docker registry.
3. used custom image while running pack build command with custom image name.

##### Current behavior

pack build command always referring latest run image.

##### Expected behavior

pack build command not referring old run which we are passing via --run-image command during the build

---

##### pack info

$ pack report
Pack:
  Version:  0.29.0+git-95c8060.build-4209
  OS/Arch:  windows/amd64

Default Lifecycle Version:  0.15.2

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.10

Config:
  default-registry = "buildpack"
  experimental = true

  [[registries]]
    name = "[REDACTED]"
    type = "github"
    url = "[REDACTED]"

  [[registries]]
    name = "[REDACTED]"
    type = "github"
    url = "[REDACTED]"

##### docker info

version:
  APIVersion: 4.7.0
  Built: 1695839078
  BuiltTime: Wed Sep 27 23:54:38 2023
  GitCommit: ""
  GoVersion: go1.20.8
  Os: linux
  OsArch: linux/amd64
  Version: 4.7.0

Pratham1812:
#### Before
image
#### After
image
## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #1964

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -1085,6 +1085,12 @@ func testAcceptance(
 						assertOutput = assertions.NewOutputAssertionManager(t, output)
 						assertOutput.ReportsSuccessfulImageBuild(repoName)
 						assertOutput.ReportsSelectingRunImageMirrorFromLocalConfig(localRunImageMirror)
+						if pack.SupportsFeature(invoke.FixesRunImageMetadata) {
+							t.Log(fmt.Sprintf("image %s was NOT added into 'io.buildpacks.lifecycle.metadata' label", localRunImageMirror))
+							assertImage.HasLabelWithNoData(repoName, "io.buildpacks.lifecycle.metadata", fmt.Sprintf(`"image":"%s"`, localRunImageMirror))
+							t.Log(fmt.Sprintf("run-image %s was added into 'io.buildpacks.lifecycle.metadata' label", runImage))
+							assertImage.HasLabelWithData(repoName, "io.buildpacks.lifecycle.metadata", fmt.Sprintf(`"image":"%s"`, runImage))
+						}
 						cachedLaunchLayer := "simple/layers:cached-launch-layer"
 
 						assertLifecycleOutput := assertions.NewLifecycleOutputAssertionManager(t, output)
@@ -1807,6 +1813,10 @@ func testAcceptance(
 
 								t.Log("uses the run image as the base image")
 								assertImage.HasBaseImage(repoName, runImageName)
+								if pack.SupportsFeature(invoke.FixesRunImageMetadata) {
+									t.Log(fmt.Sprintf("run-image %s was added into 'io.buildpacks.lifecycle.metadata' label", runImageName))
+									assertImage.HasLabelWithData(repoName, "io.buildpacks.lifecycle.metadata", fmt.Sprintf(`"image":"%s"`, runImageName))
+								}
 							})
 						})
 
@@ -2842,6 +2852,14 @@ include = [ "*.jar", "media/mountain.jpg", "/media/person.png", ]
 						var localRunImageMirror string
 
 						it.Before(func() {
+							imageManager.CleanupImages(repoName)
+							pack.RunSuccessfully(
+								"build", repoName,
+								"-p", filepath.Join("testdata", "mock_app"),
+								"--builder", builderName,
+								"--pull-policy", "never",
+							)
+
 							localRunImageMirror = registryConfig.RepoName("run-after/" + h.RandString(10))
 							buildRunImage(localRunImageMirror, "local-mirror-after-1", "local-mirror-after-2")
 							pack.JustRunSuccessfully("config", "run-image-mirrors", "add", runImage, "-m", localRunImageMirror)
@@ -2872,9 +2890,16 @@ include = [ "*.jar", "media/mountain.jpg", "/media/person.png", ]
 					when("image metadata has a mirror", func() {
 						it.Before(func() {
 							// clean up existing mirror first to avoid leaking images
-							imageManager.CleanupImages(runImageMirror)
-
+							imageManager.CleanupImages(runImageMirror, repoName)
 							buildRunImage(runImageMirror, "mirror-after-1", "mirror-after-2")
+
+							pack.RunSuccessfully(
+								"build", repoName,
+								"-p", filepath.Join("testdata", "mock_app"),
+								"--builder", builderName,
+								"--pull-policy", "never",
+							)
+
 						})
 
 						it("selects the best mirror", func() {
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -801,11 +801,11 @@ func testAcceptance(
 							bpSimpleLayersDiffID := "sha256:ade9da86859fa4ea50a513757f9b242bf1038667abf92dad3d018974a17f0ea7"
 							bpReadEnvDiffID := "sha256:db0797077ba8deff7054ab5578133b8f0206b6393de34b5bfd795cf50f6afdbd"
 							// extensions
-							assertImage.HasLabelWithData(builderName, "io.buildpacks.extension.layers", `{"read/env":{"read-env-version":{"api":"0.9","layerDiffID":"`+extReadEnvDiffID+`","name":"Read Env Extension"}},"simple/layers":{"simple-layers-version":{"api":"0.7","layerDiffID":"`+extSimpleLayersDiffID+`","name":"Simple Layers Extension"}}}`)
-							assertImage.HasLabelWithData(builderName, "io.buildpacks.buildpack.order-extensions", `[{"group":[{"id":"read/env","version":"read-env-version"},{"id":"simple/layers","version":"simple-layers-version"}]}]`)
+							assertImage.HasLabelContaining(builderName, "io.buildpacks.extension.layers", `{"read/env":{"read-env-version":{"api":"0.9","layerDiffID":"`+extReadEnvDiffID+`","name":"Read Env Extension"}},"simple/layers":{"simple-layers-version":{"api":"0.7","layerDiffID":"`+extSimpleLayersDiffID+`","name":"Simple Layers Extension"}}}`)
+							assertImage.HasLabelContaining(builderName, "io.buildpacks.buildpack.order-extensions", `[{"group":[{"id":"read/env","version":"read-env-version"},{"id":"simple/layers","version":"simple-layers-version"}]}]`)
 							// buildpacks
-							assertImage.HasLabelWithData(builderName, "io.buildpacks.buildpack.layers", `{"read/env":{"read-env-version":{"api":"0.7","stacks":[{"id":"pack.test.stack"}],"layerDiffID":"`+bpReadEnvDiffID+`","name":"Read Env Buildpack"}},"simple/layers":{"simple-layers-version":{"api":"0.7","stacks":[{"id":"pack.test.stack"}],"layerDiffID":"`+bpSimpleLayersDiffID+`","name":"Simple Layers Buildpack"}}}`)
-							assertImage.HasLabelWithData(builderName, "io.buildpacks.buildpack.order", `[{"group":[{"id":"read/env","version":"read-env-version","optional":true},{"id":"simple/layers","version":"simple-layers-version","optional":true}]}]`)
+							assertImage.HasLabelContaining(builderName, "io.buildpacks.buildpack.layers", `{"read/env":{"read-env-version":{"api":"0.7","stacks":[{"id":"pack.test.stack"}],"layerDiffID":"`+bpReadEnvDiffID+`","name":"Read Env Buildpack"}},"simple/layers":{"simple-layers-version":{"api":"0.7","stacks":[{"id":"pack.test.stack"}],"layerDiffID":"`+bpSimpleLayersDiffID+`","name":"Simple Layers Buildpack"}}}`)
+							assertImage.HasLabelContaining(builderName, "io.buildpacks.buildpack.order", `[{"group":[{"id":"read/env","version":"read-env-version","optional":true},{"id":"simple/layers","version":"simple-layers-version","optional":true}]}]`)
 						}
 					})
 
@@ -1063,10 +1063,10 @@ func testAcceptance(
 						assertImage.HasBaseImage(repoName, runImage)
 
 						t.Log("sets the run image metadata")
-						assertImage.HasLabelWithData(repoName, "io.buildpacks.lifecycle.metadata", fmt.Sprintf(`"image":"pack-test/run","mirrors":["%s"]`, runImageMirror))
+						assertImage.HasLabelContaining(repoName, "io.buildpacks.lifecycle.metadata", fmt.Sprintf(`"image":"pack-test/run","mirrors":["%s"]`, runImageMirror))
 
 						t.Log("sets the source metadata")
-						assertImage.HasLabelWithData(repoName, "io.buildpacks.project.metadata", (`{"source":{"type":"project","version":{"declared":"1.0.2"},"metadata":{"url":"https://github.com/buildpacks/pack"}}}`))
+						assertImage.HasLabelContaining(repoName, "io.buildpacks.project.metadata", (`{"source":{"type":"project","version":{"declared":"1.0.2"},"metadata":{"url":"https://github.com/buildpacks/pack"}}}`))
 
 						t.Log("registry is empty")
 						assertImage.NotExistsInRegistry(repo)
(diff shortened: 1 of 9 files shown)


Pratham1812:
@jjbustamante  @natalieparellano  could you please review and suggest improvements.

jjbustamante:
@Pratham1812 could you run `make format` in your branch and push it to the repo?

Pratham1812:
@jjbustamante I have run make prepare-for-pr could you pls check

internal/builder/builder.go
```diff
@@ -108,16 +108,16 @@ type options struct {
 }
 
 // FromImage constructs a builder from a builder image
-func FromImage(img imgutil.Image) (*Builder, error) {
-	return constructBuilder(img, "", true)
+func FromImage(img imgutil.Image, runImage string, registry string) (*Builder, error) {
```
natalieparellano, line 111: Rather than adding arguments, it might be cleaner to add a `BuilderOption` to support this, e.g.,

```
type BuilderOption func(*options) error

type options struct {
	flatten bool
	exclude []string
	runImage string     <- new
}

func WithRunImage(name string) BuilderOption {
	return func(o *options) error {
		o.runImage = name
		return nil
	}
}
```

internal/builder/builder.go
```diff
@@ -108,16 +108,16 @@ type options struct {
...
+	return constructBuilder(img, runImage, registry, "", true)
 }
 
 // New constructs a new builder from a base image
-func New(baseImage imgutil.Image, name string, ops ...BuilderOption) (*Builder, error) {
-	return constructBuilder(baseImage, name, false, ops...)
+func New(baseImage imgutil.Image, runImage string, registry string, name string, ops ...BuilderOption) (*Builder, error) {
```
natalieparellano, line 116: Same comment as above

internal/builder/builder.go
```diff
@@ -108,16 +108,16 @@ type options struct {
...
+	return constructBuilder(baseImage, runImage, registry, name, false, ops...)
 }
 
-func constructBuilder(img imgutil.Image, newName string, errOnMissingLabel bool, ops ...BuilderOption) (*Builder, error) {
+func constructBuilder(img imgutil.Image, runImage string, registry string, newName string, errOnMissingLabel bool, ops ...BuilderOption) (*Builder, error) {
```
natalieparellano, line 120: Same comment as above

internal/builder/builder.go
```diff
@@ -141,6 +141,10 @@ func constructBuilder(img imgutil.Image, newName string, errOnMissingLabel bool,
 		return nil, err
 	}
 
+	if registry != "" && runImage != "" {
+		metadata.RunImages[0].Image = registry + "/" + runImage
```
natalieparellano, line 145: Rather than replacing the first element, should we override the whole array? So like:

`metadata.RunImages = []RunImageMetadata{{Image: "blah"}}`

pkg/client/build.go
```diff
@@ -332,7 +332,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "getting builder architecture")
 	}
 
-	bldr, err := c.getBuilder(rawBuilderImage)
+	bldr, err := c.getBuilder(rawBuilderImage, opts.RunImage, imgRegistry)
```
natalieparellano, line 335: Do we need to assume the registry from the output image? Why not just pass `opts.RunImage` only (I think we already require the `--run-image` to be fully qualified)?

pkg/client/build_test.go
```diff
@@ -1045,7 +1045,7 @@ api = "0.2"
 					h.AssertNil(t, err)
 					h.AssertEq(t, fakeLifecycle.Opts.Builder.Name(), defaultBuilderImage.Name())
 
-					bldr, err := builder.FromImage(defaultBuilderImage)
+					bldr, err := builder.FromImage(defaultBuilderImage, "", "")
```
natalieparellano, line 1048: It would be awesome to see a test that passes in `some-value` for the run image, and then verifies that `some-value` is present in the image metadata.

natalieparellano (commented):
@Pratham1812 thanks a lot for this! I left a few comments

pkg/client/build_test.go
Pratham1812: Hey! I did the changes as per the above comments, I am not sure how to write tests for this, if you could  guide me further, it would be great.

internal/builder/builder.go
```diff
@@ -103,17 +103,26 @@ type moduleWithDiffID struct {
...
 // FromImage constructs a builder from a builder image
-func FromImage(img imgutil.Image) (*Builder, error) {
-	return constructBuilder(img, "", true)
+func FromImage(img imgutil.Image, runImage string) (*Builder, error) {
+	return constructBuilder(img, "", true, WithRunImage(runImage))
 }
 
 // New constructs a new builder from a base image
-func New(baseImage imgutil.Image, name string, ops ...BuilderOption) (*Builder, error) {
+func New(baseImage imgutil.Image, runImage string, name string, ops ...BuilderOption) (*Builder, error) {
```
jjbustamante, line 124: I think you need to add the `runImage` argument, the `BuilderOption` already allow you to passthrough `WithRunImage` option.

internal/builder/builder.go
```diff
@@ -103,17 +103,26 @@ type moduleWithDiffID struct {
...
+		o.runImage = name
+		return nil
+	}
 }
 
 // FromImage constructs a builder from a builder image
-func FromImage(img imgutil.Image) (*Builder, error) {
-	return constructBuilder(img, "", true)
+func FromImage(img imgutil.Image, runImage string) (*Builder, error) {
+	return constructBuilder(img, "", true, WithRunImage(runImage))
```
jjbustamante, line 120: I think we don't need to add this argument here, that's why you need to change a lot of places in the tests with an empty string

pkg/client/build.go
```diff
@@ -457,7 +457,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		buildEnvs[k] = v
 	}
 
-	ephemeralBuilder, err := c.createEphemeralBuilder(rawBuilderImage, buildEnvs, order, fetchedBPs, orderExtensions, fetchedExs, usingPlatformAPI.LessThan("0.12"))
+	ephemeralBuilder, err := c.createEphemeralBuilder(rawBuilderImage, buildEnvs, order, fetchedBPs, orderExtensions, fetchedExs, usingPlatformAPI.LessThan("0.12"), opts.RunImage, imgRegistry)
```
jjbustamante, line 460: I think this is fine, but I will change the two arguments with just one of them the `runImage String`

pkg/client/build.go
```diff
@@ -1324,9 +1324,11 @@ func (c *Client) createEphemeralBuilder(
 	orderExtensions dist.Order,
 	extensions []buildpack.BuildModule,
 	validateMixins bool,
+	runImage string,
+	registry string,
```
jjbustamante, line 1328: remove this one

pkg/client/build.go
```diff
@@ -332,7 +332,7 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "getting builder architecture")
 	}
 
-	bldr, err := c.getBuilder(rawBuilderImage)
+	bldr, err := c.getBuilder(rawBuilderImage, opts.RunImage)
```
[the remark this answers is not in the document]
jjbustamante: I pulled your branch, built the binary and tested out removing this, and it is still working so I think we don't need it.

jjbustamante:
@natalieparellano could you give me your thoughts here.

Testing the binary, I can see the following output

```bash
pack build -B paketobuildpacks/builder-jammy-base -p ./java-maven  --run-image paketobuildpacks/run-jammy-base:0.0.2  java-app-3
```

In this case we are replacing the `run-image` with `paketobuildpacks/run-jammy-base:0.0.2`, checking the docker config file:

```bash
 docker inspect java-app-3  | jq '.[] | .Config.Labels | ."io.buildpacks.lifecycle.metadata" | fromjson' | jq .
```

```json
{
  "runImage": {
    "topLayer": "sha256:be9eb0a7d50b0438c62376a29b5557f85496a55ae83d41a49f934160ffa09421",
    "reference": "945a5e4052d9b9b024f69bf89f8cf8bd72bc727882ab0b78df9592b37119a31c",
    "image": "paketobuildpacks/run-jammy-base:0.0.2". // This is ok!
  },
  "stack": {
    "runImage": {
      "image": "paketobuildpacks/run-jammy-base:0.0.2".  // Is it ok to change this?
    }
}
```

- Is it ok to change the `stack.runImage.image`? I think we shouldn't/

jjbustamante:
Hi @Pratham1812 could try to solve the merge conflicts and the DCO check ?

internal/builder/builder.go
```diff
@@ -102,10 +102,21 @@ type moduleWithDiffID struct {
...
+  toFlatten buildpack.FlattenModuleInfos
 	labels    map[string]string
 }
 
+func WithRunImage(name string) BuilderOption {
+	return func(o *options) error {
+		o.runImage = name
+		return nil
+	}
+
```
natalieparellano, line 117:
```suggestion
}
```

This should fix the linter

internal/builder/builder.go
```diff
@@ -102,10 +102,21 @@ type moduleWithDiffID struct {
 type BuilderOption func(*options) error
 
 type options struct {
-	toFlatten buildpack.FlattenModuleInfos
+	flatten  bool
+	exclude  []string
```
natalieparellano, line 106:
```suggestion
```
Unused

internal/builder/builder.go
jjbustamante: @Pratham1812 could you do this fix?

jjbustamante:
I tried to fix the acceptance test failures. I focused in [one](https://github.com/buildpacks/pack/blob/main/acceptance/acceptance_test.go#L2872-L2892) of them and here are my findings.

The test case is trying to rebase an **application image** saved in the docker **daemon** with **mirrors metadata** saved with the application image.  

## Behavior before this PR

- First we created a **builder**, and the builder has a `pack-test/run` as run image
- Then we built the application image with a command similar to `pack build localhost:32801/some-org/xeyagowjzc --run-image localhost:32801/run-before/vgdpkyhhne` 
- The output from the previous command is an **application image** `localhost:32801/some-org/xeyagowjzc` with the following `io.buildpacks.lifecycle.metadata` 
```json
"runImage": {
    "topLayer": "sha256:47746006a138ccf6c120beaf5ec55a5589fc6a9b93777082ee9e3b722f3ddf73",
    "reference": "92761540a511123b9c76b9e4fe8c4826c3d19ef1dbfd67ed6b825d6b43546bb6",
    "image": "pack-test/run",
    "mirrors": [
      "localhost:32801/pack-test/run"
    ]
  },
  "stack": {
    "runImage": {
      "image": "pack-test/run",
      "mirrors": [
        "localhost:32801/pack-test/run"
      ]
    }
  }
```
-  Finally we run `pack rebase localhost:32801/some-org/xeyagowjzcand` and the **Test** expects the run image to be selected from the mirror. 

If we take a look at the previous metadata, the `runImage.Image` is using the value from the **Builder** and the `--run-image` used during pack build is not being saved (which is actually the issue this PR is trying to solve)

## Behavior with this PR

The current PR is updating the metadata when `pack build` is invoked, as I mentioned above, we are running `pack build --run-image` to setup the test fixture. The `io.buildpacks.lifecycle.metadata` looks like this with the new code.

```json
"runImage": {
    "topLayer": "sha256:74275d5fce8e3c3343321b75529e7751bd5966664f6af3aec1bb50890c7b6e39",
    "reference": "5d17ab3bdb1b4764f429b059f7a8a1a7f820c4553d2a3f3dfeff69668a5c881a",
    "image": "localhost:32804/run-before/byxrrrztji",
    "mirrors": [
      "localhost:32804/pack-test/run"
    ]
  },
  "stack": {
    "runImage": {
      "image": "localhost:32804/run-before/byxrrrztji",
      "mirrors": [
        "localhost:32804/pack-test/run"
      ]
    }
  }
```
- When running the rebase, the run image selected is `localhost:32804/run-before/byxrrrztji` and not the expected `localhost:32804/pack-test/run`.
I think the metadata is correct, and the problem is with the test case and how to set up the fixture to use the mirrors. I think we need to do **something** to make the [access_checker](https://github.com/buildpacks/pack/blob/main/pkg/image/access_checker.go#L28-L41) to return false when `run-before` image is used, I tried removing the image in the `Before` but it didn't work

internal/builder/builder.go
```diff
@@ -140,6 +148,11 @@ func constructBuilder(img imgutil.Image, newName string, errOnMissingLabel bool,
 		return nil, err
 	}
 
+	if opts.runImage != "" {
+		metadata.RunImages = []RunImageMetadata{{Image: opts.runImage}}
+		metadata.Stack.RunImage.Image = opts.runImage
```
natalieparellano, line 153: Looking at this again, we need to be consistent in how we handle mirrors information. Right now it is being zeroed out for `metadata.RunImages` but preserved for `metadata.Stack.RunImage` (as `metadata.Stack.RunImage.Mirrors` could still have data).

I propose we look in the existing known run images (`RunImageMetadata.Mirrors`) and if the provided `opts.runImage` is found there then preserve the run image information, otherwise zero it out.

natalieparellano:
@jjbustamante since the linked acceptance test is validating rebase behavior we should probably just update the test setup to produce a test image (`repoName`) with the correct metadata. I think if we update [these lines](https://github.com/buildpacks/pack/blob/c36dcf8b668dd17b7db23499f9fa532ae332d885/acceptance/acceptance_test.go#L2783-L2789) in the test setup to remove `--run-image` it might just work, as this particular test doesn't have any expectations around `runBefore`. We might have to duplicate the setup for this test to avoid breaking others in this block.

internal/builder/builder.go
jjbustamante: Yeah, I did that locally, trying to fix the issues with the acceptance but I didn't push it

internal/builder/builder_test.go
```diff
@@ -1846,6 +1846,32 @@ func testBuilder(t *testing.T, when spec.G, it spec.S) {
...
+				var newBuilder *builder.Builder
+				newRunImage := "another/run"
+
+				it.Before(func() {
+					var err error
+					newBuilder, err = builder.New(builderImage, "newBuilder/image", builder.WithRunImage(newRunImage))
+					h.AssertNil(t, err)
+				})
+
+				it("Overrides de RunImageMetadata", func() {
```
natalieparellano, line 1863: Maybe we should clarify somewhere (in a code comment or something) that pack always recreates stack.toml and run.toml from the metadata when creating the ephemeral builder. Because we're not just overriding the metadata, we're changing what the lifecycle encounters on the filesystem.

jjbustamante:
Hi @Pratham1812! 

We made some changes to your code to get it into the last mile! but we need you to fix the DCO issues to get this PR merged.  Is it possible for you to do that?

jjbustamante:
@Pratham1812 

The steps required to sign-off your commits are describe [here](https://github.com/buildpacks/pack/pull/2022/checks?check_run_id=22350701199), the most important are:

To add your Signed-off-by line to every commit in this branch:

1. Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
2. In your local branch, run: git rebase HEAD~16 --signoff
3. Force push your changes to overwrite the branch: git push --force-with-lease origin alpha/patch

It is very important that you sign-off your commits in order for us to merge the PR! Otherwise we can't do it

jjbustamante:
Oh man! I think you didn't update your main fork before running the rebase right? and we still have some commits without signing.

internal/builder/builder_test.go
natalieparellano, line 1863:
```suggestion
				it("overrides the run image metadata (which becomes run.toml)", func() {
```

acceptance/assertions/image.go
```diff
@@ -69,6 +69,15 @@ func (a ImageAssertionManager) HasLabelWithData(image, label, data string) {
 	a.assert.Contains(label, data)
 }
 
+func (a ImageAssertionManager) HasLabelWithNoData(image, label, data string) {
```
natalieparellano, line 72:
```suggestion
func (a ImageAssertionManager) HasLabelNotContaining(image, label, data string) {
```

Maybe we could change the above `HasLabelWithData` -> `HasLabelContaining` for parity. `HasLabelWithNoData` makes me think that we're asserting the label is completely empty.

acceptance/acceptance_test.go
```diff
@@ -1085,6 +1085,12 @@ func testAcceptance(
 						assertOutput = assertions.NewOutputAssertionManager(t, output)
 						assertOutput.ReportsSuccessfulImageBuild(repoName)
 						assertOutput.ReportsSelectingRunImageMirrorFromLocalConfig(localRunImageMirror)
+						if pack.SupportsFeature(invoke.FixesRunImageMetadata) {
+							t.Log(fmt.Sprintf("image %s was NOT added into 'io.buildpacks.lifecycle.metadata' label", localRunImageMirror))
```
natalieparellano, line 1089:
```suggestion
							t.Log(fmt.Sprintf("run image mirror %s was NOT added into 'io.buildpacks.lifecycle.metadata' label", localRunImageMirror))
```

natalieparellano (approved):

PR #2028, opened 2024-01-20
feat: added a mac-address flag to build option, so that user can manually set the MAC-address.

### LINKED ISSUE #1779: Make specifying a MAC address for the container used for building possible
### Description
Some software licenses are tied to a MAC address of a network interface. It would be nice to use these in company-internal buildpacks.

### Proposed solution
`docker run` allows using the `--mac-address` parameter to specify the MAC address for the network device of the container. `pack` could pass this on.

Build the stuff manually and just have a buildpack that creates the container

Edit: Here is a SO post that explains a workaround how to get a custom MAC address in the `docker build` phase:
https://stackoverflow.com/questions/48675505/is-it-possible-to-set-a-mac-address-for-docker-build/48676884#48676884

Edit 2: Not trying to get around the license, these buildpacks are only used by the legit license owners, but I want to streamline and automate the build process.

Parthiba-Hazra:
Resolves #1779


jjbustamante:
I am going to revert this PR based on the following:

- After it was merged, @matejvasek pointed me out that it probably wont work and he created this [PR](https://github.com/buildpacks/pack/pull/2063) to do some changes 
- I added an acceptance test to check the new flag added, locally it worked but in our CI system we have some constraints summarized in this [comment](https://github.com/buildpacks/pack/pull/2063#issuecomment-2038046319)

We will need to wait for the ubuntu runner to update the docker >= 25.0 and we also need to update the windows runner with the same version, Once that is done, I think we can merge again this code + the changes made by @matejvasek  and the acceptance test.

PR #2034, opened 2024-01-24
Adding `--label` flag to `pack builder create` command

### LINKED ISSUE #2033: Ability to add image labels to buildpacks during `pack builder create`
### Description
We would like to be able to add metadata to our builder via image labels at build time but there is no way for us to add custom image labels.

### Proposed solution
Add a flag `--label <label>=<value>` to `pack builder create` command that will allow buildpack authors to add custom metadata to the builder image at build time.

This can be achieved with `crane mutate` but that requires you to upload the image without metadata to a registry and then mutate the file. This also means that image labels cannot be added to the `.cnb` files at all which feels bad as the OCI image in registry and in `.cnb` file format should be identical. 

### Additional context
- This request is a follow up a similar feature added to `pack buildpack package` command. See issue [1867](https://github.com/buildpacks/pack/issues/1867)
-  Discussion in slack [here](https://cloud-native.slack.com/archives/C0331B61A1Y/p1692112371262099)

jjbustamante:
## Summary
This PR adds a new `--label` flag to the `pack builder create` command 

#### Before

The flag didn't exist before

#### After

New flag will look like this, when running `pack builder create --help`

```bash
Flags:
  -l, --label stringToString        Labels to add to the builder image, in the form of '<name>=<value>' (default [])
```

Running a local builder creation like 

```bash
> out/pack builder create cnbs/local-builder:jammy --config builder.toml --label "foo=bar"
# then we inspect the labels
>  docker inspect cnbs/local-builder:jammy | jq '.[] | .Config.Labels' | jq .
```

output

```json
{
  "foo": "bar",
  "io.buildpacks.base.distro.name": "ubuntu",
  "io.buildpacks.base.distro.version": "\"22.04\"",
  "io.buildpacks.builder.metadata": "...."
}
```

As we can see the label "foo=bar" was set to the builder image

## Documentation

    - [x] Yes, see #__
    - [ ] No

## Related

Resolves #2033

### DIFF
--- a/internal/builder/builder.go
+++ b/internal/builder/builder.go
@@ -103,6 +103,7 @@ type BuilderOption func(*options) error
 
 type options struct {
 	toFlatten buildpack.FlattenModuleInfos
+	labels    map[string]string
 }
 
 // FromImage constructs a builder from a builder image
@@ -139,6 +140,13 @@ func constructBuilder(img imgutil.Image, newName string, errOnMissingLabel bool,
 		return nil, err
 	}
 
+	for labelKey, labelValue := range opts.labels {
+		err = img.SetLabel(labelKey, labelValue)
+		if err != nil {
+			return nil, errors.Wrapf(err, "adding label %s=%s", labelKey, labelValue)
+		}
+	}
+
 	bldr := &Builder{
 		baseImageName:        img.Name(),
 		image:                img,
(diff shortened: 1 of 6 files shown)


jjbustamante:
@matejvasek  could you take a look at the [binaries](https://github.com/buildpacks/pack/actions/runs/7643303416)

@natalieparellano could you give me a review here?

matejvasek:
@jjbustamante I do not use this as a binary but rather as a library. Regardless this PR seems to work for me.

natalieparellano (approved):

PR #2063, opened 2024-02-14
[TESTING ONLY]

matejvasek:


jjbustamante:
I created simple acceptance test with this code, and @matejvasek you were right I got the following error:

```bash
  Host Settings:
          Binds: 'pack-cache-some-org_gbmqdqekqf_latest-bc166514d2fd.build:/cache pack-layers-smuiylvmuq:/layers pack-app-ohcxidehgd:/workspace'
          Network Mode: 'host'
        ERROR: failed to build: executing lifecycle: failed to create 'creator' container: "specify mac-address per network" requires API version 1.44, but the Docker daemon API version is 1.38
``` 

We are actually using [Docker API version ](https://github.com/buildpacks/pack/blob/main/pkg/client/client.go#L197) I will try to update and push the acceptance test to this branch

jjbustamante:
Damn, the linux [runner](https://github.com/actions/runner-images/blob/ubuntu22/20240324.2/images/ubuntu/Ubuntu2204-Readme.md#tools) is using Docker Server 24.0.9, and this version only [support](https://docs.docker.com/engine/api/#api-version-matrix) version 1.43

jjbustamante:
This table shows the Docker version used by pack and the runners, we need to move to 25.0 in order to support setting the `mac-address` but, the ubuntu latest runner doesn't support that version yet, also, we will need to update the self hosted windows runner

| Docker version | Maximum API version | Runner Version | Note |
| --------------- | --------------------- | ----------------| ------|
| 26.0 | 1.45 | | | 
| 25.0 | 1.44 | | Expected version |
| 24.0 | 1.43 | Linux runner | |
| 23.0 | 1.42 | | |
| 20.10 | 1.41 | Windows runner | <- I upgrade till this one in this PR |
| 19.03 |  1.40 | | |
| 18.09 | 1.39 | | |
| 18.06 | 1.38 | | current pack version |

PR #2075, opened 2024-02-20
Add feature to `pull-policy`  for pulling images according to user specified pulling interval (e.g., hourly, daily, weekly, interval=5d4h30m )

### LINKED ISSUE #1368: Add support for a pull policy that only pulls new images periodically
### Description
Pack's default pull policy is `always`, which means that on every `pack build`, it attempts to pull images (eg the builder image).

This is great for ensuring users don't accidentally build against outdated images, however:
- It adds several seconds to every `pack build` (even for a no-op when images are up to date)
- It counts against the container registry's rate limits (even for a no-op), which is inconvenient given Docker Hub and ECR's fairly low rate limits for unauthenticated requests.

These issues are particularly problematic when running integration tests, where `pack build` will be run many times.

Using the `if-not-present` pull policy (either via Pack global config, or by passing `--pull-policy if-not-present` to `pack build`), prevents these issues, however:
- For CI: It's easy to forget to set the global pack pull policy to `if-not-present`, which can mean much slower integration tests (which can be hard to spot, given the Pack stdout mentioning pulling would typically be hidden in the test runner output unless the test fails).
- When using pack locally: It risks using out of date images.

### Proposed solution
- Initially: Add support for a new Pack pull policy option (alongside the existing `always` and `if-not-present` options), that only pulls new images if the time since last pull of those images is greater than N hours/days.
- Later on: Make this new pull policy the default, instead of `always`. This would (a) mean users get the performance benefit locally without having to know that the pull policy option exists, (b) reduce the boilerplate needed (and chance of forgetting to set it) in CI configs.

### Open questions
- Should the new pull policy allow the user to configure the time interval?
  - If yes, via what syntax? eg `interval=1d`
  - If no, what should the hardcoded interval be? 1 day?
- What should the new pull policy be called? Examples:
  - `--pull-policy interval=1d`
  - `--pull-policy if-not-recent`
  - `--pull-policy if-not-pulled-recently`
- How should the state of last pull time for each image be tracked?

- Modify integration test runners to pass `--pull-policy if-not-present` to `pack build` **both locally and in CI**:
  - Pros: This will speed up integration tests everywhere.
  - Cons: Users can potentially be integration testing against out of date images locally (if they don't also run `pack build` manually from time to time, which would pull). Doesn't help speed up manual repeat `pack build`s locally when developing.
- Modify integration test runners to pass `--pull-policy if-not-present` to `pack build` **only in CI**:
  - Pros: This will speed up integration tests in CI. No risk of running integration tests locally against stale images.
  - Cons: Doesn't help speed up integration tests locally, or running manual repeat `pack build`s locally when developing. Means having to add vendor-specific CI checks (eg known env vars) in integration test runners.
- In the CI config only, set the Pack global pull policy to `if-not-present`:
  - Pros: This will speed up integration tests in CI. No risk of running integration tests locally against stale images. No vendor-specific checks in integration test runners.
  - Cons: Doesn't help speed up integration tests locally, or running manual repeat `pack build`s locally when developing. Easy to forget to configure (particularly when the end user projects are owned by different users to the test runner, so less familiar with best practices).

### Additional context
- Earlier discussion in the original pull-policy RFC: https://github.com/buildpacks/rfcs/pull/80#issuecomment-644120544
- https://github.com/Malax/libcnb.rs/issues/306

Parthiba-Hazra:
## Summary
This is the minimal PR for issue - #1368 .We need to discuss about the correctness and quality of changes. I didn't add any test cases yet as some functionalities might change after some reviews, will add those tests gradually as this PR keeps progressing.
## Output
Screenshot from 2024-02-21 00-14-15
Screenshot from 2024-02-21 00-17-07
Screenshot from 2024-02-21 00-02-43
Screenshot from 2024-02-21 00-03-36

## Documentation
Right now it's not documented yet.

    - Yes
 
## Related
#1368 

Resolves #1368


Makefile
```diff
@@ -146,6 +150,12 @@ package: out
 install:
 	mkdir -p ${DESTDIR}${BINDIR}
 	cp ./out/$(PACK_BIN) ${DESTDIR}${BINDIR}/
+	@if [ ! -d "$(IMAGE_JSON_DIR)" ]; then \
+		mkdir -p $(IMAGE_JSON_DIR); \
+	fi
+	@if [ ! -f "$(IMAGE_JSON_FILE)" ]; then \
+		echo '{ "interval": { "pulling_interval": "", "pruning_interval": "7d", "last_prune": "" }, "image": {} }' > $(IMAGE_JSON_FILE); \
+	fi
```
jjbustamante, line 158: Is it possible to initialized the `image.json` file during the configuration? I mean when `pack config pull-policy hourly | daily | weekly | interval=<_d_h_m>` is enabled?

pkg/image/fetcher.go
```diff
@@ -120,7 +137,17 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
 	}
 
-	return f.fetchDaemonImage(name)
+	image, err := f.fetchDaemonImage(name)
+	if err != nil {
+		return nil, err
+	}
+
+	// Update image pull record in the JSON file
+	if err := f.updateImagePullRecord(name, time.Now().Format(time.RFC3339)); err != nil {
```
jjbustamante, line 146: Do we want to keep track of every pulling even if the **interval policy** is not enable?

pkg/image/fetcher.go
```diff
@@ -106,6 +107,22 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
 		if err == nil || !errors.Is(err, ErrNotFound) {
 			return img, err
 		}
+	case PullWithInterval, PullDaily, PullHourly, PullWeekly:
+		img, err := f.fetchDaemonImage(name)
```
jjbustamante, line 111: Shouldn't we check the interval first before deciding to fetch from the daemon? 

```go
if pull, _ := f.CheckImagePullInterval(name); !pull {
   img, err := f.fetchDaemonImage(name)
   if err == nil {
      // removes it from the image.json file, maybe it was deleted
   }
   return img, err
} 
```

Makefile
Parthiba-Hazra: yea, I think so.

pkg/image/fetcher.go
Parthiba-Hazra: I also thought about this, and I think keeping the pulling track even when interval policy is not enable will be good. cause let's suppose user enable it with some x days, and user might thought as the image pull recently(when interval policy was not enabled) it will not pull again, but if we don't track pulling then it will pull again. I mean the effect of policy will start after one pull of each image.

pkg/image/fetcher.go
Parthiba-Hazra: I think it's related to this [message](https://github.com/buildpacks/pack/issues/1368#issuecomment-1956723355)
I will update it soon.

Makefile
Parthiba-Hazra: @jjbustamante I think the current implementation already has this support, cause when ever we are calling the `ReadImageJSON` function in [pull_policy.go](https://github.com/buildpacks/pack/pull/2075/files/3bd5c2dc25db28e2e773ed00cb02e7f1b4f1ce01#diff-10ca3b182bbdbfb8cc63caaa60de64594482471c61f73144f6ecf03fcdb751bf) we are checking image.json is present or not if not then we create it with a minimal JSON, and eventually whenever we configure `pack config pull-policy hourly | daily | weekly | interval=<_d_h_m>` the `ReadImageJSON` function will be called via `updateImageJSONDuration` function.

So should I remove those changes from the `MAKEFILE` ?

jjbustamante: Yeah, I think we can then remove it from the make file, and pack itself can take care of initialized the filesystem with everything it needs

pkg/image/fetcher.go
jjbustamante: My only concern is to break some current behaivor for end-users and they don't know why, let's suppose for some reason we introduced a bug with the interval policy code, maybe with the hard drive space on environments that is limited. 
It could affect the users and they don't even using the feature, that's why I like to keep things as they are and only if they enable the feature we start keeping track of the data.

Parthiba-Hazra: Hmm, that's a fair point, thank you for addressing it.

Parthiba-Hazra: @jjbustamante  should I go for the test cases now? or wait some more time to get reviews from you or other maintainers?

jjbustamante: I think you can start working in your test coverage yes.

pkg/image/fetcher.go
```diff
@@ -120,7 +162,19 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
 		return nil, err
 	}
 
-	return f.fetchDaemonImage(name)
+	image, err := f.fetchDaemonImage(name)
+	if err != nil {
+		return nil, err
+	}
+
+	if options.PullPolicy == PullWithInterval || options.PullPolicy == PullHourly || options.PullPolicy == PullDaily || options.PullPolicy == PullWeekly {
```
jjbustamante, line 170: Maybe we can externalize the guard to a method.

```go 
if intervalPolicy(options) {
  // Update image pull record in the JSON file
  if err := f.updateImagePullRecord(name, time.Now().Format(time.RFC3339)); err != nil {
    return nil, err
  }
}
```

jjbustamante:
@Parthiba-Hazra this is looking really good! thanks for your amazing work!!

I think your tests are failing because you need to run `make format` to fix some formatting issues

Screenshot 2024-03-04 at 10 37 10 AM

Parthiba-Hazra:
> @Parthiba-Hazra this is looking really good! thanks for your amazing work!!
> 
> I think your tests are failing because you need to run `make format` to fix some formatting issues
> 
> Screenshot 2024-03-04 at 10 37 10 AM
> 
> You can run `make verify` in your local environment before pushing, usually we forget about it

@jjbustamante currently I'm still working on test cases .. actually getting some errors while running test cases of `fetcher_test.go`, it's saying it can't connect to docker daemon cause it's not running although it is running. Let me resolve this and add the all necessary test cases ... I will try to finish it withing this week.

pkg/image/pull_policy.go
```diff
@@ -1,41 +1,312 @@
...
+
+	if err := WriteFile(updatedJSON); err != nil {
+		return errors.Wrap(err, "failed to write updated image.json")
+	}
+
+	return nil
+}
+
+func WriteFile(data []byte) error {
+	cmd := exec.Command("sudo", "sh", "-c", "echo '"+string(data)+"' > "+imagePath)
```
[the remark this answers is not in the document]
Parthiba-Hazra: @jjbustamante I was trying it like this - 
```
func WriteFile(data []byte, path string) error {
	file, err := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0644)
	if err != nil {
		return errors.New("failed to open file: " + err.Error())
	}
	defer file.Close()

	_, err = file.Write(data)
	if err != nil {
		return errors.New("failed to write data to file: " + err.Error())
	}

	return nil
}
```
but there is some kind of bug, it can't detect the actual path (`$HOME/.pack/image.json`) of `image.json` it's trying to write the file under `/root/.pack/ `

internal/commands/config_prune_interval.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/buildpacks/pack/internal/config"
+	"github.com/buildpacks/pack/internal/style"
+	"github.com/buildpacks/pack/pkg/image"
+	"github.com/buildpacks/pack/pkg/logging"
+)
+
+func ConfigPruneInterval(logger logging.Logger, cfg config.Config, cfgPath string) *cobra.Command {
```
jjbustamante, line 17: I think you can use 
```go
cfg config.Config 
cfgPath string
```
To get the path where you want to save the `image.json`,  `cfgPath` has the path to the `config.toml` file, then you can use `filepath.Dir(cfgPath)` to get the parent directory.

pkg/image/pull_policy.go
```diff
@@ -1,41 +1,288 @@
...
+
+	updatedJSON, err := json.MarshalIndent(imageJSON, "", "    ")
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal updated records")
+	}
+
+	return WriteFile(updatedJSON)
+}
+
+func ReadImageJSON(l logging.Logger) (*ImageJSON, error) {
```
jjbustamante, line 146: What about creating a handler for these methods? I think this `ReadImageJSON` has a lot of responsibilities:
- It is trying to discover the user's home directory
- it creates the whole folder structure to save the `image.json` on the disk
- it reads and writes! 

```golang
type ImagePullPolicyManager struct {
	Logger logging.Logger
}

func (i *ImagePullPolicyManager) Read(path string) (*ImageJSON, error) {
  	imagePath := filepath.Join(path, "image.json")
	// Check if the file exists, if not, create it with minimum JSON
	if _, err := os.Stat(imagePath); os.IsNotExist(err) {
                // We don't need to write on disk, we are reading, right? 
                // We return our default 
		return &ImagePullPolicy{
			Interval: &Interval{
				PullingInterval: "7d",
				LastPrune:       "",
			},
			Image: &ImageData{},
		}, nil
	}
        // do what you are doing today
}

func (i *ImagePullPolicyManager) Write(imagePullPolicy ImagePullPolicy, path string) error {
	// I will move all the marshaling logic into these methods, I don't want the method caller to worry about them
        updatedJSON, err := json.MarshalIndent(imagePullPolicy, "", "    ")
	if err != nil {
		return errors.Wrap(err, "failed to marshal updated records")
	}
	return WriteFile(updatedJSON, filepath.Join(path, "image.json"))
}

func WriteFile(data []byte, path string) error {
        // Try this one, it will create the file if it doesn't exits.
	file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
	if err != nil {
		return errors.New("failed to open file: " + err.Error())
	}
	defer file.Close()
	_, err = file.Write(data)
	if err != nil {
		return errors.New("failed to write data to file: " + err.Error())
	}
	return nil
}

```

Parthiba-Hazra: Yes, this looks way better than what I was currently doing.

pkg/image/pull_policy.go
```diff
@@ -1,41 +1,296 @@
...
+		return "", errors.Wrap(err, "getting pack home")
+	}
+	return filepath.Join(home, "image.json"), nil
+}
+
+var nameMap = map[string]PullPolicy{"always": PullAlways, "hourly": PullHourly, "daily": PullDaily, "weekly": PullWeekly, "never": PullNever, "if-not-present": PullIfNotPresent, "": PullAlways}
+
+// ParsePullPolicy from string with support for interval formats
+func ParsePullPolicy(policy string, logger logging.Logger) (PullPolicy, error) {
+	pullPolicyManager := NewPullPolicyManager(logger)
```
Parthiba-Hazra, line 79: @jjbustamante I'm not sure about spawning the manager here. should we create the manager in the caller function of `ParsePullPolicy` and then pass it to the function?

jjbustamante: Yeah, let add some suggestions about this. 

First of all, a simplify class diagram of your current implementation looks like this:

```mermaid
classDiagram
    Client o-- ImageFetcher
    Fetcher  ..|>  ImageFetcher
    Fetcher o-- ImagePullChecker 
    ImagePullPolicyManager ..|> ImagePullChecker

    Client : +bool experimental
    Client : +map[string]string registryMirrors
    Client : +string version
    Client : +authn.Keychain

    Client --> ImagePullPolicyManager: Remove it!
    ImagePullChecker ..> Fetcher : Remove it!
  
    class  ImageFetcher {
          <<interface>>
    }

    class Fetcher {
    }

    class ImagePullChecker {
        <<interface>>
        CheckImagePullInterval()
	Read()
	PruneOldImages(f *Fetcher) error
	UpdateImagePullRecord()
        Write()
    }

    class ImagePullPolicyManager {
    }
```

The first thing I noticed that doesn't look very health is the circular dependency between `ImagePullChecker` and `Fetcher` in the method `PruneOldImages(f *Fetcher) error`, we need to remove that, I noticed what you are trying to do is to load a local image from the daemon to check if the image exists, maybe you can use the `imgutil.Local` image to do that, but we need to remove that circular dependency.

The other thing I will suggest is:
- You are creating an instance of `NewPullPolicyManager(logger)` in 4 places right now, I think we can:
   -  hold a reference into the Client to an `image.ImagePullChecker` and add an option to set it up
   ```go
type Client struct {
   imagePullChecker    image.ImagePullChecker
}

func WithImagePullChecker(i image.ImagePullChecker) Option {
       return func(c *Client) {
               c.imagePullChecker = i
       }
}
```
- After that, you can move the creation of the PullPolicyManager to the `cmd/cmd.go` just right before creating the client. Update the method to create the client to use the `Option` we declared before
```go
imagePullChecker := image.NewPullPolicyManager(logger)
packClient, err := initClient(logger, cfg, imagePullChecker)

func initClient(logger logging.Logger, cfg config.Config, imagePullChecker image.ImagePullChecker) (*client.Client, error) {
 return client.NewClient(client.WithLogger(logger), client.WithExperimental(cfg.Experimental), ... client.WithImagePullChecker(imagePullChecker))
}
```
- I will move the `ParsePullPolicy` method to be part of the `ImagePullChecker` interface
```go
func (i *ImagePullPolicyManager) ParsePullPolicy(policy string) (PullPolicy, error) {
  // Do what we have today
}
```
- Because of the previous change, all the commands are broken now, I will update the commands that need to call  `ParsePullPolicy` and passthrough the `ImagePullChecker`, for example in `cmd/cmd.go`
```go
 // remember we have the imagePullChecker at this point, because cmd is creating it.
 rootCmd.AddCommand(commands.Build(logger, cfg, packClient, imagePullChecker))
```
- I will repeat the same thing for all the other commands that are broken

pkg/image/fetcher.go
```diff
@@ -35,6 +36,22 @@ type LayoutOption struct {
...
+	PruneOldImages(f *Fetcher) error
+	UpdateImagePullRecord(path string, imageID string, timestamp string) error
+	Write(imageJSON *ImageJSON, path string) error
+}
+
+func intervalPolicy(options FetchOptions) bool {
+	return options.PullPolicy == PullWithInterval || options.PullPolicy == PullHourly || options.PullPolicy == PullDaily || options.PullPolicy == PullWeekly
+}
+
+func NewPullPolicyManager(logger logging.Logger) *ImagePullPolicyManager {
```
jjbustamante, line 51: I think this method should return the interface `ImagePullChecker`, this will break the `ParsePullPolicy` but if you do the refactoring I suggested you will be able to call the internal methods

pkg/image/fetcher.go
```diff
@@ -35,6 +36,22 @@ type LayoutOption struct {
 	Sparse bool
 }
 
+type ImagePullChecker interface {
```
jjbustamante, line 39: I am not totally convinced about this name, I do not have any suggestion, but not sure. I will try to think about it

Parthiba-Hazra: @jjbustamante what about `ImagePullPolicyService` or `ImagePullPolicyHandler`

jjbustamante: I like ** ImagePullPolicyHandler**. Also, can you move the interface definition to the [client](https://github.com/buildpacks/pack/blob/main/pkg/client/client.go#L89) file? I think it must be along with the others

Parthiba-Hazra: @jjbustamante please one more question- should we need to create a mock of `NewPullPolicyManager` method for test cases (I think we should) - 
```
func NewPullPolicyManager(logger logging.Logger) ImagePullChecker {
	return &ImagePullPolicyManager{Logger: logger}
}
```
or can use it directly during tests.

Parthiba-Hazra: > I like ** ImagePullPolicyHandler**. Also, can you move the interface definition to the [client](https://github.com/buildpacks/pack/blob/main/pkg/client/client.go#L89) file? I think it must be along with the others

if we move the `ImagePullChecker` interface to the client file then there will be a import cycle between client and image package cause `Fetcher` struct has a field `imagePullChecker` type `ImagePullChecker`   interface.

jjbustamante: Oh! .. ok, let's keep it there for now, we will come back to that later

internal/commands/build_test.go
```diff
@@ -38,21 +38,23 @@ func TestBuildCommand(t *testing.T) {
...
+		cfg                    config.Config
+		imagePullPolicyHandler image.ImagePullPolicyHandler
 	)
 
 	it.Before(func() {
 		logger = logging.NewLogWithWriters(&outBuf, &outBuf)
 		cfg = config.Config{}
 		mockController = gomock.NewController(t)
 		mockClient = testmocks.NewMockPackClient(mockController)
+		imagePullPolicyHandler = image.NewPullPolicyManager(logger)
```
Parthiba-Hazra, line 55: @jjbustamante  currently we are spawning `imagePullPolicyHandler` for every test cases. Is it ok? or do we need some kind of mock here ?

jjbustamante: I think we can create a mock, at this point we don't want to use the real component because we want to test only the build logic.

internal/commands/extension_package.go
```diff
@@ -45,7 +45,7 @@ func ExtensionPackage(logger logging.Logger, cfg config.Config, packager Extensi
 				stringPolicy = cfg.PullPolicy
 			}
 
-			pullPolicy, err := image.ParsePullPolicy(stringPolicy)
+			pullPolicy, err := imagePullPolicyHandler.ParsePullPolicy(stringPolicy, logger)
```
jjbustamante, line 48: I think the `imagePullPolicyHandler` already holds a logger instance, we can remove the `logger` parameter from this method.

internal/commands/package_buildpack_test.go
```diff
@@ -226,7 +229,9 @@ func testPackageBuildpackCommand(t *testing.T, when spec.G, it spec.S) {
 					configPath: "/path/to/some/file",
 				}
 
-				cmd := commands.PackageBuildpack(config.logger, config.clientConfig, config.buildpackPackager, config.packageConfigReader)
+				imagePullChecker := image.NewPullPolicyManager(config.logger)
```
jjbustamante, line 232: rename the variable from `imagePullChecker` to `pullPolicyManager`

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/buildpacks/imgutil v0.0.0-20240118145509-e94a1b7de8a9
 	github.com/buildpacks/lifecycle v0.18.5
 	github.com/docker/cli v25.0.3+incompatible
-	github.com/docker/docker v25.0.3+incompatible
```
jjbustamante, line 11: I think you need to update your fork, and rebase your branch to get the latest dependencies updates

Parthiba-Hazra: @jjbustamante codecov/patch demanding some test cases yet.. should we work on those or we can ignore that ?

Parthiba-Hazra:
> Hi! Thank you so much for working on this :-D (I'm the person that asked for this in #1368)
> 
> I notice that the implementation here currently supports both:
> 
> * named durations (eg `hourly`, `daily`, `weekly`)
> * and custom durations (`interval=<_d_h_m>`)
> 
> I wonder if we could simplify things and support only one or the other?
> 
> My instinct would be to say the named values would cover 95% of needs and so perhaps we don't need the `interval=<_d_h_m>` form? (Bearing in mind, the main use-case here is to prevent excessive pulls during CI runs, or to reduce hassle/download requirements when building locally and the upstream builder rapidly changes; both of which are handled by the named `hourly`, `daily`, `weekly` durations. Anyone who needs anything more advanced can use a pull policy of `never` and handle their own pulling or similar.) Removing `interval=<_d_h_m>` would also mean the parsing implementation can be simplified, and from an end-user perspective means the config options are more consistent (ie: all words, not some words, some key-value pairs etc.)
> 
> Also, I see that a new `prune-interval` config option was added to control when to remove image entries from the `.pack/image.json` file. If we were to set an upper bound for the pull interval duration (eg say the interval is not allowed to be longer than N duration), we could potentially simplify things further, and remove `prune-interval`, and instead set pruning to be that max value?
> 
> Combining all of the above, if we were to remove support for `interval=<_d_h_m>`, then we know the max possible pull-policy duration is `weekly`, and as such the auto-pruning can simply prune any entry older than a week, meaning we can then remove the `prune-interval` config option.
> 
> Thoughts? :-)

@edmorley thank you for your review, sure we can do your cahanges easily, actually I implement this after going through all the maintainer's comments on that issue and I found out this would be the most flexible solution. @jjbustamante @jkutner any thought on this ?

jjbustamante:
> @edmorley thank you for your review, sure we can do your cahanges easily, actually I implement this after going through all the maintainer's comments on that issue and I found out this would be the most flexible solution. @jjbustamante @jkutner any thought on this ?

I am happy with @edmorley suggestions if we can support 95% of the use cases and keep the code simpler, let's go for it!

Parthiba-Hazra:
> > @edmorley thank you for your review, sure we can do your cahanges easily, actually I implement this after going through all the maintainer's comments on that issue and I found out this would be the most flexible solution. @jjbustamante @jkutner any thought on this ?
> 
> I am happy with @edmorley suggestions if we can support 95% of the use cases and keep the code simpler, let's go for it!

Sure, I'm gonna work on this. Thanks.

pkg/image/fetcher.go
```diff
@@ -246,3 +306,46 @@ func (w *colorizedWriter) Write(p []byte) (n int, err error) {
...
+	if err != nil {
+		return err
+	}
+
+	if imageJSON.Image.ImageIDtoTIME == nil {
+		imageJSON.Image.ImageIDtoTIME = make(map[string]string)
+	}
+	imageJSON.Image.ImageIDtoTIME[imageID] = timestamp
+
+	err = i.Write(imageJSON, path)
```
Parthiba-Hazra, line 321: @jjbustamante I think it's not a good idea to call `Read` and `Write` under every function. Should we handle it in the `Fetch` method only? What do you think?

natalieparellano:
Hi there - just checking in on this one. Do we have a sense of the effort left to complete it?

Parthiba-Hazra:
> Hi there - just checking in on this one. Do we have a sense of the effort left to complete it?

@natalieparellano actually I implemented a complex version of this feature. and as @edmorley suggestion we should keep it simpler as that will enough for the actual use cases, and as far I can remember I request some suggestion from @jjbustamante and I think he must be busy on something else. anyway what should we do here, I can still work on it to get it done.

jjbustamante:
I moved this one to milestone 0.35 because 0.34 was already big enough,  but I  think we can start merging the latest changes into it

Issue #1916, opened 2023-09-21
Support Insecure Registries


dlion:
I'm opening this issue to track the work that needs to be done on pack to enable the support for [insecure registries](https://github.com/buildpacks/rfcs/blob/main/text/0111-support-insecure-registries.md).

## Sample Command
```sh
pack build test_img --path apps/test-app --insecure-registry=insecure.host --builder insecure.host/my-builder 
```

## Dependencies

Once [this PR](https://github.com/buildpacks/lifecycle/pull/1140) is merged the work on this issue can be started

Main tracking issue: https://github.com/buildpacks/rfcs/issues/246

jjbustamante:
Waiting for  https://github.com/buildpacks/lifecycle/pull/1140

jjbustamante:
We reviewed the Platform 0.13 scope yesterday during our WG meeting, I think we will unblock this issue soon

natalieparellano:
It should be possible to work on this one already - though there isn't a released lifecycle that supports it, we should be able to verify that the flag is passed all the way down from the command line invocation to the "lifecycle execution"

jjbustamante:
Thank you very much @natalieparellano, all those hints are great for new contributors.

> Questions:
> Do we need to add warnings anywhere when insecure registry options are provided?

I will say **Yes**, using an insecure registry is probably something just for development and it is valuable to add some warnings

jjbustamante:
@prashantrewar what about this one? do you want to give it a try?

prashantrewar:
> @prashantrewar what about this one? do you want to give it a try?

Sure, I will try.

Thank you so much @jjbustamante

prashantrewar:
Hey @jjbustamante and @natalieparellano I have a question related to `image.NewFetcher`.

[pkg/client/client.go](https://github.com/buildpacks/pack/commit/bd2adb94f0865972c89e6e16fce52f5b96279e55#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R242)

How should I pass the `insecureRegistries` values to the client?

jjbustamante:
hi @prashantrewar!

I think for the rebase we need the following:

- Add a new string slice flag to `pack rebase` command to get the insecure registry values
- When we are fetching the application images and the base image we need to pass through this value when invoking the [fetching](https://github.com/buildpacks/pack/blob/main/pkg/client/rebase.go#L59-L113).

prashantrewar:
[...]

PR #2077, opened 2024-02-21
Support Insecure Registries

prashantrewar:
## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

 
Resolves #___
#1916


internal/build/lifecycle_execution.go
```diff
@@ -328,6 +328,10 @@ func (l *LifecycleExecution) Create(ctx context.Context, buildCache, launchCache
 		flags = append(flags, "-uid", strconv.Itoa(l.opts.UID))
 	}
 
+	for _, reg := range l.opts.InsecureRegistries {
```
jjbustamante, line 331: I think we need to add a validation for platform API version, something like

```go
if l.platformAPI.AtLeast("0.13") {
  // we can add the `-insecure-registry` flag when calling the lifecycle.
}
```

pkg/client/client.go
```diff
@@ -102,9 +102,10 @@ type Client struct {
 	lifecycleExecutor   LifecycleExecutor
 	buildpackDownloader BuildpackDownloader
 
-	experimental    bool
-	registryMirrors map[string]string
-	version         string
+	experimental       bool
+	registryMirrors    map[string]string
+	version            string
+	insecureRegistries []string
```
jjbustamante, line 108: I think we are not using the `insecureRegistries` values when fetching images. In theory the values configured here must be use during the `Fetch` operation.

During the [Fetch](https://github.com/buildpacks/pack/blob/8ed450d44de15cefe5875c22ae87293cc625e69d/pkg/image/fetcher.go#L86) operation we can try to access images on different sources:
- Daemon
- Registry
- On disk in OCI layout format
The `insecureRegistries` make sense for remote registry, right? we need to update our remote fetching with the correct configuration from imgutil.

```go

func (f *Fetcher) fetchRemoteImage(name string) (imgutil.Image, error) {
       // maybe here, we need to check if we have insecureRegistries and for each of them
       // we need to create a options with 
       // remote.WithRegistrySetting(<the insecure registry>, true)
       // passthrough the options  to the creation of the image

	image, err := remote.NewImage(name, f.keychain, remote.FromBaseImage(name))
	if err != nil {
		return nil, err
	}

	if !image.Found() {
		return nil, errors.Wrapf(ErrNotFound, "image %s does not exist in registry", style.Symbol(name))
	}

	return image, nil
}

```

internal/build/lifecycle_execution.go
prashantrewar: Do we need to add validation only for create or do we need to add for all like `analyze`, `restore`, `export`, and `rebase`?

jjbustamante: I think in all places where we are sending the values to lifecycle, basically because only those lifecycle will be able to handle the insecure registry

prashantrewar: Done!

pkg/client/client.go
prashantrewar: Done!

internal/commands/rebase.go
```diff
@@ -51,7 +51,7 @@ func Rebase(logger logging.Logger, cfg config.Config, pack PackClient) *cobra.Co
 	cmd.Flags().StringVar(&policy, "pull-policy", "", "Pull policy to use. Accepted values are always, never, and if-not-present. The default is always")
 	cmd.Flags().StringVar(&opts.ReportDestinationDir, "report-output-dir", "", "Path to export build report.toml.\nOmitting the flag yield no report file.")
 	cmd.Flags().BoolVar(&opts.Force, "force", false, "Perform rebase operation without target validation (only available for API >= 0.12)")
-
+	cmd.Flags().StringSliceVarP(&opts.InsecureRegistries, "insecure-registry", "", nil, "List of insecure registries")
```
jjbustamante, line 54: I think, I will add something similar to the `--force` flag. 


```go
cmd.Flags().StringSliceVarP(&opts.InsecureRegistries, "insecure-registry", "", nil, "List of insecure registries (only available for API >= 0.13)")
```

pkg/client/client.go
```diff
@@ -231,7 +239,7 @@ func NewClient(opts ...Option) (*Client, error) {
 	}
 
 	if client.imageFetcher == nil {
-		client.imageFetcher = image.NewFetcher(client.logger, client.docker, image.WithRegistryMirrors(client.registryMirrors), image.WithKeychain(client.keychain))
+		client.imageFetcher = image.NewFetcher(client.logger, client.docker, image.WithRegistryMirrors(client.registryMirrors), image.WithKeychain(client.keychain), image.WithInsecureRegistries(client.insecureRegistries))
```
jjbustamante, line 242: I think there is a misunderstood with this code, for what I can see, `client.insecureRegistries` is always going to be empty. because [this](https://github.com/buildpacks/pack/pull/2077/files#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R192) option is never called when we initialized the `client`.

Usually, the client is being [initialized](https://github.com/buildpacks/pack/blob/main/cmd/cmd.go#L141-L151) with configuration we save in the pack configuration file, see `pack config` command. In this case, the `insecure-registries` are configured at runtime when the user invokes the `pack build` or `pack rebase` commands, but at that point in time, the client was already initialized and the image fetcher was already created. 

I think we need a way to passthrough the `insecure-registries` to the `Fetch` method at runtime

jjbustamante (changes requested):

pkg/client/client.go
jjbustamante: @prashantrewar Do you have a chance to take a look at this?

prashantrewar: I apologize @jjbustamante, right now I am busy with something, on the weekend I will try to fix it.

natalieparellano:
Any further updates here?

jjbustamante:
Hi, @prashantrewar! Do you have some bandwidth to take a look at this?

prashantrewar:
> Hi, @prashantrewar! Do you have some bandwidth to take a look at this?

@jjbustamante Sorry, recently I'm busy with something else.

Issue #2079, opened 2024-02-25
When creating a buildpack with `docker buildx --platform linux/arm64` pack does not set the architecture


jericop:
### Summary
The following occurs when running pack from a dockerfile using `docker buildx build`. When creating a buildpack with `pack buildpack package` the architecture is always an empty string, regardless of the architecture you run on.

This issue is similar to https://github.com/buildpacks/pack/issues/1958, except that it affects `pack buildpack package` instead of `pack builder create`.

### Reproduction

Note that docker needs to be configured to build multi-arch images for this to work as described [here](https://docs.docker.com/build/building/multi-platform/).

##### Steps

1. Create the following dockerfile in a directory.
```
FROM buildpacksio/pack as pack

FROM bitnami/git
USER root
WORKDIR /workspace

COPY --from=pack /usr/local/bin/pack /usr/local/bin/pack

RUN <<RUN_EOF

image_arch=amd64
if [ $(arch) = "aarch64" ]; then
  image_arch=arm64
fi

git clone --depth 1 --branch main https://github.com/buildpacks/samples.git

pack buildpack package ttl.sh/jericop-hello-world-buildpack:${image_arch} --publish \
  --path samples/buildpacks/hello-world

RUN_EOF

```
2. Run the following `docker buildx` command to create and publish buildpacks to ttl.sh, an ephemeral registry.
```
docker buildx build --platform linux/amd64,linux/arm64 .
```
3. Inspect the image config with crane to confirm the architecture is an empty string.
```
➜  pack-arch-issue crane ls ttl.sh/jericop-hello-world-buildpack
amd64
arm64
➜  pack-arch-issue crane config ttl.sh/jericop-hello-world-buildpack:amd64 | jq .architecture
""
➜  pack-arch-issue crane config ttl.sh/jericop-hello-world-buildpack:arm64 | jq .architecture
""
➜  pack-arch-issue 
```

##### Current behavior

The architecture is empty, which prevents building multi-arch images without modifying the image or doing some `docker manifest` magic.

##### Expected behavior

The architecture should match the architecture that pack is running on.

---

##### pack info

```
Pack:
  Version:  0.33.2+git-f2cffc4.build-5562
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.18.5

Supported Platform APIs:  0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.10, 0.11, 0.12
```

##### docker info

```
Client:
[...]

PR #2080, opened 2024-02-26
Add architecture for images created with`pack buildpack package`

jericop:
## Summary
This changes the behavior of `pack buildpack package` to include the platform architecture when the docker daemon is available, or from `runtime.GOARCH` otherwise. It also has support for user-supplied architecture, but this is currently not supported through any pack cli arguments.

#### Before

The current behavior is described in https://github.com/buildpacks/pack/issues/2079

#### After

You can verify the change in this PR by following the steps below.

Note that docker needs to be configured to build multi-arch images for this to work as described [here](https://docs.docker.com/build/building/multi-platform/).

##### Steps

1. Create the following dockerfile in a directory.
```
FROM golang:1.21 as builder
WORKDIR /pack

RUN git clone --depth 1 --branch add-arch-in-buildpack-package https://github.com/jericop/pack.git /pack
RUN make mod-tidy build

FROM bitnami/git
USER root
WORKDIR /workspace

COPY --from=builder /pack/out/pack /usr/local/bin/pack

RUN <<RUN_EOF

image_arch=amd64
if [ $(arch) = "aarch64" ]; then
  image_arch=arm64
fi

git clone --depth 1 --branch main https://github.com/buildpacks/samples.git

pack buildpack package ttl.sh/jericop-buildpack-add-arch-in-buildpack-package:${image_arch} --publish \
  --path samples/buildpacks/hello-world

RUN_EOF

```
2. Run the following `docker buildx` command to create and publish buildpacks to ttl.sh, an ephemeral registry.
```
docker buildx build --platform linux/amd64,linux/arm64 .
```
3. Inspect the image config with crane to confirm the architecture is an empty string.
```
➜  pack-arch-issue crane ls ttl.sh/jericop-buildpack-add-arch-in-buildpack-package
amd64
arm64
➜  pack-arch-issue crane config ttl.sh/jericop-buildpack-add-arch-in-buildpack-package:amd64 | jq .architecture
"amd64"
➜  pack-arch-issue crane config ttl.sh/jericop-buildpack-add-arch-in-buildpack-package:arm64 | jq .architecture
"arm64"
➜  pack-arch-issue 
```

## Documentation

    - [ ] Yes, see #___
    - [x] No

It is generally expected that when you build an image on a certain platform and architecture, the image you produce should be meant to run on that platform and architecture. As a result I don't believe this change needs to be documented.

## Related

Resolves https://github.com/buildpacks/pack/issues/2079
It should also resolve https://github.com/buildpacks/pack/issues/1968, which was fixed by retry logic.

### DIFF
--- a/pkg/dist/dist.go
+++ b/pkg/dist/dist.go
@@ -35,7 +35,9 @@ func (c *ImageOrURI) DisplayString() string {
 }
 
 type Platform struct {
-	OS string `toml:"os"`
+	OS           string `toml:"os"`
+	Architecture string `toml:"architecture"`
+	Platform     string `toml:"platform,omitempty" json:"platform,omitempty" yaml:"platform,omitempty"`
 }
 
 type Order []OrderEntry
--- a/pkg/buildpack/builder.go
+++ b/pkg/buildpack/builder.go
@@ -27,7 +27,7 @@ import (
 )
 
 type ImageFactory interface {
-	NewImage(repoName string, local bool, imageOS string) (imgutil.Image, error)
+	NewImage(repoName string, local bool, platform dist.Platform) (imgutil.Image, error)
 }
 
 type WorkableImage interface {
@@ -343,12 +343,12 @@ func (b *PackageBuilder) resolvedStacks() []dist.Stack {
 	return stacks
 }
 
-func (b *PackageBuilder) SaveAsFile(path, imageOS string, labels map[string]string) error {
+func (b *PackageBuilder) SaveAsFile(path string, platform dist.Platform, labels map[string]string) error {
 	if err := b.validate(); err != nil {
 		return err
 	}
 
-	layoutImage, err := newLayoutImage(imageOS)
+	layoutImage, err := newLayoutImage(platform)
 	if err != nil {
 		return errors.Wrap(err, "creating layout image")
 	}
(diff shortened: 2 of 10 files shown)


pkg/dist/dist.go
```diff
@@ -35,7 +35,9 @@ func (c *ImageOrURI) DisplayString() string {
 }
 
 type Platform struct {
-	OS string `toml:"os"`
+	OS           string `toml:"os"`
+	Architecture string `toml:"architecture"`
+	Platform     string `toml:"platform,omitempty" json:"platform,omitempty" yaml:"platform,omitempty"`
```
natalieparellano, line 40: This is a bit weird - why is Platform a field on Platform?

jjbustamante:
This issue is related to the following [comment](https://github.com/buildpacks/rfcs/pull/295#discussion_r1476722481) in the multi-arch RFC

pkg/dist/dist.go
jjbustamante: This Platform struct is used by the `package.toml`, as part of the multi-arch RFC, we are suggesting deprecating this field and adding a set of Targets. 

Maybe we can add a single `Target` as part of this PR? [here](https://github.com/buildpacks/pack/pull/2086/commits/a27ebb3fd627a6669609f9894a76ff912cc56494#diff-66c276bbc91058d8cc79d886cc7ed9d1020973d3cc3620cef91c6695e41451b8) is what I did on the multi-arch PoC.

@jericop what about getting the information you need from there for buildpack packages and extensions?

edmorley:
This would resolve the missing arch we're seeing in our (new) multi-arch buildpack images, eg:
https://hub.docker.com/repository/docker/heroku/buildpack-go/tags

jjbustamante:
Closing this PR, because we decided to include the implementation of the RFC 0128 with `pack 0.34.0`, that implementation solves the issues this PR is trying to fix.

See my [comment](https://github.com/buildpacks/pack/issues/2079#issuecomment-2115605706)

PR #2081, opened 2024-02-26
Add validation support for empty/nil `[[stacks]]`

### LINKED ISSUE #2047: `pack buildpack package` fails when using `[[targets]]` with `ERROR: no compatible stacks among provided buildpacks`
### Summary

When I use `pack buildpack new` with default options, the resulting generated buildpack fails to package with `pack buildpack package`.

Whilst the error can presumably be fixed by passing additional `--targets` options to `pack buildpack new`, I feel like the defaults should at least "just work" when immediately packaged on the same machine as that which generated the new buildpack.

---

##### Steps

1. `pack buildpack new test-buildpack`
2. `pack buildpack package --verbose --path test-buildpack/ --format file test-buildpack.cnb`

##### Current behavior

```
$ pack buildpack package --verbose --path test-buildpack/ --format file test-buildpack.cnb
Downloading buildpack from URI: file:///Users/emorley/src/test/test-buildpack
ERROR: no compatible stacks among provided buildpacks
```

```
$ cat test-buildpack/buildpack.toml
api = "0.8"
WithWindowsBuild = false
WithLinuxBuild = false

[buildpack]
  id = "test-buildpack"
  version = "1.0.0"

[[targets]]
  os = "darwin"
  arch = "arm64"
```

##### Expected behavior

For the `pack buildpack package` command to succeed with no error.

---

##### pack info

```
$ pack report
Pack:
  Version:  0.33.0+git-e3f8bc2.build-5475
  OS/Arch:  darwin/arm64

Default Lifecycle Version:  0.18.4

Config:
  experimental = true

  [[trusted-builders]]
    name = "[REDACTED]"

  [[trusted-builders]]
    name = "[REDACTED]"
```

##### docker info
N/A

joshwlewis:
## Summary

A `buildpack.toml` without `[[order]]` and `[[stacks]]` tables fails to pass validation during `pack buildpack package` as mentioned in #2047.

`[[stacks]]` was [deprecated in API `0.10`](https://github.com/buildpacks/spec/blob/buildpack/v0.10/buildpack.md#buildpacktoml-toml-stacks-array). Based on my experience with many other software deprecations, deprecations are usually hints to stop using that feature. It's surprising to me that pack would ask to specify `[[stacks]]` while supporting a Buildpack API that marks `[[stacks]]` as deprecated.

This PR adds a new test (which fails on the main branch) that illustrates the problem. It also changes the validation a bit so that `pack` supports non-meta buildpacks without this table.

This change will add support for empty stacks for Buildpack API versions prior to `0.10`. I believe this makes sense. Previous versions of the Buildpack API did not mark `[[stacks]]` as mandatory, though it looks like this validation may have been accidentally doing so for `pack`.

#### Before

```bash
pack buildpack package procfile-test --config ../procfile-cnb/packaged/x86_64-unknown-linux-musl/debug/heroku_procfile/package.toml
ERROR: saving image: no compatible stacks among provided buildpacks
```

#### After

```bash
go run cmd/pack/main.go buildpack package procfile-test --config ../procfile-cnb/packaged/x86_64-unknown-linux-musl/debug/heroku_procfile/package.toml
Successfully created package procfile-test and saved to docker daemon
```
## Documentation

    - [ ] Yes, see #___
    - [ ] No
    - [x] Maybe. Should we be clearer about what deprecated means here? https://github.com/buildpacks/spec/blob/buildpack/v0.10/buildpack.md#buildpacktoml-toml-stacks-array

## Related

Resolves #2047

### DIFF
--- a/pkg/buildpack/builder.go
+++ b/pkg/buildpack/builder.go
@@ -339,7 +339,7 @@ func (b *PackageBuilder) resolvedStacks() []dist.Stack {
 		bpdStacks := bp.Descriptor().Stacks()
 		if len(bpdStacks) == 0 && len(bpd.Order()) == 0 {
 			// For non-meta-buildpacks using targets, not stacks: assume any stack
-			bpdStacks = append(stacks, dist.Stack{ID: "*"})
+			bpdStacks = append(bpdStacks, dist.Stack{ID: "*"})
 		}
 
 		if len(stacks) == 0 {
--- a/pkg/buildpack/builder.go
+++ b/pkg/buildpack/builder.go
@@ -330,13 +330,22 @@ func (b *PackageBuilder) validate() error {
 
 func (b *PackageBuilder) resolvedStacks() []dist.Stack {
 	stacks := b.buildpack.Descriptor().Stacks()
+	if len(stacks) == 0 && len(b.buildpack.Descriptor().Order()) == 0 {
+		// For non-meta-buildpacks using targets, not stacks: assume any stack
+		stacks = append(stacks, dist.Stack{ID: "*"})
+	}
 	for _, bp := range b.AllModules() {
 		bpd := bp.Descriptor()
+		bpdStacks := bp.Descriptor().Stacks()
+		if len(bpdStacks) == 0 && len(bpd.Order()) == 0 {
+			// For non-meta-buildpacks using targets, not stacks: assume any stack
+			bpdStacks = append(stacks, dist.Stack{ID: "*"})
+		}
 
 		if len(stacks) == 0 {
-			stacks = bpd.Stacks()
-		} else if len(bpd.Stacks()) > 0 { // skip over "meta-buildpacks"
-			stacks = stack.MergeCompatible(stacks, bpd.Stacks())
+			stacks = bpdStacks
+		} else if len(bpdStacks) > 0 { // skip over "meta-buildpacks"
+			stacks = stack.MergeCompatible(stacks, bpdStacks)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/buildpack/builder.go
```diff
@@ -330,13 +330,22 @@ func (b *PackageBuilder) validate() error {
 
 func (b *PackageBuilder) resolvedStacks() []dist.Stack {
 	stacks := b.buildpack.Descriptor().Stacks()
+	if len(stacks) == 0 && len(b.buildpack.Descriptor().Order()) == 0 {
+		// For non-meta-buildpacks using targets, not stacks: assume any stack
+		stacks = append(stacks, dist.Stack{ID: "*"})
```
jjbustamante, line 335: Should we need to add `b.buildpack.Descriptor().API().AtLeast("0.10")` to the condition?

jjbustamante:
> This change will add support for empty stacks for Buildpack API versions prior to 0.10. I believe this makes sense. Previous versions of the Buildpack API did not mark [[stacks]] as mandatory, though it looks like this validation may have been accidentally doing so.

I agree that it seems the validation was enforcing the use of `[[stacks]]` although the [spec](https://github.com/buildpacks/spec/blob/buildpack/v0.9/buildpack.md#stacks) doesn't explicitly said it is **mandatory**

pkg/buildpack/builder.go
joshwlewis: My opinion: no. I'm happy to do so if y'all want; just let me know. My reasoning: old versions of the Buildpack API spec do not specify that `[[stacks]]` is required or mandatory. It seems like `[[stacks]]` has been mandatory in `pack` because of this logic, and perhaps that was unintentional.

natalieparellano: > Should we need to add b.buildpack.Descriptor().API().AtLeast("0.10") to the condition?

I would be okay leaving this out. Older buildpacks should be able to work on newer platforms (where stacks are not required) without having to make changes.

jjbustamante (approved):

PR #2010, opened 2023-12-14
Ensure read access when resolving run image location

pbusko:
## Summary

`pack` performs it's own run-image resolution logic, and must be compliant with the platform spec, in particular with the https://github.com/buildpacks/spec/pull/357
## Output

```console
$ docker logout unauthorized.registry.io
Removing login credentials for unauthorized.registry.io
$ pack build --verbose --publish private-registry.corp/my-image:latest
...
CheckReadAccess failed for the run-image unauthorized.registry.io/run:latest, error: GET ***: : Authentication is required
CheckReadAccess succeeded for the run-image authorized.registry.io/run:latest
Selected run image mirror authorized.registry.io/run:latest
...
```

#### Before

`pack` is not compliant with https://github.com/buildpacks/spec/pull/357/files and avoids the run-image resolution baked into the `lifecycle` (https://github.com/buildpacks/lifecycle/pull/1024)

#### After

`pack` will ensure read access during the run-image resolution

## Documentation

- Should this change be documented?
    - [ ] Yes
    - [X] No

## Open Questions

Currently the `lifecycle` already has read-access and preferable registry resolution for the run-image. What would be the correct place for this? Should it be kept in both places or be left only in a single one?

PR #2088, opened 2024-03-04
Return an image even if inaccessible

AidanDelaney:
When all run images are inaccessible, still return a run image.  Pack will then provide a better error message to the end-user.

## Summary
When all run images are inaccessible, still return a run image.  Pack will then provide a better error message to the end-user.

#### Before
Produces the  panic listed in #2087

#### After
Produces a build, avoiding the panic.  Restores the previous behaviour.

## Documentation

- Should this change be documented?
    - [ ] Yes, see #___
    - [X] No

## Related

Resolves #2087

PR #2092, opened 2024-03-08
Adding a nop-op when trying to check access for run-image against the daemon

### LINKED ISSUE #2078: v0.33.0 bug: CheckReadAccess fails when using a local run image
### Summary

---

See https://cloud-native.slack.com/archives/C0331B61A1Y/p1708468320706469 for more context.
When upgrading to pack v0.33.0 from v0.32.1 in our stack implementation test suite, we run into the following error during a `pack build`:
```
 CheckReadAccess failed for the run image stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4:latest, error: HEAD https://index.docker.io/v2/library/stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4/manifests/latest: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)
        panic: runtime error: index out of range [0] with length 0
```

We are creating a stack build/run image **locally**, then creating a builder with those stack images, and then using the builder in a `pack build`. As a result of the changes in [this PR](https://github.com/buildpacks/pack/pull/2010/files#diff-a0c732d7543d7450a4886f604425cdd9195ca3da77338f9a0f4b6980a6ad5eb6), this no longer works due to issues with accessing the local run image on the docker daemon.

As a workaround, we are able to create a local docker registry, and push the stack images, which helps us to avoid this issue. We would still like to to be able to build and test stack images locally without needing to do this, however.

##### Steps

1. Install pack v0.33.0
2. Create a stack build/run image locally
3. Create a builder with the local stacks using `pack`
4. Use the builder in a `pack build` of a simple app

OR more simply
1. Checkout the Paketo Jammy Base Stack at [this commit](https://github.com/paketo-buildpacks/jammy-base-stack/commit/3d526b629bac0cf8c4bc37742de9e0af8984cdec)
2. Run the integration tests via `scripts/test.sh` which will do the above steps
3. See the failure above in the `buildpack_integration_test.go` test case

##### Current behavior

Build failed due to inability to access the local run image

##### Expected behavior

Successful build

---

##### pack info

```
 ‣ .bin/pack report
Pack:
  Version:  0.33.0+git-e3f8bc2.build-5475
  OS/Arch:  darwin/amd64

Default Lifecycle Version:  0.18.4

Config:
  experimental = true
  layout-repo-dir = "/Users/swigmore/.pack/layout-repo"

  [[trusted-builders]]
    name = "[REDACTED]"

  [[trusted-builders]]
    name = "[REDACTED]"

  [[trusted-builders]]
    name = "[REDACTED]"

  [[trusted-builders]]
    name = "[REDACTED]"
```

jjbustamante:
## Summary
In a previous [PR](https://github.com/buildpacks/pack/pull/2010) we added a new logic to calculate the run-image according to the [spec](https://github.com/buildpacks/spec/pull/357), but we also introduced an error when trying to check `read access` for run-images that are saved only in the daemon. In those cases, the `checker` was returning false and panic was thrown in this [method](https://github.com/buildpacks/pack/blob/v0.33.2/pkg/client/common.go#L111-L124) because the `runImageList` was empty.

Following a similar [behavior](https://github.com/buildpacks/lifecycle/blob/main/platform/lifecycle_inputs.go#L160-L175
) we already have in the lifecycle, this PR uses the `publish` value to check if we are running against the daemon or the registry, when running against the daemon, we don't need to check for `read access`

#### Before

Panic was thrown

#### After

The correct run-image is resolved

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #2078


pkg/client/common_test.go
```diff
@@ -131,9 +131,30 @@ func testCommon(t *testing.T, when spec.G, it spec.S) {
...
+					}
+					accessChecker.RegistriesToFail = []string{
+						stackInfo.RunImage.Image,
+					}
+				})
+
+				it("selects the given run-image", func() {
+					// issue: https://github.com/buildpacks/pack/issues/2078
+					runImageName := subject.resolveRunImage("", "", "", stackInfo.RunImage, nil, false, accessChecker)
+					assert.Equal(runImageName, "stack/run-image")
```
jjbustamante, line 156: if we removed the lines:

```go 
if !publish {
	return true
}
```

From the `FakeAccessChecker` and removed this [fix](https://github.com/buildpacks/pack/pull/2088/) from Aidan this test reproduces the panic error

` panic: runtime error: index out of range [0] with length 0`

c0d1ngm0nk3y:
> this PR uses the `publish` value to check if we are running against the daemon or the registry

Are you sure that using the `publish` flag for this is the right approach? The problem is that the `stack` is local and not that the result should not be published, right? It would be valid to use a remote stack that should be checked.

jjbustamante:
> > this PR uses the `publish` value to check if we are running against the daemon or the registry
> 
> Are you sure that using the `publish` flag for this is the right approach? The problem is that the `stack` is local and not that the result should not be published, right? It would be valid to use a remote stack that should be checked.

The problem was: when running `pack build` without `--publish` which is the flow to run against the daemon, pack was throwing the error:

```bash
 CheckReadAccess failed for the run image stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4:latest, error: HEAD https://index.docker.io/v2/library/stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4/manifests/latest: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)
        panic: runtime error: index out of range [0] with length 0
```   

The reason is we are trying to check the read access in a remote location, pack was converting the image reference and by default it tries to hit dockerhub. When we use the `publish` flag, yes it means we want the final image to be saved into a registry, but it also defines the place from where `pack` is going to read the images, it means which `image` from imgutil is going to be instantiated (remote, local, layout)  in this case, we want pack to instantiate a `imgutil.local` image and that's the actual problem

jjbustamante:
> Having publish as a criteria how to interpret the run image (docker or daemon) sounds wrong to me. I is also a use case to use a run image from docker, but not publish the result, right? Would it be possible to identify daemon via daemon:// for example?

We've been using the `publish` flag to configure the [Fetcher](https://github.com/buildpacks/pack/blob/main/pkg/image/fetcher.go#L86-L124) from where it should get the image, maybe as Natalie [said](https://github.com/buildpacks/pack/issues/2078#issuecomment-1969969248) the best way to solve it could be to move the `AccessChecker` to be a method exposed by the Fetcher, but even if we do that, well be implementing a logic that checks `if !options.Daemon` do nothing. 

By adding `daemon://` you mean adding it to the `run-image metadata`? I do not see why we should request end-users to add a extra information for something that was working (from their perspective) in the previous release? 

> Passing the publish flag to the ReadChecker to then decide not to check anything defeats the purpose I guess. So maybe we can find a way to inject a NoOpReadChecker like in the unit tests. But I think we should have a better criteria to distinguish references to docker.io from references to the local daemon.

Yeah, I think ideally the no-op should be the best way,  that's how I think it is working in the [lifecycle](https://github.com/buildpacks/lifecycle/blob/main/platform/lifecycle_inputs.go#L160-L175) but again, at end lifecycle is also checking if it hitting the daemon or the registry to decide if it needs to do the check or not.

c0d1ngm0nk3y:
But adding `publish` to the signature of `AccessChecker.Check` just to return `true` feels wrong. Especially since we have to pass `publish` down the the actual call. Should/could we maybe just exchange the `AccessChecker` in this case? Similar like it is done in the tests.

I still think that using `publish` as indicator if the run image is daemon only or default the location to `docker.io` is strange.

jjbustamante:
@natalieparellano @c0d1ngm0nk3y 

I just did a little refactoring and converted the check read access into a `function` and the `Fetcher` is responsible for returning this function. The logic behind is actually the same, if you try to fetch an image from the Daemon, then we skip the validation.

pkg/image/fetcher_test.go
```diff
@@ -404,4 +403,47 @@ func testFetcher(t *testing.T, when spec.G, it spec.S) {
...
+
+	when("#CheckReadAccessValidator", func() {
+		var daemon bool
+
+		when("Daemon is true", func() {
+			it.Before(func() {
+				daemon = true
+			})
+
+			it("read access is always valid", func() {
```
natalieparellano, line 415: I think it's actually a bit more complicated - it is also dependent on the pull policy. If the pull policy is "always" then it's still essentially a registry operation; conversely if the pull policy is "never" then we could check if the image is already in the daemon; "if not present" is the special case where we fall back to the registry. This logic is already contained (and can be extracted from) [here](https://github.com/buildpacks/pack/blob/6a9e3d193abca7fff82ce2df10d181c63fbde832/pkg/image/fetcher.go#L100-L109).

pkg/client/client.go
```diff
@@ -52,6 +52,8 @@ type ImageFetcher interface {
 	// PullIfNotPresent and daemon = false, gives us the same behavior as PullAlways.
 	// There is a single invalid configuration, PullNever and daemon = false, this will always fail.
 	Fetch(ctx context.Context, name string, options image.FetchOptions) (imgutil.Image, error)
+
+	CheckReadAccessValidator(options image.FetchOptions) image.CheckReadAccess
```
natalieparellano, line 56: I think you are making this a method that returns a function because you want to provide an "access checker" to `getBestRunMirror`. But why not just pass the whole image fetcher to that function? You don't need to expect an image fetcher, just an interface that satisfies `CheckReadAccess`.

pkg/client/common_test.go
```diff
@@ -56,46 +64,83 @@ func testCommon(t *testing.T, when spec.G, it spec.S) {
...
-			})
-
-			it("prefers config defined run image mirror to stack defined run image mirror", func() {
-				configMirrors := map[string][]string{
-					runImageName: {defaultRegistry + "/unique-run-img"},
-				}
-				runImageName := subject.resolveRunImage("", defaultRegistry, "", stackInfo.RunImage, configMirrors, true, accessChecker)
-				assert.NotEqual(runImageName, defaultMirror)
-				assert.Equal(runImageName, defaultRegistry+"/unique-run-img")
+			when("desirable run-image are accessible", func() {
```
natalieparellano, line 82: I could be wrong, but I'm not sure if these tests need to care if publish is true or not. The access checker is abstracting away that part (via the pull policy).

pkg/client/client.go
jjbustamante: Well, I wanted to preserve the signature for CheckReadAccess function `func(repo string) bool`, but yes I think we can pass through the `Fetcher` instance

pkg/image/fetcher_test.go
jjbustamante: I am not totally sure about it, we will always have access to read from the daemon but it is possible that the image is not available there, but for me, that's a different thing. For example: 

The error reported by Sophie in this issue is:

```bash
 CheckReadAccess failed for the run image stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4:latest, error: HEAD https://index.docker.io/v2/library/stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4/manifests/latest: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)
```

- run image ref: stack-run-845695c9-5147-42a2-8cb3-c0ea0500c6a4
- Daemon: true
- Pull-Policy: "always" (default one)

according to your statement
> it is also dependent on the pull policy. If the pull policy is "always" then it's still essentially a registry operation

`CheckReadAccessValidator` should return:
-  **false**? because it must try to check read access against a registry that doesn't exist? 
-  **true**? because although it can't find the image a remote registry the image is available in the daemon? 

My confusion is: if we want `CheckReadAccessValidator` success because I do have read access from the source or success because the image exists in the source?

jjbustamante: I [changed](https://github.com/buildpacks/pack/pull/2092/commits/2d8f1bb15626e578289e6edf1a718e75d2dfa871#diff-76bea0285076919ef2d39e8bbb47824f3f7c991ff0a04302eb496f244f3380dc) the implementation, when `opts.Daemon` is true, we will try to check if the image is available in the docker daemon and if **it's not** we will try to check the remote only when Pull is not **Never**. I think this should cover the cases.

pkg/client/common_test.go
jjbustamante: Yeah, we are using mocks and the parameters to the method are not relevant in these tests, I removed some of them

internal/builder/image_fetcher_wrapper.go
```diff
@@ -13,6 +13,14 @@ type ImageFetcher interface {
...
 	// attempt to pull a remote image first.
 	Fetch(ctx context.Context, name string, options image.FetchOptions) (imgutil.Image, error)
+
+	// CheckReadAccessValidator verifies if an image accessible with read permissions
+	// When FetchOptions.Daemon is true and the image doesn't exist in the daemon,
+	// the behavior is dictated by the pullPolicy, which can have the following behavior
+	//   - PullNever: returns false
+	//   - PullAlways Or PullIfNotPResent: it will check read access for the remote image.
+	// When FetchOptions.Daemon is false it will check read access for the remote image.
+	CheckReadAccessValidator(repo string, options image.FetchOptions) bool
```
c0d1ngm0nk3y, line 23: What about calling it `CheckReadAccess`? Since it is a function, it should be a verb and no noun.

pkg/image/fetcher.go
```diff
@@ -123,6 +123,37 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
 	return f.fetchDaemonImage(name)
 }
 
+func (f *Fetcher) CheckReadAccessValidator(repo string, options FetchOptions) bool {
+	if !options.Daemon {
+		return f.checkRemoteReadAccess(repo)
+	}
+	if _, err := f.fetchDaemonImage(repo); err != nil {
```
c0d1ngm0nk3y, line 130: We should probably check for `ErrNotFound` and not assume that every error mean that the image is not available.

pkg/image/fetcher.go
```diff
@@ -123,6 +123,37 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
+		// Image doesn't exist in the daemon
+		// 	Pull Never: should failed
+		// 	Pull Always or Pull If Not Present: need to check the registry
+		if options.PullPolicy == PullNever {
+			return false
+		}
+		return f.checkRemoteReadAccess(repo)
+	}
+	// no-op
+	return true
```
c0d1ngm0nk3y, line 140: But if `PullPolicy == PullAlways`, shouldn't we always check the remote access? Even if the image is readable in the daemon? Or would we just silently take the daemon one?

pkg/image/fetcher.go
```diff
@@ -123,6 +123,37 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
+}
+
+func (f *Fetcher) checkRemoteReadAccess(repo string) bool {
+	img, err := remote.NewImage(repo, f.keychain)
+	if err != nil {
+		return false
```
c0d1ngm0nk3y, line 146: Don't we potentially hide an error here?

c0d1ngm0nk3y (changes requested):

pkg/image/fetcher.go
natalieparellano: > shouldn't we always check the remote access

I would think so

jjbustamante: I changed guard to be

```go
	if !options.Daemon || options.PullPolicy == PullAlways {
		return f.checkRemoteReadAccess(repo)
	}
```

pkg/image/fetcher.go
jjbustamante: Maybe, I just added a debug message into the console

pkg/image/fetcher.go
jjbustamante: I added a new validation to check if it is `ErrNotFound` execute the actual logic, otherwise just return `false`, also I added a few unit test to check the behavior

jjbustamante:
@natalieparellano @c0d1ngm0nk3y could you give me one more review on this one?

pkg/image/fetcher.go
```diff
@@ -123,6 +123,42 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
+		if errors.Is(err, ErrNotFound) {
+			// Image doesn't exist in the daemon
+			// 	Pull Never: should failed
+			// 	Pull If Not Present: need to check the registry
+			if options.PullPolicy == PullNever {
+				return false
+			}
+			return f.checkRemoteReadAccess(repo)
+		}
+		f.logger.Debugf("failed reading image from the daemon %s, error: %s", repo, err.Error())
```
c0d1ngm0nk3y, line 140:
```suggestion
		f.logger.Debugf("failed reading image %s from the daemon, error: %s", repo, err.Error())
```

pkg/image/fetcher.go
```diff
@@ -123,6 +123,42 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
+		return false
+	}
+	// no-op
```
c0d1ngm0nk3y, line 143:
```suggestion
```

c0d1ngm0nk3y (commented):
lgtm

internal/builder/image_fetcher_wrapper.go
```diff
@@ -13,6 +13,14 @@ type ImageFetcher interface {
 	// If daemon is true, it will look return a `local.Image`. Pull, applicable only when daemon is true, will
 	// attempt to pull a remote image first.
 	Fetch(ctx context.Context, name string, options image.FetchOptions) (imgutil.Image, error)
+
+	// CheckReadAccess verifies if an image accessible with read permissions
```
natalieparellano, line 17:
```suggestion
	// CheckReadAccess verifies if an image is accessible with read permissions
```

internal/builder/image_fetcher_wrapper.go
```diff
@@ -13,6 +13,14 @@ type ImageFetcher interface {
...
+	// When FetchOptions.Daemon is true and the image doesn't exist in the daemon,
+	// the behavior is dictated by the pullPolicy, which can have the following behavior
```
natalieparellano, line 19:
```suggestion
	// the behavior is dictated by the pull policy, which can have the following behavior
```

internal/builder/image_fetcher_wrapper.go
```diff
@@ -13,6 +13,14 @@ type ImageFetcher interface {
...
+	//   - PullNever: returns false
+	//   - PullAlways Or PullIfNotPResent: it will check read access for the remote image.
```
natalieparellano, line 21:
```suggestion
	//   - PullAlways Or PullIfNotPresent: it will check read access for the remote image.
```

pkg/client/client.go
```diff
@@ -52,6 +52,14 @@ type ImageFetcher interface {
 	// PullIfNotPresent and daemon = false, gives us the same behavior as PullAlways.
 	// There is a single invalid configuration, PullNever and daemon = false, this will always fail.
 	Fetch(ctx context.Context, name string, options image.FetchOptions) (imgutil.Image, error)
+
+	// CheckReadAccess verifies if an image is accessible with read permissions
+	// When FetchOptions.Daemon is true and the image doesn't exist in the daemon,
+	// the behavior is dictated by the pullPolicy, which can have the following behavior
```
natalieparellano, line 58:
```suggestion
	// the behavior is dictated by the pull policy, which can have the following behavior
```

pkg/client/client.go
```diff
@@ -52,6 +52,14 @@ type ImageFetcher interface {
...
+	//   - PullNever: returns false
+	//   - PullAlways Or PullIfNotPResent: it will check read access for the remote image.
```
natalieparellano, line 60:
```suggestion
	//   - PullAlways Or PullIfNotPresent: it will check read access for the remote image.
```

pkg/image/fetcher.go
```diff
@@ -123,6 +123,42 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
 }
 
+func (f *Fetcher) CheckReadAccess(repo string, options FetchOptions) bool {
+	if !options.Daemon || options.PullPolicy == PullAlways {
+		return f.checkRemoteReadAccess(repo)
+	}
+	if _, err := f.fetchDaemonImage(repo); err != nil {
+		if errors.Is(err, ErrNotFound) {
+			// Image doesn't exist in the daemon
+			// 	Pull Never: should failed
```
natalieparellano, line 133:
```suggestion
			// 	Pull Never: should fail
```

natalieparellano (approved):
LGTM. Nice work @jjbustamante

PR #2093, opened 2024-03-11
chore: fix comment

tgolang:
## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___

### DIFF
--- a/internal/paths/paths.go
+++ b/internal/paths/paths.go
@@ -126,7 +126,7 @@ func WindowsToSlash(p string) string {
 }
 
 // WindowsPathSID returns the appropriate SID for a given UID and GID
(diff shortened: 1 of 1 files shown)


internal/paths/paths.go
```diff
@@ -126,7 +126,7 @@ func WindowsToSlash(p string) string {
 }
 
 // WindowsPathSID returns the appropriate SID for a given UID and GID
-// This the basic logic for path permissions in Pack and Lifecycle
```
jjbustamante, line 129: I think the comment was suppose to say: 

This **is** the basic logic for path permissions in Pack and Lifecycle

jjbustamante (approved):

PR #2112, opened 2024-04-01
Support docker uri for lifecycle

### LINKED ISSUE #2076: Support `docker://` URIs for builder lifecycle parameter
### Description
Right now we can provide a `docker://` URI for the buildpacks listed in a builder inside the `builder.toml` but we cannot do the same thing with the `lifecycle`. 

### Proposed solution
```toml
[lifecycle]
  uri = "docker://buildpacksio/lifecycle:0.18.5"
```

Right now you need to provide a link to the tar.gz and cannot use docker images.

### Additional context
The reason this would be a great quality of life feature is primarily because almost all the other fields in the specification allow `docker://` archives. This would be consistent and not require a separate location for storage of the lifecycle binaries. It is a bit inconsistent that at build time you are able to specify a custom lifecycle image, but you can't do the same when you are packaging a builder.

rashadism:
## Summary

Now can provide a docker uri for the lifecycle image in builder.toml

## Documentation

    - [x] Yes, the possibility of using a docker uri in builder.toml should be documented
    - [ ] No

## Related

Resolves #2076


pkg/client/create_builder.go
```diff
@@ -212,20 +219,75 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
...
+		var blob blob.Blob
+		imageName := buildpack.ParsePackageLocator(config.URI)
+		c.logger.Debugf("Downloading lifecycle image: %s", style.Symbol(imageName))
+
+		lifecycleImage, err = c.imageFetcher.Fetch(ctx, imageName, image.FetchOptions{Daemon: false})
+		if err != nil {
+			return nil, err
+		}
+
+		lifecycleImageTar := filepath.Join(relativeBaseDir, "lifecycle-image.tar")
```
WYGIN, line 233: @rashadism Thank You for contributing to Buildpacks, I want to take it to your notice that `"lifecycle-image.tar"` might not going to be static, **I think it will be based on `imageName` variable,** 
can you please take a look into it, and **please ignore this message** if I am wrong(as I am just a contributor like you and never worked on imageFetcher)
I recommend adding tests, and manually testing code as far as possible, I really appreciate your effort.

pkg/client/create_builder.go
```diff
@@ -212,20 +219,75 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
 	var uri string
 	var err error
 	switch {
+	case isDockerURI(config.URI):
```
jjbustamante, line 222: I think you can use the following [function](https://github.com/buildpacks/pack/blob/main/pkg/buildpack/locator_type.go#L84-L86) that is doing the same thing

pkg/client/create_builder.go
rashadism: Hi, thank you for going through this. Can you explain a bit more on 

> "lifecycle-image.tar" might not going to be static, I think it will be based on imageName 
> variable

I think I did not understand properly what you mentioned

WYGIN: > I think I did not understand properly what you mentioned

```go
lifecycleImage, err = c.imageFetcher.Fetch(ctx, imageName, image.FetchOptions{Daemon: false})
if err != nil {
	return nil, err
}

lifecycleImageTar := filepath.Join(relativeBaseDir, "lifecycle-image.tar")
```
I feel like here the `imageName` in call `c.imageFetcher.Fetch` might decide whether the lifecycle file name should be `lifecycle-image.tar` or something else, 
try changing the lifecycle uri in builder.toml to something else if you have provided `lifecycle-image` as lifecycle uri

jjbustamante: I think what Sai was highlighting your hardcoded variable "lifecycle-image.tar" but checking your code I noticed you are just extracting the image from the docker daemon and saving it on disk using that hardcoded name. That's ok.

Now, I took sometime trying to think about this and I am going to do one suggestion (but I am not sure if it is going to work).

Your current code looks like this:

```go
switch {
	case buildpack.HasDockerLocator(config.URI):
            var lifecycleImage imgutil.Image
            var blob blob.Blob
            imageName := buildpack.ParsePackageLocator(config.URI)
            c.logger.Debugf("Downloading lifecycle image: %s", style.Symbol(imageName))
            
            lifecycleImage, err = c.imageFetcher.Fetch(ctx, imageName, image.FetchOptions{Daemon: false})
            if err != nil {
	            return nil, err
            }
            // Good, till this point, we have imgItil.Image and we need to create a 
            // Blob from it
}
```  

Your are going through a tricky path of trying to read/write the image from the daemon and saving it into a tar. We also have a `layout` implementation of the Image interface that **maybe** can help us here, could you try something like this:

```go
       
	lifecycleLayoutImage, err := layout.NewImage("<path-to-safe-the-lifecyce>", layout.FromBaseImage(lifecycleImage.UnderlyingImage()))
	lifecycle.Save()
	blob := blob.NewBlob("<path-to-safe-the-lifecyce>")
         lifecycle, err := builder.NewLifecycle(blob)
         if err != nil {
	          return nil, errors.Wrap(err, "invalid lifecycle")
         }
         return lifecycle, nil

``` 

- `Save()` is going to write the image on disk in OCI layout format at the specified path
- Then we create the blob from that path

if this works, maybe you can avoid extracting all the other methods in the `build.go` file and simplify your current implementation. The initial part is exactly what I think needs to be done, the part of getting the lifecycle out of the daemon is where I have some doubts.

pkg/client/create_builder.go
```diff
@@ -212,20 +219,75 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
 	var uri string
 	var err error
 	switch {
+	case buildpack.HasDockerLocator(config.URI):
+		var lifecycleImage imgutil.Image
+		var blob blob.Blob
+		imageName := buildpack.ParsePackageLocator(config.URI)
+		c.logger.Debugf("Downloading lifecycle image: %s", style.Symbol(imageName))
+
+		lifecycleImage, err = c.imageFetcher.Fetch(ctx, imageName, image.FetchOptions{Daemon: false})
```
jjbustamante, line 228: I think, we need to passthrough some key fetch options: `Daemon`, `PullPolicy`, `Platform`, those values should be available from method caller and we need to configure the Fetcher with them

pkg/client/create_builder.go
```diff
@@ -201,7 +208,7 @@ func (c *Client) createBaseBuilder(ctx context.Context, opts CreateBuilderOption
 	return bldr, nil
 }
 
-func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleConfig, relativeBaseDir, os string, architecture string) (builder.Lifecycle, error) {
+func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleConfig, relativeBaseDir, o string, architecture string) (builder.Lifecycle, error) {
```
jjbustamante, line 211: Do no change the `os` variable name, it is more descriptive that `o`

pkg/client/create_builder.go
```diff
@@ -212,20 +219,75 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
 	var uri string
 	var err error
 	switch {
+	case buildpack.HasDockerLocator(config.URI):
```
jjbustamante, line 222: Maybe we also need to validate the given URI is a valid image reference we do this when checking buildpacks

```go
if _, err := name.ParseReference(config.URI); err == nil {
     // the URI is a valid image reference.
}
```

pkg/client/create_builder.go
rashadism: Okay, that is smart and more convenient, I did not know about the layout implementation, will look into it thank you.

pkg/client/create_builder.go
rashadism: Changed because I needed to use a function from the os package. Will try an alternative way

pkg/client/create_builder.go
WYGIN: > I think what Sai was highlighting your hardcoded variable "lifecycle-image.tar" but checking your code I noticed you are just extracting the image from the docker daemon and saving it on disk using that hardcoded name. That's ok.

Oops, I haven't gone through those lines, Thanks for the catch :)

@jjbustamante why isn't it possible to use LLB in Phase 2, I feel like we can [create](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L386) Buildpacks and Extensions that returns State that we can [cache](https://github.com/moby/buildkit/tree/master/cache) and [export](https://github.com/moby/buildkit/tree/master/exporter) them to the format that the pack currently supports and Buildkit does supports a wide range of ways to create an Image including From 

1. [Registry](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L100), 
2. [Git](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L249), 
3. [Scratch](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L381), 
4. [Local (which is loading from a directory, can be used at `pack buildpack package` and `pack extension package` commands)](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L386), 
5. [OCILayout](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L490) and 
6. [HTTP](https://github.com/moby/buildkit/blob/cbb116dcc38acb40408743697b575f0ff2a7df6a/client/llb/source.go#L570)

Are we not using LLB because of the metadata that we use during lifecycle phases like analyzer?

rashadism: > * `Save()` is going to write the image on disk in OCI layout format at the specified path
> * Then we create the blob from that path
> 
> if this works, maybe you can avoid extracting all the other methods in the `build.go` file and simplify your current implementation. The initial part is exactly what I think needs to be done, the part of getting the lifecycle out of the daemon is where I have some doubts.

WYGIN:
@jjbustamante If this feature ships isn't it possible for platform operators to create their own lifecycles ?, existed to see this feature shipping, and I think one should publish an article on this

pkg/client/create_builder.go
rashadism: > image The lifecycle.toml of [latest lifecycle version](https://hub.docker.com/r/buildpacksio/lifecycle) file exists under `cnb` folder

I see. I was having problems inspecting and traversing the file system using code. Are there any functions for that? @WYGIN @jjbustamante

WYGIN: @rashadism  you can [install dive](https://github.com/wagoodman/dive) and you can inspect any image filesystem using `dive [image-referance]`
here for lifecycle run 
```sh
dive buildpacksio/lifecycle`
```
Taking a look into [NewLifecycle](https://github.com/buildpacks/pack/blob/6fc092be641b5db554dfa0b8d82cd94653de5f3e/internal/builder/lifecycle.go#L40-L68) might help you

rashadism: @WYGIN , what I meant was after doing `lifecycleLayoutImage.Save()`, the image is saved in OCI layout format. Couldn't find a way to inspect the file system of that(from within golang codebase)

jjbustamante: @rashadism 

I think I understand your problem, that's why I said I was not sure if it was going to work 😄 

Checking the code, I have the following idea:


```go

switch {
	case buildpack.HasDockerLocator(config.URI):
            var lifecycleImage imgutil.Image
            var blob blob.Blob
            imageName := buildpack.ParsePackageLocator(config.URI)
            c.logger.Debugf("Downloading lifecycle image: %s", style.Symbol(imageName))
            
            lifecycleImage, err = c.imageFetcher.Fetch(ctx, imageName, image.FetchOptions{Daemon: false})
            if err != nil {
	            return nil, err
            }
            // Good, till this point, we have imgItil.Image and we need to create a 
            // Blob from it
          lifecycleLayoutImage, err := layout.NewImage("<path-to-safe-the-lifecyce>", layout.FromBaseImage(lifecycleImage.UnderlyingImage()))
          lifecycle.Save()

          // You probably need to convert the <path-to-safe-the-lifecyce> to an URI
          // Not sure if it is relative or not
          uri, err = paths.FilePathToURI("<path-to-safe-the-lifecyce>", "")
}

blob, err := c.downloader.Download(ctx, URI)

// blob is going to be OCI layout format, we need to make NewLifecycle method capable of handling it
lifecycle, err := builder.NewLifecycle(blob)
	
return lifecycle, nil
```

As Sai mentioned, it's time to understand how to look for a lifecycle in OCI layout format. We do something similar for Buildpack Packages [here](https://github.com/buildpacks/pack/blob/main/pkg/buildpack/oci_layout_package.go) I think that is what you want.

jjbustamante: The OCI Layout format spec is [here](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) I think the tricky part is to open the blobs and find where the lifecycle binaries are located. 

As @WYGIN mentioned, if you install [dive](https://github.com/wagoodman/dive) tool, and then you pill the latest image from dockerhub

```bash
dive buildpacksio/lifecycle:latest
```
We can see that the blob containing the lifecycle is at the end, and also as mentioned by @WYGIN , it is under `/cnb`

Screenshot 2024-04-02 at 12 16 05 PM

I also used `skopeo` to copy the lifecycle from the daemon to disk in OCI layout

```bash
> skopeo copy docker-daemon:buildpacksio/lifecycle:latest oci:local-lifecycle
Getting image source signatures
Copying blob ac805962e479 done   |
Copying blob 1df9699731f7 done   |
Copying blob 70c35736547b done   |
Copying blob af5aa97ebe6c done   |
Copying blob 4d049f83d9cf done   |
Copying blob 6fbdf253bbc2 done   |
Copying blob bbb6cacb8c82 done   |
Copying blob 2a92d6ac9e4f done   |
Copying blob 1a73b54f556b done   |
Copying blob c048279a7d9f done   |
Copying blob 2388d21e8e2b done   |
Copying blob a03079e61f77 done   |
Copying config e5c54e7a25 done   |
Writing manifest to image destination
> ls
local-lifecycle
> tree local-lifecycle
local-lifecycle
├── blobs
│   └── sha256
│       ├── 02c91f6b395a6b06d17b8985a2db90aef7b09feedff77eff1f7c269161263a9b
│       ├── 0b2215fb07602fba31abe9852102a9a15baebc13afe0a1c8904bfaec08abb99b
│       ├── 12aa63c6434d9d6411f6f5316b053d1c869c83e24c974e31ca8491401e137e90
│       ├── 18c59b7472ef6bc22a3919f76e0928d6a595ba3350201be6b473c3f2fbc9488b
│       ├── 1e45aa110bce64887bed49d9216b825d63437fdca5d53a1f879e17bc35691e0e
│       ├── 50d2b2cd67880cbdac647edf0295c59ca5c7a54ae2512547960132c4861741d1
│       ├── 80f00805faa81bccdbddf9458369d8dede6a30de1dced521f32a6237aeca64c2
│       ├── 8284fef10e55c7fbd563239f69e0a94cbf76468489d74e7bab59c34ed61cccd9
│       ├── a97b9dcac04afbc295431c6a36a5ba65f42b2804647d34e4f578412d4475dca6
│       ├── b4aa04aa577f2b2f4b4a930e905d091b68b0719ec302b9abca710ffae50ebcaa
│       ├── e5c54e7a259ab941a22c38591dd8a0f08366003e0176e8723bcf0506cd71b154
│       ├── efd880768fbced8b0e7e6c514ee6fb69f2207088a5c5f15c9dbae25c45d2f852
│       ├── f5ceaaa2f4b47256c44e2cdae5cbbf663f8269e6ee345e31b75da5a97974ca62
│       └── ffdf8cfdbafeabe2762d2dde33e36f1b87dfa967ae34811ff5342ebe8233eb4b
├── index.json
└── oci-layout

3 directories, 16 files

```

If I am not wrong, each `blob` is a **tar** file, what **I think** you need to do is to find the blob containing the lifecycle binaries and then probably create another `Blob` instance to passthrough  `builder.NewLifecycle(blob)`

jjbustamante: Another idea is:

Once you fetch the Lifecycle image:

```go
lifecycleImage, err = c.imageFetcher.Fetch(ctx, imageName, image.FetchOptions{Daemon: false})
```

Try to find the `v1.Layer` containing the binaries, maybe, iterating over the Layers and inspecting them to find the `cnb` folder

```go
lifecycleImage.UnderlyingImage().Layers()
``` 

Once you have the Layer, I think you can write it on disk, just that blob, and then keep doing all the other process maybe we don't need to save it as OCI layout

rashadism: Thanks for the detailed insights @jjbustamante @WYGIN , will look into it

braunsonm:
@WYGIN you could always create your own lifecycle binaries, you just had to provide them in a tar if creating your own builder.

The pack CLI during build has supported `--lifecycle-image` for custom lifecycles with existing builders before this as well.

WYGIN:
> The pack CLI during build has supported `--lifecycle-image` for custom lifecycles with existing builders before this as well.

never noticed this flag, Thanks a lot for the information

rashadism:
@jjbustamante @WYGIN can u have a look

pkg/client/create_builder.go
```diff
@@ -212,6 +216,43 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
...
+
+		file, err := stripTopDirAndWrite(layerReader, lifecyclePath)
+		if err != nil {
+			return nil, err
+		}
+
+		defer file.Close()
+		// defer OS.Remove(lifecyclePath)
+
+		uri, err = paths.FilePathToURI(lifecyclePath, "")
```
jjbustamante, line 252: Maybe you can refactor all these lines to a `uriFromLifecycleImage` method?

pkg/client/create_builder.go
```diff
@@ -212,6 +216,43 @@ func (c *Client) fetchLifecycle(ctx context.Context, config pubbldr.LifecycleCon
...
+		}
+		defer layerReader.Close()
+
+		file, err := stripTopDirAndWrite(layerReader, lifecyclePath)
+		if err != nil {
+			return nil, err
+		}
+
+		defer file.Close()
+		// defer OS.Remove(lifecyclePath)
```
jjbustamante, line 250:
```suggestion
Remove this comment
```

pkg/client/create_builder.go
```diff
@@ -361,3 +402,39 @@ func uriFromLifecycleVersion(version semver.Version, os string, architecture str
...
+
+func stripTopDirAndWrite(layerReader io.ReadCloser, outputPath string) (*OS.File, error) {
+	file, err := OS.Create(outputPath)
+	if err != nil {
+		return nil, err
+	}
+
+	tarWriter := tar.NewWriter(file)
+	tarReader := tar.NewReader(layerReader)
+	tarReader.Next()
```
jjbustamante, line 414: Why do you need this?

pkg/client/create_builder.go
rashadism: Will do

pkg/client/create_builder.go
rashadism: Its to make the tarReader skip the first entry, so that the top level directory can be discarded

jjbustamante: Good, maybe you can just add a comment, you know, for your future self 😄

AidanDelaney:
Why do we want to support a docker URI for lifecycle images?

natalieparellano:
> Why do we want to support a docker URI for lifecycle images?

@AidanDelaney simply for convenience. We publish the buildpacksio/lifecycle image as well as test images for every commit in development and it would be nice to simply refer to that image instead of having to build the tarball.

Feel free to opine further if you think this should go in another direction.

jjbustamante:
Hi, @rashadism are you still interested in working on this issue to get it into the last mile?

jjbustamante (approved):
I added some unit tests

PR #2117, opened 2024-04-04
Exposing the logic for processing the Docker Context to be public

### LINKED ISSUE #2113: Expose the logic to process Docker Context outside the internal package
### Description
With pack 0.31.0 we added [support](https://github.com/buildpacks/pack/pull/1878) for reading the configuration from the docker context and set the environment variables required by pack accordingly. That logic was added into the **internal** package which means it can't be reuse by the community who uses pack as library.

Double checking this code, I think it is save if we move it outside the **internal** package, in that case, end-uses can decide to call the method before creating a pack client if they want.

Currently the way we create a pack client can be checked [here](https://github.com/buildpacks/pack/blob/main/cmd/cmd.go#L141-L151)

### Proposed solution

Move the files: `internal/docker/context.go` and `internal/docker/context_test.go` to `pkg/client`

I thought on moving the context processing method inside the `client.NewClient` method, in this way it will simplify the pack client creation process.

### Additional context

jjbustamante:
## Summary
With pack 0.31.0 we added [support](https://github.com/buildpacks/pack/pull/1878) for reading the configuration from the docker context and set the environment variables required by pack accordingly. That logic was added to the **internal** package meaning it can't be reused by the community that uses pack as a library.

This PR is moving the docker context logic into `pkg/client` to make the logic available for external consumers

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #2113


natalieparellano (approved):
Looks good to me - @jabrown85 as the originator of this code, did you want to have a look?

PR #2123, opened 2024-04-09
Fix pack buildpack new --targets

### LINKED ISSUE #2120: `pack buildpack new --targets` creates `buildpack.toml` with wrong field/table names for distro name/version
### Summary
The `pack buildpack new` command has a `--targets` option that allows for pre-populating the `[[targets]]` table in the generated `buildpack.toml` file.

However, the command currently uses the wrong field and table names if a distro name/version is specified:
- `targets.distributions` (instead of `targets.distros`)
- `versions` of type array (instead of `version` of type string)

xref:
- https://github.com/buildpacks/spec/issues/401
- https://github.com/buildpacks/rfcs/pull/310

cc @natalieparellano 

---

##### Steps

1. `pack buildpack new testcase-targets-distros --api 0.10 --targets "linux/amd64:ubuntu@123"`
2. `cat testcase-targets-distros/buildpack.toml`

##### Current behavior

```toml
# ...
[[targets]]
  os = "linux"
  arch = "amd64"

  [[targets.distributions]]
    name = "ubuntu"
    versions = ["123"]
```

##### Expected behavior

```toml
# ...
[[targets]]
  os = "linux"
  arch = "amd64"

  [[targets.distros]]
    name = "ubuntu"
    version = "123"
```

---

### Environment

```
$ pack --version
0.33.2+git-f2cffc4.build-5562
```

### LINKED ISSUE #2121: `pack buildpack new --targets` with distro name+version gives incorrect warning about the version not being specified
### Summary
The `pack buildpack new` command has a `--targets` option that allows for pre-populating the `[[targets]]` table in the generated `buildpack.toml` file.

This `--targets` option supports also specifying the distro name and version.

However, if I specify **both** a name and version, I get an incorrect warning about there being no version specified.

---

##### Steps

1. `pack buildpack new testcase-targets-distros --api 0.10 --targets "linux/amd64:ubuntu@123"`

##### Current behavior

A warning is shown about there being no version specific, but one was specified:

```
$ pack buildpack new testcase-targets-distros --api 0.10 --targets "linux/amd64:ubuntu@123"
Warning: distro with name ubuntu has no specific version!
    create  buildpack.toml
    create  bin/build
    create  bin/detect
Successfully created testcase-targets-distros
```

##### Expected behavior

No warning message.

---

### Environment

```
$ pack --version
0.33.2+git-f2cffc4.build-5562
```

natalieparellano:
## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #2120
Resolves #2121


pkg/dist/buildmodule.go
```diff
@@ -56,10 +56,10 @@ type Target struct {
 	OS            string         `json:"os" toml:"os"`
 	Arch          string         `json:"arch" toml:"arch"`
 	ArchVariant   string         `json:"variant,omitempty" toml:"variant,omitempty"`
-	Distributions []Distribution `json:"distributions,omitempty" toml:"distributions,omitempty"`
+	Distributions []Distribution `json:"distros,omitempty" toml:"distros,omitempty"`
```
jjbustamante, line 59: Not sure how many people started using this new fields, but I will need to leave some **Note** in the release note for pack 0.34.0

jjbustamante (approved):
I ran the scenario from @edmorley 

Note: `pack` is the binary compiled from the branch

```bash
>  pack  buildpack new testcase-targets-distros --api 0.10 --targets "linux/amd64:ubuntu@123"
    create  buildpack.toml
    create  bin/build
    create  bin/detect
Successfully created testcase-targets-distros
➜  pack-2120 cat testcase-targets-distros/buildpack.toml
api = "0.10"
WithWindowsBuild = false
WithLinuxBuild = false

[buildpack]
  id = "testcase-targets-distros"
  version = "1.0.0"

[[targets]]
  os = "linux"
  arch = "amd64"

  [[targets.distros]]
    name = "ubuntu"
    version = "123"
```

Everything looks fine

internal/target/parse.go
```diff
@@ -66,11 +67,15 @@ func ParseDistro(distroString string, logger logging.Logger) (distro dist.Distri
...
+	if len(d) < 2 {
 		logger.Warnf("distro with name %s has no specific version!", style.Symbol(d[0]))
+		return distro, err
 	}
-	distro.Name = d[0]
-	distro.Versions = d[1:]
+	if len(d) > 2 {
+		return distro, fmt.Errorf("invalid distro: %s", distroString)
+	}
+	distro.Version = d[1]
```
jjbustamante, line 78: Maybe we could add a test case for this line?

jjbustamante:
I did the following test, I tried to package the Buildpack generated. 

Using the branch binary everything is fine, as expected

```bash
> $PACK_BINARY_FROM_BRANCH buildpack package testcase-targets-distros --format file
Successfully created package testcase-targets-distros.cnb and saved to file
```

And then, I tried to use the current `pack 0.33.2` version

```bash
> pack buildpack package testcase-targets-distros-2 --format file
ERROR: no compatible stacks among provided buildpacks
```

As expected an error must be thrown because we changed the toml schema, my only concern is, should we add a helpful message for end-users to point them out to update:
- `distributions` -> `distros`
- `versions` -> `version` (single value instead of an array)

jjbustamante:
Sorry about all the messages. 

The new pack version containing this feature is not going to fail with a buildpack using the previous schema, so things could change for end-users and they wouldn't notice

These examples uses the binary for this branch and package the same buildpack with the previous and the new schema

```bash
# Metadata is fine, it has the expected values
> docker inspect targets-distros-new-schema-new-version  | jq '.[] | .Config.Labels | ."io.buildpacks.buildpack.layers" | fromjson' | jq
{
  "testcase-targets-distros": {
    "1.0.0": {
      "api": "0.10",
      "targets": [
        {
          "os": "linux",
          "arch": "amd64",
          "distros": [
            {
              "name": "ubuntu"
            }
          ]
        }
      ],
      "layerDiffID": "sha256:5274312d546cfc8a8803d2ba932df72a27c8e65fb5015f4a6870dfef030764bf"
    }
  }
}

# Metadata is missing distros/name but the package is created.
> docker inspect targets-distros-old-schema-new-version  | jq '.[] | .Config.Labels | ."io.buildpacks.buildpack.layers" | fromjson' | jq
{
  "testcase-targets-distros": {
    "1.0.0": {
      "api": "0.10",
      "targets": [
        {
          "os": "linux",
          "arch": "amd64"
        }
      ],
      "layerDiffID": "sha256:60a056247c8b9513c22737dbca36a092836a9cd4bbedb4cf64f7f2172938df21"
    }
  }
}
```

I think we should:
- Failed creating the buildpack package if we detect the `distributions` and `versions`
- Or at least throw an **warning** message to get end-users attention, I am not sure about this, because if they are using a CI pipeline and the command runs successfully they will not pay attention to the warning

edmorley:
> I think we should:
> 
>     * Failed creating the buildpack package if we detect the `distributions` and `versions`
> 
>     * Or at least throw an **warning** message to get end-users attention, I am not sure about this, because if they are using a CI pipeline and the command runs successfully they will not pay attention to the warning

I'd love to see a warning (or error) added if possible, since it would have allowed me to realise sooner that `targets.distributions` wasn't the correct table name sooner (I'd been using `targets.distributions` since both the RFC and `pack buildpack new` used it).

However, I wonder if this would be best implemented as more strict `buildpack.toml` schema validation in general? ie: Warn (or error) on any unknown table/field names. Doing this would also catch cases like the `version` vs `versions` issue seen in https://github.com/buildpacks/spec/issues/401

edmorley:
> Fixes #2120

From skim reading, I think this PR might also fix #2121, thanks to the switch from `if len(d) <= 2` to `if len(d) < 2`?

jjbustamante:
> From skim reading, I think this PR might also fix #2121, thanks to the switch from `if len(d) <= 2` to `if len(d) < 2`?

Yeap, it is also fixing #2121

natalieparellano:
I can update this PR to add a warning if we encounter wrong stuff in buildpack.toml

internal/target/parse.go
natalieparellano: Added in 7949931b

natalieparellano:
@jjbustamante @edmorley I updated the PR to add the warnings. We should be good to go!

edmorley:
Amazing - looks great, thank you!

jjbustamante:
Awesome work @natalieparellano !!! ❤️

pkg/dist/buildmodule.go
robdimsdale: The `paketo` team (and Broadcom commercial teams) ran into an issue around this. Specifically, newer versions of the lifecycle require `[[targets.distributions]]` in `buildpack.toml` but `pack build` does not pick up those fields correctly.

PR #2125, opened 2024-04-15
Removing experimental configuration for extensions starting with API 0.13

### LINKED ISSUE #2066: Support Platform API 0.13
### Description

See spec milestone: https://github.com/buildpacks/spec/milestone/27

* [Platform API changes for image extension build contexts](https://github.com/buildpacks/spec/pull/384)
  * This [function](https://github.com/buildpacks/pack/blob/ded78f30bd5a0f0658f68f65b9532d2594ab3797/internal/build/lifecycle_execution.go#L889-L899) can no longer rely on the presence of a "build" directory to indicate build extensions - instead, we must traverse the directory and look for a `<dir>/<buildpack-id>/build.Dockerfile` 
* [Making builpack image extensions feature non-experimental](https://github.com/buildpacks/spec/issues/378)
  * Let's remove the [guard](https://github.com/buildpacks/pack/blob/ded78f30bd5a0f0658f68f65b9532d2594ab3797/pkg/client/build.go#L473-L475) in `pack build`
  * We don't have to set the env var [here](https://github.com/buildpacks/pack/blob/ded78f30bd5a0f0658f68f65b9532d2594ab3797/internal/build/lifecycle_execution.go#L427), [here](https://github.com/buildpacks/pack/blob/ded78f30bd5a0f0658f68f65b9532d2594ab3797/internal/build/lifecycle_execution.go#L721), [here](https://github.com/buildpacks/pack/blob/ded78f30bd5a0f0658f68f65b9532d2594ab3797/internal/build/lifecycle_execution.go#L742), and [here](https://github.com/buildpacks/pack/blob/ded78f30bd5a0f0658f68f65b9532d2594ab3797/internal/build/lifecycle_execution.go#L778) anymore
* [[RFC #0111] Support insecure registries](https://github.com/buildpacks/spec/issues/364) - see other [issue](https://github.com/buildpacks/pack/issues/1916)
* [[RFC 0125] Lifecycle parallel export](https://github.com/buildpacks/spec/issues/379) - maybe it doesn't make sense to expose this in pack just set; we might wait until this feature is requested

### Proposed solution

See above

jjbustamante:
## Summary

This PR adds some logic to determine when the experimental mode must be enable when running extensions. Since Platform API 0.13, extensions are no more experimental and pack doesn't need to set the env variable `CNB_EXPERIMENTAL_MODE=warn`, With this PR pack will evaluate if the API version is less than 0.13 and in such as case it will add the environment variable

#### Before

Pack enables experimental mode when running against lifecycle < 0.18.x

#### After

When running with platform API <= 0.12

```bash
===> EXTENDING (BUILD)
Running the extender on OS linux from image pack.local/builder/6463666b7974676b7763:latest with:
Container Settings:
  Args: /cnb/lifecycle/extender -app /workspace -log-level debug
  System Envs: CNB_EXPERIMENTAL_MODE=warn CNB_PLATFORM_API=0.12
```

When running with platform API >= 0.13 the `CNB_EXPERIMENTAL_MODE=warn` env is not set

```bash
===> EXTENDING (BUILD)
Running the extender on OS linux from image pack.local/builder/6e647a67686d6278736d:latest with:
Container Settings:
  Args: /cnb/lifecycle/extender -app /workspace -log-level debug
  System Envs: CNB_PLATFORM_API=0.13
```

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #2066


natalieparellano (commented):
This looks good @jjbustamante

PR #2126, opened 2024-04-16
[Buildkit] phase3 don't merge

WYGIN:
## Summary

Proof of concept for [LFX mentorship: BuildKit explore](https://mentorship.lfx.linuxfoundation.org/project/2c5ced86-d23b-41f5-aec3-59730e29f092)

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


pkg/dist/targets.go
```diff
@@ -0,0 +1,48 @@
+package dist
+
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+)
+
+type Target struct {
```
jjbustamante, line 9: Avoid trying to use Target for now, if you want, just hardcode the `os/arch` values for now. There is a bug with the implementation of the spec around `Distributions` and @natalieparellano is fixing it [here](https://github.com/buildpacks/pack/pull/2123/) once this PR is merged, then probably you can sync those changes and use it. Just want to avoid merge conflicts in the near future

Issue #1395, opened 2022-03-20
File ownership on files copied to the <app> dir are set to root when using Podman


frayer:
### Summary

Workspace files copied into the `<app>` directory of the Builder image have permissions set to `root` when `pack` is used in combination with Podman.  When doing the same against a Docker daemon, the `user:group` permissions match `${CNB_USER_ID}:${CNB_GROUP_ID}` of the selected Builder image.

This causes `permission denied` errors for buildpacks which attempt to write to the `<app>` directory such as `paketo-buildpacks/maven` since they are running as `CNB_USER_ID` and don't have permissions to write to the `<app>` directory.

---

##### Steps

I've created a minimal buildpack at https://github.com/frayer/details-buildpack to demonstrate the issue.

1. Install and configure Podman per the [Building on Podman](https://buildpacks.io/docs/app-developer-guide/building-on-podman/) instructions.
2. Clone the repo at https://github.com/frayer/details-buildpack
3. Execute the `pack build` once using against Docker, and once against Podman to observe the difference

> against Docker

```bash
git clone https://github.com/frayer/details-buildpack.git
cd details-buildpack
pack build test-app -B paketobuildpacks/builder:base -b ./buildpack
```

> against Podman

```bash
export DOCKER_HOST="unix://$(podman info -f "{{.Host.RemoteSocket.Path}}")"
git clone https://github.com/frayer/details-buildpack.git
cd details-buildpack
pack build test-app -B paketobuildpacks/builder:base -b ./buildpack --docker-host=inherit
```

##### Current behavior

Look in the log output of `pack build` for the `---> Working directory details` section.

> when executed against Docker

```
---> Working directory details
pwd=/workspace
ls -al
total 28
drwsrwsrwt 4 cnb  cnb  4096 Jan  1  1970 .
drwxr-xr-x 1 root root 4096 Mar 20 18:38 ..
drwxrwxr-x 8 cnb  cnb  4096 Mar 20 18:08 .git
-rw-rw-r-- 1 cnb  cnb   709 Mar 20 17:40 README.md
drwxrwxr-x 3 cnb  cnb  4096 Mar 20 17:40 buildpack
-rw-rw-r-- 1 cnb  cnb    22 Mar 20 18:09 main.sh
```

> when executed against Podman

```
---> Working directory details
pwd=/workspace
ls -al
total 28
drwxr-xr-x 4 root root 4096 Mar 20 18:36 .
[...]

PR #2129, opened 2024-04-17
Add ephemeral lifecycle image, enabling podman rootless support

BarDweller:
## Summary

Podman rootless will set ownership for a volume based on ownership of the first mountpoint it is mounted to, or the executing uid/gid if the mountpoint is absent. 

`pack` runs untrusted builds by running the first step (either detect or analyze depending on selected platform version) as root against a matched lifecycle image. Lifecycle images cannot contain a mountpoint for the workspace as lifecycle images are common across many builds that may not use the same uid/gid. 

Thus with podman rootless, pack ends up _causing the workspace volume to end up owned by root._ 

Subsequent mounts of the volume are unable to undo the ownership applied by this initial mount, leading to builds failing against podman rootless as it becomes impossible for the build uid/gid to create or remove files within the workspace directory. 

This change introduces an ephemeral lifecycle container, akin to the ephemeral builder image already in use. 

The ephemeral lifecycle image adds only the workspace directory to the selected lifecycle image, when a lifecycle image will be in use. The workspace directory is added owned by the current build uid/gid, and the ephemeral lifecycle image is used instead of the chosen lifecycle image for all container invocations that would have used the chosen lifecycle image.

#### Before
Example errors from the build would indicate permissions/ownership of `/workspace` to be insufficient for the build to proceed.
```
[extender (build)]   Removing source code
[extender (build)] unable to invoke layer creator
[extender (build)] unable to remove /workspace/.dockerignore
[extender (build)] unlinkat /workspace/.dockerignore: permission denied
```

#### After
Afterwards, builds succeed as expected
```
Successfully built image '127.0.0.1:5000/bardweller/stilettos'
```

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___


BarDweller:
PR is draft, as ideally some of the code that was duplicated into pkg/client/build.go here should become common utility code. 

- userAndGroupIDs on internal/builder/builder.go would benefit from either becoming public, or moving to a util class that can be driven from both places. 
- mountPathsForOS from internal/build/mount_paths.go would be awesome to be accessible as a utility too.

BarDweller:
This will help with 
 - https://github.com/buildpacks/pack/issues/1395

pkg/client/build.go
```diff
@@ -1320,6 +1351,109 @@ func (c *Client) processExtensions(ctx context.Context, builderImage imgutil.Ima
...
+
+func workspacePathForOS(os, workspace string) string {
+	if workspace == "" {
+		workspace = "workspace"
+	}
+	if os == "windows" {
+		// note we don't use ephemeral lifecycle when os is windows..
+		return "c:\\" + workspace
+	}
+	return "/" + workspace
```
jjbustamante, line 1391: I think this can be replaced by:

```go
return filepath.Join(paths.RootDir, workspace)
```

BarDweller: I don't think so, as filepath.Join will use the executing environments path separator, and the intent here is to use the separator for the os of the target image (eg, you could be building a linux container on windows).. this is based on how the ephemeral builder image uses mount_paths.go which performs similar path goop to handle paths-not-of-the-executing platform.

acceptance/acceptance_test.go
```diff
@@ -846,7 +848,7 @@ func testAcceptance(
 								assertions.NewOutputAssertionManager(t, output).ReportsSuccessfulImageBuild(repoName)
 
 								assertOutput := assertions.NewLifecycleOutputAssertionManager(t, output)
-								assertOutput.IncludesLifecycleImageTag(lifecycle.Image())
+								assertOutput.IncludesLifecycleImageTag(origLifecycle)
```
natalieparellano, line 851: nit: maybe we should call this `IncludesTagOrEphemeralLifecycle` (or something) to indicate the fallback

pkg/client/build.go
```diff
@@ -639,7 +664,13 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
...
 				return "", errors.New("missing manifest entries")
 			}
-			return manifestContents[0].Layers[len(manifestContents[0].Layers)-1], nil // we can assume the lifecycle layer is the last in the tar
+			// we can assume the lifecycle layer is the last in the tar, except if the lifecycle has been extended as an ephemeral lifecycle
+			layerOffset := 1
+			if strings.Contains(lifecycleOpts.LifecycleImage, "pack.local/lifecycle") {
+				layerOffset = 2
+			}
+
+			return manifestContents[0].Layers[len(manifestContents[0].Layers)-layerOffset], nil
```
natalieparellano, line 673: nit: this didn't exist before, but we should probably also do some check on `len(manifestContents[0].Layers)` to avoid a panic

jjbustamante:
@BarDweller I think I am fine with this one, if you can address the comments from Natalie I will merge it on Monday :)

jjbustamante (approved):
LGTM

PR #2136, opened 2024-04-18
Updating to Docker 26.x & lifecycle 0.19.4

jjbustamante:
## Summary
With docker 26.0.0 some refactoring was done and the parsing logic for volumes is not available on `darwin`, as a consequence pack failed to compile on mac machines. See this [discussion](https://github.com/moby/moby/discussions/47750) created for more details.

This PR in addition to update the dependencies to docker 26.0.0 also reimplement the `processVolumes` method for `darwin` os.

#### Before

when compiling pack after updating the dependencies to docker 26.0.0

```bash
=====> Linting code...
WARN [linters_context] rowserrcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
pkg/client/build.go:26:2: could not import github.com/docker/docker/volume/mounts (-: # github.com/docker/docker/volume/mounts
../../../../pkg/mod/github.com/docker/docker@v26.0.1+incompatible/volume/mounts/mounts.go:187:30: undefined: safepath.Join) (typecheck)
        "github.com/docker/docker/volume/mounts"
        ^
make: *** [lint] Error 1
```

#### After

The code compiles and the unit tests are passing

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #___


pkg/client/process_volumes_darwin.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func parseVolume(volume string) (types.ServiceVolumeConfig, error) {
+	// volume format: '<host path>:<target path>[:<options>]'
+	split := strings.Split(volume, ":")
+	if len(split) == 3 {
+		if split[2] != "ro" && split[2] != "rw" && !strings.Contains(split[2], "volume-opt") {
+			return types.ServiceVolumeConfig{}, errors.New(fmt.Sprintf("platform volume %q has invalid format: invalid mode: %s", volume, split[2]))
+		}
+	}
```
natalieparellano, line 45: Just curious, what happens if we don't check this? It seems that `loader.ParseVolume` handles at least some of this logic (https://github.com/docker/cli/blob/0bdc20ecbeb510083b98344be99c2e4f66b2ec9f/cli/compose/loader/volume.go#L16)

natalieparellano (commented):
This looks good @jjbustamante. We'll work on getting a lifecycle release out soon so that you can update to that version in go.mod

pkg/client/process_volumes_darwin.go
jjbustamante: If I remove the validation, this [test](https://github.com/buildpacks/pack/pull/2136/files#diff-03220a2f3a516ece374582fcc687eb7c47a3c6e2dcd11e0f5a117c5a9c0decc6R2753) fails. 

The internal parsing method [does not](https://github.com/docker/cli/blob/0bdc20ecbeb510083b98344be99c2e4f66b2ec9f/cli/compose/loader/volume.go#L82) throw an error with an invalid format volume option

natalieparellano: Ah, makes sense. Okay, seems fine to me

natalieparellano:
@jjbustamante I pushed 7cd4e0bd to bring the lifecycle up to a released version. When CI goes green we should be good to merge :)

natalieparellano (approved):

PR #2145, opened 2024-04-29
Platform 0.13: look for build Dockerfiles in <layers>/generated/<buildpack-id>/Dockerfile.build

natalieparellano:
Newer platforms don't copy the Dockerfile from where extensions output them

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


jjbustamante:
I was testing #2125 with the samples extensions and I realized I need this one first! :), when platform 0.13 is used I was getting:

```bash
ERROR: open /var/folders/l7/zgqmtk951vx4dy6q8zl_9js80000gn/T/pack.tmp279107406/build: no such file or directory
```

I will merge than one after this.

internal/build/lifecycle_execution.go
```diff
@@ -892,10 +892,21 @@ func (l *LifecycleExecution) hasExtensionsForBuild() bool {
 	}
 	// the directory is <layers>/generated/build inside the build container, but `CopyOutTo` only copies the directory
 	fis, err := os.ReadDir(filepath.Join(l.tmpDir, "build"))
+	if err == nil && len(fis) > 0 {
```
jjbustamante, line 895: This one is for API < 0.12 right? I think a comment here will be valuable

jjbustamante (approved):
Thank you @natalieparellano

PR #2151, opened 2024-05-10
Do not initialize `http.Client` during each request to a remote location

pbusko:
## Summary

Instead of initialising the `http.Client` during every call to the `downloader.Download`, use the `http.DefaultClient`. Also custom client can be provided with the `blob.WithClient` option (for example with a custom `RoundTripper` for injecting credentials)

## Documentation

    - [ ] Yes, see #___
    - [X] No

## Related

Resolves #___

### DIFF
--- a/pkg/blob/downloader.go
+++ b/pkg/blob/downloader.go
@@ -28,6 +28,14 @@ type Logger interface {
 	Writer() io.Writer
 }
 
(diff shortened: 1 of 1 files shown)


jjbustamante (approved):
LGTM

Thanks @pbusko

PR #2042, opened 2024-01-30
Warn if project.toml contains keys not supported by schema

colincasey:
## Summary

As detailed in #1922, a common source of confusion with `project.toml` is failing to declare the `schema-version`. The `schema-version` then silently defaults to `0.1` which can lead to confusing and hard to debug errors when the user is using keys from the `0.2` schema.

This PR adds messaging to the build output notifying the user if:
- the schema version defaults to `0.1`
- the project descriptor contains unsupported keys for the target schema version and these keys will be ignored

#### Before

```
Using project descriptor located at '/var/folders/r4/xq3q_hjj2v1ghqlnr1k1zrdw0000gp/T/unsupported-schema-keys-app2010798657/project.toml'
Builder 'localhost:55020/test/builder-meaprzkkef' is trusted
Warning: Ignoring the provided lifecycle image as the builder is trusted, running the creator in a single container using the provided builder
Pulling image 'localhost:55020/test/builder-meaprzkkef:latest'
latest: Pulling from test/builder-meaprzkkef
Digest: sha256:b23b53c8d15aefb31136b9c2cda830a62a801bfb3fad413c3496a775c3ca9bc1
Status: Image is up to date for localhost:55020/test/builder-meaprzkkef:latest
CheckReadAccess failed for the run image pack-test/run, error: HEAD https://index.docker.io/v2/pack-test/run/manifests/latest: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)
CheckReadAccess succeeded for the run image localhost:55020/pack-test/run
Selected run image mirror 'localhost:55020/pack-test/run'
Pulling image 'localhost:55020/pack-test/run'
latest: Pulling from pack-test/run
Digest: sha256:39b229916f00ebfd4e824e7d55f55d31679ab856f0ce6590ab156c85088f09f8
Status: Image is up to date for localhost:55020/pack-test/run:latest
Creating builder with the following buildpacks:
-> 'read/env@read-env-version'
-> 'noop.buildpack@noop.buildpack.version'
-> 'noop.buildpack@noop.buildpack.later-version'
-> 'simple/layers@simple-layers-version'
Using build cache volume 'pack-cache-some-org_yrjscyouuj_latest-82c320b56cc2.build'
Running the 'creator' on OS 'linux' from image 'pack.local/builder/616d78677a73746b7975:latest' with:
Container Settings:
[...]

PR #2155, opened 2024-05-15
Warn misconfiguration keys in project.toml only for tables we own

### LINKED ISSUE #1922: Pack silently fails to use `project.toml` if the `schema-version` is not specified
### Summary
If a `project.toml` file does not contain an appropriately set `schema-version`, Pack silently ignores values set in the `project.toml` file, leading to a frustrating to debug experience.

---

##### Steps

1. `mkdir testcase && cd $_ `
2. `echo -e "[io.buildpacks]\nbuilder = \"some-builder\"" > project.toml`
3. `pack build testcase`

##### Current behavior

The `pack build` tries to perform a build using whatever is set as the default builder.

The cause of this is that the `builder` property is only supported as of `project.toml` schema version 0.2, but Pack CLI silently defaults to v0.1 if no schema version was specified, and ignores knowns keys rather than failing with a schema validation error.

If I add this, then the expected behaviour occurs:

```
[_]
schema-version = "0.2"
```

##### Expected behavior

For either:
1. Pack to default The build to be performed using the specified builder (which would fail with an invalid builder error, but this is just a testcase)
2. Pack to output a warning or error saying "no `schema-version` specified, defaulting to schema v0.1", along with perhaps a schema v0.1 validation error about the `builder` property not being recognised.

---

##### pack info

```
$ pack report
Pack:
  Version:  0.31.0+git-3a994bd.build-5086
  OS/Arch:  darwin/arm64

Default Lifecycle Version:  0.17.1

Config:
  experimental = true
  
  [[trusted-builders]]
    name = "[REDACTED]"
  
  [[trusted-builders]]
    name = "[REDACTED]"
```

jjbustamante:
## Summary

With pack [PR 2042](https://github.com/buildpacks/pack/pull/2042) we added a warning message when unsupported keys are declared in `project.toml`, the problem was that our [spec](https://buildpacks.io/docs/reference/config/project-descriptor/) allow the end-users to extend the file outside all `non-_` table and this shouldn't print a warning 

#### Before

Any `project.toml` version 0.1 or 0.2 with something like declared

```toml
[unsupported-table]
unsupported-key = "some value"
``` 

Will print a warning message

#### After

We will only print a warning message for keys under `_.*` that are not supported or if `buildpack.io` is being used and not valid, everything else will be ignore

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #1922

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -80,7 +80,7 @@ func warnIfTomlContainsKeysNotSupportedBySchema(schemaVersion string, tomlMetaDa
 		for _, unsupportedKey := range unsupportedKeys {
 			logger.Warnf("- %s\n", unsupportedKey)
 		}
-		logger.Warn("The above keys will be ignored. If this is not intentional, maybe try updating your schema version.\n")
+		logger.Warn("The above keys will be ignored. If this is not intentional, try updating your schema version.\n")
 	}
 }
 
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -68,10 +68,9 @@ func ReadProjectDescriptor(pathToFile string, logger logging.Logger) (types.Desc
 func warnIfTomlContainsKeysNotSupportedBySchema(schemaVersion string, tomlMetaData toml.MetaData, logger logging.Logger) {
 	unsupportedKeys := []string{}
 
-	// filter out any keys from [_]
 	for _, undecodedKey := range tomlMetaData.Undecoded() {
 		keyName := undecodedKey.String()
-		if keyName != "_" && !strings.HasPrefix(keyName, "_.schema-version") {
+		if unsupportedKey(keyName, schemaVersion) {
 			unsupportedKeys = append(unsupportedKeys, keyName)
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/project/project.go
```diff
@@ -68,10 +68,9 @@ func ReadProjectDescriptor(pathToFile string, logger logging.Logger) (types.Desc
 func warnIfTomlContainsKeysNotSupportedBySchema(schemaVersion string, tomlMetaData toml.MetaData, logger logging.Logger) {
 	unsupportedKeys := []string{}
 
-	// filter out any keys from [_]
 	for _, undecodedKey := range tomlMetaData.Undecoded() {
 		keyName := undecodedKey.String()
-		if keyName != "_" && !strings.HasPrefix(keyName, "_.schema-version") {
+		if unsopportedKey(keyName, schemaVersion) {
```
jkutner, line 73:
```suggestion
		if unsupportedKey(keyName, schemaVersion) {
```

pkg/project/project.go
```diff
@@ -85,6 +84,17 @@ func warnIfTomlContainsKeysNotSupportedBySchema(schemaVersion string, tomlMetaDa
 	}
 }
 
+func unsopportedKey(keyName, schemaVersion string) bool {
```
jkutner, line 87:
```suggestion
func unsupportedKey(keyName, schemaVersion string) bool {
```

pkg/project/project_test.go
```diff
@@ -416,17 +427,36 @@ unsupported-key = "some value"
 
 			_, err = ReadProjectDescriptor(tmpProjectToml.Name(), logger)
 			h.AssertNil(t, err)
-
-			h.AssertContains(t, readStdout(), "Warning: The following keys declared in project.toml are not supported in schema version 0.1:\nWarning: - unsupported-table\nWarning: - unsupported-table.unsupported-key\nWarning: The above keys will be ignored. If this is not intentional, maybe try updating your schema version.\n")
+			h.AssertContains(t, readStdout(), "Warning: The following keys declared in project.toml are not supported in schema version 0.1:\nWarning: - project.authors\nWarning: - io.buildpacks.build.env\nWarning: - io.buildpacks.build.env.name\nWarning: - io.buildpacks.build.env.value\nWarning: The above keys will be ignored. If this is not intentional, maybe try updating your schema version.\n")
```
natalieparellano, line 430:
```suggestion
			h.AssertContains(t, readStdout(), "Warning: The following keys declared in project.toml are not supported in schema version 0.1:\nWarning: - project.authors\nWarning: - io.buildpacks.build.env\nWarning: - io.buildpacks.build.env.name\nWarning: - io.buildpacks.build.env.value\nWarning: The above keys will be ignored. If this is not intentional, try updating your schema version.\n")
```

natalieparellano: Slightly more professional :)

PR #2161, opened 2024-05-17
When validating buildpack.toml, don't warn for keys that are actually defined

natalieparellano:
## Summary

Fixes the overly aggressive validation that was added in https://github.com/buildpacks/pack/pull/2123

### DIFF
--- a/pkg/client/example_fetcher_test.go
+++ b/pkg/client/example_fetcher_test.go
@@ -54,6 +54,6 @@ func (f *fetcher) Fetch(_ context.Context, imageName string, _ image.FetchOption
 	return nil, errors.New("not implemented")
 }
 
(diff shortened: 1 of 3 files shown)


pkg/client/example_fetcher_test.go
```diff
@@ -54,6 +54,6 @@ func (f *fetcher) Fetch(_ context.Context, imageName string, _ image.FetchOption
 	return nil, errors.New("not implemented")
 }
 
-func (f *fetcher) CheckReadAccess(_ string, _ FetchOptions) bool {
+func (f *fetcher) CheckReadAccess(_ string, _ image.FetchOptions) bool {
```
natalieparellano, line 57: Not sure how this happened...

Issue #2154, opened 2024-05-14
Add a new `--platform` or `--target` flag to `pack build` command to avoid breaking M1 workflow users after multi-platform is released


jjbustamante:
### Description

Right now, multi-arch builders aren't too prevalent in the CNB ecosystem. Most builders are linux/amd64. But with the release of pack 0.34 (enables multi-arch builder creation) this might change.

Today, it is possible to specify the os/arch for a build by specifying the (not multi-arch) builder that supports the desired platform. When more builders are multi-arch, this selection mechanism goes away.

Specifically, M1 users with emulation enabled might lost the ability to build linux/amd64 images because when using a multi-arch builder, the linux/arm64 version will be automatically selected.

### Proposed solution

We should support `pack build --platform <my desired platform>`

### Describe alternatives you've considered

`pack build --target <my desired platform>` - but this probably makes sense to add when CNB supports cross-platform builds. As an example, `pack build --platform <my desired build platform> --target <my desired runtime platform>` could become a thing.

### Additional context
- [ ] This feature should be documented somewhere

jjbustamante:
cc @natalieparellano 

We need to fill the description.

PR #2086, opened 2024-03-01
Implementation of the multi-platform support for builders and buildpack packages RFC 0128

jjbustamante:
## Summary
The purpose of this PR is to implement the RFC [0128](https://github.com/buildpacks/rfcs/blob/main/text/0128-multiarch-builders-and-package.md). 

It adds the capability to the `pack buildpack package` and `pack builder create` to generate multi-platform OCI images and create an image index to combine them.

A draft version of this PR was demo during KubeCon EU 24. See the recording [here](https://youtu.be/cenTw6WzQv8?si=_vi0P4aJujT6pZSS)

#### Before

-  We were not able to create multi-platform buildpacks and builders 

#### After

- Follow the RFC documentation to specify `Targets` entries in the `buildpack.toml` or `package.toml` files.

Check the document mentioned above, it has a detailed output

## Documentation

- Should this change be documented?
    - [X] Yes, see 
    - [ ] No

## Related

Resolves #1459
Resolves #2079

PR #2162, opened 2024-05-20
Support `pack build --platform`

natalieparellano:
Fixes https://github.com/buildpacks/pack/issues/2154

Has a merge conflict with https://github.com/buildpacks/pack/pull/2086 - let's get that one in first and then update this one


pkg/client/build.go
```diff
@@ -335,6 +345,11 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "getting builder architecture")
 	}
 
+	platformToUse := opts.Platform
+	if platformToUse == "" {
+		platformToUse = fmt.Sprintf("%s/%s", builderOS, builderArch) // TODO: what about arch variant, etc.
```
natalieparellano, line 350: This question existed before this PR I am just wondering what we plan to do about it

pkg/client/build_test.go
```diff
@@ -2261,6 +2261,158 @@ api = "0.2"
...
+
+				fakePackage = fakes.NewImage("example.com/some/package", "", nil)
+				h.AssertNil(t, dist.SetLabel(fakePackage, "io.buildpacks.buildpack.layers", bpLayers))
+				h.AssertNil(t, dist.SetLabel(fakePackage, "io.buildpacks.buildpackage.metadata", md))
+
+				h.AssertNil(t, fakePackage.AddLayer(metaBuildpackTar))
+				h.AssertNil(t, fakePackage.AddLayer(childBuildpackTar))
+
+				fakeImageFetcher.LocalImages[fakePackage.Name()] = fakePackage
+			})
```
natalieparellano, line 2344: Setup just for buildpack package, maybe we can find a way to re-use this code from other tests

pkg/image/fetcher.go
```diff
@@ -110,9 +110,12 @@ func (f *Fetcher) Fetch(ctx context.Context, name string, options FetchOptions)
...
 	if err = f.pullImage(ctx, name, options.Platform); err != nil {
-		// sample error from docker engine:
-		// image with reference <image> was found but does not match the specified platform: wanted linux/amd64, actual: linux
-		if strings.Contains(err.Error(), "does not match the specified platform") {
+		// FIXME: this matching is brittle and the fallback should be removed when https://github.com/buildpacks/pack/issues/2079
+		// has been fixed for a sufficient amount of time.
+		// Sample error from docker engine:
+		// `image with reference <image> was found but does not match the specified platform: wanted linux/amd64, actual: linux`
+		if strings.Contains(err.Error(), "does not match the specified platform") &&
+			strings.HasSuffix(strings.TrimSpace(err.Error()), "actual: linux") {
```
natalieparellano, line 118: I added this to make the acceptance test work, but I think it's actually closer to what we want anyway

pkg/client/build.go
```diff
@@ -335,6 +345,11 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 		return errors.Wrapf(err, "getting builder architecture")
 	}
 
+	platformToUse := opts.Platform
+	if platformToUse == "" {
+		platformToUse = fmt.Sprintf("%s/%s", builderOS, builderArch) // FIXME: what about arch variant, etc.
```
natalieparellano, line 350: This question existed before this PR I am just wondering what we plan to do about it

acceptance/managers/image_manager.go
```diff
@@ -62,6 +62,13 @@ func (im ImageManager) HostOS() string {
 	return daemonInfo.OSType
 }
 
+func (im ImageManager) HostArch() string {
+	im.testObject.Helper()
+	daemonInfo, err := im.dockerCli.Info(context.Background())
```
jjbustamante, line 67: I think I have had some issues with this in the past. Shouldn't be `dockerCli.ServerVersion` ?

natalieparellano: Updated in f8a10339 (I realized we don't need it)

natalieparellano: For posterity, the version API is safer because:

```
/info os is: linux
/info arch is: x86_64
/version os is: linux
/version arch is: amd64
```

Version arch is what we want when pulling images

pkg/client/build.go
natalieparellano: Fixed in 863eb981

pkg/client/build.go
```diff
@@ -491,6 +516,11 @@ func (c *Client) Build(ctx context.Context, opts BuildOptions) error {
 	}
 	defer c.docker.ImageRemove(context.Background(), ephemeralBuilder.Name(), types.RemoveOptions{Force: true})
 
+	builderOS, err := rawBuilderImage.OS()
+	if err != nil {
+		return fmt.Errorf("failed to get builder OS: %w", err)
+	}
```
natalieparellano, line 522: I think we could also do `targetToUse.OS`

pkg/client/build.go
```diff
@@ -1206,19 +1256,10 @@ func (c *Client) fetchBuildpack(ctx context.Context, bp string, relativeBaseDir
...
-
-		builderArch, err := builderImage.Architecture()
-		if err != nil {
-			return nil, nil, errors.Wrapf(err, "getting builder architecture")
-		}
-
 		downloadOptions := buildpack.DownloadOptions{
 			RegistryName:    registry,
-			Target:          &dist.Target{OS: builderOS, Arch: builderArch},
+			ImageOS:         targetToUse.OS,
```
natalieparellano, line 1261: This seemed to be needed by unit tests

jjbustamante: I ran locally, and removing that line it's ok for me

pkg/client/build.go
jjbustamante: Interesting, you moved the validation way too far, is it for any particular reason?

jjbustamante:
This looks good to me, the only thing is adding back `ImageOS: targetToUse.OS` for testing because I didn't get any failure locally

pkg/client/build.go
natalieparellano: Removed in 08995083

pkg/client/build.go
natalieparellano: It's because we are calling `getTargetFromBuilder` where we were calling `rawBuilderImage.OS()` before

pkg/client/build_test.go
natalieparellano: Simplified in 89cb6173

PR #2166, opened 2024-05-24
Building darwin arm64 binaries by default on CI and amd64 during releases

### LINKED ISSUE #2167: Publishing wrong architecture for `macos` intel binaries
### Description
Once we released pack [0.34.0-rc1](https://github.com/buildpacks/pack/releases/tag/v0.34.0-rc1) I noticed the binaries for macos intel machine is actually a `arm64` binary, the reason is because the latest macos runner from github is now `macos14` and than runner is based on arm64 architecture.  Github [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) 

- When CI runs, we build binaries for linux, macos and windows 
- When a Release is created, build binaries for other architectures, like: linux-arm64, darwin-arm64, linux-s390x

The problem was that we were expecting the darwin binary built during CI to be based on amd64 and since the runner was updated this not true. We ended up shipping the same binary (arm64) twice

### Proposed solution

The easiest way to fix it, is to build by default the `arm64` binary during CI, and build the `amd64` when a release is done. The problem with this solution is that sometimes we ask people from the community to download binaries from a PR and in this case, if they running macos on intel, we will not have binaries available.

jjbustamante:
## Summary

By default we will built `darwin/arm64` binaries during CI and `darwing/amd64` binaries during release, because the new macos runner is based on `arm` architecture

### Before

This is a image from our last PR before releasing `0.34.0-rc1` the `macos` binary is actually an `arm64` binary instead of `amd64`

Screenshot 2024-05-24 at 10 40 41 AM

### After

Now, we actually saved the binary as `arm64` and the `amd64` will be built later

Screenshot 2024-05-24 at 10 39 05 AM

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #2167

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,8 @@ jobs:
         config: [macos, linux, windows-lcow, windows-wcow]
         include:
           - config: macos
-            os: macos
+            # since macos-14 the latest runner is arm64
+            os: macos-arm64
             runner: macos-latest
             no_docker: "true"
             pack_bin: pack
@@ -170,8 +171,9 @@ jobs:
           - name: linux-arm64
             goarch: arm64
             goos: linux
-          - name: macos-arm64
-            goarch: arm64
+          - name: macos
+            # since macos-14 default runner is arm, we need to build for intel architecture later
+            goarch: amd64
             goos: darwin
           - name: linux-s390x
             goarch: s390x


.github/workflows/build.yml
```diff
@@ -23,7 +23,8 @@ jobs:
         config: [macos, linux, windows-lcow, windows-wcow]
         include:
           - config: macos
-            os: macos
+            # since macos-14 the latest runner is arm64
+            os: macos-arm64
```
natalieparellano, line 27: Can we be certain that we'll always land on a `macos-14` runner? Is it deterministic?

jjbustamante: well, until the latest runner is updated I think the answer is yes.

Issue #2159, opened 2024-05-17
Feat: PR Title Validation


sanmarg:
The purpose of this issue is to ensure that all pull requests (PRs) submitted to our repository follow a consistent and meaningful naming convention. This will improve the readability and maintainability of our project by making it easier to understand the purpose of each PR at a glance. I came across this project and was exploring it and had difficulty in understanding the PR's at glance So I wish to add this. Below is more information for it.

### Problem:
Currently, there is no standardized format for PR titles in our repository. This can lead to confusion and difficulty in understanding the intent and scope of changes introduced by each PR. Inconsistent PR titles also make it harder to generate meaningful release notes and track changes over time.

### Solution:
To address this issue, we will implement a GitHub Action that automatically checks the titles of PRs against a predefined format. This action will ensure that PR titles follow a specific pattern, making them more informative and standardized.

PR #2170, opened 2024-05-29
Added the Pr title checker workflow and config

sanmarg:
Added for #2159


AidanDelaney:
Do we really want to place more constraints on PR titles?  If we do, might https://github.com/mergeability/mergeable be a more generic alternative implementation that supports PR title checking and other functionality?

I would need to be convinced that putting more checks on contributions is useful.

PR #2178, opened 2024-06-05
Add `heroku/builder:24` as a trusted + suggested builder

edmorley:
## Summary

We have just completed our new `heroku/builder:24` builder, which is based upon Ubuntu 24.04 and is also our first multi-architecture (AMD64+ARM64) builder.

As such, I have:
- Added `heroku/builder:24` as a trusted+suggested builder
- Changed the existing `heroku/builder:22` builder entry so that it is now just trusted, and not also suggested (since we want to encourage users to use the newer builder, especially since it's now multi-arch).
- Re-synced the builder description metadata here with what's in our `builder.toml` configs. (The descriptions in this repo are what gets used in offline mode, when Pack CLI can't pull the latest image description from the image labels on the registry.)

See:
https://github.com/heroku/cnb-builder-images/pull/534
https://github.com/heroku/cnb-builder-images/blob/main/builder-20/builder.toml 
https://github.com/heroku/cnb-builder-images/blob/main/builder-22/builder.toml
https://github.com/heroku/cnb-builder-images/blob/main/builder-24/builder.toml

## Documentation

    - [ ] Yes, see #___
    - [x] No


jjbustamante (approved):
LGTM

natalieparellano:
Manually verified the DCO

edmorley:
Thank you for the review/merge :-)

Issue #2195, opened 2024-07-02
pack should not create an ephemeral builder if it isn't truly needed


natalieparellano:
### Description

Today, we always create an ephemeral builder, because we assume that it's an inexpensive operation if nothing changed.

When containerd storage is enabled, we can no longer omit base layers for images when saving (see https://github.com/buildpacks/imgutil/pull/222#discussion_r1372167387). This unfortunately makes creating ephemeral builders expensive. We should only do this when it's absolutely necessary.

### Proposed solution

Return the original builder if no buildpacks were added, order was not modified, etc.

### Additional context
- [ ] This feature should be documented somewhere

PR #2196, opened 2024-07-02
Don't create an ephemeral builder if it isn't truly needed

natalieparellano:
Fixes https://github.com/buildpacks/pack/issues/2195

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


jjbustamante:
Do we have an acceptance test covering this scenario? The only thing that comes to my mind is we are not trying to save the  ephemeral builder at some point during the build and we end up introducing some bug

jjbustamante (approved):
LGTM

My only concern is to double-check the acceptance test coverage

natalieparellano:
> Do we have an acceptance test covering this scenario?

I didn't add an explicit test for this, because I am lazy (and also because our current suite is long enough) but the tests were failing before I added https://github.com/buildpacks/pack/pull/2196/commits/423f59667110a4c602955debf835a6b15220de57 because we were cleaning up the original builder, making it unavailable for other tests.

jjbustamante:
> > Do we have an acceptance test covering this scenario?
> 
> I didn't add an explicit test for this, because I am lazy (and also because our current suite is long enough) but the tests were failing before I added [423f596](https://github.com/buildpacks/pack/commit/423f59667110a4c602955debf835a6b15220de57) because we were cleaning up the original builder, making it unavailable for other tests.

Then I am ok with it

PR #2043, opened 2024-01-30
Separate suggested builders from default trusted builders

colincasey:
## Summary

This PR replaces the `SuggestedBuilder` struct with a `KnownBuilder` struct that contains the same details + the following boolean flags: 
- `Suggested` 
- `Trusted`

These flags can be set on a `KnownBuilder` to control if it should be displayed as a suggested builder and/or be added as a default trusted builder.

With these changes, `heroku/builder:20` has also been removed from the suggested list but remains as a default trusted builder.

## Output

`pack builder suggest`

#### Before

```
Suggested builders:
        Google:                gcr.io/buildpacks/builder:v1                            Ubuntu 18 base image with buildpacks for .NET, Go, Java, Node.js, and Python                                                                                     
        Heroku:                heroku/builder:20                                       Heroku-20 (Ubuntu 20.04) base image with buildpacks for Go, Java, Node.js, PHP, Python, Ruby & Scala.                                                            
        Heroku:                heroku/builder:22                                       Heroku-22 (Ubuntu 22.04) base image with buildpacks for Go, Java, Node.js, PHP, Python, Ruby & Scala.                                                            
        Paketo Buildpacks:     paketobuildpacks/builder-jammy-base                     Ubuntu 22.04 Jammy Jellyfish base image with buildpacks for Java, Go, .NET Core, Node.js, Python, Apache HTTPD, NGINX and Procfile                               
        Paketo Buildpacks:     paketobuildpacks/builder-jammy-buildpackless-static     Static base image (Ubuntu Jammy Jellyfish build image, distroless-like run image) with no buildpacks included. To use, specify buildpacks at build time.         
        Paketo Buildpacks:     paketobuildpacks/builder-jammy-full                     Ubuntu 22.04 Jammy Jellyfish full image with buildpacks for Apache HTTPD, Go, Java, Java Native Image, .NET, NGINX, Node.js, PHP, Procfile, Python, and Ruby     
        Paketo Buildpacks:     paketobuildpacks/builder-jammy-tiny                     Tiny base image (Ubuntu Jammy Jellyfish build image, distroless-like run image) with buildpacks for Java, Java Native Image and Go                               

Tip: Learn more about a specific builder with:
        pack builder inspect <builder-image>
```

#### After

```
Suggested builders:
        Google:                gcr.io/buildpacks/builder:v1                            Ubuntu 18 base image with buildpacks for .NET, Go, Java, Node.js, and Python                                                                                     
[...]

PR #2205, opened 2024-07-04
Trusted builders fix

### LINKED ISSUE #2198: pack build not correctly treating trusted builders as trusted
### Summary

The `pack build` command is not correctly treating builders as trusted after #2043. Specifically, if a builder is "trusted but not suggested" then `pack build` treats it as untrusted, even though the builder still shows up as trusted in the output of `pack config trusted-builders`.

---

##### Steps

1. Update to latest Pack CLI (0.34.2) 
2. Remove the Pack CLI config (to reset any manually added trusted builders): `rm ~/.pack/config.toml`
3. `mkdir testcase && touch testcase/requirements.txt`
4. `pack build --builder heroku/builder:22 --path testcase testapp --verbose`

##### Current behavior

The `heroku/builder:22` builder is treated as untrusted by `pack build`:

```
$ pack build --builder heroku/builder:22 --path testcase testapp --verbose
Builder heroku/builder:22 is untrusted
As a result, the phases of the lifecycle which require root access will be run in separate trusted ephemeral containers.
For more information, see https://medium.com/buildpacks/faster-more-secure-builds-with-pack-0-11-0-4d0c633ca619
Pulling image index.docker.io/heroku/builder:22
22: Pulling from heroku/builder
...
```

##### Expected behavior

For `heroku/builder:22` to be treated as a trusted builder, given that the `pack config trusted-builders` Pack CLI command lists it under trusted builders:

```
$ pack config trusted-builders
Trusted Builders:
  gcr.io/buildpacks/builder:v1
  heroku/builder:20
  heroku/builder:22
  heroku/builder:24
...
```

...and that it's marked as trusted here:
https://github.com/buildpacks/pack/blob/ce8db3c5d11b8e040fd565632668a697e53e4f87/internal/builder/known_builder.go#L28-L31

### Notes

This appears to be caused by a bug here:
https://github.com/buildpacks/pack/blob/ce8db3c5d11b8e040fd565632668a697e53e4f87/internal/commands/commands.go#L108-L116

In that function, the `return isSuggestedBuilder(builder)` line should instead be `return isTrustedBuilder(builder)`.

It looks like when the concept of "suggested" vs "trusted" builders was added in https://github.com/buildpacks/pack/commit/1b68d1261eb8e8f01aade2737d5268e54f4a4604 some places in the codebase weren't updated along with the others.

This is why `pack config trusted-builders` says the builder is trusted, but `pack build` says it is not.

In addition to fixing this bug, it seems worth reducing the number of places that implement the same "is this a trusted builder" check to avoid issues like this (where the behaviour of different Pack subcommands for common functionality can diverge). Plus [this comment](https://github.com/buildpacks/pack/blob/ce8db3c5d11b8e040fd565632668a697e53e4f87/pkg/client/build.go#L412) needs updating too.

cc @colincasey @schneems

---

##### pack info

```
$ pack report
Pack:
  Version:  0.34.2+git-ce8db3c.build-6005
  OS/Arch:  darwin/arm64

Default Lifecycle Version:  0.19.6

Config:
```

##### docker info

N/A

colincasey:
## Summary

With the changes introduced in #2043 for separating suggested builders and trusted builders, there were several places that still had logic referencing suggested builders in the trusted context. This PR updates those code paths to only consider trusted builders and extracts out a shared function `IsKnownTrustedBuilder` that can be used for "is this a trusted builder" checks.

## Output

There is a small output change when calling `pack config trusted-builders remove <x>` with a builder in the known trusted list.  The previous output gave the following message:

```
Builder <x> is a suggested builder, and is trusted by default. Currently pack doesn't support making these builders untrusted
```

Which now says:

```
Builder <x> is a known trusted builder. Currently pack doesn't support making these builders untrusted
```

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #2198


edmorley (commented):
Thank you for fixing this! The changes look good to me to fix the issue, though it would be good to have a test for this. I presume something similar to the tests here?
https://github.com/buildpacks/pack/blob/ce8db3c5d11b8e040fd565632668a697e53e4f87/internal/commands/build_test.go#L93-L128

internal/commands/build_test.go
```diff
@@ -126,6 +126,19 @@ func testBuildCommand(t *testing.T, when spec.G, it spec.S) {
 					h.AssertContains(t, outBuf.String(), "Builder 'heroku/builder:24' is trusted")
 				})
 			})
+
+			when("the builder is not trusted", func() {
```
edmorley, line 130: Thank you for adding a test, however I don't believe the test changes mean the bug is now tested.

There are a number of different scenarios:
1. The builder in a built-in known builder that is both trusted and suggested
2. The builder in a built-in known builder that is trusted but not suggested
3. The builder has been manually trusted via the user provided config
4. The builder is untrusted

The existing tests in this file appear to test (1) + (3).

The new test you have added tests (4).

However, the bug was in the behaviour for (2), and there isn't a test for that.

jjbustamante:
I am ok with the change, I agree with Ed, add the missing test case

internal/commands/build_test.go
colincasey: Added test for (2)

internal/commands/build_test.go
```diff
@@ -126,6 +126,32 @@ func testBuildCommand(t *testing.T, when spec.G, it spec.S) {
...
+					command.SetArgs([]string{"image", "--builder", "heroku/builder:22"})
+					h.AssertNil(t, command.Execute())
+					h.AssertContains(t, outBuf.String(), "Builder 'heroku/builder:22' is trusted")
+				})
+			})
+
+			when("the builder is not trusted", func() {
+				it("warns the user that the builder is untrusted", func() {
+					mockClient.EXPECT().
+						Build(gomock.Any(), EqBuildOptionsWithImage("org/builder:unknown", "image")).
```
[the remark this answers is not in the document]
colincasey: I think you're right @edmorley. I tried `EqBuildOptionsWithTrustedBuilder(false)` initially but the assertion would always fail. Debugging through that assertion showed that it was never checking the expected value, just asserting that the builder was trusted so it would only pass if that was `true`. I've correct the matcher logic and switched back to `EqBuildOptionsWithTrustedBuilder(false)`.

natalieparellano (approved):

Issue #2219, opened 2024-04-23
Security review: launch build containers in a separate ephemeral Docker bridge network


natalieparellano:
### Description

In the security review, this is `MED-2: Docker permissive inter-container connectivity`. The action plan asks us to ensure that

> Launch the Docker build containers in a separate ephemeral Docker bridge network

Reference: https://docs.docker.com/network/network-tutorial-standalone/#use-user-defined-bridge-networks

### Proposed solution

If `pack build --network` is NOT set, we should
* create an ephemeral bridge network
* start containers with this network
* tear the network down after the build (or if the build fails for any reason)

~We'll probably need to pass `--network=host` to more acceptance tests, so that build containers can reach the test registry.~ This ended up not being needed (we are already setting the network for these tests)

### Additional context
- [ ] This feature should be documented somewhere

modulo11:
This change broke setups using podman. Apparently slashes are not allowed while it's perfectly fine for docker:

```
Using build cache volume pack-cache-library_node-npm-sample_latest-a3b5329b03b5.build
ERROR: failed to build: executing lifecycle: failed to create ephemeral bridge network: Error response from daemon: network name pack.local/network/7176696877677273646a invalid: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
```

natalieparellano:
> Apparently slashes are not allowed while it's perfectly fine for docker

That is unfortunate. I'll put up a fix for this. We can release it in the next patch.

PR #2241, opened 2024-07-31
Fix ephemeral bridge network name for podman

natalieparellano:
See https://github.com/buildpacks/pack/issues/2219#issuecomment-2260923340

### DIFF
--- a/internal/build/lifecycle_execution.go
+++ b/internal/build/lifecycle_execution.go
@@ -211,7 +211,7 @@ func (l *LifecycleExecution) Run(ctx context.Context, phaseFactoryCreator PhaseF
 		if l.os == "windows" {
 			driver = "nat"
 		}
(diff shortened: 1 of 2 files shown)


edmorley (approved):

natalieparellano:
Hmm... the LCOW failures are weird:

```
ERROR: failed to build: executing lifecycle: failed to create ephemeral bridge network: Error response from daemon: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network
```

Could that really be stemming from `/` vs `-`?

natalieparellano:
The LCOW worker was full of ephemeral networks. Likely some containers are still being deleted when we try to delete the network, so I added retries.

edmorley:
I think the milestone for this PR should be `0.35.1` rather than `0.36.0`, since it made the release:
https://github.com/buildpacks/pack/compare/v0.35.0...v0.35.1

PR #2248, opened 2024-08-12
added the condition for args [""] and no args[]

keshavdalmia10:
## Summary

Added the condition for args [""] and no args[]

## Output

The CLI will display <None> if there are no arguments.

#### Before
There was no difference when args[""] or arg[]
#### After
Now the <None> will display when no args.

## Documentation

    - [ ] Yes
    - [ ] No

## Related

Resolves ##1283


natalieparellano:
Apologies for the delay in getting to this one, we will look at it soon

natalieparellano (commented):
@keshavdalmia10 thank you for taking this on, and I really apologize for the long delay in getting back to you about this PR. I think this is headed in the right direction, but we still need to pass on the `EnhancedProcess` so that `ArgsDisplay` is shown to the user.

Actually, I'm not sure that we need `EnhancedProcess` in pkg/client/inspect_image.go. I would make the changes here: https://github.com/buildpacks/pack/blob/4ffdb5e09c7c1c42f0d1d17de65d0f39bc8d517f/internal/inspectimage/writer/human_readable.go#L190-L202

My templating is a little rusty, but you could check out https://gobyexample.com/text-templates as a place to get started inserting the needed logic into the template

PR #2250, opened 2024-08-13
Add .grype.yaml file to ignore known false positives from scans

natalieparellano:
We do this for the lifecycle: https://github.com/buildpacks/lifecycle/blob/main/.grype.yaml

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -1,0 +1,3 @@
+ignore:
+  - vulnerability: CVE-2015-5237 # false positive, see https://github.com/anchore/grype/issues/558
+  - vulnerability: CVE-2021-22570 # false positive, see https://github.com/anchore/grype/issues/558


.grype.yaml
```diff
@@ -0,0 +1,3 @@
+ignore:
+  - vulnerability: CVE-2015-5237 # false positive, see https://github.com/anchore/grype/issues/558
+  - vulnerability: CVE-2021-22570 # false positive, see https://github.com/anchore/grype/issues/558
```
edmorley, line 3: I'm curious why this is needed since the linked issues were supposedly fixed by Grype adding these to the default ignore list in https://github.com/anchore/grype/pull/1062? Are we using an older Grype version, or does is this another variation not covered by their exclusion? 

I tried looking at the CI logs (https://github.com/buildpacks/pack/actions/runs/10344879230/job/28631066568#step:6:20), but the full details from Grype don't seem to be printed?

natalieparellano: Ah, good point - we're still on an older version of the action: https://github.com/buildpacks/pack/blob/3a22a7ffef9e69b33ba32fcbfbe6491a09d5397c/.github/workflows/check-latest-release.yml#L73

jjbustamante: Should we try to update the `grype` version first?

jjbustamante: Oh! we have this [PR](https://github.com/buildpacks/pack/pull/2215) for that

natalieparellano:
Closing in favor of https://github.com/buildpacks/pack/pull/2215

Issue #2255, opened 2024-08-29
Ephemeral builder image not deleted at the end of the build


edmorley:
### Summary

When a `pack build` is performed that creates an ephemeral builder image, that builder image is now left behind at the end of the build when it was previously deleted. After lots of builds in a row this results in many stray images left behind.

---

##### Steps

1. `docker system prune --all --volumes --force`
2. `docker images`
3. `mkdir testcase && cd $_ && touch Procfile`
4. `pack build testcase --builder heroku/builder:24 --buildpack heroku/procfile@3.1.1`
5. `docker images | rg builder`

##### Current behavior

Unexpected leftover ephemeral builder image:

```
$ docker images | rg builder
pack.local/builder/796e7271616672796575   latest    92e6cc20ac67   44 years ago   1.05GB
...
```

##### Expected behavior

No leftover ephemeral builder image.

---

##### pack info

- Pack CLI v0.35.1+git-3a22a7f.build-6099
- lifecycle v0.20.1

##### docker info

<details>

<summary>docker info</summary>

```
Client:
 Version:    27.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1-desktop.1
    Path:     /Users/emorley/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1-desktop.1
    Path:     /Users/emorley/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.34
    Path:     /Users/emorley/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Alpha) (Docker Inc.)
    Version:  v0.0.14
    Path:     /Users/emorley/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/emorley/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.25
    Path:     /Users/emorley/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.5
    Path:     /Users/emorley/.docker/cli-plugins/docker-feedback
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.3.0
[...]

PR #2260, opened 2024-09-05
Fix deletion of ephemeral builder

natalieparellano:
`createEphemeralBuilder` mutates the provided `rawBuilderImage`, so we must save the image name before this method is called.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


natalieparellano:
Confirmed manually that with this change we don't see a `pack.local/builder/<uid>` hanging around after the build

edmorley (approved):
@natalieparellano Thank you for looking into this - I've tested locally using the artifact attached to the GitHub Actions run and confirm it resolves the issue (using the STR from #2255) :-)

edmorley:
Would it be possible to release a new version of Pack CLI that includes this fix?

PR #2266, opened 2024-09-25
Allow trusted builders by repo or tag

modulo11:
## Summary

With this change, the tag part of a trusted builder is ignored. Trusting a builder `my/registry/builder` at the moment does not also trust `my/registry/builder:latest`, which should be the same image. Following up on a Slack [discussion](https://cloud-native.slack.com/archives/C0331B61A1Y/p1726214408262059), it might be more intuitive for users to trust an image regardless of any tag.

I took the opportunity to extract trusted/known builder handling into a common location.

## Documentation

    - [ ] Yes, see #___
    - [x] No

### DIFF
--- a/internal/builder/trusted_builder.go
+++ b/internal/builder/trusted_builder.go
@@ -1,5 +1,11 @@
 package builder
 
+import (
+	"github.com/google/go-containerregistry/pkg/name"
+
+	"github.com/buildpacks/pack/internal/config"
+)
+
 type KnownBuilder struct {
 	Vendor             string
 	Image              string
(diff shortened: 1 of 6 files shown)


natalieparellano (commented):
Thank you for the contribution @modulo11 - the implementation looks solid.

I do want to circulate this a bit more to ensure that folks are on board with it - I've added it to the agenda for the next working group. There are two things to note:
* We'll auto trust deprecated images (e.g., heroku/builder:22) in this case, but we do that already (because heroku/builder:22 is on the trusted list though it's not suggested)
* We could end up auto-trusting "dev" images, which worries me a little. Folks might not be placing the same care into the creation of those images compared to ones that get promoted to production. But that seems more of a risk for bugs vs malicious content.

modulo11:
@natalieparellano any updates from the working group meeting?

natalieparellano:
> @natalieparellano any updates from the working group meeting?

I missed it 🙀 - @jabrown85 do you recall if this was discussed?

natalieparellano:
Awesome - thank you @jabrown85! @modulo11 would you be able to update the implementation to reflect that?

modulo11:
I updated the PR according to the discussion. The current behavior (also seen in the tests) is:

- Trusting `my/trusted/builder-jammy` will trust any tag, e.g., `my/trusted/builder-jammy:1.2.3` and `my/trusted/builder-jammy:latest`
- Trusting `my/trusted/builder-jammy:1.2.3` will only trust this specific tag
- Trusting `my/trusted/builder-jammy:latest` will only trust the `latest` tag, so `my/trusted/builder-jammy:1.2.3` will be untrusted

jjbustamante:
@modulo11  Could you update it to reflect the latest changes from the main branch?

PR #2273, opened 2024-10-28
Bump deps

natalieparellano:
## Summary

* Updates go
* Supersedes:
  * https://github.com/buildpacks/pack/pull/2262
  * https://github.com/buildpacks/pack/pull/2233
  * https://github.com/buildpacks/pack/pull/2271
  * https://github.com/buildpacks/pack/pull/2094

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


natalieparellano:
The TOML bump is biting us, taking a look: https://github.com/buildpacks/pack/actions/runs/11561373034/job/32180217662?pr=2273#step:11:8473

natalieparellano:
This should be blocked on the next lifecycle minor (ETA: first week of November)

.github/workflows/benchmark.yml
```diff
@@ -17,7 +17,8 @@ jobs:
       - name: Set up go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.22"
+          check-latest: true
```
jjbustamante, line 20: If we always pull the latest go version, I think we could have some similar that we had with the github actions, where our [Dockerfile](https://github.com/buildpacks/pack/pull/2273/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557) could stay behind right?

natalieparellano:
> Do we want to update to lifecycle 0.20.3?

Ehh, sure - strictly speaking, it's kinda orthogonal to this PR which is just library changes. But I don't see why pack shouldn't default to the latest binary so let's add it.

jjbustamante (approved):
Awesome @natalieparellano

PR #2282, opened 2024-11-13
Cleanup after running acceptance tests

natalieparellano:
While we try to cleanup after ourselves, it doesn't always work.
This will save us from having to RDP into the LCOW worker every so often


jjbustamante (approved):
Awesome!!!

PR #2287, opened 2024-11-18
Sign buildpacksio/pack images using `cosign` and attach `sbom` generated from binaries

### LINKED ISSUE #2193: buildpacksio/pack images should be signed
### Description

See here: https://github.com/buildpacks/pack/issues/934#issuecomment-2070153200

The lifecycle does this [here](https://github.com/buildpacks/lifecycle/blob/7b5a8ec1cf69ddcbc2ed544dbefc98fdd8af49e4/.github/workflows/build.yml#L245-L252) and [here](https://github.com/buildpacks/lifecycle/blob/7b5a8ec1cf69ddcbc2ed544dbefc98fdd8af49e4/.github/workflows/post-release.yml#L33-L53) and [here](https://github.com/buildpacks/lifecycle/blob/7b5a8ec1cf69ddcbc2ed544dbefc98fdd8af49e4/.github/workflows/post-release.yml#L79-L96)

This should be relatively easy to add for pack.

adwait-godbole:
Fixes #2193 

Generate JSON SBOM for release binaries just like how lifecycle is doing right now and use `cosign` for signing, attaching SBOM and verifying container images.

### DIFF
--- a/.github/workflows/delivery-docker.yml
+++ b/.github/workflows/delivery-docker.yml
@@ -19,7 +19,7 @@ env:
   REGISTRY_NAME: 'docker.io'
   USER_NAME: 'buildpacksio'
   IMG_NAME: 'pack'
-  ARCHITECTURES: ["arm64", "amd64", "s390x", "ppc64le"]
+  ARCHITECTURES: arm64,amd64,s390x,ppc64le
 
 jobs:
   deliver-docker:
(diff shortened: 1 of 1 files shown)


adwait-godbole:
Hi @jjbustamante , @natalieparellano, before you go ahead with reviewing this PR, I had a query to clear off my head. Since we are doing a multi-platform image build using `buildx`, should the `cosign sign` and `cosign attest (for sbom)` be done individually for every platform arch image manifest digest or the entire multi-platform index digest as a whole?

natalieparellano:
@adwait-godbole thank you for this. I'd recommend we do this similarly to how we do it in the lifecycle, you can find the workflow file here: https://github.com/buildpacks/lifecycle/blob/main/.github/workflows/build.yml

The sbom that we attach to the release on GitHub is from linux-amd64. But the sbom that we embed within the lifecycle image is for the particular os/arch that we're shipping (see [here](https://github.com/buildpacks/lifecycle/blob/a058e93f6011d11a617cfddb374b33d6af7bff07/Makefile#L300)). But, since we don't have the concept of sbom-within-the-image for pack, we can probably skip over this.

Issue #2104, opened 2024-03-20
Remove validation of deprecated `io.buildpacks.stack.id`


matthewrobertson:
### Summary

Pack still has logic that validates the builder and run image have a matching value set for the `io.buildpacks.stack.id` [here](https://github.com/buildpacks/pack/blob/95b5aef857d6f878c1ec9cdaedd445f9ef782b36/pkg/client/build.go#L819). This label is deprecated as of platform v0.12 so we should probably weaken this validation.

---

### Reproduction

Run a pack build using a packeto builder and google run image:

```
pack build stack-validation \
  --builder paketobuildpacks/builder-jammy-base
  --run-image gcr.io/buildpacks
```

##### Current behavior
Build fails with an error message about mistmatched stack IDs

> run-image stack "google.22" does not match builder stackID...

##### Expected behavior

The builder and run image are both derived from Ubuntu Jammy so we should allow the build proceed

---

### Environment
Linux
##### pack info
All versions of pack impacted

##### docker info
N/A

natalieparellano:
This might benefit from some synchronous discussion so I've added it as a "future topic" for one of our Working Groups

natalieparellano:
Discussed in 3/28 Working Group: we think we can replace the error with a warning. The warning should probably note that stacks are deprecated so that folks who are not using stacks can simply ignore it.

PR #2119, opened 2024-04-07
Removing validation of deprecated io.buildpacks.stack.id

pglushko:
## Summary
Replaces the error with a warning.

## Output
Warning: No schema version declared in project.toml, defaulting to schema version 0.1
latest: Pulling from paketobuildpacks/builder-jammy-base
Digest: sha256:9bb4d01f8d42ccb9aa3b6b10a5876c8913e3ba9709882f18c9bd8c54f46d862b
Status: Image is up to date for paketobuildpacks/builder-jammy-base:latest
latest: Pulling from buildpacks/builder/php
Digest: sha256:91065e07a53b1c048fcaf8e49936a9d19c388b2dfdf2e7e589885aab2f5aae79
Status: Image is up to date for gcr.io/buildpacks/builder/php:latest
Warning: deprecated usage of stack

#### Before
Error message that interrupted build process
#### After
Warning message, build continues

## Documentation
https://buildpacks.io/docs/for-app-developers/concepts/base-images/stack/

- Should this change be documented?
    - [ ] Yes, see #___
    - [ x] No, already documented in  [docs](https://buildpacks.io/docs/for-app-developers/concepts/base-images/stack/) 

## Related
#2104

Resolves #2104

PR #2295, opened 2024-11-27
Removing validation of deprecated io.buildpacks.stack.id

joeybrown:
## Summary
This PR picks up where https://github.com/buildpacks/pack/pull/2119 left off.

#### Before
Screenshot 2024-11-27 at 4 17 14 PM

#### After
Screenshot 2024-11-27 at 4 26 28 PM

## Related

Resolves https://github.com/buildpacks/pack/issues/2104

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -2185,7 +2185,24 @@ func testAcceptance(
 								imageManager.CleanupImages(runImageName)
 							})
 
+							it("fails with a message", func() {
+								h.SkipIf(t, pack.SupportsFeature(invoke.StackWarning), "stack is validated in prior versions")
+								output, err := pack.Run(
+									"build", repoName,
+									"-p", filepath.Join("testdata", "mock_app"),
+									"--run-image", runImageName,
+								)
+								assert.NotNil(err)
+
+								assertOutput := assertions.NewOutputAssertionManager(t, output)
+								assertOutput.ReportsRunImageStackNotMatchingBuilder(
+									"other.stack.id",
+									"pack.test.stack",
+								)
+							})
+
 							it("succeeds with a warning", func() {
+								h.SkipIf(t, !pack.SupportsFeature(invoke.StackWarning), "stack is no longer validated")
 								output, err := pack.Run(
 									"build", repoName,
 									"-p", filepath.Join("testdata", "mock_app"),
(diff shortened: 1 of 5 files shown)


acceptance/acceptance_test.go
```diff
@@ -2181,19 +2185,16 @@ func testAcceptance(
 								imageManager.CleanupImages(runImageName)
 							})
 
-							it("fails with a message", func() {
+							it("succeeds with a warning", func() {
```
jjbustamante, line 2188: @joeybrown 

This test is failing when running the acceptance tests because in the previous `pack` release we were expecting the error message.

```bash
        --- FAIL: TestAcceptance/acceptance_suite/p_previous_cb_current_lc_current/stack_is_created/builder_is_created/build/builder_is_trusted_(and_set_as_default)/--run-image/the_run_image_has_the_wrong_stack_ID/succeeds_with_a_warning (3.02s)
```

Usually, when a new behavior,  we add a [Support](https://github.com/buildpacks/pack/blob/main/acceptance/invoke/pack.go#L200-L225) flag and use [SkipIf](https://github.com/buildpacks/pack/blob/main/acceptance/acceptance_test.go#L1099-L1101) with previous pack versions to avoid running that test.

In this case, maybe we can do the following:
- Add a new Support flag, valid from versions greater or equal to `0.37.0`
- Keep the current test, skip if the version is greater or equal to `0.37.0`
- Add the new test and skip it if the version is less than 0.37.0

joeybrown: sounds great can do!

joeybrown-sf:
@jjbustamante this is now ready for your review

pkg/client/build_test.go
```diff
@@ -531,14 +532,15 @@ func testBuild(t *testing.T, when spec.G, it spec.S) {
 					h.AssertNil(t, fakeRunImage.SetLabel("io.buildpacks.stack.id", "other.stack"))
 				})
 
-				it("errors", func() {
-					h.AssertError(t, subject.Build(context.TODO(), BuildOptions{
+				it("warning", func() {
+					subject.version = "0.37.0"
```
jjbustamante, line 536: @joeybrown do you need this one? I think we didn't do anything related to the version in `build.go`

jjbustamante (approved):
LGTM
Thanks @joeybrown !!!

PR #2385, opened 2025-05-11
Do not select target for daemon only from os if arch is specified

### LINKED ISSUE #2384: `pack builder create` does not accurately build for targets when not publishing
### Summary

I expect that the current behavior of `pack builder create` when `publish` is false is: if targets are specified in `builder.toml`, find the target matching with the daemon. It does not do this.

It will instead find the first target in the list that matches the os of the docker daemon. 

https://github.com/buildpacks/pack/blob/8f3b1d1e6ba7b5616b7cab61602a5ace75bcd5b0/pkg/client/package_buildpack.go#L284-L290

This makes it very difficult to make a builder that produces an image for both `linux/arm64` and `linux/amd64` when not publishing.
I think it should just not select based on only os if arch is specified.

---

**For context:**

I want to build a single builder image and push to multiple repositories. Because  pack does not allow publishing to multiple repositories, and pack does not allow building multi-arch images when not publishing, I am building the image on two separate machines (arm64 and amd64) and then merging the images with ` docker manifest create`.
But because of this issue, both machines are producing `amd64`images.

I think there is a larger issue with the usage of `pack builder create` when not publishing. Why cant it produce multi-arch packages when not publishing? Why does the arch have to match the daemon?

---

##### pack info

```
Pack:
  Version:  0.36.1
  OS/Arch:  linux/amd64

Default Lifecycle Version:  0.20.3

```

##### Steps
1. Create a builder that has both a `linux/amd64` and `linux/arm64` target
```toml
[[targets]]
os = "linux"
arch = "amd64"

[[targets]]
os = "linux"
arch = "arm64"
```

2) run `pack builder create dummy --config builder.toml` on a `arm64` machine
3) run `pack builder create dummy --config builder.toml` on a `amd64` machine

##### Current behavior
Output image on both hosts will be the same. The resulting arch will be whatever target is first on the list.
In the example above, both machines produce a `amd64` image.

##### Expected behavior
The machine should produce the image of its architecture.

pigeonhands:
## Summary
PR: #2384

Currently when not publishing, pack will find the first target that matches the os of the docker daemon, rather than the os **and** the arch. This makes it so that is not the case.

for old behavior, a target with only os can be added as the last item in the targets array.

```toml
[[targets]]
os = "linux"
arch = "amd64"

[[targets]]
os = "linux"
arch = "arm64"

[[targets]]
os = "linux"
```

## Documentation

    - [x] Yes, see #2384
    - [ ] No

## Related

Resolves #2384

### DIFF
--- a/pkg/client/package_buildpack.go
+++ b/pkg/client/package_buildpack.go
@@ -281,10 +281,11 @@ func (c *Client) daemonTarget(ctx context.Context, targets []dist.Target) (dist.
 	if err != nil {
 		return dist.Target{}, err
 	}
+
 	for _, t := range targets {
 		if t.Arch != "" && t.OS == info.Os && t.Arch == info.Arch {
 			return t, nil
-		} else if t.OS == info.Os {
+		} else if t.Arch == "" && t.OS == info.Os {
 			return t, nil
 		}
 	}


jjbustamante:
@pigeonhands Could yoi fix the DCO signing your contribution?

jjbustamante (approved):
### Issue Analysis

The current daemonTarget function in pack has a flawed target selection logic that prevents proper multi-architecture builds when not publishing. The issue occurs in
  /pkg/client/package_buildpack.go:279-292.

  Current Problematic Behavior

```go
  if t.Arch != "" && t.OS == info.Os && t.Arch == info.Arch {
      return t, nil
  } else if t.OS == info.Os {
      return t, nil  // This is the problem - it returns first OS match ignoring arch
  }
```
## This causes:
  - On an ARM64 machine, if linux/amd64 is listed first in targets, it will be selected
  - Users cannot build architecture-specific images on their respective machines

## RFC-0128 Alignment

  ✅ The proposed fix aligns with RFC-0128 because:
  - It respects the multi-arch principles by matching the daemon's actual architecture
  - It maintains the expected behavior for local builds using Docker's OSType
  - It doesn't break the --target flag functionality

PR #2386, opened 2025-05-21
  Fix default userns=host behavior to be opt-in

### LINKED ISSUE #2369: Add option to turn off userns=host configuration.
### Description
We are using no-privilege-root plugin to protect our Docker configuration, which doesn't allow to use userns as host. Due to this, our customer can only use Pack CLI version, which is older than 0.35.0.

### Proposed solution
A switch or flag could be added to the "pack" command, allowing the user to decide, which mode to be used.

### Additional context

jjbustamante:
## Summary

The flag `--userns-host` is now required to enable the user namespace isolation for build containers. By default, this is disabled, which resolves the incompatibility with Docker plugins that don't support "userns as host" functionality.

This reverts the default behavior from pack 0.35.0 where user namespace isolation was enabled by default, while still allowing users who want this feature to opt in.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #2369

### DIFF
--- a/internal/build/fakes/enable_userns_host.go
+++ b/internal/build/fakes/enable_userns_host.go
@@ -1,0 +1,1 @@
+package fakes
--- a/internal/build/fakes/enable_userns_host.go
+++ b/internal/build/fakes/enable_userns_host.go
@@ -1,1 +1,0 @@
-package fakes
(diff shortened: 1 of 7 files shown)


jjbustamante:
@jkutner could you give me a review here? I got [confirmation](https://github.com/buildpacks/pack/issues/2369#issuecomment-2908639699) that is working as expected

PR #2390, opened 2025-05-31
Make `go build` work without Makefile

abitrolly:
## Summary

This merges `version.go` into `pkg/client/client.go`, which is the only file that uses the it. Then it moves `cmd/pack/main.go` into root. Now `go build` produces executable `pack` instead of bogus `pack` with just `version.go`.

No need to specify `./cmd/pack` while compiling anymore. The version is injected into client directly without extra files.

#### Before

`go build` works, but produces nothing.

#### After

`go build` produces `pack` executable.

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related

Resolves #___


jjbustamante (commented):
LGTM
Thanks for this change @abitrolly

drupol:
Feel free to ignore me, the issue has been fixed, we replaced:

```nix
  subPackages = [ "cmd/pack" ];
```

with

```nix
  subPackages = [ "." ];
```

And we got it compiling again.

PR #2394, opened 2025-06-11
Added the first tag argument

### LINKED ISSUE #2367: Add support for --tag to ease automated release processes
### Description
`pack build` has the nice (multiple, array) `-t`- or `--tag`-argument, where one can specify alternative tags when building the app. This is also necessary for buildpacks, extensions and builders. If one wants to setup pipelines to automate releases with qualifiers like "stable", "latest", aso. it would be very beneficial to handle those in one command.

### Proposed solution
Add (multiple, array) `--tag`-argument to 
  - `pack buildpack package` 
  - `pack extension package` 
  - `pack builder create` 

Packing and pushing again several times.

dgannon991:
## Summary
Adding support for --tag into additional commands.

## Documentation

    - [x] Yes, see buildpacks/docs#834
    - [ ] No

## Related

Resolves #2367

### DIFF
--- a/.github/workflows/delivery-docker.yml
+++ b/.github/workflows/delivery-docker.yml
@@ -61,7 +61,7 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
       - uses: docker/setup-qemu-action@v3
       - uses: docker/setup-buildx-action@v3
-      - uses: buildpacks/github-actions/setup-tools@v5.9.1
+      - uses: buildpacks/github-actions/setup-tools@v5.9.2
       - name: Buildx Build/Publish
         run: |
           docker buildx build . \
--- a/pkg/client/register_buildpack.go
+++ b/pkg/client/register_buildpack.go
@@ -68,6 +68,7 @@ func (c *Client) RegisterBuildpack(ctx context.Context, opts RegisterBuildpackOp
 		params := url.Values{}
 		params.Add("title", issue.Title)
 		params.Add("body", issue.Body)
+		params.Add("template", "add-buildpack.md")
 		issueURL.RawQuery = params.Encode()
 
 		c.logger.Debugf("Open URL in browser: %s", issueURL)


internal/builder/builder.go
```diff
@@ -463,7 +463,7 @@ func (b *Builder) SetValidateMixins(to bool) {
 }
 
 // Save saves the builder
-func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata) error {
+func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata, additionalTags []string) error {
```
dgannon991, line 466: I'm not happy with this additional argument, but when I scanned the codebase I found a mixture of positional arguments like this, structs and also the `opts...` variadic approach, so I wasn't sure which one to pick. I figured the best approach was to pick the simplest implementation and then ask for specific feedback in the code review, so here we are :)

jjbustamante (changes requested):
## 🎯 **Great Feature Implementation, But Needs Design Refinement**

Thank you @dgannon991 for this comprehensive implementation of the `--tag` feature\! The overall approach is excellent, but I've identified a significant design issue that should be addressed before merging.

## ✅ **What's Working Well**

- ✅ Complete feature coverage across all three target commands
- ✅ Consistent API design and flag implementation  
- ✅ Comprehensive test coverage
- ✅ Clean data flow through the architecture
- ✅ Proper integration with underlying image libraries

## 🚨 **Breaking Change Concern**

The current implementation introduces a **breaking change** to the core `Builder.Save` method signature:

**Current Change:**
```go
// Before
func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata) error

// After  
func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata, additionalTags []string) error
```

**Impact Analysis:**
- 🔴 Breaks **every existing caller** of `Builder.Save`
- 🔴 Forces updates to **many test files** (requiring `[]string{}` parameter)
- 🔴 Creates maintenance burden for downstream consumers
- 🔴 Violates Go's compatibility promise for minor versions

**Evidence:**
```bash
# All these test calls now require empty slice parameter
internal/builder/builder_test.go:304: subject.Save(logger, builder.CreatorMetadata{}, []string{})
internal/builder/builder_test.go:312: subject.Save(logger, builder.CreatorMetadata{}, []string{})
internal/builder/builder_test.go:330: subject.Save(logger, builder.CreatorMetadata{}, []string{})
```

## 💡 **Suggested Improvement: Variadic Parameters**

**Replace the current signature with:**
```go
func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata, additionalTags ...string) error {
    return b.image.Save(additionalTags...)
}
```

### **Benefits:**
- ✅ **Zero breaking changes** - all existing calls continue to work unchanged
- ✅ **Clean API** - no need to pass empty slices 
- ✅ **Intuitive usage** - follows Go idioms
- ✅ **Future-proof** - easily extensible

### **Usage Examples:**
```go
// Existing calls continue to work unchanged
bldr.Save(logger, metadata)

// New functionality for additional tags
bldr.Save(logger, metadata, "tag1", "tag2", "tag3")

// From slice (when needed)
bldr.Save(logger, metadata, tags...)
```

### **Required Changes:**

**1. Update `internal/builder/builder.go:466`:**
```go
func (b *Builder) Save(logger logging.Logger, creatorMetadata CreatorMetadata, additionalTags ...string) error {
    // ... existing implementation ...
    return b.image.Save(additionalTags...)
}
```

**2. Update `pkg/client/create_builder.go:140`:**
```go
err = bldr.Save(c.logger, builder.CreatorMetadata{Version: c.version}, opts.AdditionalTags...)
```

**3. Revert all test file changes:**
- Remove `, []string{}` from all test calls
- Tests will automatically pass with the variadic signature

## 🔄 **Implementation Notes**

The same pattern should be applied to:
- `pkg/buildpack/builder.go` `SaveAsImage` method (if it has similar issues)
- Any other methods that were modified to accept additional tags

## 🏆 **Why This Matters**

This approach:
- Maintains backward compatibility for existing pack users
- Follows Go best practices for optional parameters  
- Reduces the diff size significantly
- Makes the feature adoption smoother for the community

## 📋 **Action Items**

1. [x] Refactor `Builder.Save` to use variadic parameters
2. [x] Update client code to use spread operator
3. [x] Revert unnecessary test file changes
4. [ ] Verify all existing tests still pass without modification
5. [ ] Add tests for the new variadic functionality

This is excellent work overall\! The variadic approach will make this feature even better by ensuring seamless adoption. Looking forward to the updated implementation\! 🚀

---
**Note**: This is the only significant issue I found. Once addressed, this implementation will be production-ready.

internal/builder/builder.go
jjbustamante: Hi @dgannon991, I suggest the `opts...` variadic approach for this particular one. Sorry for the extensive feedback, I am playing with Claude Code :smile:  and this PR was ideal to do that!!

dgannon991:
Hi @jjbustamante,
Oh perfect, that was exactly the sort of suggestion I was looking for! Thank you for the very thorough feedback. I'm travelling with work for the next day or two, so realistically I should be able to complete the refactor on either Tuesday or Wednesday evening.
Many thanks,
David

PR #2404, opened 2025-06-21
Fix project.toml directory exclusion behavior

### LINKED ISSUE #2402: Unexpected `project.toml` include/exclude behavior
### Summary
When specifying a [list of files to include or exclude in a build](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksinclude-optional-and-iobuildpacksexclude-optional) in `project.toml`, the behavior is somewhat surprising when specifying directories (e.g. entries ending with a trailing slash). In particular, an excluded directory `foo/` will still be present (albeit empty) during build and in the final image. Conversely, an included, empty `foo/` directory will not be present during build not the final image.

The spec only describes the behavior for *files*, and the current behavior might make sense considering `.gitignore` patterns are used in both cases. However, it seems reasonable to expect that directory entries (ending in a trailing slash) would also `include`/`exclude` the actual directories?

It's possible that this should be handled by the lifecycle instead (if you agree the current behavior is unexpected). Either way the Project Descriptor spec should likely be updated to reflect how directory entries are handled.

---

##### Steps

These steps only cover the `exclude` case.

With an input directory:

```term
$ find .
.
./foo
./foo/bar.baz
./project.toml
```

And a `project.toml`:

```toml
[_]
schema-version = "0.2"

[io.buildpacks]
exclude = [
  "foo/"
]

[[io.buildpacks.group]]
id = "foo/finders"

  [io.buildpacks.group.script]
  api = "0.10"
  inline = "find ."
```

##### Current behavior

Running `pack build foo` returns:

```term
...
===> BUILDING
.
./foo
./project.toml

...
```

(Showing the now empty excluded directory `foo/` is still present during build).

##### Expected behavior

Running `pack build foo` should (probably) return:

```term
...
===> BUILDING
.
./project.toml

...
```

(Excluding both the `foo/*` files as well as the `foo` directory itself)

---

##### pack info

```term
Pack:
  Version:  0.38.1+git-9bd06a9.build-6494
  OS/Arch:  darwin/arm64

Default Lifecycle Version:  0.20.8

...
```

##### docker info

```term
Client:
 Version:    28.2.2
 Context:    desktop-linux
 Debug Mode: false
 ...
 ```

### PR DESCRIPTION
**Pull request by jjbustamante**
## Summary
<\!-- Provide a high-level summary of the change. -->

Fixes unexpected behavior when excluding directories with trailing slash patterns in `project.toml`. Previously, patterns like `"foo/"` would exclude files within the directory but still create empty directory entries in the final image. This change ensures that when users specify `exclude = ["foo/"]`, the directory `foo` and all its contents are completely excluded from the build process.

## Output
<\!-- If applicable, please provide examples of the output changes. -->

#### Before

When using `project.toml` with:
```toml
[io.buildpacks]
exclude = [
  "foo/"
]
```

The build would show:
```
===> BUILDING
.
./foo
./project.toml
```

The empty `foo/` directory was still present during build and in the final image.

#### After

With the same `project.toml` configuration, the build now shows:
```
===> BUILDING
.
./project.toml
```

The `foo/` directory is completely excluded, with no empty directory artifacts.

## Documentation
<\!-- If this change should be documented, please create an issue or PR on https://github.com/buildpacks/docs and link below. -->
<\!-- NOTE: This can be added (by editing the issue) after the PR is opened. -->

    - [ ] No

## Related
<\!-- If this PR addresses an issue, please provide issue number below. -->

Resolves #2402


CLAUDE.md
```diff
@@ -0,0 +1,153 @@
+# CLAUDE.md
```
edmorley, line 1: Hi! Could you move the unrelated changes to a separate PR? :-)

jjbustamante: Yeah.

PR #2432, opened 2025-08-22
fix: update lifecycle API validation for experimental flag solve issue #2414

### LINKED ISSUE #2414: ERROR: builder config contains image extensions; support for image extensions is currently experimental
I am trying to run pack builder create with image extensions in my builder.toml, but i am getting this error: 

`ERROR: builder config contains image extensions; support for image extensions is currently experimental`

My specification setup include - 

buildpack API - 0.11
Platform API - 0.13
lifecycle - 0.20.11

In the release notes of buildpack API and platform API we see that with these versions image extensions is no longer experimental - https://github.com/buildpacks/spec/releases, but i am still seeing this error.

There is no check on API versions so looks like kind of a bug - ```https://github.com/buildpacks/pack/blob/main/internal/commands/builder_create.go#L69-L73``

vky5:
## Summary
This PR refines builder configuration validation for `lifecycle` API versions, aligning behavior with the intended experimental/stable boundary.

#### Before
- Lifecycle APIs marked experimental were **only accepted if `cfg.experimental = true`**.  
- This blanket rule applied to all versions, even the newer lifecycle APIs that are now considered stable.  
- As a result, users were forced to enable experimental mode unnecessarily when using lifecycle APIs >= 0.20.

#### After
- For lifecycle API **>= 0.20**, the version is treated as stable and accepted by default.  
- For lifecycle API **< 0.20**, the builder must explicitly set `experimental = true`, otherwise validation fails.  
- This prevents accidental use of unstable/experimental APIs while removing the unnecessary restriction for modern lifecycle versions.

## Implementation Notes
- Parsed lifecycle API version using semantic versioning.  
- Adjusted validation logic:  
  - **If version >= 0.20 → always allowed**.  
  - **If version < 0.20 → require `cfg.experimental = true`**.  
- Preserves explicit opt-in for older experimental APIs, while simplifying use of newer stable ones.  
- Added tests to verify both acceptance and rejection cases.

## Documentation
    - [ ] Yes, see #___  
    - [x] No  

## Related
Resolves #2414

### DIFF
--- a/pkg/client/create_builder.go
+++ b/pkg/client/create_builder.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/Masterminds/semver"
 	"github.com/buildpacks/imgutil"
-	"github.com/buildpacks/lifecycle/api"
 	"github.com/pkg/errors"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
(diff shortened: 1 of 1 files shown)


jjbustamante (changes requested):
Hi @vky5,

I apologize for my delayed response!

While your approach is not incorrect, I believe we should implement it in a different location. To provide some context, the experimental validation was added a while ago as a quick gatekeeper.

Here’s my suggestion:

1. Remove the entire code block from the `builder_create.go` and `create_builder.go` commands:
    ```go
    if hasExtensions(builderConfig) {
        if !cfg.Experimental {
            return errors.New("builder config contains image extensions; support for image extensions is currently experimental")
        }
    }
    ```

2. Navigate to the `client/create_builder.go` file [here](https://github.com/buildpacks/pack/blob/main/pkg/client/create_builder.go#L270). Once you have the fetched lifecycle, you can do the following:
    - Move the `hasExtensions` function into `client/create_builder.go`.
    - Create a function called `validateLifecycleVersion` that checks the Lifecycle Platform API version:
        - Validate if the builder has extensions AND if the lifecycle Platform API version is less than 0.13.
        - Check if the client experimental flag is not set; if that’s the case, throw an error.

3. Update any broken tests or add new ones as necessary.

**Tips:**
- The client has the experimental attribute initialized upon creation.
- The Lifecycle Version object has a method called `LessThan("")` that allows you to avoid creating the parsing logic that you added to the pull request.
- The overall idea is similar to what you proposed but should be done in a different location.

Please let me know if you have some time to make these changes!

PR #2471, opened 2025-11-19
bump Docker API version to 1.44

### LINKED ISSUE #2464: Docker API Version 1.44
### Description
Docker Engine latest version requires a minimum API version of 1.44. Several CI/CD pipelines seem to be affected in the last week. pack hardcodes using the 1.38 API version and it seems we cannot override it which makes it incompatible with updated CI/CD environments. I am not an expert here so I may be wrong.

### Proposed solution
Update client DockerAPIVersion constant to 1.44 together with several tests that hardcode 1.38 or a mechanism to override.

Attempted to use a DOCKER_API_VERSION environment variable hoping it would override the default but it did not work.

### Additional context

Image

hone:
## Summary
This bumps the Docker API version from `1.38` to `1.44`. This API version was introduced in Docker Engine is now `25`.

## Output
None

## Documentation
Do we document the Docker API version anywhere?

    - [ ] Yes
    - [ ] No

## Related

Resolves #2464

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -262,7 +262,7 @@ func WithKeychain(keychain authn.Keychain) Option {
 	}
 }
 
-const DockerAPIVersion = "1.38"
+const DockerAPIVersion = "1.44"
 
 // NewClient allocates and returns a Client configured with the specified options.
 func NewClient(opts ...Option) (*Client, error) {
--- a/benchmarks/build_test.go
+++ b/benchmarks/build_test.go
@@ -32,7 +32,7 @@ var (
 
 func BenchmarkBuild(b *testing.B) {
 	setEnv()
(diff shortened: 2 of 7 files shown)


pkg/client/client.go
```diff
@@ -262,7 +262,7 @@ func WithKeychain(keychain authn.Keychain) Option {
 	}
 }
 
-const DockerAPIVersion = "1.38"
+const DockerAPIVersion = "1.44"
```
natalieparellano: Lifecycle uses `client.WithAPIVersionNegotiation()` - maybe that's safer? https://github.com/buildpacks/lifecycle/blob/73739826a85bc909937c87f0feb3c23e3667c9b5/priv/docker.go#L27

jjbustamante: Natalie! Yeah, we did that [here](https://github.com/buildpacks/pack/pull/2474)

PR #2474, opened 2025-11-21
Use Docker API version negotiation instead of hardcoded version

### LINKED ISSUE #2464: Docker API Version 1.44

jjbustamante:
## Summary
Replaces all hardcoded Docker API version 1.38 references with `WithAPIVersionNegotiation()`, allowing pack to automatically adapt to the Docker daemon's supported API version.

Fixes #2464

## Context
This is an alternative approach to PR #2471 that provides a more future-proof solution. Instead of bumping the hardcoded API version from 1.38 to 1.44, this PR uses Docker's built-in API version negotiation mechanism as suggested in https://github.com/buildpacks/pack/pull/2471#issuecomment-3553577419.

## Technical Details

[`WithAPIVersionNegotiation()`](https://pkg.go.dev/github.com/docker/docker/client#WithAPIVersionNegotiation) works by:
1. Client queries daemon for its supported API version
2. Automatically uses the highest mutually supported version
3. Falls back gracefully to older versions when needed
4. Eliminates maintenance burden of hardcoded version constants

## Notes

The `DockerAPIVersion` constant in `pkg/client/client.go` has been kept for backward compatibility (it's exported), though it's no longer used internally.
## Output
None

## Documentation
Do we document the Docker API version anywhere?

    - [ ] Yes
    - [ ] No

## Related

Resolves #2464


baobaab:
@jjbustamante I tested your branch with `builder create` and `build` commands on my projects, and it works for me. Now waiting for the code to be reviewed :)

hone (approved):

PR #2489, opened 2025-11-25
fix: semgrep-missing-unlock-before-return

rahulkhinchi-wq:
Pull Request — Semgrep Rule Fix
Rule ID: missing-unlock-before-return
Rule Message: Missing mutex unlock (c.outputMux variable) before returning from a function. This could result in panics resulting from double lock operations
File Path: /tools/scanResult/unzipped-1969552506/acceptance/invoke/pack.go
Line: 163

## Summary
This pull request resolves a concurrency issue detected by Semgrep, where a mutex (`outputMux`) was not unlocked before an early return. Failing to unlock a mutex can lead to deadlocks and unpredictable behavior.

### Modified File
`acceptance/invoke/pack.go`

### Update
- Added a missing `c.outputMux.Unlock()` immediately before the early `return` statement to ensure the mutex is always released.

## Rationale
The Semgrep check identified that the function could return without unlocking the mutex, which can result in blocked goroutines and test instability. This PR ensures proper lock management and prevents potential deadlocks.

### DIFF
--- a/pkg/image/fetcher.go
+++ b/pkg/image/fetcher.go
@@ -257,6 +257,17 @@ func (f *Fetcher) FetchForPlatform(ctx context.Context, name string, options Fet
 		return nil, err
 	}
 
+	platformStr := options.Target.ValuesAsPlatform()
+
+	// When PullPolicy is PullNever, skip platform-specific digest resolution as it requires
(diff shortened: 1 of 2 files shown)


rahulkhinchi-wq:
Hi @jjbustamante, can you please look into it and let me know if any changes are required

PR #2513, opened 2026-01-20
feat: add support to use issue template from registry index for pack yank

### LINKED ISSUE #2461: pack buildpack yank command fails
### Summary

##### Steps

It generates:

```

id = "foo-ns/foo-bp"
version = "1.2.3"

```

It should generate...

````
```
id = "foo-ns/foo-bp"
version = "1.2.3"
yank = true
```
````

### PR DESCRIPTION
**Pull request by burnerlee**
## Summary
This PR adds support to read the issue template from [registry-index](https://github.com/buildpacks/registry-index/blob/main/.github/ISSUE_TEMPLATE/yank-buildpack.md) while running `pack buildpack yank ...`.

In case the template doesn't exist it falls back to the harcoded template. The hardcoded template has been updated to include `yank=true` when `.yank` is true.

#### Before
Issue Body:
```
id = "foo-ns/foo-bp"
version = "1.2.3"
```

#### After
Issue Body:
````
```
id = "foo-ns/foo-bp"
version = "1.2.3"
yank = true
```
````

## Documentation

    - [ ] Yes, see #___
    - [ ] No

## Related
resolves #2461

### DIFF
--- a/internal/registry/github.go
+++ b/internal/registry/github.go
@@ -4,7 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"net/url"
+	"os"
 	"os/exec"
+	"path/filepath"
+	"regexp"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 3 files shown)


burnerlee:
@dmikusa @jjbustamante could you PTAL?

jjbustamante:
Thanks for kicking this off, @burnerlee. Heads up that the original bug from #2461 — the missing `yank = true` and code fence — is now fixed and merged in #2631, so that part is covered.

What's left here is the more interesting idea: reading the issue template from registry-index at runtime instead of hardcoding it. I'd like to keep this open and think it through — I'm not sold either way yet. A few things on my mind:

- It refreshes the registry cache on every `yank` just to read one template file — worth weighing that cost against the benefit.
- The frontmatter/placeholder parsing is hand-rolled and only handles a couple of placeholders, so it could break quietly if the upstream template changes.
- Sending both a `template=` param and a prefilled `body` may conflict on GitHub's side.

No action needed right now — leaving this open while I think about whether the runtime-template approach is worth the added coupling.

PR #2518, opened 2026-01-27
Ship FreeBSD binaries

gogolok:
## Summary

Since `pack` is usable on FreeBSD it would be nice to build a FreeBSD binary right away in the future. 

## Output

I tested this by removing some other files from the build.yaml to gain speed in testing, but the following URLs provide an idea it's working:
https://github.com/gogolok/buildpacks-pack/releases/tag/v0.40.0-rc2
https://github.com/gogolok/buildpacks-pack/actions/runs/21378904452

I've tested it quickly on FreeBSD arm64 VM without sudo (which can be installed of course) to test the tar command and whether it's a valid binary.

```shell
root@server:/home/gogo # (curl -sSL "https://github.com/gogolok/buildpacks-pack/releases/download/v0.40.0-rc2/pack-v0.40.0-rc2-freebsd-arm64.tgz" | tar -C /usr/local/bin/ --no-same-owner -xzv pack )
x pack
root@server:/home/gogo # pack --version
0.40.0-rc2+git-985ac2f.build-7
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,6 +135,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - name: freebsd
+            goarch: amd64
+            goos: freebsd
+          - name: freebsd-arm64
+            goarch: arm64
+            goos: freebsd
           - name: linux-arm64
             goarch: arm64
             goos: linux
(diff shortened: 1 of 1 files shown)


jjbustamante (approved):
Thank you @gogolok

PR #2571, opened 2026-04-15
Fix `builder inspect` showing known builders as untrusted

runesoerensen:
## Summary

`builder_inspect.go` was missing the `IsKnownTrustedBuilder` check that other commands already include, causing known trusted builders to always show as untrusted.

- Add `IsKnownTrustedBuilder` check in `builder_inspect.go`, matching the pattern in `build.go` and `config_trusted_builder.go`

#### Before
```
$ pack builder inspect heroku/builder:24
...
Trusted: No
```

#### After
```
$ pack builder inspect heroku/builder:24
...
Trusted: Yes
```

## Documentation
- Should this change be documented?
    - [ ] No

## Related
Resolves #2570

PR #2573, opened 2026-04-15
Consolidate trust logic into `IsTrustedBuilder`

### LINKED ISSUE #2572: Trust checks are fragmented and known builders without a tag don't match tagged queries
### Summary

`IsTrustedBuilder` only checks user-configured trusted builders (`cfg.TrustedBuilders`), so callers that need to also recognize known trusted builders must separately call `IsKnownTrustedBuilder()`. This split makes it easy for trust checks to diverge across commands (see #2570).

Additionally, `IsKnownTrustedBuilder` uses exact string matching, so known builders listed without a tag (e.g. `paketobuildpacks/builder-jammy-base`) don't match when a specific tag is used (e.g. `paketobuildpacks/builder-jammy-base:latest`). `IsTrustedBuilder` already supports matching by repository for user-configured builders (via `name.ParseReference`), but this is not applied to known builders.

Some builder repositories (e.g. `heroku/builder`) don't publish a `latest` tag, so builders must be specified with a tag (e.g. `pack build --builder heroku/builder:24`). With exact string matching, a tagless `heroku/builder` entry in the known builders list wouldn't match these tagged references.

---

##### Steps

1. `pack builder inspect paketobuildpacks/builder-jammy-base:latest`

##### Current behavior

Output shows `Trusted: No`, because `IsKnownTrustedBuilder` performs an exact string match and `paketobuildpacks/builder-jammy-base:latest` does not equal `paketobuildpacks/builder-jammy-base`.

##### Expected behavior

A known trusted builder listed without a tag should match any tag in that repository, consistent with the behavior for user-configured trusted builders.

##### Proposed solution

Make `IsTrustedBuilder` check both known and user-configured trusted builders using `name.ParseReference` for consistent reference matching: entries without a tag match any tag in the repository; entries with a tag require an exact match. This removes the need for callers to separately check `IsKnownTrustedBuilder()`.

Re: [#2266 (comment)](https://github.com/buildpacks/pack/pull/2266#issuecomment-2438351422) - a concern was raised that trusting by repository could implicitly trust unstable/RC tags (e.g. `builder:24-rc3.dev`). I verified via `crane ls` that all tagless known builders (the Paketo `paketobuildpacks/*` repositories) only publish semver version tags and `latest` - no RC, dev, alpha, beta, or pre-release tags exist in any of these repositories.

---

### Environment

Presumably all versions since PR #2266 was merged (which introduced this functionality.

runesoerensen:
## Summary

Consolidate all trust checking into `IsTrustedBuilder` so callers don't need to separately check `IsKnownTrustedBuilder()`, and use `name.ParseReference` for consistent reference matching across both known and user-configured builders.

- `IsTrustedBuilder` now checks both known trusted builders and user-configured trusted builders
- Uses `name.ParseReference` for consistent reference matching across both sources - entries without a tag match any tag in the repository, entries with a tag require an exact match
- Remove now-unnecessary `|| IsKnownTrustedBuilder()` calls from `build.go`, `config_trusted_builder.go`, and `builder_inspect.go`

#### Before
```
$ pack builder inspect paketobuildpacks/builder-jammy-base:latest
...
Trusted: No
```

#### After
```
$ pack builder inspect paketobuildpacks/builder-jammy-base:latest
...
Trusted: Yes
```

## Documentation
    - [ ] No

## Related
Resolves #2572


runesoerensen:
@jjbustamante thanks for reviewing/merging #2571! This branch was based on that work, so I've rebased and marked the PR ready for review :)

internal/builder/trusted_builder_test.go
```diff
@@ -30,6 +30,23 @@ func trustedBuilder(t *testing.T, when spec.G, it spec.S) {
...
 
 	when("IsTrustedBuilder", func() {
+		it("trusts known trusted builders", func() {
+			// Known builder with exact tag match
+			isTrusted, err := bldr.IsTrustedBuilder(config.Config{}, "heroku/builder:24")
+			h.AssertNil(t, err)
+			h.AssertTrue(t, isTrusted)
+
+			// Known builder without tag should match any tag
+			isTrusted, err = bldr.IsTrustedBuilder(config.Config{}, "paketobuildpacks/builder-jammy-base:latest")
```
jjbustamante, line 47: Should this test case reference: `paketobuildpacks/builder-jammy-base`?

runesoerensen: The `:latest` here is intentional - this case exercises the _"known entry has no tag, match any tag in the same repository"_ branch of the new `name.ParseReference` logic, which is what #2572 is about and is also the example shown in the Before/After in the PR description.

The `heroku/builder:24` case above covers the exact-tag branch. Switching to a bare `paketobuildpacks/builder-jammy-base` would collapse this into a bare-vs-bare repository match and stop exercising the "any tag" path.

Happy to expand the inline comment if it'd read more clearly, or add a second assertion with a different tag (e.g. `:anything`) to make the "any tag" intent more obvious - let me know what you'd prefer (or if I'm missing something obvious :)).

TIR44:
I accidentally opened duplicate PR #2619 after auditing #2572, then noticed this PR already covers the core fix. I closed #2619 in favor of this one.

While doing that audit, I found one extra test angle that may be useful here: caller coverage in addition to `internal/builder` unit tests. The branch in #2619 added checks that:

- `build` treats `paketobuildpacks/builder-jammy-base:latest` as trusted;
- `builder inspect` reports the same tagged known builder as trusted;
- `config trusted-builders add paketobuildpacks/builder-jammy-base:latest` is a no-op because it is already trusted through the tagless known entry.

The reason I think caller coverage is useful is that `pkg/client/build.go` defaults `opts.TrustBuilder` to `builder.IsKnownTrustedBuilder`, while command code used to compose `IsTrustedBuilder(...) || IsKnownTrustedBuilder(...)`. So the regression is not just the helper result; it is also keeping the command and client paths aligned.

runesoerensen:
Thanks @TIR44 - good catch on the SDK default.

You're right that `pkg/client/build.go:453` still pointed at `IsKnownTrustedBuilder`, which kept exact-string matching. The CLI worked because `internal/commands/build.go` overrides `TrustBuilder` with a closure built from `IsTrustedBuilder`, but a programmatic caller passing `paketobuildpacks/builder-jammy-base:latest` without setting `TrustBuilder` would still hit the original regression.

Pushed two follow-up commits:

- 6bb75db5: `IsKnownTrustedBuilder` now uses the same `name.ParseReference` matching as `IsTrustedBuilder`, sharing a `referencesMatch` helper. Tagless known entries match any tag in the repo; tagged entries still require an exact tag match. This also closes the same hole in the "can't untrust a known builder" guard in `removeTrustedBuilder`.
- 3dc1da2d: Caller-level tests for `build`, `builder inspect`, and `config trusted-builders add` against `paketobuildpacks/builder-jammy-base:latest`.

Could you take a look and let me know if this covers what you had in mind?

edmorley:
@jjbustamante Please could we have a re-review? :-)

edmorley:
@jjbustamante @jabrown85 Could we have a re-review here? This blocks a stack of PRs which leads to us being able to trust `heroku/builder` (without a tag), which saves us from having to manually add every new Heroku stack to the trusted list.

PR #2575, opened 2026-04-15
Separate `TrustedBuilders` from `SuggestedBuilders`

### LINKED ISSUE #2574: Separate trusted builders from suggested builders
### Description

The `KnownBuilder` struct combines trust decisions with display metadata (`Trusted`, `Suggested`, `Vendor`, `DefaultDescription`) in a single type. Separating these concerns would make trust logic easier to reason about and reduce the surface area for issues like those described in #2570 and #2572.

### Proposed solution

Split `KnownBuilder` / `KnownBuilders` into two separate concepts:
- **`TrustedBuilders`** - a flat `[]string` of image names used solely for trust decisions
- **`SuggestedBuilder` / `SuggestedBuilders`** - a struct with `Vendor`, `Image`, and `DefaultDescription` fields used for display in `builder suggest`

We should also add an invariant test ensuring every suggested builder is also a trusted builder, to prevent future omissions.

### Additional context

Since `KnownBuilder` is in an `internal/` package, there shouldn't be any external consumers to worry about.

runesoerensen:
## Summary

_Note: This draft PR builds on unmerged changes introduced in #2573 - see https://github.com/runesoerensen/pack/compare/runesoerensen/consolidate-trusted-builder-logic...runesoerensen:pack:runesoerensen/separate-trusted-suggested-builders to inspect only the commits introduced since that branch._

Separate trust decisions from display metadata by splitting `KnownBuilder` into dedicated data structures for each concern.

- Replace `KnownBuilder` struct with separate `TrustedBuilders` (flat `[]string` for trust decisions) and `SuggestedBuilder` / `SuggestedBuilders` (display metadata for `builder suggest`)
- Add invariant test ensuring every suggested builder is also a trusted builder
- Remove `Trusted` and `Suggested` boolean fields that previously required filtering

## Output

N/A - no user-facing output changes.

## Documentation

    - [x] No

## Related
Resolves #2574

### DIFF
--- a/internal/builder/trusted_builder.go
+++ b/internal/builder/trusted_builder.go
@@ -1,113 +1,92 @@
 package builder
 
 import (
+	"slices"
+
 	"github.com/google/go-containerregistry/pkg/name"
 
 	"github.com/buildpacks/pack/internal/config"
 )
 
-type KnownBuilder struct {
+// TrustedBuilders is a flat list of builder image names that are trusted by default.
+var TrustedBuilders = []string{
+	"gcr.io/buildpacks/builder:google-22",
+	"heroku/builder:24",
+	"heroku/builder:22",
+	"heroku/builder:20",
+	"paketobuildpacks/builder-jammy-base",
+	"paketobuildpacks/builder-jammy-full",
+	"paketobuildpacks/builder-jammy-tiny",
+	"paketobuildpacks/builder-jammy-buildpackless-static",
+	"paketobuildpacks/ubuntu-noble-builder",
+	"paketobuildpacks/builder-ubi8-base",
+	"paketobuildpacks/ubi-9-builder",
+	"paketobuildpacks/ubi-10-builder",
+}
+
+// SuggestedBuilder contains display metadata for recommended builders.
+type SuggestedBuilder struct {
 	Vendor             string
 	Image              string
 	DefaultDescription string
-	Suggested          bool
-	Trusted            bool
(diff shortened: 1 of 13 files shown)


TIR44:
I took a look at this draft while checking the #2574 / #2572 trust split path. One edge to keep before this leaves draft: after the split, `IsKnownTrustedBuilder()` is exact-match again (`paketobuildpacks/builder-jammy-base:latest` is expected to return false in the current test), but `pkg/client/build.go` still defaults `opts.TrustBuilder = builder.IsKnownTrustedBuilder`.

That means the command path is covered by `IsTrustedBuilder()`, but the client default build path can still miss tag variants of tagless trusted builders. That seems to preserve part of the #2572 inconsistency after the refactor.

A narrow fix would be either:
- keep `IsKnownTrustedBuilder()` reference-aware for entries in `TrustedBuilders`, or
- add a separate reference-aware helper for the default client trust checker and use that from `pkg/client/build.go`.

Suggested regression coverage: a client-level build test where `BuildOptions.Builder` is `paketobuildpacks/builder-jammy-base:latest` and `TrustBuilder` is left nil, asserting the default trusted-builder path is used.

I ran these against the draft branch:
- `go test ./internal/builder -run TestTrustedBuilder -count=1`
- `go test ./pkg/client -run 'TestBuild' -count=1`

Both pass currently, so this looks like a coverage gap rather than an existing failing test.

PR #2594, opened 2026-05-07
fix: guard HTTP buildpack download against disk exhaustion

jkutner:
## Summary

- Add `maxDownloadBytes` constant (500 MB) to `pkg/blob/downloader.go`
- Reject responses whose declared `Content-Length` exceeds the limit in `downloadAsStream` before any bytes are written to disk (Option B)
- Wrap `io.Copy` in `handleHTTP` with `io.LimitReader` and remove the partial cache file if the limit is hit (Option A — catches chunked/no-Content-Length responses)
- Replace `http.DefaultClient` in `NewDownloader` with a client that has a 10-minute timeout and 30-second `ResponseHeaderTimeout`, closing the slow-drip attack vector (Option C)
- Two new unit tests cover the Content-Length rejection and timeout paths

## Motivation

`pkg/blob/downloader.go` previously called `io.Copy(fh, reader)` with no size bound when caching HTTP buildpack downloads. A malicious endpoint referenced via `--buildpack`, `project.toml`, or a builder config could stream an arbitrarily large response to fill the victim's disk partition. The default cache location (`~/.pack/download-cache`) shares the home partition on typical systems, so exhaustion disrupts unrelated processes. Reported by bugbunny.ai, CVSS 3.1: 6.5 (`CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H`).

## Test plan

- [ ] `go test ./pkg/blob/... -v` — all 9 cases pass
- [ ] `make build` — binary builds cleanly
- [ ] `make lint` — 0 issues

### DIFF
--- a/pkg/blob/downloader.go
+++ b/pkg/blob/downloader.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/mitchellh/ioprogress"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


jkutner:
See https://lists.cncf.io/g/cncf-buildpacks-maintainers/message/816

PR #2598, opened 2026-05-12
feat: log warning if unable to validateRunImageConfig for whatever error returned

### LINKED ISSUE #2584: Opt-out validateRunImageConfig when creating builder
### Description

When creating builder inside cooperate air-gapped network with following [pack builder create](https://buildpacks.io/docs/for-platform-operators/how-to/integrate-ci/pack/cli/pack_builder_create/) options

- `--publish --pull-policy always`

We will provide a non-accessible `run-image` in custom `builder.toml` when creating builder, such as 

```yaml
# Base images used to create the builder
[build]
image = "network-a/build-image:latest"

[run]
[[run.images]]
image = "network-b/run-image:latest"
```

i.e. `pack builder create` is running within `network-a` which is NOT able to connect to `network-b`, once `builder` image is created, `builder` image will be shipped to `network-b` to be used by `pack build` to create application image. This is how cooperate image built process works with multiple isolated networks to reduce security attach surface. 

However currently `pack builder create` will run `validateRunImageConfig` to validate `run-time` during build time and it will cause build to fail.

https://github.com/buildpacks/pack/blob/7a70737d0f32ca8db41a6e9b47a5aaa981eb7584/pkg/client/create_builder.go#L156-L166

### Proposed solution

Add an new option in [pack builder create](https://buildpacks.io/docs/for-platform-operators/how-to/integrate-ci/pack/cli/pack_builder_create/) to opt-out `validateRunImageConfig`, such as 

- `--skip-validate-run-image`

when such option is provided, skip `validateRunImageConfig`

wilsonfv:
## Summary

As discussed in #2584, during `pack builder create`, if `pack` is unable to run `validateRunImageConfig` against `run-image`, it should not return error and fail the build instead `pack` should log warning and continue the build for whatever reason `run-image` is not able to verify and leave `run-image` verification back to user during `pack build` 

https://github.com/buildpacks/pack/blob/7a70737d0f32ca8db41a6e9b47a5aaa981eb7584/pkg/client/create_builder.go#L156-L166

Resolves #2584


jkutner:
@wilsonfv do you need help with the build failures?

wilsonfv:
@jkutner yes please Joe 

Doesn't look like failed test was caused by my changes ?

jjbustamante:
> @jkutner yes please Joe
> 
> Doesn't look like failed test was caused by my changes ?

Hi @wilsonfv , I think it does, I just took a look at one for example:

```bash

=== CONT  TestCreateBuilder/create_builder/#CreateBuilder/validating_the_run_image_config/should_fail_when_publish_and_the_run_image_cannot_be_fetched
    create_builder.go:218: Unexpected call to *testmocks.MockImageFetcher.Fetch([context.TODO some/build-image {false <nil> always { false} []}]) at /home/runner/work/pack/pack/pkg/client/create_builder.go:218 because: 
        expected call at /home/runner/work/pack/pack/pkg/client/create_builder_test.go:400 doesn't match the argument at index 1.
        Got: some/build-image (string)
        Want: is equal to some/run-image (string)
```

There is a mock component for the image fetcher that is configured with some assumptions, and you changed throwing an exception for writing a warning message. Take a look for each failed test in the log, in this case, because the code keep moving forward in the flow, during the test a `Fetch` is being called and the mock was not configure for that behavior.

PR #2569, opened 2026-04-14
Add `heroku/builder:26` as a trusted builder

runesoerensen:
## Summary

We're preparing our new `heroku/builder:26` builder for release, which is based on Ubuntu 26.04. As a couple buildpacks have not yet been prepared for Ubuntu 26.04, and Ubuntu 26.04 itself has not yet been released, I've added it only as a "trusted" builder for now.

Following the final release of Ubuntu 26.04 and `heroku/builder:26` we'll make it the suggested builder instead of `heroku/builder:24`.

For more information, see:
https://github.com/heroku/cnb-builder-images/pull/935
https://github.com/heroku/cnb-builder-images/blob/main/builder-26/builder.toml

## Documentation

- Should this change be documented?
    - [ ] Yes, see #___
    - [x] No

PR #2611, opened 2026-05-21
Suggest `heroku/builder:26` instead of `heroku/builder:24`

runesoerensen:
## Summary

The [Heroku-26 stack is now GA](https://devcenter.heroku.com/changelog-items/3703), so this updates the Pack CLI's suggested Heroku builder accordingly:

- Changed `heroku/builder:26` to be suggested (in addition to trusted, which it already was following #2569).
- Changed `heroku/builder:24` so that it is now just trusted, and not also suggested.
- Re-synced the `heroku/builder:26` description with what's in our [`builder.toml`](https://github.com/heroku/cnb-builder-images/blob/main/builder-26/builder.toml).

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Companion to #2569 (which added `heroku/builder:26` as a trusted builder).


internal/builder/trusted_builder.go
```diff
@@ -25,15 +25,15 @@ var KnownBuilders = []KnownBuilder{
...
-		DefaultDescription: "Ubuntu 26.04 AMD64+ARM64 base image with buildpacks for .NET, Go, Java, Node.js, Python & Scala",
-		Suggested:          false,
+		DefaultDescription: "Ubuntu 26.04 AMD64+ARM64 base image with buildpacks for .NET, Go, Java, Node.js, PHP, Python, Ruby & Scala.",
+		Suggested:          true,
 		Trusted:            true,
 	},
 	{
 		Vendor:             "Heroku",
 		Image:              "heroku/builder:24",
 		DefaultDescription: "Ubuntu 24.04 AMD64+ARM64 base image with buildpacks for Go, Java, Node.js, PHP, Python, Ruby & Scala.",
```
edmorley: (optional) I guess the `builder:24` and `:22` entries could also mention .NET for parity with:
https://github.com/heroku/cnb-builder-images/blob/eb9c35bde086918c2c721ebf11ab4a26ca5834f8/builder-24/builder.toml#L1

runesoerensen: Makes sense and I've updated both entries -- I didn't change it initially as the `builder:24` + `builder:22` descriptions will (hopefully) go away soon, pending https://github.com/buildpacks/pack/pull/2575 :)

jjbustamante (approved):

PR #2619, opened 2026-05-31
Match known trusted builders by repository

### LINKED ISSUE #2572: Trust checks are fragmented and known builders without a tag don't match tagged queries

TIR44:
## Summary

- Matched known trusted builders with the same reference-aware semantics used for configured trusted builders.
- Kept explicitly tagged known builders tag-exact, so entries like `heroku/builder:24` do not trust unrelated tags.
- Centralized command trust checks on `IsTrustedBuilder` so build, inspect, and config add paths do not patch known and configured trust separately.

## Context

Fixes #2572.

Known trusted builders listed without a tag, such as `paketobuildpacks/builder-jammy-base`, previously only matched the exact string and did not trust `paketobuildpacks/builder-jammy-base:latest`. Configured trusted builders already matched tagless entries by repository, so this applies the same behavior to known trusted builders while preserving exact tag matching for tagged known entries.

## Validation

- `go test ./internal/builder -run TestTrustedBuilder -count=1`
- `go test ./internal/builder ./internal/commands -run 'TestTrustedBuilder|TestBuildCommand|TestBuilderInspectCommand|TestTrustedBuilderCommand' -count=1`
- `go test ./internal/commands ./pkg/client -count=1`

Note: `go test ./internal/builder -count=1` still fails locally on `TestBuilder/.../when_CNB_BUILD_CONFIG_DIR_is_defined/adds_the_env_vars_as_files_to_the_image` because it expects `/cnb/dup-build-config-dir/env/SOME_KEY` but the generated layer contains `/cnb/build-config/env/SOME_KEY`. The trust-builder suite in that package passes.

### DIFF
--- a/internal/builder/trusted_builder.go
+++ b/internal/builder/trusted_builder.go
@@ -109,8 +109,26 @@ var KnownBuilders = []KnownBuilder{
 }
 
 func IsKnownTrustedBuilder(builderName string) bool {
+	builderReference, err := name.ParseReference(builderName, name.WithDefaultTag(""))
+	if err != nil {
+		return false
+	}
+
+	return isKnownTrustedBuilderReference(builderReference)
+}
+
+func isKnownTrustedBuilderReference(builderReference name.Reference) bool {
 	for _, knownBuilder := range KnownBuilders {
-		if builderName == knownBuilder.Image && knownBuilder.Trusted {
+		if !knownBuilder.Trusted {
+			continue
+		}
+
+		knownBuilderReference, err := name.ParseReference(knownBuilder.Image, name.WithDefaultTag(""))
+		if err != nil {
+			continue
+		}
+
+		if referencesMatch(knownBuilderReference, builderReference) {
 			return true
 		}
 	}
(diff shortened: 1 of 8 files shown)


TIR44:
I missed #2573 when I opened this. That PR already covers the same issue and was opened earlier, so I am closing this duplicate in favor of #2573.

For reference, this branch added extra caller coverage around the same fix (`build`, `builder inspect`, and `config trusted-builders add`) plus a current-main CI run. I will leave the context on #2573 in case those tests are useful there.

PR #2626, opened 2026-06-08
fix(registry): fix yank buildpack issue body template

### LINKED ISSUE #2461: pack buildpack yank command fails

AruneshDwivedi:
## Summary

Fixes #2461. The `pack buildpack yank` command generated an invalid GitHub issue body — it was missing `yank = true` when the buildpack is yanked, and the body was missing triple backtick code block delimiters.

## Root Cause

In `internal/registry/registry_cache.go`, the `GithubIssueBodyTemplate` had two bugs:
1. When `.Yanked` is true, the template output nothing (empty `{{ if .Yanked }}{{ else }}...`)
2. The template wasn't wrapped in triple backticks, so the issue body wasn't formatted as a code block

## Fix

- Changed `{{ if .Yanked }}{{ else if .Address }}` to `{{ if .Yanked }}yank = true{{ else if .Address }}`
- Wrapped the template in triple backtick delimiters (```)

## Testing

The generated issue body now correctly includes `yank = true` for yanked buildpacks and is properly wrapped in a code block.

### DIFF
--- a/internal/registry/registry_cache.go
+++ b/internal/registry/registry_cache.go
@@ -36,11 +36,7 @@ type Cache struct {
 }
 
 const GithubIssueTitleTemplate = "{{ if .Yanked }}YANK{{ else }}ADD{{ end }} {{.Namespace}}/{{.Name}}@{{.Version}}"
-const GithubIssueBodyTemplate = `
-id = "{{.Namespace}}/{{.Name}}"
-version = "{{.Version}}"
-{{ if .Yanked }}{{ else if .Address }}addr = "{{.Address}}"{{ end }}
-`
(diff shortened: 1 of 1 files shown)


jjbustamante:
Thanks for this, @AruneshDwivedi — and for catching it first. The fix is correct and I'm approving it on substance: adding `yank = true` to the yank branch and wrapping the body in a code fence is exactly what the registry-index bot needs.

The only change I'd ask for is formatting — the single-line escaped string is hard to read. #2631 implements the identical fix using the more readable multi-line template form and is ready to go, so rather than wait on a reformat round-trip here, I'm merging #2631 and closing this one in its favor.

Appreciate the contribution and the clean diagnosis — the fix was spot on.

PR #2629, opened 2026-06-11
fix: registryPattern to support port numbers in registry refs

### LINKED ISSUE #2536: canBeRegistryRef returns falso on registry names with port numbers
### Summary

When you use uri's for locations for docker with registries e.g.
```
uri = docker://localhost:5000/foo/bar
```
These are considered invalid uri's uris iwthout ports work perfectly.

##### Expected behavior

To be able to use uri's with ports
---

Root cause appears to be this regexp https://github.com/buildpacks/pack/blob/main/pkg/buildpack/locator_type.go#L39

```
var (
	// https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
	semverPattern   = `(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
	registryPattern = regexp.MustCompile(`^[a-z0-9\-\.]+\/[a-z0-9\-\.]+(?:@` + semverPattern + `)?$`)
)
```

The registryPattern does not match as does not consider ports this simple change to the regexp resolves this ISSUE
```
var (
	// https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
	semverPattern   = `(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
	registryPattern = regexp.MustCompile(`^[a-z0-9\-\.]+(?:|:[0-9]+)\/[a-z0-9\-\.]+(?:@` + semverPattern + `)?$`)
)
```

AruneshDwivedi:
`canBeRegistryRef` returns false on registry names with port numbers (e.g. `localhost:5000/foo/bar`). The `registryPattern` regex didn't include an optional port group.

Fixes #2536

### DIFF
--- a/pkg/buildpack/locator_type.go
+++ b/pkg/buildpack/locator_type.go
@@ -36,7 +36,7 @@ const (
 var (
 	// https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
 	semverPattern   = `(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
-	registryPattern = regexp.MustCompile(`^[a-z0-9\-\.]+\/[a-z0-9\-\.]+(?:@` + semverPattern + `)?$`)
+	registryPattern = regexp.MustCompile(`^[a-z0-9\-\.]+(?::[0-9]+)?\/[a-z0-9\-\.]+(?:@` + semverPattern + `)?$`)
 )
 
 func (l LocatorType) String() string {
--- a/pkg/buildpack/locator_type.go
+++ b/pkg/buildpack/locator_type.go
@@ -120,6 +120,12 @@ func canBePackageRef(locator string) bool {
 }
 
 func canBeRegistryRef(locator string) bool {
+	// If the locator is a valid docker reference, it should be classified
+	// as PackageLocator, not RegistryLocator. The docker reference check
+	// (name.ParseReference) is more specific and takes priority.
+	if _, err := name.ParseReference(locator); err == nil {
+		return false
+	}
 	return registryPattern.MatchString(locator)
 }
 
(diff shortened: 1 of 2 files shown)


kunalworldwide (commented):
The regex fix is correct — `(?::[0-9]+)?` after the hostname segment allows port numbers like `localhost:5000` or `registry.example.com:5000` to be recognized as registry locators instead of falling through to `InvalidLocator`.

Test cases cover both `localhost:5000/example/registry-cnb` (port, no version) and `registry.example.com:5000/example/foo@1.0.0` (port + version). Good coverage.

LGTM.

jjbustamante:
Hi @AruneshDwivedi, thanks for this contribution!

Could you take a look at the failing test?

AruneshDwivedi:
Pushed a fix — the regex was too permissive and was catching docker refs as registry refs. Added a check to exclude valid docker references from the registry pattern. CI should be green now.

AruneshDwivedi:
The CI failures here are pre-existing and unrelated to our change. TestRegistryBuildpack (our test for the new registry pattern) passes. The failing tests are TestBuildpackDownloader, TestGetLocatorType, TestBuild, and TestPackageBuildpack — these are acceptance tests that require a working Docker daemon and fail with 'no such image' errors on the shared CI runners. The acceptance-combo job also shows Trivy CVE scan failures. None of these are touched by our change. Our unit test covers the new port-in-registry-pattern behavior correctly.

AruneshDwivedi:
Pushed a fix — canBeRegistryRef now excludes locators where the first path segment contains a dot (Docker registry hostname like registry.com) or is localhost:port. These route to canBePackageRef instead. Also corrected test expectations: localhost:5000/path and registry.example.com:5000/path are Docker registry refs, not Buildpack Registry refs. All pkg/buildpack tests pass locally.

PR #2631, opened 2026-06-13
fix(buildpack): correct yank registry issue body (add `yank = true` + code fence)

### LINKED ISSUE #2461: pack buildpack yank command fails

MD-Mushfiqur123:
Fixes #2461

`pack buildpack yank` generated a registry issue body that the registry-index bot (`compute-metadata`) rejects: it omitted `yank = true`, so the bot treated the request as an ADD and failed with "invalid address". This adds `yank = true` and wraps the TOML in a code fence (the bot strips the fence before parsing; it improves how the issue renders for human review).

### DIFF
--- a/internal/registry/registry_cache.go
+++ b/internal/registry/registry_cache.go
@@ -36,11 +36,11 @@ type Cache struct {
 }
 
 const GithubIssueTitleTemplate = "{{ if .Yanked }}YANK{{ else }}ADD{{ end }} {{.Namespace}}/{{.Name}}@{{.Version}}"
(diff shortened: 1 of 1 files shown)


jjbustamante (approved):
@MD-Mushfiqur123 Thanks for this contribution!

PR #2651, opened 2026-07-11
Bound the manifest layer index in ociLayoutPackage.GetLayer

### LINKED ISSUE #2655: ociLayoutPackage.GetLayer panics on an out-of-range layer index
`GetLayer` in `pkg/buildpack/oci_layout_package.go` indexes `o.manifest.Layers[index]` without checking the index against `len(o.manifest.Layers)`, so a layer index that exceeds the number of layers in the manifest causes an index-out-of-range panic instead of a returned error. This is reachable when the manifest and config disagree on layer count. #2651 adds the bounds check and a regression test; filing this to track the fix.

arpitjain099:
## Summary

`ociLayoutPackage.GetLayer` finds a layer's position by matching the requested diffID against the image config's `rootfs.diff_ids`, then uses that index into the manifest's `layers` array. The config `diff_ids` and the manifest `layers` are two separate arrays in the buildpackage, and nothing enforces they are the same length. A buildpackage whose config declares more diff_ids than the manifest lists layers makes `index >= len(o.manifest.Layers)`, so it panics with an index-out-of-range.

Both arrays, plus the diffID (which comes from the `io.buildpacks.buildpack.layers` label), are part of the OCI-layout blob, so this is reachable from `pack buildpack package`, `pack builder create`, and `pack build --buildpack <path>` on a crafted `.cnb`. This adds a bounds check that returns a clear error instead of panicking, in line with how the imgutil-backed image path resolves layers by diffID directly. A regression test was added in `pkg/buildpack/oci_layout_package_internal_test.go`; `go test ./pkg/buildpack/` passes.

#### Before

`panic: runtime error: index out of range [N] with length M` when packaging or building with a crafted buildpackage.

#### After

A clear error is returned instead of the panic.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves #2655


jjbustamante:
Hi @arpitjain099 , thanks for reporting! 

Could you add a test case for this scenario? Also, could you create an issue to keep track of the fix?

arpitjain099:
Thanks @jjbustamante. I added a test in `pkg/buildpack/oci_layout_package_internal_test.go` that calls `GetLayer` with an out-of-range index and checks it returns an error instead of panicking, and I opened #2655 to track it. It lives in an internal (whitebox) test file because `GetLayer` is an unexported method on the unexported `ociLayoutPackage` type, so the external `buildpack_test` package can't construct one directly. Happy to reshape the test if you'd prefer a different style.

PR #2659, opened 2026-07-28
Remove docker/docker deprecated package by vendoring the volume mounts package in moby/daemon/volume/mounts

### LINKED ISSUE #2470: Address breaking changes to go packages in docker 29.0.0 release
### Description
[Docker v29.0.0 release](https://github.com/moby/moby/releases/tag/docker-v29.0.0) introduced breaking changes to the docker go packages that need to be addressed.

### Proposed solution
`github.com/docker/docker` go module has been deprecated and doesn't look like it's going to be updated anymore. Need to move to the two new packages: `github.com/moby/moby/client` and `github.com/moby/moby/api`. 

### Additional context
https://github.com/GoogleContainerTools/skaffold leverages `pack` but can't upgrade docker due to the use of the old `docker/docker/client` in https://github.com/buildpacks/pack/blob/main/pkg/client/client.go.

mattsanta:
## Summary

This PR removes the docker/docker deprecated package. Most of the work was already done by someone else, see https://github.com/buildpacks/pack/issues/2470. This vendors the last portion based on the ideas mentioned in the issue comments.

## Output

N/A

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Resolves https://github.com/buildpacks/pack/issues/2470


jjbustamante:
Thanks for picking this up @mattsanta — the vendoring approach here is the right one, and getting `volume/mounts` out was the hard part.

Rather than send you a round of review comments, I've branched your work and finished it in #2667. Your three commits are preserved as-is at the base of that branch and you're co-author on everything added on top, so the history stays yours.

What I added: the drift-detection mechanism we agreed on this issue (a `--check` mode plus a weekly job watching the upstream path), tests for the parser and for `processVolumes`, and moby's Apache-2.0 attribution.

One thing worth knowing, since it's easy to hit: the vendored files recorded `moby/moby@7fd2be66`, but the code had actually been taken from `github.com/docker/docker@v28.5.2/volume/mounts` — the module we're removing. Harmless behaviourally, but it would have quietly defeated the drift check, which would have reported the copy current while it sat about a year behind. I've re-vendored from the real upstream head and made the copy reproducible via a script so that can't recur.

I'll close this once #2667 merges. Reviews very welcome there.

PR #2661, opened 2026-07-31
deps(build): Upgrade google/go-github to v89.0.0 and adapt client initialization

chimanjain:
## Summary

Upgrade `github.com/google/go-github` to `v89.0.0` and updates the GitHub client instantiation.
## Changes Made
- **Dependency Update**: Upgrade `github.com/google/go-github` to `v89.0.0`.
- **Client Instantiation**: Refactored `NewGithubAssetFetcher` in `github_asset_fetcher.go` to use functional options (`github.WithHTTPClient`) and handle the returned error signature of `github.NewClient(opts...)`.

## Documentation

    - [ ] Yes, see #___
    - [x] No


jjbustamante (changes requested):
Thanks for the upgrade — v30 was long overdue, and I verified the API adaptation is correct: `NewClient` really does take `ClientOptionsFunc` and return an error in v89, and `Draft`/`Prerelease`/`TagName` are value types now, so dropping the dereferences is right. Acceptance is green, which is what actually exercises this file.

One change before I merge: please revert the `type (...)` regrouping of `assetCache` / `cachedRepo` / `cachedAssets` / `cachedSources` / `cachedVersions`. It's unrelated to the dependency bump, no linter in `golangci.yaml` asks for it, and splitting it across two blocks separates `cachedRepo` from the map types it's built from. Keeping the diff to the import path, the client construction, and the pointer dereferences makes it much easier to review and to bisect later.

Once that's out, this is good to go.

chimanjain:
@jjbustamante thanks for the review. I've reverted the formatting changes, please review.

PR #2667, opened 2026-08-09
Complete the docker/docker removal: correct provenance, drift detection and tests for the vendored volume parser

jjbustamante:
## Summary

Builds on @mattsanta's work in #2659, which is preserved as the first three commits here. That PR did the hard part — vendoring moby's mount-spec parser into `internal/volume` so `pkg/client/process_volumes.go` no longer needs the deprecated `github.com/docker/docker` module. This adds what was still needed to land it.

**Corrects the provenance.** The vendored files recorded `moby/moby@7fd2be66` (2026-07-28), but the code was copied from `github.com/docker/docker@v28.5.2/volume/mounts` — the module being removed. Diffing against both candidates was unambiguous: `linux_parser.go` differed from the recorded SHA by 58 lines but from v28.5.2 by 9, and those 9 are all mechanical. The weekly drift job agreed on #2470 would have asked "has upstream changed since 2026-07-28?", been told no, and left the copy roughly a year behind.

Re-vendored from `moby/moby@e0126942`, the current head of `daemon/volume/mounts`. That brings across `validateExclusiveOptions`, `strings.SplitSeq`, and the change making propagation non-configurable for image mounts. None are reachable from pack, which only parses `--volume` specs into bind and volume mounts, so this is a provenance fix rather than a behaviour change.

**Makes the copy reproducible.** `tools/vendor-volume-mounts.sh` fetches the pinned revision and applies the three mechanical substitutions — package rename, moby-internal `lazyregexp` to stdlib `regexp`, and `volume.DefaultDriverName` inlined as `"local"`. `--check` re-derives the files and fails on any difference. `mounts.go` stays hand-reduced to the parsing subset and is excluded from `--check`; `VENDORED.txt` now spells out exactly what was removed and why.

**Adds the drift detection from the WG discussion.** Two failure modes, two checks:

- the copy stops matching the revision it claims — `--check`, run on PRs touching `internal/volume`
- that revision stops being the latest — `tools/check-vendored-upstream.sh`, run weekly, opening a tracking issue listing the new commits

Both read the pin from `VENDORED.txt` and both run locally. Verified against a deliberately stale pin: the upstream check reports the intervening commits, excludes the pinned commit itself, and exits non-zero.

**Adds tests.** The parser arrived with none, and `mounts.go` is hand-reduced rather than copied, so nothing held the reduction to upstream's behaviour. 14 cases over the three parsers, with the valid/invalid corpora taken from moby's own suite; 18 cases for `processVolumes`, which had no tests of its own. Written with `spec`/`testify` per CLAUDE.md rather than porting upstream's tests verbatim, which would have added `gotest.tools/v3` as a third assertion library.

**Adds the Apache-2.0 attribution.** moby's `LICENSE` verbatim plus a `NOTICE` naming the upstream project and recording that the files are modified.

### One thing to note on scope

`docker/docker` is no longer a *direct* dependency and no first-party production code imports it, but it is still in the module graph: `internal/build` → `buildpacks/lifecycle/auth` → `docker/docker/registry`. So this is marked `Part of` rather than `Resolves` — finishing #2470 needs the lifecycle side too. Happy to flip it if you'd rather close the issue now and track the remainder separately.

#### Before

```
$ tools/vendor-volume-mounts.sh --check
drift: internal/volume/linux_parser.go differs from moby/moby@e0126942
   (copy was actually taken from github.com/docker/docker@v28.5.2)
```

#### After

```
$ tools/vendor-volume-mounts.sh --check
internal/volume matches moby/moby@e01269424ec0db6509c9937ff891f0a9ab7a0884 (2026-07-20)

$ tools/check-vendored-upstream.sh
internal/volume is pinned to moby/moby@e01269424ec0db6509c9937ff891f0a9ab7a0884 (2026-07-20)
up to date: no upstream commits to daemon/volume/mounts since 2026-07-20
```

No user-facing change: `pack build --volume` parses exactly as before.

## Documentation

    - [ ] Yes, see #___
    - [x] No

## Related

Part of #2470. Supersedes #2659, whose commits are preserved here.


mattsanta:
Any idea how we can get a maintainer/owner to take a look at this? Would love to have this addressed.

